PR #18, opened 2021-09-28
Add State type, methods and tests

### LINKED ISSUE #13: Add types and utils for states

geoknee:
Transcription of https://github.com/statechannels/statechannels/blob/f3095a4b6ec48ba7a07035e06d6f3841620a5ae5/packages/nitro-protocol/src/contract/state.ts.

Closes #13 

Adds:
- computation of channel id
- state encoding
- state hashing
- state signing
- signer recovery

I'd like to invite feedback in particular on 0d773e7. I decided that less is more so ripped out some structs; although I do think they may aid clarity overall.

### DIFF
--- a/channel/state/outcome/destination.go
+++ b/channel/state/outcome/destination.go
@@ -6,10 +6,7 @@ import (
 
 // IsExternalDestination returns true if the destination has the 12 leading bytes as zero, false otherwise
 func IsExternalDestination(destination types.Bytes32) bool {
-	for i, b := range destination[0:12] {
-		if i > 11 {
-			break
-		}
+	for _, b := range destination[0:12] {
 		if b != 0 {
 			return false
 		}


channel/state/state.go
```diff
@@ -105,11 +105,12 @@ func (s State) ChannelId() (types.Bytes32, error) {
 }
 
 func (a appPart) hash() (types.Bytes32, error) {
+
 	encodedAppPart, error := abi.Arguments{
 		{Type: uint256},
 		{Type: address},
 		{Type: bytesTy},
-	}.Pack(a.ChallengeDuration, a.AppDefinition, a.AppData)
+	}.Pack(a.ChallengeDuration, a.AppDefinition, []byte(a.AppData))
```
andrewgordstewart: @geoknee can you explain WHY this works?

geoknee: I'm not sure if I fully understand your question -- but I recall that the `.Pack` method is very picky about having exactly the right type fed into it. Here,  `a.AppData` is a custom type `types.Bytes` defined in the `types` package of our `go-nitro` module. Although the _underlying_ type is still `[]byte`, we have to explicitly convert the type to `[]byte` in order for `.Pack()` to work.

andrewgordstewart: https://github.com/statechannels/go-nitro/pull/130/commits/94e39a929839288adec7b891e51ed7e019fabe41

PR #26, opened 2021-10-14
Add `Decode` method to `Exit`, tests

NiloCK:
Adds:
Equals function for Exit type + tests
Decode function + test

Changes:
type of `Destination` on an `Allocation` from `types.Bytes32` to `[32]byte`. This was to enable the type assertion on the literal type returned from geth's `abi.Unpack` function (outcome.go, ln 133).


channel/state/outcome/outcome_test.go
```diff
@@ -58,3 +153,25 @@ func TestExitEncode(t *testing.T) {
...
+	var decodedExit, err = Decode(encodedExit)
+	if err != nil {
+		t.Error(err)
+	}
+
+	var a = Allocations{{
+		Destination:    common.HexToHash("0x00000000000000000000000096f7123E3A80C9813eF50213ADEd0e4511CB820f"),
+		Amount:         big.NewInt(1),
+		AllocationType: 0,
+		Metadata:       zeroBytes}}
```
geoknee, line 170: Could we / should we dry out this test by putting reused variables at the top of the file (outer scope)?

geoknee (approved):
LGTM, left two questions.

PR #28, opened 2021-10-15
Flowchart methods

NiloCK:
WIP, sharing because I won't be available much today.


protocols/common.go
```diff
@@ -50,6 +51,7 @@ type DirectFundingObjectiveState struct {
 	ChannelId types.Bytes32
 
 	ParticipantIndex map[types.Address]uint // the index for each participant
+	ExpectedStates   []state.State          // indexed by turn number
```
andrewgordstewart, line 54: 💡 it's worth considering storing a list of expected hashes (which are computed during initialization)

NiloCK:
Superseded by #41

PR #33, opened 2021-10-20
Objective.Crank() should return an Objective

geoknee:
Because cranking may update the state of the objective. This fix follows on from PR comments on #29.


andrewgordstewart:
@NiloCK are you suggesting that `crank` returns a `(CrankedObjective, err)` for some `crankedObjective` struct?

If so, I agree, as this appears more maintainable to me. But perhaps it's relatively unimportant right now. Should it be determined by whomever implements `crank`?

PR #34, opened 2021-10-21
Pass a (pointer to a) secret key from the store to objective.Crank()

geoknee:
We need to decide whether objectives synchronously sign state updates, or whether they declare them as side effects. 

My view is, if something *can* be performed synchronously during a crank, it should.

Therefore this PR adds a method to get a pointer to a secret key from the store. The engine calls this method and passes the pointer to an objective when it cranks it.

By passing the secret key by reference in this way, we avoid having multiple copies of the key in memory.


NiloCK (approved):
:+1:

PR #35, opened 2021-10-21
Remove initialize method from Objective interface

geoknee:
we will instead use a `pkg.New()` function, where each protocol is in it's own package


NiloCK (approved):
:+1:

Issue #14, opened 2021-09-15
Settle on the format / signature for off-chain protocols


geoknee:
They should be 
- pure functions
- possibly implement a state chart
   * https://statecharts.dev/concepts.html
   * possibly use a golang finite state machine / state chart library
- possibly follow the server-wallets "cranker" pattern

PR #36, opened 2021-10-21
Introduce chainservice and messageservice interfaces

geoknee:
Along with filling in some of the imperative code that interacts with those services. 

The implementation of these services work really belongs to Milestone 2 (and even then we may not need a _real_ chain service) ; but by considering their interfaces now, we can:
*  refine the `SideEffects` type (which is part of the `Objective` / protocols interface #14)
* form a more concrete model of information flow in the `client`


client/engine/engine.go
```diff
@@ -132,6 +125,11 @@ func (e *Engine) handleAPIEvent(apiEvent APIEvent) {
...
 // executeSideEffects executes the SideEffects declared by cranking an Objective
-func (e *Engine) executeSideEffects(protocols.SideEffects) {
-	// TODO
+func (e *Engine) executeSideEffects(sideEffects protocols.SideEffects) {
+	for _, message := range sideEffects.MessagesToSend {
+		e.toMsg <- message
+	}
+	for _, tx := range sideEffects.TransactionsToSubmit {
+		e.toChain <- tx
+	}
```
geoknee, line 134: This is the nub of the changeset: the engine now has outbound channels which are connected to services that run in their own goroutine.

client/engine/engine.go
```diff
@@ -2,22 +2,24 @@
...
-	"github.com/statechannels/go-nitro/types"
 )
 
 // Engine is the imperative part of the core business logic of a go-nitro Client
 type Engine struct {
 	// inbound go channels
-	API   chan APIEvent
-	Chain chan ChainEvent
-	Inbox chan Message
+	FromAPI   chan APIEvent // This one is exported so that the Client can send API calls
```
NiloCK, line 14: todo: ToAPI chan?

client/engine/chainservice/chainservice.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"math/big"
+
+	"github.com/statechannels/go-nitro/protocols"
+	"github.com/statechannels/go-nitro/types"
+)
+
+// ChainEvent is an internal representation of a blockchain event
+type Event struct {
```
NiloCK, line 12: As I understand it, this overly generic name may only be referred to in other packages as `chainservice.Event`, which is fine.

client/engine/engine.go
```diff
@@ -29,30 +31,21 @@ type APIEvent struct {
...
-	Holdings           map[types.Address]big.Int // indexed by asset
-	AdjudicationStatus protocols.AdjudicationStatus
-}
-
-// Message is an internal representation of a message from another client
-type Message struct {
-	ObjectiveId protocols.ObjectiveId
-	Sigs        map[types.Bytes32]state.Signature // mapping from state hash to signature
-}
-
```
NiloCK, line 44: This is de-duped from (dropped in favor of) `protocols/interfaces.go`?

client/engine/engine.go
NiloCK: Useful comment, thanks!

client/engine/engine.go
NiloCK: Another note: this is public to serve as a channel for a client to send events, but client code will also see the exposed named `FromAPI`, where `ToEngine` might make more sense from that perspective.

This seems like a common problem that has a conventional solution.

NiloCK (approved):
LGTM. Engine.go remains extremely comprehensible.

client/engine/engine.go
geoknee: About `ToAPIchan`: I think we don't want just a single `chan` for the nitro client to send to its parent application (or even for the `engine` to send to the `client`). 

The pattern instead is as follows. The application calls `client.SomeMethod()`.  The `client` then creates a dedicated `chan engine.Response` and embeds it an `APIEvent`. When the `APIEvent` is  sent it to the `engine`:
https://github.com/statechannels/go-nitro/blob/7f8e3c5cc3e7b60ff1b2891a50987094686a136b/client/client.go#L27-L36

, the engine can use it to communicate the response to `SomeMethod`. The `SomeMethod` call returns the `chan engine.Response` so that the parent application has a handle on it.

geoknee: In terms of naming the channel at each "end" (one routine sends on it, the other receives on it). I think in the above example `c.engine.FromAPI <- apiEvent` is pretty reasonable. But there are plenty of other tricks we could use, which may or may not make things more readable:

1. A simple alias
```go
	toEngine := c.engine.FromAPI
	toEngine <- apiEvent
```
2. A method on the engine that wraps the send in a function (example here uses a type switch)
```go
c.engine.SendTo(apiEvent)


func (e *Engine) SendTo(i interface{}) {
	switch i := i.(type) {
	case APIEvent:
		e.FromAPI <- i
// etc
	default:
	}
}
```
This approach would also allow us to make all of the engines channels private.

geoknee: It's also possible to get greater type safety on `chans`, by stipulating their direction https://gobyexample.com/channel-directions. AFAICT you can only do that by passing `chans` as function parameters. There's probably a way we can do that.

PR #38, opened 2021-10-22
Objective methods mutate the receiver

geoknee:
* by allowing objective.Crank to mutate the objective, we don’t need to return it. This sidesteps some typing headaches (see https://www.notion.so/statechannels/Golang-onboarding-218e0e14a51d4aed93b5acd119df86c4#714c8e2ffbbe4a64936712a5e8660e50 )

- we can maintain an overall functional approach by having the store objective getter always returning a copy (deep clone) of the objective. This will happen "for free" in most store implementations (with the exception of a memory store)
- another advantage is that now cloning is done in far fewer places overall (just in the store, not in several objective methods). We can write one test to ensure the store passes a copy and not a reference

### DIFF
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -86,10 +86,11 @@ func (e *Engine) handleMessage(message Message) {
 	objective := e.store.GetObjectiveById(message.ObjectiveId)
 	event := protocols.ObjectiveEvent{Sigs: message.Sigs}
 	secretKey := e.store.GetChannelSecretKey()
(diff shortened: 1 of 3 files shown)


geoknee:
I'll close this for now, since I think it's not a priority to tackle immediately. Let's move the discussion [to notion](https://www.notion.so/statechannels/Protocol-interface-e076fe46c0774122818d1abb7cfd0c5a#328ebcc6b85348a8b44bd9d0ea60c7f2).

Issue #30, opened 2021-10-18
Write a direct funding protocol


geoknee:
Comply with the interface chosen when #14 resolves.

This is being worked on in https://github.com/statechannels/go-nitro/tree/direct-funder-rewrite.

# TODO
Left undone are:

- [ ] #50
- [ ] #51
- [ ] #57
- [ ] #52
- [ ] #53

NiloCK:
Outstanding Qs:

For filecoin, presumably the only _asset_ in question is FIL. go-nitro has an exit-format style outcome defined in [channel/state/outcome/outcome.go](https://github.com/statechannels/go-nitro/blob/main/channel/state/outcome/outcome.go). The `DirectFundingObjectiveState` has single-asset style `MyDepositSafetyThreshold`, `MyDepositTarget`, and `FullyFundedThreshold` properties.

- Do we intend / want to use the outcome as-is, with the 0-address indicating whatever chain's native currency?
- Should the above properties be re-written as `[tokenAddress]: amount` maps to align with the exit-format?

(I hadn't thought much about the intersection of deposit safety thresholds and multi-asset channels - I guess the reasonable approach is to make one deposit tx when all assets have reached the safety threshold.)

geoknee:
> * Do we intend / want to use the outcome as-is, with the 0-address indicating whatever chain's native currency?

I would say yes -- that's a relatively nice way to preserve compatibility with our solidity code which ought to handle FIL, too.

> * Should the above properties be re-written as `[tokenAddress]: amount` maps to align with the exit-format?

I would say also yes. At the cost of a bit more boilerplate now, we can preserve the option of having a multi-asset system in the future. I think it's worth it -- how do you feel?

 
> (I hadn't thought much about the intersection of deposit safety thresholds and multi-asset channels - I guess the reasonable approach is to make one deposit tx when all assets have reached the safety threshold.)

Yes I think this is reasonable. I think if you look at our[ existing code targeting Ethereum](https://docs.statechannels.org/contract-api/natspec/MultiAssetHolder#deposit), we only support one asset type per deposit. So it may be a) `n` transactions when all assets reach their safety threshold. Or it could be b) one transaction when each of the `n` asset reaches its safety threshold.

andrewgordstewart:
This is being worked on in https://github.com/statechannels/go-nitro/tree/direct-funder-rewrite.

NiloCK:
Closing, with follow-on issues created.

PR #41, opened 2021-11-02
Direct-funder

NiloCK:
Work todards #30 plus removal of prior speculative work (StateChart style objectives).


channel/channel.go
```diff
@@ -0,0 +1,24 @@
+package channel
+
+import (
+	"math/big"
+
+	"github.com/statechannels/go-nitro/channel/state"
+)
+
+type Channel interface {
```
geoknee, line 9: I believe we may indeed want something like this interface -- but I think it is unreferenced at the moment, so should probably be cut?

protocols/common.go
```diff
@@ -0,0 +1,112 @@
+package protocols
```
geoknee, line 1: As a part of the tidy up (i.e. removing mention of state charts), I would have thought this file should be combined with `direct-funding.go`.

protocols/readme.md
```diff
@@ -0,0 +1,30 @@
+This package currently contains two prototype implementations of the direct funding protocol.
```
geoknee, line 1: This file also should be tidied up a bit.

geoknee: Optional: we could record the way we *aren't* doing it (and why) in an architectural decision record md file in this directory?

types/types.go
```diff
@@ -18,3 +18,47 @@ type Bytes []byte
...
+
+// IsNonZero returns true if the Holdings structure has any non-zero asset
+func (h Holdings) IsNonZero() bool {
+	for _, v := range h {
+		if v.Cmp(big.NewInt(0)) == 1 {
+			return true
+		}
+	}
+	return false
+}
```
geoknee, line 33: Should these methods live in this file?

protocols/direct-funding.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+func (s DirectFundingObjectiveState) FundingUpdated(amount types.Holdings, finalized bool) (DirectFundingObjectiveState, error) {
+	updated := s.Clone()
+	updated.OnChainHolding.Add(amount)
+	return updated, nil
+}
+
+// SignatureReceived updates the objective's cache of which participants have signed which states
+func (s DirectFundingObjectiveState) SignatureRecieved(signature state.Signature, turnNum int) (DirectFundingObjectiveState, error) {
```
geoknee, line 139: I think we want to combine all of these named methods into the `Update` method. (You may want to keep them as private methods which are executed inside of a loop).

protocols/direct-funding.go
```diff
@@ -0,0 +1,198 @@
...
+// (it assumes that events are handled by pushing information into the store)
+
+// Crank inspects the extended state and declares a list of Effects to be executed
+// It's like a state machine transition function where the finite / enumerable state is returned (computed from the extended state)
+// rather than being independent of the extended state; and where there is only one type of event ("the crank") with no data on it at all
+func (s DirectFundingObjectiveState) Crank(secretKey *[]byte) (Objective, SideEffects, WaitingFor, error) {
+
+	// Input validation
+	if s.Status != Approved {
+		return s, NoSideEffects, WaitingForNothing, ErrNotApproved
```
geoknee, line 20: We should return a clone of `s`, I think.

(Even if cloning is trivial, it's nice to have it written out explicitly).

The pattern I anticipate is to clone `s` immediately, and then go about mutating the clone before returning it.

geoknee:
I'll close this for now as this is being taken forward in #42 .

PR #44, opened 2021-11-09
Add unit test file for direct-funder

geoknee:
This PR:

* adds a test file for the direct-funding protocol
* adds a test for the constructor
* ... which reveals the constructor currently panics...
* ...fixes the panics
* factors out a `TestState` object so tests do not need to construct new literal `state.State` objects.

### DIFF
--- a/channel/state/test-fixtures.go
+++ b/channel/state/test-fixtures.go
@@ -1,0 +1,27 @@
+package state
+
+import (
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/statechannels/go-nitro/channel/state/outcome"
+	"github.com/statechannels/go-nitro/types"
+)
+
+var chainId, _ = big.NewInt(0).SetString("9001", 10)
+
+var TestState = State{
+	ChainId: chainId,
+	Participants: []types.Address{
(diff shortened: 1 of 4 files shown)


channel/state/test-fixtures.go
```diff
@@ -0,0 +1,27 @@
...
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/statechannels/go-nitro/channel/state/outcome"
+	"github.com/statechannels/go-nitro/types"
+)
+
+var chainId, _ = big.NewInt(0).SetString("9001", 10)
+
+var TestState = State{
```
NiloCK, line 13: I see that this is imported in `/protocols`, so must be an export.

Should globally useful test-fixture code like this live in a text-fixtures package, to avoid diluting the namespace of each package?

(ie, now `state.TestState` is visible everywhere `state` is imported)

NiloCK (approved):
:+1:

channel/state/test-fixtures.go
geoknee: My naive attempt at this has fallen foul of 
```
package github.com/statechannels/go-nitro/channel/state
	imports github.com/statechannels/go-nitro/fixtures: import cycle not allowed in test`
```

NiloCK: Right... Not immediately clear to me if there's a way out of that loop.

PR #55, opened 2021-11-16
Fixes from review on 42

NiloCK:
Responses to 4 comments from @andrewgordstewart in PR #42.

Recommend reading per-commit; commit messages contain links to original context.

### DIFF
--- a/channel/state/state.go
+++ b/channel/state/state.go
@@ -2,6 +2,7 @@ package state
 
 import (
 	"errors"
+	"math/big"
 
 	"github.com/ethereum/go-ethereum/accounts/abi"
 	"github.com/ethereum/go-ethereum/crypto"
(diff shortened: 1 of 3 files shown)


channel/state/state.go
```diff
@@ -165,3 +166,24 @@ func (s State) RecoverSigner(sig Signature) (types.Address, error) {
 	}
 	return RecoverEthereumMessageSigner(stateHash[:], sig)
 }
+
+// Clone returns a clone of the state
+func (s State) Clone() State {
+	clone := State{}
+
+	// shallow clone of fixed part
```
andrewgordstewart, line 174: Is there an easy way for me to know that the shallow clone here is a safe clone?

protocols/direct-funding.go
```diff
@@ -103,6 +103,8 @@ func NewDirectFundingObjectiveState(initialState state.State, myAddress types.Ad
 			init.MyDepositSafetyThreshold[assetAddress] = threshold
 			init.MyDepositTarget[assetAddress] = myShare.Add(myShare, threshold)
 		}
+	} else {
+		return init, err
```
andrewgordstewart, line 107: Here, I would also prefer the "check errors, then do stuff".
```
channelOutcome, err := outcome.Decode(initialState.VariablePart().EncodedOutcome);
if err != nil {
  // PANIC!
}

...
```

Is there a general guideline for golang developers which addresses this style choice? (I understand the verbosity of my snippet above is one of the pain points of golang.)

andrewgordstewart (approved):
LGTM!

protocols/direct-funding.go
NiloCK: In this case, there's a more concrete benefit of `if [initialize]; condition { [imperative] } else { [handleError] }`: it limits the scope of the initialized variables to just the `if` block.

Maybe it is better for the (much shorter) error condition to occur first: `if [init]; err != nil { handleError() } else { doStuff() }`

All of this aside, I think the entire condition here can / should be scrapped. More on that in another PR.

channel/state/state.go
NiloCK: These are shallow clones of values that _should never change_.

PR #56, opened 2021-11-16
Embed `VariablePart` and `FixedPart` in `State`

NiloCK:
A thought that occurred when seeing `state.go`, given previous discussion of struct embedding.

This PR can be viewed and discarded, as it will conflict in more-trouble-than-it's-worth ways with #55.


andrewgordstewart (commented):
My general feeling is, I prefer embedding things in `State`, rather than having a flat state.

PR #60, opened 2021-11-16
Remove unnecessary `encode`/`decode` of Outcome

NiloCK:
This `if` block amounts to a unit-test of the `encode` and `decode` functions. As the code stood, it was checking the well-formed-ness of the initialState.Outcome by (implicitly) encoding it and then decoding it.

### DIFF
--- a/protocols/direct-funding.go
+++ b/protocols/direct-funding.go
@@ -6,7 +6,6 @@ import (
 	"math/big"
 
 	"github.com/statechannels/go-nitro/channel/state"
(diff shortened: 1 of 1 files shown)


andrewgordstewart (approved):
> This if block amounts to a unit-test of the encode and decode functions. As the code stood, it was checking the well-formed-ness of the initialState.Outcome by (implicitly) encoding it and then decoding it.

FYI it really helps discoverability if this is in the _commit_ message rather than the _PR_ body. (Since this PR has a single commit, the PR will automatically be created with the commit message as the PR body.)

PR #61, opened 2021-11-17
run golangci-lint first in github action, instead of setup-go

### LINKED ISSUE #31: reduce noise from linting in our github action
There seem to be a lot of errors coming from setting up the linter, which distracts from the much smaller number of actual linting problems: 
https://github.com/statechannels/go-nitro/runs/3938205972

geoknee:
golangci-lint already sets up go and is designed to be run first

Closes #31 . This PR drastically improves the output of the linting stage of our CI. 

This was manually tested by introducing an intentional linting error: https://github.com/statechannels/go-nitro/runs/4238244289?check_suite_focus=true . I have since removed the offending commit.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -12,16 +12,11 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Set up Go
-        uses: actions/setup-go@v2
-        with:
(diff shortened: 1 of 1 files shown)


NiloCK (approved):
Great :+1:

Issue #49, opened 2021-11-12
Consider: refactor `types.Destination` as embedded type


NiloCK:
https://github.com/statechannels/go-nitro/pull/42/commits/8c473b3218f6bab3219d6b74cb4e940f241901e5

`types.Destination` is motivated by providing a semantically meaningful narrower subtype to geth's `common.Hash`, which itself is (in essence) a decorated `[32]byte`. Originally, it was an alias to `common.Hash`, but this didn't allow appending the helper functions `ToAddress` and `IsAddress`.

As implemented (a defined type `type Destination common.Hash`) it is necessary to explicitly define forwarding functions (`String`, `Bytes`) that exist on `common.Hash`.

A third option is as a defined type with `common.Hash` as an [embedded type](https://golang.org/doc/effective_go#embedding).

```golang
type Bytes32 = common.Hash


// previous - func (d Destination) destinationHelper(){} not allowed
type Destination = Bytes32

// current - requires reimplementation of methods on Bytes32
type Destination Bytes32

// to investigate - what annoyances lie down this path?
type Destination struct {
  Bytes32
}
```

NiloCK:
I tried this out on branch `embedded-destination-type` - 42ae59d.

**Pro**: forwarding methods are removed from `Destination`, but still available via the embedded `Bytes32` struct
**Con**: constructing **nested** literals becomes still-more verbose, because the linter wants the field `Bytes32` to be labelled. See, for example, `outcome_test.go` in the linked commit.
**Blocker**: the current code fails `go test ./...` because of the abi unmarshaling:

```text
2021/11/16 11:21:29 error unmarshallingjson: cannot unmarshal non-string into Go struct field Allocation.Allocations.Destination of type common.Hash
```

My initial fiddling with `allocationsTy` and `rawAllocationsType` were not successful. @geoknee: you had more success with me in the marshal / unmarshal department. If there is a quick fix that you see we may prefer to merge this. If not, I think the current implementation is not so awful.

NiloCK:
Closed (or deferred) via #62

PR #62, opened 2021-11-17
Convert `Destination` to embed `Bytes32`...

NiloCK:
**NOTE**: this branch is broken. Draft PR opened for comment.

From https://github.com/statechannels/go-nitro/issues/49#issuecomment-970409722 :

> I tried this out on branch `embedded-destination-type` - [42ae59d](https://github.com/statechannels/go-nitro/commit/42ae59d0c875a4562a506c273514d64d1c35b3af).
> 
> **Pro**: forwarding methods are removed from `Destination`, but still available via the embedded `Bytes32` struct **Con**: constructing **nested** literals becomes still-more verbose, because the linter wants the field `Bytes32` to be labelled. See, for example, `outcome_test.go` in the linked commit. **Blocker**: the current code fails `go test ./...` because of the abi unmarshaling:
> 
> ```
> 2021/11/16 11:21:29 error unmarshallingjson: cannot unmarshal non-string into Go struct field Allocation.Allocations.Destination of type common.Hash
> ```
> 
> My initial fiddling with `allocationsTy` and `rawAllocationsType` were not successful. @geoknee: you had more success with me in the marshal / unmarshal department. If there is a quick fix that you see we may prefer to merge this. If not, I think the current implementation is not so awful.

### DIFF
--- a/channel/state/outcome/outcome_test.go
+++ b/channel/state/outcome/outcome_test.go
@@ -13,13 +13,13 @@ import (
 func TestEqualAllocations(t *testing.T) {
 
 	var a1 = Allocations{{ // [{Alice: 2}]
-		Destination:    types.Destination(common.HexToHash("0x0a")),
+		Destination:    types.Destination{Bytes32: common.HexToHash("0x0a")},
 		Amount:         big.NewInt(2),
 		AllocationType: 0,
 		Metadata:       make(types.Bytes, 0)}}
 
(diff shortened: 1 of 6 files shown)


geoknee:
I couldn't find a quick fix, so perhaps we should stick with the current implementation for now?

NiloCK:
Agreed.

Closes #49

PR #63, opened 2021-11-18
Export R,S,V components of `state.Signature`

geoknee:
And swap out participant 0 in TestState (we would like to use an address we know the private key for)

These changes allow us to:
* construct literal signatures in packages other than `state`
* produce valid signatures by participants of the `TestState` (because we have recorded a private key for one of the participants

Both of these abilities are necessary for completing #51.

### DIFF
--- a/channel/state/state_test.go
+++ b/channel/state/state_test.go
@@ -8,15 +8,15 @@ import (
 )
 
 // The following constants are generated from our ts nitro-protocol package
-var correctChannelId = common.HexToHash(`b79270eb4cf4d11dcd5cf44c6337b1c9e4730dc3c26f022567bcf2eb63557a72`)
-var correctStateHash = common.HexToHash(`3e460a311caf589f1cf80036adfd092d05a30ff72f234918c5cdbc6c4333343a`)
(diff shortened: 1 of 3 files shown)


channel/state/state_test.go
```diff
@@ -8,15 +8,15 @@ import (
 )
 
 // The following constants are generated from our ts nitro-protocol package
-var correctChannelId = common.HexToHash(`b79270eb4cf4d11dcd5cf44c6337b1c9e4730dc3c26f022567bcf2eb63557a72`)
-var correctStateHash = common.HexToHash(`3e460a311caf589f1cf80036adfd092d05a30ff72f234918c5cdbc6c4333343a`)
+var correctChannelId = common.HexToHash(`cd578811171bb0291c7dc59081b9a910960b78001dece1954c00da9d8c12a925`)
+var correctStateHash = common.HexToHash(`e77a7a870087be40371b8c9b9f50a7c0e410387ce7ea8a22c9c61ea837277420`)
```
geoknee, line 12: These changes follow-on from the change in participants of the `TestState`.

NiloCK (approved):
:+1:

PR #64, opened 2021-11-18
Add unit tests for `Funds` methods

NiloCK:
This PR

- adds pretty comprehensive testing of accounting arithmetic around the `Funds` account->balance map
- fixes an implementation bug in `Sum`
- fixes the `String` method which had an invalid base (64) (funds.go:22)

### DIFF
--- a/types/funds.go
+++ b/types/funds.go
@@ -51,17 +51,25 @@ func Sum(a ...Funds) Funds {
 	return sum
 }
 
-// Equal returns true if reciever `f` and input `g` are identical.
+// Equal returns true if reciever `f` and input `g` are identical in value.
 //
-// Note that a zero-balance does NOT equal a non-balance: {[0x0a,0x00]} != {}
+// Note that a zero-balance equals a non-balance: {[0x0a,0x00],[0x0b,0x01]} == {[0x0b,0x01]}
 func (f Funds) Equal(g Funds) bool {
-	if len(f) != len(g) {
-		return false
-	}
+	return f.isMatchedBy(g) && g.isMatchedBy(f)
+}
+
+// isMatchedBy returns true if each of `f`'s non-zero asset balances is matched by the
+// same asset-balance in `g`
+func (f Funds) isMatchedBy(g Funds) bool {
+	zero := big.NewInt(0)
 
 	for asset, amount := range f {
-		if g[asset].Cmp(amount) != 0 {
-			return false
+		// only check g for non-zero f balances
+		if amount.Cmp(zero) > 0 {
+			gAmount, ok := g[asset]
+			if !ok || gAmount.Cmp(amount) != 0 {
+				return false
+			}
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


types/funds_test.go
```diff
@@ -0,0 +1,200 @@
...
+	},
+	"e": {
+		common.HexToAddress("0x00"):  big.NewInt(1),
+		common.HexToAddress("0xabc"): big.NewInt(0),
+	},
+	"f": {
+		common.HexToAddress("0x00"): big.NewInt(2),
+	},
+	// manually calculated sums
+	"ab": {
```
geoknee, line 41: Nit: could you key this as "a+b" (and so on for the other manually calculated entries)?

types/funds.go
```diff
@@ -42,11 +42,40 @@ func (h Funds) Add(a ...Funds) Funds {
...
 
 	return sum
 }
+
+// Equal returns true if reciever `f` and input `g` are identical in value.
+//
+// Note that a zero-balance equals a non-balance: {[0x0a,0x00],[0x0b,0x01]} == {[0x0b,0x01]}
+func (f Funds) Equal(g Funds) bool {
+	return f.canAfford(g) && g.canAfford(f)
+}
```
geoknee, line 63: 1. We discussed whether we want to consider zero-balances with non-balances. I think we know about some situations where we wouldn't that kind of behaviour -- for example where the respective hashes of each side in the "equality"  are important.  That said, I'm fine with this implementation for now. It's well documented and we can amend or extend it later.
2. The implementation is pretty elegant (calling into `canAfford`). But I guess it's also a bit surprising, and probably also a bit less efficient than a more naive approach? Interested to know your thoughts.

geoknee (approved):

types/funds.go
andrewgordstewart: Re. 1, as long as there's a mapping from `Funds` to `Outcome` such that `f.Equals(g)` implies `hash(f.AsOutcome) == hash(g.AsOutcome)`, we're fine.

Considering this, defining `f.Equals(g)` as `hash(f.AsOutcome) == hash(g.AsOutcome)` may make sense, but is probably _less_ efficient, as it relies on cryptography.

We are currently not concerned with efficiency, so IMO it could go either way. I like the idea of avoiding cryptography if there's a safe alternative implementation, but we can always implement the safe, efficient alternative at a later point, and so I would lean towards the _obviously correct_ implementation.

geoknee: I think `Funds` is primarily there to represent on-chain `holdings` rather than on chain `Outcomes`.  As long as we keep that separation clear, we shouldn't run into trouble with the current notion of `Funds.Equal`.

If we want to convert from `Funds` to `Outcomes`, then I think we need the stricter / more opinionated notion that @andrewgordstewart mentions.

NiloCK: Seeing now that a response I had written was not submitted and no longer exists... BUT, @geoknee has cleanly articulated the separation between `Funds` and `Outcome`s.

The other point was about **2)**. This `Equal` method naturally fell out of the more naive approach. Golang doesn't have friendly access to map keys, unions of collections, etc, so most any implementation will involve looping over the keys from both sides.  After doing that, it was natural to refactor the symmetric operation into its own function, which happens to be the semantically meaningful `canAfford`.

Issue #65, opened 2021-11-18
Consider whether `direct-funding.Update()` should depend on the current state


geoknee:
The following code can result in a valid, relevant signature being discarded:

https://github.com/statechannels/go-nitro/blob/f70aaf40a74e0cd1468ebab36ff4cc956c5dc8d1/protocols/direct-funding.go#L143-L149

The `turnNum` is derived from the current state, and then used to choose which state hash to recover the signature against.

It violates the principle that _objectives should update independently of their current state_*. If a peer sends me a postfund signature before I have all of the prefund signatures -- this implementation of the objective would result in my wallet dropping that signature.

An alternative would be to try and recover the signature against *both* turn numbers (in turn, say). This might imply a bit of a performance hit in some cases.

*I think this principle is not necessarily one we want to always subscribe to.... but I think it feels mostly right, here. Perhaps the real principle is: _don't drop updates if they arrive at an unexpected point_.

geoknee:
This was addressed in #115

PR #66, opened 2021-11-18
Add unit tests for direct-funding.Update()

geoknee:
Adds unit tests for the `.Update()` method on the direct funding objective. 

Recall that the [direct funding objective is a struct holding data particular to the off-chain goal of opening a directly funded channel](https://github.com/statechannels/go-nitro/blob/f60dce55abf9faee7f8fecf90a9446664ad45a65/protocols/direct-funding.go#L30-L48). As such, it needs to update that data based on events that arrive (for example, a digital signature on a state update, or an event fired by the chain).  The `Update()` method is performing this function.

The approach I took here was to:
1. prepare an `event`
2. call `DirectFundingObjectiveState.Update(event)`
3. assert on the various data fields on the returned `DirectFundingObjectiveState` , as well as the returned error
4. progressively modify the event, and GOTO 2

Towards #51.

It's easy enough to describe the intended behaviour for expected events. For unexpected events, there is a little freedom. This test can be used to *define* the behaviour we want. Right now, an incorrect channel id on the event returns an error. An unexpected or faulty signature causes errors internally, but these are logged and ignored. In the latter case, this PR includes the changes that implement that behaviour (so that the test passes). For now, logging is just ` fmt.Print`, but we have an issue to improve that (#70) .

### DIFF
--- a/protocols/direct-funding.go
+++ b/protocols/direct-funding.go
@@ -148,7 +148,10 @@ func (s DirectFundingObjectiveState) Update(event ObjectiveEvent) (Objective, er
 
 		err := updated.applySignature(sig, turnNum)
 		if err != nil {
-			return s, err
+			// If there was an error applying the signature, log it and swallow it
+			// This is a conscious choice (to ignore signatures we don't expect)
+			// Examples include faulty signatures, signatures by non-participants, signatures on unexpected states, etc
+			fmt.Println(err)
 		}
 	}
 
@@ -295,15 +298,16 @@ func (s DirectFundingObjectiveState) applySignature(signature state.Signature, t
 	index, ok := s.ParticipantIndex[signer]
 	if !ok {
 		return fmt.Errorf("signature recieved from unrecognized participant 0x%x", signer)
-	}
+	} else {
 
-	if turnNum == 0 {
-		s.PreFundSigned[index] = true
-	} else if turnNum == 1 {
-		s.PostFundSigned[index] = true
-	}
+		if turnNum == 0 {
+			s.PreFundSigned[index] = true
+		} else if turnNum == 1 {
+			s.PostFundSigned[index] = true
+		}
 
-	return nil
+		return nil
+	}
 }
 
 // todo: is this sufficient? Particularly: s has pointer members (*big.Int)
--- a/protocols/direct-funding.go
+++ b/protocols/direct-funding.go
@@ -298,16 +298,15 @@ func (s DirectFundingObjectiveState) applySignature(signature state.Signature, t
 	index, ok := s.ParticipantIndex[signer]
 	if !ok {
 		return fmt.Errorf("signature recieved from unrecognized participant 0x%x", signer)
-	} else {
-
-		if turnNum == 0 {
-			s.PreFundSigned[index] = true
-		} else if turnNum == 1 {
-			s.PostFundSigned[index] = true
-		}
-
-		return nil
 	}
+
+	if turnNum == 0 {
+		s.PreFundSigned[index] = true
+	} else if turnNum == 1 {
+		s.PostFundSigned[index] = true
+	}
+	return nil
+
 }
 
 // todo: is this sufficient? Particularly: s has pointer members (*big.Int)
--- a/protocols/direct-funding_test.go
+++ b/protocols/direct-funding_test.go
@@ -81,8 +81,8 @@ func TestUpdate(t *testing.T) {
 	if err != nil {
 		t.Error(err)
 	}
-	if updated.(DirectFundingObjectiveState).OnChainHolding[common.Address{}].Cmp(big.NewInt(3)) != 0 {
-		t.Error(`Objective data not updated as expected`)
+	if !updated.(DirectFundingObjectiveState).OnChainHolding.Equal(e.Holdings) {
+		t.Error(`Objective data not updated as expected`, updated.(DirectFundingObjectiveState).OnChainHolding, e.Holdings)
 	}
 
 }
--- a/protocols/direct-funding_test.go
+++ b/protocols/direct-funding_test.go
@@ -31,10 +31,8 @@ var privateKeyOfParticipant0 = common.Hex2Bytes(`caab404f975b4620747174a75f08d98
 var correctSignatureByParticipant, _ = stateToSign.Sign(privateKeyOfParticipant0)
 
 func TestUpdate(t *testing.T) {
-	// First, prepare a new objective using the constructor:
-	s, _ := NewDirectFundingObjectiveState(state.TestState, state.TestState.Participants[0])
 
-	// Next, prepare an event with a mismatched channelId
+	// Prepare an event with a mismatched channelId
 	e := ObjectiveEvent{
 		ChannelId: types.Destination{},
 	}
--- a/protocols/direct-funding_test.go
+++ b/protocols/direct-funding_test.go
@@ -1,14 +1,93 @@
 package protocols
 
 import (
+	"math/big"
 	"testing"
 
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/statechannels/go-nitro/channel/state"
+	"github.com/statechannels/go-nitro/types"
 )
 
+// TestNew tests the constructor using a TestState fixture
 func TestNew(t *testing.T) {
 	_, err := NewDirectFundingObjectiveState(state.TestState, state.TestState.Participants[0])
 	if err != nil {
 		t.Error(err)
 	}
+}
+
+// Construct various variables for use in TestUpdate
+var s, _ = NewDirectFundingObjectiveState(state.TestState, state.TestState.Participants[0])
+var dummySignature = state.Signature{
+	R: common.Hex2Bytes(`49d8e91bd182fb4d489bb2d76a6735d494d5bea24e4b51dd95c9d219293312d9`),
+	S: common.Hex2Bytes(`22274a3cec23c31e0c073b3c071cf6e0c21260b0d292a10e6a04257a2d8e87fa`),
+	V: byte(1),
+}
+var dummyStateHash = common.Hash{}
+var stateToSign state.State = s.ExpectedStates[0]
+var stateHash, _ = stateToSign.Hash()
+var privateKeyOfParticipant0 = common.Hex2Bytes(`caab404f975b4620747174a75f08d98b4e5a7053b691b41bcfc0d839d48b7634`)
+var correctSignatureByParticipant, _ = stateToSign.Sign(privateKeyOfParticipant0)
+
+func TestUpdate(t *testing.T) {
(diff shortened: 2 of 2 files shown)


protocols/direct-funding.go
```diff
@@ -148,7 +148,10 @@ func (s DirectFundingObjectiveState) Update(event ObjectiveEvent) (Objective, er
 
 		err := updated.applySignature(sig, turnNum)
 		if err != nil {
-			return s, err
+			// If there was an error applying the signature, log it and swallow it
+			// This is a conscious choice (to ignore signatures we don't expect)
+			// Examples include faulty signatures, signatures by non-participants, signatures on unexpected states, etc
+			fmt.Println(err)
```
NiloCK, line 154: I think this is appropriate, but it raises the question of logging, which I've given about zero thought to.

protocols/direct-funding.go
```diff
@@ -295,15 +298,16 @@ func (s DirectFundingObjectiveState) applySignature(signature state.Signature, t
...
-	if turnNum == 0 {
-		s.PreFundSigned[index] = true
-	} else if turnNum == 1 {
-		s.PostFundSigned[index] = true
-	}
+		if turnNum == 0 {
+			s.PreFundSigned[index] = true
+		} else if turnNum == 1 {
+			s.PostFundSigned[index] = true
+		}
```
NiloCK, line 307:
```
if !ok {
 return err
}

//regular flow
```
vs
```
if !ok {
 return err
} else {
  // regular flow
}
```

The "implicit else" style of error-handling and control flow is not generally my favorite, but it seems ubiquitous in go. This is probably because the (comma, ok) and `func() (returnType, error)` idioms are so common that the alternative to implicit-else style is for everything to move one-indent-to-the-right every time something doesn't fail.

protocols/direct-funding_test.go
```diff
@@ -1,14 +1,93 @@
...
+func TestUpdate(t *testing.T) {
+	// First, prepare a new objective using the constructor:
+	s, _ := NewDirectFundingObjectiveState(state.TestState, state.TestState.Participants[0])
+
+	// Next, prepare an event with a mismatched channelId
+	e := ObjectiveEvent{
+		ChannelId: types.Destination{},
+	}
+	// Assert that Updating the objective with such an event returns an error
+	// TODO is this the behaviour we want? Below with the signatures, we prefer a log + NOOP (no error)
```
NiloCK, line 42: Are there concrete practical differences between a noop that returns an error and a noop that logs one? Each puts the handling / investigation duties back to calling code. Returning the error forces calling code to explicitly handle things.

protocols/direct-funding_test.go
```diff
@@ -1,14 +1,93 @@
...
+	_, err := s.Update(e)
+	if err == nil {
+		t.Error(`ChannelId mismatch -- expected an error but did not get one`)
+	}
```
NiloCK, line 46:
```suggestion
	if _, err := s.Update(e); err == nil {
		t.Error(`ChannelId mismatch -- expected an error but did not get one`)
	}
```

This is a recommended style in order to minimize scope of variables. ie, prevent accidentally referring to `err` later in the same function when its meaning has become stale.

protocols/direct-funding_test.go
```diff
@@ -1,14 +1,93 @@
...
+}
+var dummyStateHash = common.Hash{}
+var stateToSign state.State = s.ExpectedStates[0]
+var stateHash, _ = stateToSign.Hash()
+var privateKeyOfParticipant0 = common.Hex2Bytes(`caab404f975b4620747174a75f08d98b4e5a7053b691b41bcfc0d839d48b7634`)
+var correctSignatureByParticipant, _ = stateToSign.Sign(privateKeyOfParticipant0)
+
+func TestUpdate(t *testing.T) {
+	// First, prepare a new objective using the constructor:
+	s, _ := NewDirectFundingObjectiveState(state.TestState, state.TestState.Participants[0])
```
NiloCK, line 35: This `s` is shadowing the file-scoped `s`, while other variables used through the function were all defined at file-scope and not shadowed.

Probably either:
- drop this shadowing declaration & just use the file-scoped variables, OR
- move all of the declarations inside of `TestUpdate`

protocols/direct-funding_test.go
```diff
@@ -1,14 +1,93 @@
...
+	// Updating the objective with this event should overwrite the holdings that are stored
+	e.Holdings = types.Funds{}
+	e.Holdings[common.Address{}] = big.NewInt(3)
+	updated, err = s.Update(e)
+	if err != nil {
+		t.Error(err)
+	}
+	if updated.(DirectFundingObjectiveState).OnChainHolding[common.Address{}].Cmp(big.NewInt(3)) != 0 {
+		t.Error(`Objective data not updated as expected`)
+	}
```
NiloCK, line 91: post-merges, this looks like a place for the shiny new `Equal` function of `Funds`.

NiloCK (approved):
This is great progress. Are we still missing:

- tests that a DFOS correctly transitions from handling prefund to postfund signatures
- ... anything else ?

protocols/direct-funding.go
geoknee: 4c0506a

protocols/direct-funding.go
geoknee: #70

protocols/direct-funding_test.go
geoknee: c1ec08f

protocols/direct-funding_test.go
geoknee: I think the option of not returning an error is making the explicit choice that the calling code does *not* need to explicitly handle the error generated deeper in the call stack. So for a malformed signature, some library code is going to return an error that our protocol needs to explicitly handle. But our protocol is able to say "this is not an error that the calling function -- i.e. the engine -- needs to be concerned with". I do not expect the engine to inspect the logs -- they are there only for debugging or analytics purposes. 

Essentially, I'm considering the pattern where we swallow / ignore an error at run time, but still record it for other non runtime needs.

protocols/direct-funding_test.go
geoknee: d02b4de

geoknee:
> * tests that a DFOS correctly transitions from handling prefund to postfund signatures

I created https://github.com/statechannels/go-nitro/issues/65 which speaks to this issue somewhat.

geoknee:
> * ... anything else ?

Possibly a check that `Update()` is indeed pure?

NiloCK:
> Possibly a check that `Update()` is indeed pure?

This seems like a separate issue maybe? Can this test  be written all-at-once sometime in the future for all implementations of `Objective`?

This PR looks pretty good to me now!

geoknee:
Requesting a review from @hannahhoward :-) If you wouldn't mind concentrating on the Go language side of things, that would be great -- but of course all and any feedback welcome.

protocols/direct-funding_test.go
```diff
@@ -1,14 +1,88 @@
...
 
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/statechannels/go-nitro/channel/state"
+	"github.com/statechannels/go-nitro/types"
 )
 
+// TestNew tests the constructor using a TestState fixture
 func TestNew(t *testing.T) {
 	_, err := NewDirectFundingObjectiveState(state.TestState, state.TestState.Participants[0])
 	if err != nil {
```
hannahhoward, line 15: if you find yourselves getting annoyed by go's built-in very minimal testing assertions, I recommend checking out https://github.com/stretchr/testify . I can source other recommendations from team if you need (I think I remember maybe something that is more lightweight?)

protocols/direct-funding_test.go
```diff
@@ -1,14 +1,88 @@
 package protocols
```
hannahhoward, line 1: something to consider here: there are two options for module here. you can use:
1. "protocols"
2. "protocols_test"

Which you choose has a subtle effect on your tests. Both will get dropped from compiled code. However, the second will mean you are in a different module, which means you'll only be able to access public vars/types. Sometimes this can be useful to make sure you're only testing public interfaces. (which may force important decisions). Other times, it may be really hard to test public interfaces without having access to a couple private vars. There's no right answer -- just a helpful language feature to know about.

protocols/direct-funding_test.go
hannahhoward: If I read your code right,  `NewDirectFundingObjectiveState` errors if TestState.ChannelID() errors. I wonder if you want to make an InvalidTestState and test both conditions.

protocols/direct-funding_test.go
```diff
@@ -1,14 +1,88 @@
...
+	V: byte(1),
+}
+var dummyStateHash = common.Hash{}
+var stateToSign state.State = s.ExpectedStates[0]
+var stateHash, _ = stateToSign.Hash()
+var privateKeyOfParticipant0 = common.Hex2Bytes(`caab404f975b4620747174a75f08d98b4e5a7053b691b41bcfc0d839d48b7634`)
+var correctSignatureByParticipant, _ = stateToSign.Sign(privateKeyOfParticipant0)
+
+func TestUpdate(t *testing.T) {
+
```
hannahhoward, line 34: This test looks like it is essentially constructing various events and sending them to the update method on an initial test state, and asserting expected results. A common pattern for golang for this is table tests -- https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

The above blog lays out a good explanation. It can take a tiny bit of getting used to but I find myself using them a lot.

hannahhoward (commented):
Overall this is great. I put a bunch of suggestions but mostly they are pointers to help you out.

protocols/direct-funding_test.go
NiloCK: This is great, thanks. Go compiler makes really liberal use of magic strings (`init` functions, `main` package, `internal` packages, etc) and this is yet-another example I hadn't been aware of.

Will check these comments out more thoroughly tomorrow.

andrewgordstewart: I would lean towards `protocols_test` to place the constraints on the tests. Sounds we can switch at the drop of a hat with 1LOC change in case that adds too much of a burden.

Switching in the other direction is more burdensome unless we are adhering to the "only test public interfaces" rule. And if we are, why not lean on the compiler?

protocols/direct-funding_test.go
andrewgordstewart: We love table tests: https://github.com/statechannels/statechannels/search?q=test.each

geoknee: We do, and we've now started writing them in Go: https://github.com/statechannels/go-nitro/pull/86/commits/859a1eb635beba8a8677d61cb7e94576f2ae43c8 .

PR #72, opened 2021-11-24
Improve testing of DirectFundingObjectiveState constructor

### LINKED ISSUE #71: Test direct funding objective constructor with a faulty state
If I read your code right,  `NewDirectFundingObjectiveState` errors if TestState.ChannelID() errors. I wonder if you want to make an InvalidTestState and test both conditions.

_Originally posted by @hannahhoward in https://github.com/statechannels/go-nitro/pull/66#discussion_r755484868_

geoknee:
Closes #71 .

There are two ways which the constructor can return an error (at the top layer of the stack)

* Error calculating the channel id
    - because we get an extremely rare hash collision. _This is so rare its actually a pretty big challenge to generate the appropriate test data_
    - because of some error thrown in go-ethereum's encoding library code. _I tried and failed to trigger this error, and have left the git history on this branch showing how_
* Error because of the use of a final state _I added a test for this_

### DIFF
--- a/protocols/direct-funding_test.go
+++ b/protocols/direct-funding_test.go
@@ -11,10 +11,16 @@ import (
 
 // TestNew tests the constructor using a TestState fixture
 func TestNew(t *testing.T) {
-	_, err := NewDirectFundingObjectiveState(state.TestState, state.TestState.Participants[0])
-	if err != nil {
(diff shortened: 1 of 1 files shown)


NiloCK (approved):
Yup

PR #74, opened 2021-11-24
add AddressToDestination helper and test

geoknee:
This is useful in tests, and likely also in production code, too.

### DIFF
--- a/types/destination.go
+++ b/types/destination.go
@@ -38,7 +38,7 @@ func (d Destination) Bytes() []byte {
 
 func AdddresstoDestination(a Address) Destination {
 	d := Destination{0}
-	for i := range d[12:] {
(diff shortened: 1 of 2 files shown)


types/destination.go
```diff
@@ -35,3 +35,11 @@ func (d Destination) String() string {
 func (d Destination) Bytes() []byte {
 	return Bytes32(d).Bytes()
 }
+
+func AdddresstoDestination(a Address) Destination {
+	d := Destination{0}
+	for i := range d[12:] {
```
NiloCK, line 41:
```suggestion
	for i := range a {
```

Requires less cognition I think. (Test still works).

NiloCK (approved):
:+1:

PR #75, opened 2021-11-24
Introduce a nontrivial `TestOutcome`

geoknee:
⚠️  This PR is stacked on #74 -- I would like it to be reviewed after that PR has been merged to main and the base branch of this PR changes.

### DIFF
--- a/protocols/direct-funding.go
+++ b/protocols/direct-funding.go
@@ -79,6 +79,8 @@ func NewDirectFundingObjectiveState(initialState state.State, myAddress types.Ad
 	}
 
 	init.FullyFundedThreshold = types.Funds{}
+	init.MyDepositSafetyThreshold = types.Funds{}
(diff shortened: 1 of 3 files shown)


protocols/direct-funding.go
```diff
@@ -79,6 +79,8 @@ func NewDirectFundingObjectiveState(initialState state.State, myAddress types.Ad
 	}
 
 	init.FullyFundedThreshold = types.Funds{}
+	init.MyDepositSafetyThreshold = types.Funds{}
```
geoknee, line 82: I wonder if it is better/worse/the same as to write `make(types.Funds)` ? The `make` syntax allows us to initialize the size of the map, if we want to (it defaults to a "small" size).

NiloCK: I wonder this as well.

NiloCK (approved):
:+1:

Issue #53, opened 2021-11-15
Return `SideEffects` from `DirectFundObjective.Crank()`


NiloCK:
The `DirectFundObjective.Crank()` method does not currently return any nontrivial `SideEffects`. This prevents an integration test from succeeding. 

To close this issue, we should return proper side effects (rather than `NoSideEffects`). The types / interfaces for those side effects may or may not need to change. 

Example: 

https://github.com/statechannels/go-nitro/blob/543c6dcf0e000e6de96b5f3f3871cddb7e8f37c1/protocols/direct-fund/direct-fund.go#L136-L140

---
(edited by @geoknee )

geoknee:
This issue should cover adding unit tests for returning the right `SideEffects` from the `Crank()`.

lalexgap:
@geoknee/ @NiloCK  Can the direct fund objective be responsible for signing a state and adding the signature to the `channel`? Or should the signature request be a`SideEffect`?

geoknee:
> @geoknee / @NiloCK Can the direct fund objective be responsible for signing a state and adding the signature to the `channel`? Or should the signature request be a`SideEffect`?

The signature does not need to be a side effect, so it should be done synchronously during the `.Crank()`.

PR #79, opened 2021-11-24
Add unit tests for `DirectFund.Crank()`

### LINKED ISSUE #51: Add state transition tests to `DirectFundingObjectiveState.` `Update`, `Crank`

geoknee:
Brings test coverage of the `protocols` package up to ~86%.

Does not yet assert on the correct `SideEffects` -- but those are sill TODO anyway #53 .

Closes #51.

### DIFF
--- a/protocols/direct-funding_test.go
+++ b/protocols/direct-funding_test.go
@@ -96,3 +96,77 @@ func TestUpdate(t *testing.T) {
 	}
 
 }
+
+func TestCrank(t *testing.T) {
+	// Assert that cranking an unapproved objective returns an error
+	if _, _, _, err := s.Crank(&privateKeyOfParticipant0); err == nil {
+		t.Error(`Expected error when cranking unapproved objective, but got nil`)
+	}
+
+	// Approve the objective, so that the rest of the test cases can run.
+	o := s.Approve()
+
+	// To test the finite state progression, we are going to progressively mutate o
+	// And then crank it to see which "pause point" (WaitingFor) we end up at.
+
+	// Initial Crank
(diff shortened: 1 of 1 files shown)


protocols/direct-funding_test.go
```diff
@@ -96,3 +96,77 @@ func TestUpdate(t *testing.T) {
...
+		t.Error(err)
+	}
+	if waitingFor != WaitingForCompletePostFund {
+		t.Errorf(`WaitingFor: expected %v, got %v`, WaitingForCompletePostFund, waitingFor)
+	}
+
+	// Manually progress the extended state by collecting postfund signatures
+	o.(DirectFundingObjectiveState).PostFundSigned[0] = true
+	o.(DirectFundingObjectiveState).PostFundSigned[1] = true
+	o.(DirectFundingObjectiveState).PostFundSigned[2] = true
```
NiloCK, line 159: There's a comment to be made here that `PostFundSigned` should probably be `postFundSigned` - that updates to, especially, the security-related state should not be possible from outside of each individual `Objective` implementation.

That said, you've done it here only by use of the explicit cast, which we don't intend to employ outside of test code.

No suggestion - just flagging something to think about.

NiloCK (approved):
:+1:

protocols/direct-funding_test.go
geoknee: I agree. Let's make this change in a rapid, follow up PR (it will set a precedent for the other protocols).

geoknee: #84

Issue #25, opened 2021-10-14
Write a virtual funding protocol


geoknee:
Comply with the interface settled on when #14 resolves.

PR #82, opened 2021-11-25
Add GuaranteeMetadata type and methods

geoknee:
This is preparatory work towards #25 .

The virtual funding protocol will need to consume Guarantees and to be able to understand them -- for example, in answering the question "does ledger channel `L` adequately guarantee funds for virtual channel `V`". 

This work brings us toward that goal. Note that I have chosen to go with the SATP format for the guarantee. This isn't set in stone and should be easily changed. (I also chose not-the-latest format from SATP, too).

### DIFF
--- a/channel/state/outcome/guarantee.go
+++ b/channel/state/outcome/guarantee.go
@@ -1,7 +1,6 @@
 package outcome
 
 import (
-	"bytes"
 	"encoding/json"
 	"log"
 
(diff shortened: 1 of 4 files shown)


channel/state/outcome/guarantee.go
```diff
@@ -0,0 +1,60 @@
...
+
+	"github.com/ethereum/go-ethereum/accounts/abi"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/statechannels/go-nitro/types"
+)
+
+// A Guarantee is an Allocation with AllocationType == GuaranteeAllocationType and Metadata = encode(GuaranteeMetaData)
+type GuaranteeMetadata struct {
+	Left  types.Address // The peer who plays the role of Alice (peer 0)
+	Right types.Address // The peer who plays the role of Bob (peer n+1, where n=len(peers))
```
NiloCK, line 15: Is there a case for these to be `types.Destination`s instead?

NiloCK, line 15:
```suggestion
	Left  types.Address // The peer who plays the role of Alice (peer 0)
	Right types.Address // The peer who plays the role of Bob (peer n+1, where n=len(intermediaries))
```

channel/state/outcome/guarantee.go
```diff
@@ -0,0 +1,60 @@
...
+	"encoding/json"
+	"log"
+
+	"github.com/ethereum/go-ethereum/accounts/abi"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/statechannels/go-nitro/types"
+)
+
+// A Guarantee is an Allocation with AllocationType == GuaranteeAllocationType and Metadata = encode(GuaranteeMetaData)
+type GuaranteeMetadata struct {
```
NiloCK, line 13: The `Left`, `Right` guarantee is a special case of the prior, more general ordered list. SATP virtual funding uses this special case, but is it better for the package to retain the more flexible format?

My instinct: probably not, as the larger list serviced the prior recursive VC construction, and (as far as I know) no other purpose.

NiloCK (approved):
LGTM

channel/state/outcome/guarantee.go
geoknee: I think yes.

geoknee: eee4945

PR #83, opened 2021-11-26
Introduce allocations.Affords method

geoknee:
Preparatory work towards #25 

This will be useful for deciding whether a given channel is funded by another.
    
By combining information about how the other channel is funded (e.g. on chain), this method infers whether enough coins "flow" to the given channel.
    
By matching against the entire allocation (not just the channel id), it is possible to filter for the given allocation having the correct type and metadata -- for example, the given allocation might be a guarantee taking a specific form.

### DIFF
--- a/channel/state/outcome/outcome.go
+++ b/channel/state/outcome/outcome.go
@@ -69,7 +69,7 @@ func (allocations Allocations) AffordsFor(given Allocation, x *big.Int) *big.Int
 			affordsInTotal = affordsInTotal.Add(affordsInTotal, affords)
 		}
 
-		surplus = surplus.Sub(surplus, affords)
+		surplus.Sub(surplus, affords)
 
 	}
 	return affordsInTotal
--- a/channel/state/outcome/outcome.go
+++ b/channel/state/outcome/outcome.go
@@ -55,8 +55,8 @@ func (a Allocations) Total() *big.Int {
 // AffordsFor computes the amount that the allocations can afford for a given allocation, assuming it is funded with x coins
 func (allocations Allocations) AffordsFor(given Allocation, x *big.Int) *big.Int {
 	bigZero := big.NewInt(0)
-	affordsInTotal := bigZero
-	surplus := x
+	affordsInTotal := big.NewInt(0).Set(bigZero)
+	surplus := big.NewInt(0).Set(x)
 	for _, allocation := range allocations {
 
 		if surplus.Cmp(bigZero) == 0 {
--- a/channel/state/outcome/outcome.go
+++ b/channel/state/outcome/outcome.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/ethereum/go-ethereum/accounts/abi"
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/math"
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/statechannels/go-nitro/types"
 )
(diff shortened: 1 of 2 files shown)


channel/state/outcome/outcome.go
```diff
@@ -51,6 +52,29 @@ func (a Allocations) Total() *big.Int {
...
+	for _, allocation := range allocations {
+
+		if surplus.Cmp(bigZero) == 0 {
+			break
+		}
+
+		affords := math.BigMin(surplus, allocation.Amount)
+
+		if allocation.Equal(given) {
+			affordsInTotal = affordsInTotal.Add(affordsInTotal, affords)
```
NiloCK, line 69: Should we just return `affords` here? (Or else, relax the `Equal` check to `allocation.destination == given.destination`).

As is, this method returns the total afforded to a destination if each allocation to that destination is for the same amount, but less than that if the allocations to the destination are for different amounts.

IE:
`[{Alice: 2, Bob: 3, Alice: 2}].AffordsFor({Alice:2}, 100)` returns `4`, but
`[{Alice: 2, Bob: 3, Alice: 3}].AffordsFor({Alice:2}, 100)` returns `2`

channel/state/outcome/outcome_test.go
```diff
@@ -163,18 +163,53 @@ func TestExitDecode(t *testing.T) {
...
+
 	total := a.Total()
 	if total.Cmp(big.NewInt(5)) != 0 {
 		t.Errorf(`Expected total to be 5, got %v`, total)
 	}
 }
+
+func TestAffordsFor(t *testing.T) {
+
+	allocationsWithRepeatedEntry := append(a, a[0]) // [{Alice: 2, Bob: 3, Alice: 2}]
```
NiloCK, line 186: Given above comment, could also use test case `[{Alice:2, Bob:3, Alice: 4}]` to communicate intent around these distinct cases.

NiloCK (commented):
Either I don't understand the purpose of this method, or you've made too strong of an `Equals` check and this method probably doesn't (always) do what you want it to. Small fixes.

channel/state/outcome/outcome.go
```diff
@@ -51,6 +52,29 @@ func (a Allocations) Total() *big.Int {
...
+		}
+
+		surplus = surplus.Sub(surplus, affords)
```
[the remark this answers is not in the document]
geoknee: Agreed 875ace2 . I think what threw me here is that even `z.Sub(z,y)` already seems needlessly long. I would have liked to written `z.Sub(y)`, which seems to me to be a better analog for `z-=y`.

As it was, I went the other way and had `z` in there _three_ times...

channel/state/outcome/outcome.go
```diff
@@ -51,6 +52,29 @@ func (a Allocations) Total() *big.Int {
 	return total
 }
 
+// AffordsFor computes the amount that the allocations can afford for a given allocation, assuming it is funded with x coins
+func (allocations Allocations) AffordsFor(given Allocation, x *big.Int) *big.Int {
```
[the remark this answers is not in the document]
geoknee: Firstly, I think we don't actually have a discrete method or function like that right now. We do compute something like that inside of `transfer`, though. 

I think there is a benefit to checking the entire `given Allocation`, which I put into the commit message of 3d08340 (should have included in PR description, too).  I think this is something we want to do off chain but probably not on chain.

channel/state/outcome/outcome.go
geoknee: Hmm yeah this is a case I didn't give enough thought to. It seems as though want we really want is to check against the destination, the type and the metadata but *not* the amount. In the words, accumulate the total amount flowing _in a particular way_ to _a particular destination_. 

Thinking about it from the virtual funding protocol's point of view, though -- perhaps it is better to return have a `Affords` function returning a `bool`? Then, the protocol can ask of an outcome -- do you have this exact guarantee (which it will have precomputed), and can you (completely) afford it?

geoknee: a28bc0b

geoknee:
In this PR I failed to give enough context to where I was going, and also failed to come up with a method that makes sense in all cases (including edge cases with repeated but nonidentical allocations). Thanks @NiloCK for helping me realise this. 

I've tried to do better over on #86 , so closing this.

Issue #46, opened 2021-11-10
Expose `Status` on the `Objective` interface?


NiloCK:
These could be moved to the top level `interfaces.go` file? It also occurs to me that we may want to expose a `Status` method on the `Objective` interface?

_Originally posted by @geoknee in https://github.com/statechannels/go-nitro/pull/42#discussion_r745505689_

geoknee:
It's not clear if we need this at the moment.

PR #85, opened 2021-11-26
make all data private  on DirectFundingObjective

### LINKED ISSUE #84: Make each objective's data private
There's a comment to be made here that `PostFundSigned` should probably be `postFundSigned` - that updates to, especially, the security-related state should not be possible from outside of each individual `Objective` implementation.

That said, you've done it here only by use of the explicit cast, which we don't intend to employ outside of test code.

No suggestion - just flagging something to think about.

_Originally posted by @NiloCK in https://github.com/statechannels/go-nitro/pull/79#discussion_r757621344_

geoknee:
Save for `Status` (pending #46 ). 

Closes #84


NiloCK (approved):
Somehow in my head this refactor would have made the strategy used in the tests unworkable.

Thankfully the world exists outside my head.

Issue #78, opened 2021-11-24
Convert store and objective interfaces to use `Channel` abstraction


geoknee:
See https://www.notion.so/statechannels/Protocol-interface-e076fe46c0774122818d1abb7cfd0c5a#43cbff515d3d49fd99cfb3a4dcd8c821

- [ ] Store
- [x] `virtual-fund` objective
- [x] `direct-fund` objective

geoknee:
Tracking the remaining work in #191

PR #86, opened 2021-11-28
Add Allocations.Affords() method

geoknee:
This method will allow the virtual funding protocol to check the following important condition before signing a postfund state for a virtual channel: "have the ledger channel(s) I care about been updated correctly?"

"Correctly" here means both:
1. That they include a specific allocation (for each asset).This allocation will be computed when the virtual funding objective is spawned. All fields, including the allocation type and metadata (which will encode a guarantee), must be correct. Furthermore:
2. The ledger channel must itself be sufficiently funded to "cover" the allocation in question.

---
For more context, I anticipate this method will be used in the following place (this is a commit from a work in progress branch):
https://github.com/statechannels/go-nitro/blob/e1b7446150128e6eb5fcd3f9a97fa1d60e862a78/protocols/virtual-fund/virtual-funding.go#L154-L165

The method would replace the `.GuaranteesFor` placeholder that I have here. It will eventually be wired up to the protocol via a `Channel` class #78 (work in progress for that viewable on the same commit) .

### DIFF
--- a/channel/state/outcome/outcome.go
+++ b/channel/state/outcome/outcome.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/ethereum/go-ethereum/accounts/abi"
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/math"
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/statechannels/go-nitro/types"
 )
@@ -53,6 +54,32 @@ func (a Allocations) Total() *big.Int {
 		total.Add(total, allocation.Amount)
 	}
 	return total
+}
+
+// Affords returns true if the allocations can afford the given allocation given input funds of x, false otherwise.
+//
+// To afford the given allocation, the allocations must include something equal-in-value to it,
+// as well as having sufficient funds left over for it after reserving funds from x for all allocations with higher priority.
+// Note that "equal-in-value" implies the same allocation type and metadata (if any).
+func (allocations Allocations) Affords(given Allocation, x *big.Int) bool {
+	bigZero := big.NewInt(0)
+	surplus := big.NewInt(0).Set(x)
+	for _, allocation := range allocations {
+
+		if surplus.Cmp(bigZero) == 0 {
+			break
+		}
+
+		affords := math.BigMin(surplus, allocation.Amount)
+
+		if allocation.Equal(given) && affords.Cmp(allocation.Amount) >= 0 {
+			return true
+		}
+
+		surplus.Sub(surplus, affords)
+
+	}
+	return false
 }
 
 // SingleAssetExit declares an ordered list of Allocations for a single asset.
--- a/channel/state/outcome/outcome.go
+++ b/channel/state/outcome/outcome.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/ethereum/go-ethereum/accounts/abi"
 	"github.com/ethereum/go-ethereum/common"
-	"github.com/ethereum/go-ethereum/common/math"
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/statechannels/go-nitro/types"
 )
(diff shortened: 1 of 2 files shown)


channel/state/outcome/outcome_test.go
```diff
@@ -163,18 +163,52 @@ func TestExitDecode(t *testing.T) {
...
 
+var a = Allocations{{ // [{Alice: 2, Bob: 3}]
+	Destination:    types.Destination(common.HexToHash("0x0a")),
+	Amount:         big.NewInt(2),
+	AllocationType: 0,
+	Metadata:       make(types.Bytes, 0)}, {
+	Destination:    types.Destination(common.HexToHash("0x0b")),
+	Amount:         big.NewInt(3),
+	AllocationType: 0,
+	Metadata:       make(types.Bytes, 0)}}
```
NiloCK, line 174:
```suggestion
var a = Allocations{ // [{Alice: 2, Bob: 3}]
	{
		Destination:    types.Destination(common.HexToHash("0x0a")),
		Amount:         big.NewInt(2),
		AllocationType: 0,
		Metadata:       make(types.Bytes, 0)},
	{
		Destination:    types.Destination(common.HexToHash("0x0b")),
		Amount:         big.NewInt(3),
		AllocationType: 0,
		Metadata:       make(types.Bytes, 0)},
}
```

channel/state/outcome/outcome_test.go
```diff
@@ -163,18 +163,52 @@ func TestExitDecode(t *testing.T) {
...
+		"case 2": {a, a[0], big.NewInt(1), false},
+		"case 3": {a, a[1], big.NewInt(6), true},
+		"case 4": {a, a[1], big.NewInt(5), true},
+		"case 5": {a, a[1], big.NewInt(4), false},
+		"case 6": {a, a[1], big.NewInt(2), false},
+		"case 7": {a, Allocation{}, big.NewInt(2), false},
+	}
+
+	for name, testcase := range testCases {
+		t.Run(name, func(t *testing.T) {
```
NiloCK, line 203: Good catch. I have some existing tests in this repo that spit out array indices in their error messages, which is a million times worse than this.

channel/state/outcome/outcome.go
```diff
@@ -55,6 +56,32 @@ func (a Allocations) Total() *big.Int {
 	return total
 }
 
+// Affords returns true if the allocations can afford the given allocation given input funds of x, false otherwise.
+//
+// To afford the given allocation, the allocations must include something equal-in-value to it,
+// as well as having sufficient funds left over for it after reserving funds from x for all allocations with higher priority.
+// Note that "equal-in-value" implies the same allocation type and metadata (if any).
+func (allocations Allocations) Affords(given Allocation, x *big.Int) bool {
```
NiloCK, line 64: Is there a better name for `x` here? `funding`, `holding`, ... ?

channel/state/outcome/outcome.go
```diff
@@ -55,6 +56,32 @@ func (a Allocations) Total() *big.Int {
...
+
+		if allocation.Equal(given) && affords.Cmp(allocation.Amount) >= 0 {
+			return true
+		}
+
+		surplus.Sub(surplus, affords)
+
+	}
+	return false
+}
```
NiloCK, line 83: - moves the most important logical check to the top of the loop
  - ... and returns a result directly, whether T or F
- a little shorter - removes accounting step
- adds comment on the `break`
- ... passes tests

```suggestion
func (allocations Allocations) Affords(given Allocation, x *big.Int) bool {
	bigZero := big.NewInt(0)
	surplus := big.NewInt(0).Set(x)
	for _, allocation := range allocations {

		if allocation.Equal(given) {
			return surplus.Cmp(given.Amount) >= 0
		}

		surplus.Sub(surplus, allocation.Amount)

		if surplus.Cmp(bigZero) != 1 {
			break // no funds remain for further allocations
		}

	}
	return false
}
```

NiloCK (approved):
:+1:

PR #87, opened 2021-11-30
Rename objectives, place each in their own package

### LINKED ISSUE #68: Rename objectives (removing the -ing suffix)
nit: It almost sounds like `directly-fund` is a better name for `Objective`. hard to imagine this is worth the refactor though?

_Originally posted by @andrewgordstewart in https://github.com/statechannels/go-nitro/pull/59#r752514847_

geoknee:
Closes #68 

* Each objective / package shall live in its own package. That means greater isolation between different objectives and their tests
* Each objective shall be named with a verb in the base form (not continuous form). E.g. "fund' instead of "funding".
* The utility functions extending the `big` package now live in the nitro `types` package. 
* "State" has been dropped from the name of each objective instance. This makes the objective instances simply "objectives" (which is shorter). "State" was a hangover from early prototyping, and IMO is now adequately implicit in the data contained in an objective instance.


protocols/direct-fund/direct-fund.go
```diff
@@ -1,34 +1,35 @@
-package protocols
+package directfund
 
 import (
 	"errors"
 	"fmt"
 	"math/big"
 
 	"github.com/statechannels/go-nitro/channel/state"
+	"github.com/statechannels/go-nitro/protocols"
```
NiloCK, line 9: This "suggestion" isn't really a suggestion - I don't think it's better than the current version - but I just wanted to point it out.

The dot import brings the imported package's identifiers into global namespace, so that, eg, `protocols.SideEffects{}` can just be `SideEffects{}`. These packages (`protocols` and its children) are about as good a case for the dot import as I can think of, but still it brings threat of namespace collisions which I'd rather not have to think about.

```suggestion
	. "github.com/statechannels/go-nitro/protocols"
```

protocols/direct-fund/direct-fund.go
```diff
@@ -47,17 +48,17 @@ type DirectFundingObjectiveState struct {
 	onChainHolding types.Funds
 }
 
-// NewDirectFundingObjectiveState initiates a DirectFundingInitialState with data calculated from
+// New initiates a DirectFundingInitialState with data calculated from
 // the supplied initialState and client address
-func NewDirectFundingObjectiveState(initialState state.State, myAddress types.Address) (DirectFundingObjectiveState, error) {
+func New(initialState state.State, myAddress types.Address) (DirectFundObjective, error) {
```
NiloCK, line 53: A lot better.

NiloCK (approved):
Looks good.

PR #90, opened 2021-12-02
rearrange outcome package

geoknee:
The `outcome.go` file in this package is now an unwieldily 315 lines. This PR splits it up into separate files as follows:

outcome.go renamed to exit.go
allocations and tests split to separate files
single asset exits and tests split to separate files
test data split to a separate file
DepositySafetyThreshold methods and tests split off into a seperate file.

Reminder: we don't need to cross-import anything when doing this kind of refactor, since all symbols from all source files in a package are visible in all of the other source files.


geoknee:
In favour of #92

PR #97, opened 2021-12-09
testState: use only 2 participants

geoknee:
with known private keys.

This simplifies tests, particularly useful when it comes to tackling #78 and #25

### DIFF
--- a/channel/state/state_test.go
+++ b/channel/state/state_test.go
@@ -8,14 +8,14 @@ import (
 )
 
 // The following constants are generated from our ts nitro-protocol package
-var correctChannelId = common.HexToHash(`cd578811171bb0291c7dc59081b9a910960b78001dece1954c00da9d8c12a925`)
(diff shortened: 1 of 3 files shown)


NiloCK:
:+1: 

Correct me if I'm wrong, but won't pieces of the VFP require at least three signers of certain states (Alice, Irene, Bob)?

geoknee:
You are absolutely right. I'm planning on having an entirely fresh set of test data for the virtual funding protocol, though.

NiloCK:
:heavy_check_mark: Was clarifying because the PR text specifically mentioned #25 (the VFP).

geoknee:
Yes that's fair I did mention #25. The full story is that the VFP work is driving the `Channel` abstraction, which I want to test with 2-party states (for simplicity), but will also handle 3-party states (which I am making new test data for).

geoknee:
... I think the change makes sense if you look at what the current tests are covering. That's the `state` class and the `direct-fund` protocol.

NiloCK (approved):

PR #98, opened 2021-12-09
introduce `Allocations.DivertToGuarantee()`

geoknee:
Towards #25

### DIFF
--- a/channel/state/outcome/allocation.go
+++ b/channel/state/outcome/allocation.go
@@ -123,9 +123,9 @@ func (a Allocations) DivertToGuarantee(
 		return Allocations{}, errors.New(`debtees must be distinct`)
 	}
 
-	newAllocations := make([]Allocation, len(a)+1)
+	newAllocations := make([]Allocation, 0, len(a)+1)
 	for i, allocation := range a {
-		newAllocations[i] = allocation
+		newAllocations = append(newAllocations, allocation)
 		switch newAllocations[i].Destination {
 		case leftDestination:
 			newAllocations[i].Amount.Sub(newAllocations[i].Amount, leftAmount)
@@ -145,12 +145,12 @@ func (a Allocations) DivertToGuarantee(
 		return Allocations{}, errors.New(`error encoding guarantee`)
 	}
 
-	newAllocations[len(a)] = Allocation{
+	newAllocations = append(newAllocations, Allocation{
 		Destination:    guaranteeDestination,
 		Amount:         big.NewInt(0).Add(leftAmount, rightAmount),
 		AllocationType: GuaranteeAllocationType,
 		Metadata:       encodedGuaranteeMetadata,
-	}
+	})
 
 	return newAllocations, nil
 }
--- a/channel/state/outcome/allocation.go
+++ b/channel/state/outcome/allocation.go
@@ -132,7 +132,7 @@ func (allocations Allocations) DivertToGuarantee(
 		case rightDestination:
 			newAllocations[i].Amount.Sub(newAllocations[i].Amount, rightAmount)
 		}
-		if newAllocations[i].Amount.Cmp(big.NewInt(0)) < 0 {
+		if types.Gt(big.NewInt(0), newAllocations[i].Amount) {
 			return Allocations{}, errors.New(`insufficient funds`)
 		}
 	}
--- a/channel/state/outcome/allocation.go
+++ b/channel/state/outcome/allocation.go
@@ -2,6 +2,7 @@ package outcome
 
 import (
 	"bytes"
+	"errors"
 	"math/big"
 
 	"github.com/ethereum/go-ethereum/accounts/abi"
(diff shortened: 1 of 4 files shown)


NiloCK:
```diff
diff --git a/channel/state/outcome/allocation.go b/channel/state/outcome/allocation.go
index 4349033..4ca0d03 100644
--- a/channel/state/outcome/allocation.go
+++ b/channel/state/outcome/allocation.go
@@ -123,7 +123,7 @@ func (allocations Allocations) DivertToGuarantee(
 		return Allocations{}, errors.New(`debtees must be distinct`)
 	}
 
-	newAllocations := make([]Allocation, len(allocations)+1)
+	newAllocations := make([]Allocation, 0, len(allocations)+1)
 	for i, allocation := range allocations {
 		newAllocations[i] = allocation
 		switch newAllocations[i].Destination {

```

NiloCK:
```diff
diff --git a/channel/state/outcome/allocation.go b/channel/state/outcome/allocation.go
index 4349033..4ca0d03 100644
--- a/channel/state/outcome/allocation.go
+++ b/channel/state/outcome/allocation.go
@@ -123,7 +123,7 @@ func (allocations Allocations) DivertToGuarantee(
 		return Allocations{}, errors.New(`debtees must be distinct`)
 	}
 
-	newAllocations := make([]Allocation, len(allocations)+1)
+	newAllocations := make([]Allocation, 0, len(allocations)+1)
 	for i, allocation := range allocations {
 		newAllocations[i] = allocation
 		switch newAllocations[i].Destination {

```

NiloCK:
... the above comments are the result of `suggestions` pushed from staging area of VSCode w/ the GH Pull Request extension, which I am test driving. Not as user-friendly as suggestions made via the web. Probably disregard.

channel/state/outcome/allocation.go
```diff
@@ -105,3 +106,51 @@ var allocationsTy = abi.ArgumentMarshaling{
...
+	if leftDestination == rightDestination {
+		return Allocations{}, errors.New(`debtees must be distinct`)
+	}
+
+	newAllocations := make([]Allocation, len(allocations)+1)
+	for i, allocation := range allocations {
+		newAllocations[i] = allocation
+		switch newAllocations[i].Destination {
+		case leftDestination:
+			newAllocations[i].Amount.Sub(newAllocations[i].Amount, leftAmount)
```
NiloCK, line 131: I take it the introduction of `newAllocations` is meant to prevent overwrites of the calling `allocations` value, but it seems like a mixed success. Placing some logs around `DivertToGuarantee` in the test:
```
allocation_test.go:89: a prior:  [{0x000000000000000000000000000000000000000000000000000000000000000a 243 0 []} {0x000000000000000000000000000000000000000000000000000000000000000b 309 0 []}]
allocation_test.go:93: a after:  [{0x000000000000000000000000000000000000000000000000000000000000000a 238 0 []} {0x000000000000000000000000000000000000000000000000000000000000000b 304 0 []}]
allocation_test.go:94: got:      [{0x000000000000000000000000000000000000000000000000000000000000000a 238 0 []} {0x000000000000000000000000000000000000000000000000000000000000000b 304 0 []} {0x0000000000000000000000000000000000000000000000000000000000000abc 10 1 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 1
```
The existing individual `Allocation`s are modified, but the slice does not include the new guarantee.

channel/state/outcome/allocation_test.go
```diff
@@ -63,3 +64,57 @@ func TestAffords(t *testing.T) {
...
 }
+
+func TestDivertToGuarantee(t *testing.T) {
+
+	aliceDestination := types.Destination(common.HexToHash("0x0a"))
+	bobDestination := types.Destination(common.HexToHash("0x0b"))
+
+	targetChannel := types.Destination(common.HexToHash("0xabc"))
+
+	a := Allocations{ // [{Alice: 2, Bob: 3}]
```
NiloCK, line 75:
```suggestion
	a := Allocations{
```

channel/state/outcome/allocation_test.go
```diff
@@ -63,3 +64,57 @@ func TestAffords(t *testing.T) {
...
+		},
+	}
+
+	got, err := a.DivertToGuarantee(aliceDestination, bobDestination, big.NewInt(5), big.NewInt(5), targetChannel)
+
+	if err != nil {
+		t.Error(err)
+	}
+
+	want := Allocations{ // [{Alice: 2, Bob: 3}]
```
NiloCK, line 96:
```suggestion
	want := Allocations{
```

NiloCK (approved):
:+1: 

Moving back to an earlier discussion: we had talked about these structures and how it's better to have the general struct with `left` and `right`, where alice and bob would be special cases with `left == nil` and `right == nil` respectively.

This is incorrect, yes?

Each ledger channel (LC) has a left and right. What alice doesn't have is a LC where she occupies the `right` slot, and similar for bob on the `left`. Intermediaries in a VC construction will interact with two channels - one where they are in each role.

Potentially nasty word-overloading if we want to use L, R to refer both to the roles inside of a LC update and also to refer to "the intermediary on my left OR alice OR nil (if I am alice)" and "the intermediary to my right OR bob OR nil".

channel/state/outcome/allocation.go
```diff
@@ -105,3 +106,51 @@ var allocationsTy = abi.ArgumentMarshaling{
...
+	leftAmount *big.Int,
+	rightAmount *big.Int,
+	guaranteeDestination types.Destination,
+) (Allocations, error) {
+
+	if leftDestination == rightDestination {
+		return Allocations{}, errors.New(`debtees must be distinct`)
+	}
+
+	newAllocations := make([]Allocation, len(allocations)+1)
```
NiloCK, line 126:
```suggestion
	newAllocations := make([]Allocation, len(allocations), len(allocations)+1)
```
Doesn't place a blank `Allocation{}` in the last slot - a requirement of the suggestion on line 148.

I'm not sure whether
```
newAllocations := make([]Allocation, 0, len(allocations)+1)
// followed by
newAllocations = append(newAllocations, allocation) // inside the loop, instead of `newA[i] = allocations`
```
is better still.

channel/state/outcome/allocation.go
```diff
@@ -105,3 +106,51 @@ var allocationsTy = abi.ArgumentMarshaling{
 		{Name: "metadata", Type: "bytes"},
 	},
 }
+
+// DivertToGuarantee returns a new Allocations, identical to the reciever but with
+// the leftDebtee's amount reduced by leftDebit,
+// the rightDebtee's amount reduced by rightDebit,
+// and a Guarantee appended for the guaranteeDestination
+func (allocations Allocations) DivertToGuarantee(
```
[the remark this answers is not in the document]
geoknee: Would `as Allocations` be a decent compromise?

channel/state/outcome/allocation.go
geoknee: Yes I went with you "better still" suggestion. Nice use of a zero length, nonzero capacity slice.

channel/state/outcome/allocation.go
geoknee: Good catch. I added a test to catch this (which fails), and also introduces and uses `Allocation.Clone()` to make it pass.

geoknee:
About the word overloading -- I agree there is potential for nastiness and we should think about ways to avoid it. Perhaps when we get the first iteration of the virtual funding protocol merged, we can revisit the issue.

PR #100, opened 2021-12-10
add `Exit.Affords()` method

geoknee:
We already have an `Affords` method which applies to a single asset. This PR extends that to cope with a full `Exit` (i.e. many assets). 

It makes use of an `easyExit` type, also introduced here. For now this is just a helper method, but it is a hint about how we could use more ergonomic types in our Go code (and have adapters ready for when we need to interface with say some solidity code). 

Towards #25.

### DIFF
--- a/channel/state/outcome/exit.go
+++ b/channel/state/outcome/exit.go
@@ -127,7 +127,7 @@ func (e *Exit) Hash() (types.Bytes32, error) {
 	}
 }
 
-type EasyExit map[common.Address]SingleAssetExit
+type easyExit map[common.Address]SingleAssetExit
 
 // toEasyExit() convets an Exit into a more ergonomic data type called an EasyExit
 //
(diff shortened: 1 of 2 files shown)


channel/state/outcome/exit.go
```diff
@@ -126,3 +126,42 @@ func (e *Exit) Hash() (types.Bytes32, error) {
...
+	easy := make(easyExit)
+	for i := range e {
+		easy[e[i].Asset] = e[i]
+	}
+	return easy
+}
+
+// Affords returns true if every allocation in the allocationMap can be afforded by the Exit, given the funds
+//
+// Both arguments are maps keyed by the same asset
```
NiloCK, line 150:
```suggestion
// Both arguments are maps keyed by the same assets
```

channel/state/outcome/exit.go
```diff
@@ -126,3 +126,42 @@ func (e *Exit) Hash() (types.Bytes32, error) {
...
+func (e Exit) Affords(
+	allocationMap map[common.Address]Allocation,
+	funds types.Funds) bool {
+	easyExit := e.toEasyExit()
+	for asset := range allocationMap {
+		x := funds[asset]
```
NiloCK, line 156:
```suggestion
	funding types.Funds) bool {
	easyExit := e.toEasyExit()
	for asset := range allocationMap {
		x := funding[asset]
```

... is consistent with the labelling on `Allocations.Affords`

NiloCK (commented):
:+1:

NiloCK (approved):

PR #103, opened 2021-12-10
type send and receive chans as write, read

NiloCK:
... prevents, at compile time, attemps to write to a recieve chan, or to read from a send chan.


geoknee:
IIRC this kind of extra type safety (which I am strongly in favour of) may only be possible for function arguments?

NiloCK:
Update: it is possible to type these this way, but probably naive to do so as described in this PR. Will revisit.

PR #104, opened 2021-12-13
Overhaul `State.Clone()` and code it depends on

### LINKED ISSUE #96: `State.Clone()` method is not a deep clone
https://github.com/statechannels/go-nitro/blob/864538aa793e7a7bd6761b7be36fa295ceade2ce/channel/state/state.go#L170-L189

1. We do not have a unit test for `State.Clone()`
2. I believe the method is inadequate for our needs. For example, say I want to create a clone of an existing state and modify the `ChannelNonce` property. Since this property is a pointer to a `big.Int` , the cloned state points to the same `big.Int` meaning that modifying the clone modifies the original.

geoknee:
-- adds important unit tests
-- corrects cloning code where necessary to get tests to pass

Closes #96

### DIFF
--- a/channel/state/state.go
+++ b/channel/state/state.go
@@ -199,17 +199,17 @@ func (s State) Equal(r State) bool {
 func (s State) Clone() State {
 	clone := State{}
 
-	// shallow clone of fixed part
-	clone.ChainId = s.ChainId
+	// Fixed part
+	clone.ChainId = new(big.Int).Set(s.ChainId)
 	clone.Participants = s.Participants
-	clone.ChannelNonce = s.ChannelNonce
+	clone.ChannelNonce = new(big.Int).Set(s.ChannelNonce)
 	clone.AppDefinition = s.AppDefinition
-	clone.ChallengeDuration = s.ChallengeDuration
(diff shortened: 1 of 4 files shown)


channel/state/state.go
```diff
@@ -168,21 +169,47 @@ func (s State) RecoverSigner(sig Signature) (types.Address, error) {
...
+
 // Clone returns a clone of the state
 func (s State) Clone() State {
 	clone := State{}
 
-	// shallow clone of fixed part
-	clone.ChainId = s.ChainId
+	// Fixed part
+	clone.ChainId = new(big.Int).Set(s.ChainId)
 	clone.Participants = s.Participants
```
NiloCK, line 204:
```suggestion
	clone.Participants = append(clone.Participants, s.Participants...)
```
See: https://go.dev/play/p/twgDh-4MD3s

Finally it has clicked for me that fixed-length arrays are comparable (`==`) and safely copyable (`safeCopy := origininal`) but slices are not. This jarred with my notion of slices being the universally friendlier container.

NiloCK (approved):
:+1:

PR #107, opened 2021-12-13
add `VirtualFundObjective` and unit tests

### LINKED ISSUE #25: Write a virtual funding protocol

geoknee:
This PR adds a new type to our `protocols` package, that implements the `Objective` interface. 

In the same way that the existing `DirectFundObjective` implements the logic for running the "direct fund" protocol, the new type implements the logic for running the "virtual fund" protocol. It is a collection of finite states, infinite state, and methods that can be `Updated` and `Cranked` in order to generate side effects that drive progress towards the goal of virtually funding a channel between Alice and Bob, given a pair of existing ledger channels connecting each to a common intermediary. 

The finite states are essentially the same as in the `direct-fund` protocol.  `Update` and `Crank` remain *pure* functions.

The way that funding is deemed to have been completed or not, is different. When the objective is spawned (the constructor `New()` is called), some "expected guarantees" are generated and stored. When an expected guarantee(s) is appropriately included in its respective ledger channels, the implementation will consider that funding has been completed.

The SATP paper draft has been my reference when writing this implementation.

As with the `direct-fund` protocol, side effects are not yet unit tested and should be considered only partially implemented. This is a conscious choice, and we will revisit the issue when the chain and message services are more mature.

One difference in implementation is that this objective makes use of the `Channel` type in the `channel` package. That's an abstraction we plan to back-port to the `direct-fund` protocol #78.

The unit tests follow a pattern where we use sub tests to cordon off test data in nested scopes. See [this notion doc](https://www.notion.so/statechannels/Thoughts-on-test-data-in-Go-00dd2eecb715419f901ab733891deb55#ad7b6933a900420090da192a5ad87867) for more information.  For now I have concentrated on single hop tests, although the protocol has been written to handle an arbitrary number of hops. I have also written a tests "as alice", and if the test is deemed good we can replicate a test "as bob" and "as Irene".

Closes #25.

### DIFF
--- a/protocols/virtual-fund/virtual-fund.go
+++ b/protocols/virtual-fund/virtual-fund.go
@@ -85,9 +85,9 @@ func New(
 	// Initialize virtual channel
 	v, err := channel.New(initialStateOfV, false, myRole, types.Destination{}, types.Destination{})
 	if err != nil {
-		return init, err
-
+		return VirtualFundObjective{}, err
 	}
+
 	init.V = &v
 	init.n = n
 	init.MyRole = myRole
--- a/protocols/virtual-fund/virtual-fund.go
+++ b/protocols/virtual-fund/virtual-fund.go
@@ -1,0 +1,364 @@
+package virtualfund
+
+import (
+	"errors"
+	"math/big"
+
+	"github.com/statechannels/go-nitro/channel"
+	"github.com/statechannels/go-nitro/channel/state"
+	"github.com/statechannels/go-nitro/channel/state/outcome"
+	"github.com/statechannels/go-nitro/protocols"
+	"github.com/statechannels/go-nitro/types"
+)
+
+const (
+	WaitingForCompletePrefund  protocols.WaitingFor = "WaitingForCompletePrefund"  // Round 1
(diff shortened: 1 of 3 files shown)


andrewgordstewart:
Is there something like:
- documentation
- a design document
- or a PR message

that helps a reviewer understand the context of this PR?

geoknee:
I've updated the PR description.

protocols/virtual-fund/virtual-fund.go
```diff
@@ -0,0 +1,369 @@
...
+	}
+
+	var init VirtualFundObjective
+
+	// Initialize virtual channel
+	v, err := channel.New(initialStateOfV, false, myRole, types.Destination{}, types.Destination{})
+	if err != nil {
+		return init, err
+
+	}
```
NiloCK, line 90: Explicitly empty returned value (matching other err returns)

```suggestion
		return VirtualFundObjective{}, err
	}

```

NiloCK (approved):
:tada:

PR #109, opened 2021-12-14
Test message service

NiloCK:
`TestMessageService` is an implementation usable for unit and integration tests for engines. This implementation amounts to a perfect communication medium between A and B. Moving forward, the `Connect` function can receive optional middleware functions which do things like adding delays, swapping order of messages, dropping messages, etc.

### DIFF
--- a/client/engine/messageservice/test-messageservice.go
+++ b/client/engine/messageservice/test-messageservice.go
@@ -6,31 +6,27 @@ for use in
 package messageservice
 
 import (
-	"github.com/ethereum/go-ethereum/common"
+	"fmt"
+
 	"github.com/statechannels/go-nitro/protocols"
 	"github.com/statechannels/go-nitro/types"
 )
 
-type TestNetwork struct {
-	addressToPeer map[[]byte]types.Address
-}
-
-func (tn TestNetwork) Connect (alice, bob TestMessageService) {
-	ch := make(chan protocols.Message)
-
-}
-
 type TestMessageService struct {
+	address types.Address
+
 	// a map of gochans to pass messages to each speific peer
-	toPeers map[types.Address]chan protocols.Message
+	toPeers map[types.Address]chan<- protocols.Message
 	out     chan protocols.Message
 
-	// a collection of gochans to recieve messages from specific peers
-	fromPeers []chan protocols.Message
-	in      chan protocols.Message
+	in chan protocols.Message
 }
 
-func (t TestMessageService) GetReceiveChan() <-chan protocols.Message {
+func (t TestMessageService) Run() {
+	go t.routeOutgoing()
+}
+
+func (t TestMessageService) GetReceiveChan() chan protocols.Message {
 	return t.in
 }
 
@@ -39,30 +35,35 @@ func (t TestMessageService) GetSendChan() chan<- protocols.Message {
 }
 
 func (t TestMessageService) Send(message protocols.Message) {
-	t.out <- message;
+	t.out <- message
+}
 
-	
+// Connect creates a gochan for message service t to communicate with
+// the given peer.
+func (t TestMessageService) Connect(peer TestMessageService) {
+	toPeer := make(chan protocols.Message)
+
+	t.toPeers[peer.address] = toPeer
+
+	go func() {
+		for msg := range toPeer {
+			peer.in <- msg
+		}
+	}()
 }
 
 func (t TestMessageService) forward(message protocols.Message) {
 	peerChan, ok := t.toPeers[message.To]
-	if !ok {
-		t.toPeers[message.To] = make(chan<- protocols.Message)
-		t.Send(message)
+	if ok {
+		peerChan <- message
+	} else {
+		panic(fmt.Sprintf("client %v has no connection to client %v",
+			t.address, message.To))
 	}
-	peerChan <- message
-
-	addr := common.HexToAddress(string(message.To))
-	t.toPeers[addr] <- message
 }
 
-func (t TestMessageService) Run() {
-	for {
-		select {
-		case outgoing := <-t.out:
-			t.forward(outgoing)
-		case incoming :=
-		}
-
+func (t TestMessageService) routeOutgoing() {
+	for msg := range t.out {
+		t.forward(msg)
 	}
 }
--- a/client/engine/messageservice/test-messageservice.go
+++ b/client/engine/messageservice/test-messageservice.go
@@ -1,0 +1,68 @@
+/*
+TestMessageService is an implementaion of the MessageService interface
+for use in
+*/
+
+package messageservice
+
+import (
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/statechannels/go-nitro/protocols"
+	"github.com/statechannels/go-nitro/types"
+)
+
+type TestNetwork struct {
+	addressToPeer map[[]byte]types.Address
+}
+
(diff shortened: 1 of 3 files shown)


protocols/interfaces.go
```diff
@@ -7,7 +7,7 @@ import (
 
 // Message is an object to be sent across the wire. It can contain a proposal and signed state hashes, and is addressed to a counterparty.
 type Message struct {
-	To          []byte
+	To          types.Address
```
andrewgordstewart, line 10: Clarifying question: is the assumption here that a message service should determine **how and where to send a message** based on the _blockchain address_ of the recipient?

andrewgordstewart (commented):
My immediate expectation is that this PR would
1. define an interface like [the one from `server-wallet`](https://github.com/statechannels/statechannels/blob/55ab6fd28ab1bade93a7d9648b25a8b63bd9dd55/packages/server-wallet/src/message-service/types.ts#L14-L18).
2. implement a `TestMessageService` interface that satisfies the interface defined in (1).

It seems like this PR instead aims to provide a TestMessageService implementation as a means of discovering the correct interface. Is that correct?

client/engine/messageservice/test-messageservice.go
```diff
@@ -0,0 +1,75 @@
...
+func (t TestMessageService) Connect(peer TestMessageService) {
+	toPeer := make(chan protocols.Message)
+
+	t.toPeers[peer.address] = toPeer
+
+	go func() {
+		for msg := range toPeer {
+			peer.in <- msg
+		}
+	}()
```
geoknee, line 54: I'm curious about this bit. If `toPeer` was just made, won't it have 0 messages in it? Meaning this code is pointless?

geoknee: I looked into this some more and answered my own question. Recording my learnings here for posterity:

>  the `range` loop receives values from the channel repeatedly until it is closed.

Since we aren't closing the channel, this is essentially an infinite loop taking values from the channel (which is what I expected). 

Source https://go.dev/tour/concurrency/4

client/engine/messageservice/test-messageservice.go
```diff
@@ -0,0 +1,75 @@
...
+func (t TestMessageService) GetSendChan() chan<- protocols.Message {
+	return t.out
+}
+
+func (t TestMessageService) Send(message protocols.Message) {
+	t.out <- message
+}
+
+// Connect creates a gochan for message service t to communicate with
+// the given peer. This connection is one-way.
```
geoknee, line 44: Nit:
```suggestion
// Connect creates a gochan for message service t to send messages to the given peer.
```

"send" is better than "communicate one way" (IMO).

geoknee: Meta comment: the receiver variable `t` doesn't make a ton of sense when a dev hovers over the method on a test message service instance they are working with (they may not have named it `t`). So does `t` belong in the documentation here?

client/engine/messageservice/test-messageservice_test.go
```diff
@@ -0,0 +1,52 @@
...
+	Sigs:        make(map[*state.State]state.Signature),
+	Proposal:    objective,
+}
+
+func TestConnect(t *testing.T) {
+	aliceMS.Run()
+	bobMS.Run()
+	bobIn := bobMS.GetReceiveChan()
+
+	aliceMS.Connect(bobMS)
```
geoknee, line 43: In a fully connected test messaging network with N peers, I think we would need to call `Connect` N(N-1) times, creating N**2 - N `chans`. This is probably fine because `chans` are cheap. But I was wondering if you had considered having a more centralised solution -- i.e. each peer has a send and receive to a single message bus (2N `chans` in total). I think this is called pub/sub https://en.wikipedia.org/wiki/Publish%E2%80%93subscribe_pattern.

geoknee (approved):
I think this is a great start. Looking forward to seeing this power some test that integrate the `engine`. There are some outstanding questions but I would be in favour of merging this now. 

FInal Q: did you consider whether we need to `close` the go `chans` at any point? I.e. a `Disconnect` method? (I guess probably not).

client/engine/messageservice/test-messageservice.go
NiloCK: These constructs (range over a channel, select statements) are really pleasant!

client/engine/messageservice/test-messageservice.go
```diff
@@ -0,0 +1,75 @@
...
+
+import (
+	"fmt"
+
+	"github.com/statechannels/go-nitro/protocols"
+	"github.com/statechannels/go-nitro/types"
+)
+
+// TestMessageService is an implementaion of the MessageService interface
+// for use in with-peers style test environments.
```
NiloCK, line 11:
```suggestion
// for use in multi-engine test environments.
```

client/engine/messageservice/test-messageservice.go
NiloCK, line 44:
```suggestion
// Connect creates a gochan for message service to send messages to the given peer.
```

client/engine/messageservice/test-messageservice_test.go
NiloCK: Idea being that per-peer connection more closely resembles real world networking system. We expect agents in a state channel network communicate p2p, so why not mirror that here?

client/engine/messageservice/test-messageservice.go
```diff
@@ -0,0 +1,75 @@
...
+
+// forward finds the appropriate gochan for the message recipient,
+// and sends the message. It panics if no such channel exists.
+func (t TestMessageService) forward(message protocols.Message) {
+	peerChan, ok := t.toPeers[message.To]
+	if ok {
+		peerChan <- message
+	} else {
+		panic(fmt.Sprintf("client %v has no connection to client %v",
+			t.address, message.To))
```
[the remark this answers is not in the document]
geoknee: OK, perhaps we can make an issue to circle back and change it?

protocols/interfaces.go
NiloCK: Maybe - this isn't __necessarily__ intended as a final detail, but I expect:
- everyone that a client will be interested in messaging will own an address. It's probably suitable as an identifier.
- the message service can internally manage a lookup that maps from `types.Address` to transport layer data, possibly something like https://github.com/multiformats/go-multiaddr (used in Lotus / go-fil-markets)

NiloCK:
@andrewgordstewart 

> My immediate expectation is that this PR would
> 
>     1. define an interface like [the one from `server-wallet`](https://github.com/statechannels/statechannels/blob/55ab6fd28ab1bade93a7d9648b25a8b63bd9dd55/packages/server-wallet/src/message-service/types.ts#L14-L18).
>     2. implement a `TestMessageService` interface that satisfies the interface defined in (1).
> 
> 
> It seems like this PR instead aims to provide a TestMessageService implementation as a means of discovering the correct interface. Is that correct?

There is in fact an existing interface: https://github.com/statechannels/go-nitro/blob/main/client/engine/messageservice/messageservice.go. Noteworthy is that go doesn't "do" explicit interface declarations - we cannot write anything like `struct TestMsgService is MessageService { ... }`, so it's not obvious that this or that implementation targets any specific interface.

But your understanding remains correctish - writing this is meant to challenge the assumptions in the existing interface.

PR #110, opened 2021-12-17
sum funds before calculating expected guarantee

NiloCK:
prevents possible problem where `a0` and `b0` do not have identical
asset lists

IE: previous implementation would fail for initial balances `a0 = {} // no balance`, `b0 = {fil: 5}`, because it applied only the assets indexed in `a0`.

### DIFF
--- a/protocols/virtual-fund/virtual-fund.go
+++ b/protocols/virtual-fund/virtual-fund.go
@@ -262,10 +262,12 @@ func (connection *Connection) insertExpectedGuarantees(a0 types.Funds, b0 types.
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


kerzhner (approved):

kerzhner:
Are there any tests that would show this code works as expected?

NiloCK:
> Are there any tests that would show this code works as expected?

No tests exist against the spotted failure case that this PR fixes, but:
- the `Add` function introduced here is well tested
- an intentionally incorrect implementation (`channelFunds = a0.Add(b0).Add(a0)`) failed the existing tests

PR #111, opened 2021-12-17
Improve __checks__ on `VirtualFund.New`

NiloCK:
This PR fixes some of the __checks__ on creating a new VirtualFund objective. It:

- fixes a broken check
- makes the ledger-channel checks dependent on the role of the creator (alice, bob, intermediary)
- adds a check to top-of-function on a valid role declaration, which removes need for later duplicating this check

Read per-commit.

### DIFF
--- a/protocols/virtual-fund/virtual-fund.go
+++ b/protocols/virtual-fund/virtual-fund.go
@@ -59,11 +59,22 @@ func New(
 	ledgerChannelToMyRight channel.Channel,
 ) (VirtualFundObjective, error) {
 
-	var init VirtualFundObjective
-
-	if !(ledgerChannelToMyLeft.IsTwoPartyLedger && ledgerChannelToMyRight.IsTwoPartyLedger) {
-		return init, errors.New(`supplied channels are not two party ledger channels`)
+	switch myRole {
+	case 0: // Alice
+		if !ledgerChannelToMyRight.IsTwoPartyLedger {
+			return VirtualFundObjective{}, errors.New(`alice's right-channel is not a two-party ledger channel`)
+		}
+	case n + 1: // Bob
+		if !ledgerChannelToMyLeft.IsTwoPartyLedger {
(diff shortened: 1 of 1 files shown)


protocols/virtual-fund/virtual-fund.go
```diff
@@ -59,12 +59,23 @@ func New(
...
-	var init VirtualFundObjective
-
-	if !(ledgerChannelToMyLeft.IsTwoPartyLedger && ledgerChannelToMyRight.IsTwoPartyLedger) {
-		return init, errors.New(`supplied channels are not two party ledger channels`)
+	switch myRole {
+	case 0: // Alice
+		if !ledgerChannelToMyRight.IsTwoPartyLedger {
+			return VirtualFundObjective{}, errors.New(`alice's right-channel is not a two-party ledger channel`)
+		}
+	case n + 1: // Bob
```
kerzhner, line 72: Is there an off-by-one error here? Should this be `n`?

kerzhner (approved):

protocols/virtual-fund/virtual-fund.go
NiloCK: An excellent and prudent question, but: nope - this is per the updated SATP paper. `n` here refers to n intermediaries - 0 = Alice, 1, ... , `n` the intermediaries, `n+1` = Bob

protocols/virtual-fund/virtual-fund.go
```diff
@@ -61,7 +61,7 @@ func New(
 
 	var init VirtualFundObjective
 
-	if ledgerChannelToMyLeft.IsTwoPartyLedger && ledgerChannelToMyRight.IsTwoPartyLedger {
```
geoknee, line 64: Was this bug previously slipping though our unit tests? Should we make any changes to our unit tests?

geoknee:
I think these changes introduced a test "failure" -- I've raised #113 so we avoid that in future.

PR #112, opened 2021-12-17
simplify cases & de-duplicate channel setup

NiloCK:
NB: previous `default` case test not necessary given `vf-patch-2`


NiloCK:
Reading the original and then reading the updated version is probably easier than trying to read the diff.

kerzhner (approved):

PR #116, opened 2022-01-05
Return boolean from `Channel.AddSignedStates()`

geoknee:
`AddSignedStates` currently calls `AddSignedState` in a loop and discards the `ok` boolean return value. The documentation for the method doesn't give any clue about this behaviour, nor whether the loop will continue if there is an error returned in the one step through the loop. 

This PR introduces a boolean return value which is the union of the return values from each step through the loop. 

It means that less information is discarded (but still most of the information is discarded). In principle it allows calling code to  better handle state insertion errors (although this is not currently something we make use of).

### DIFF
--- a/channel/channel.go
+++ b/channel/channel.go
@@ -191,11 +191,14 @@ func (c *Channel) AddSignedState(s state.State, sig state.Signature) bool {
 	return true
 }
 
-// AddSignedStates adds each signed state in the mapping
-func (c Channel) AddSignedStates(mapping map[*state.State]state.Signature) {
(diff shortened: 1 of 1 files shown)


NiloCK (approved):
This is an improvement (so :+1: ), but given that it still discards most information is it currently in a useful state for calling code?

geoknee:
I think it is in a useful state, but perhaps not the best state it could be in. I think probably not worth worrying about unless or until we reach a point where we need more nuanced error handling (currently we just ignore things that don't work).

geoknee:
WIth hindsight, this may not have been an improvement :-P. It introduced a bug, fixed in #127 .

PR #119, opened 2022-01-05
`Funds` type: add `Clone()` method

geoknee:
Because we wish to have functional methods at the core of our software, there is a need to generate deep clones of certain objects. Such objects frequently embed a `Funds` object. By having a `Clone` method on the child, we can adopt a modular approach to cloning the parent.


NiloCK (approved):
Good eye for code re-use. :+1:

PR #120, opened 2022-01-05
place `testState` in `direct-fund` package

geoknee:
We need to use a different turn number and we do not wish to disturb other tests importing from state.testState.


NiloCK (approved):
this is:
- a duplication of data from `state/test-fixtures.go`, (other than then turnNum)
but:
- we don't care much about these test states staying "mostly equal". It's not a problem if they fall out of sync
- it's "only local test data"

LGTM

PR #125, opened 2022-01-06
Refactor myRole

andrewgordstewart:
This was mostly done as an exercise, as it's my first contribution to the project. There are two changes to the use of `myRole`:
1. Bob is now role 1
2. An enum is used for specific roles instead of magic numbers.

(See the commit messages for reasons.)

Note that due to (1) and some sloppy changes, the roles are now incorrectly set in the `single-hop` test file, but the tests still pass. For this reason, this PR is still marked as a draft.

### DIFF
--- a/protocols/virtual-fund/virtual-fund.go
+++ b/protocols/virtual-fund/virtual-fund.go
@@ -70,7 +70,7 @@ func New(
 		if !ledgerChannelToMyRight.IsTwoPartyLedger {
 			return VirtualFundObjective{}, errors.New(`alice's right-channel is not a two-party ledger channel`)
 		}
-	case n + 1: // Bob
+	case 1: // Bob
 		if !ledgerChannelToMyLeft.IsTwoPartyLedger {
 			return VirtualFundObjective{}, errors.New(`bob's left-channel is not a two-party ledger channel`)
 		}
@@ -119,7 +119,7 @@ func New(
 		}
 	}
 
-	if myRole != n+1 { // everyone other than Bob has a right-channel
+	if myRole != 1 { // everyone other than Bob has a right-channel
 		init.ToMyRight = &Connection{}
 		init.ToMyRight.Channel = ledgerChannelToMyRight
 		err = init.ToMyRight.insertExpectedGuarantees(init.a0, init.b0, init.V.Id, init.ToMyRight.Channel.MyDestination, init.ToMyRight.Channel.TheirDestination)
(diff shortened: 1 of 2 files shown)


protocols/virtual-fund/virtual-fund.go
```diff
@@ -285,16 +297,14 @@ func (s VirtualFundObjective) fundingComplete() bool {
...
-	case s.MyRole == 0: // Alice
+	case s.MyRole == Alice:
 		return s.ToMyRight.ledgerChannelAffordsExpectedGuarantees()
-	case s.MyRole < n+1: // Intermediary
-		return s.ToMyRight.ledgerChannelAffordsExpectedGuarantees() && s.ToMyLeft.ledgerChannelAffordsExpectedGuarantees()
-	case s.MyRole == n+1: // Bob
+	case s.MyRole == Bob:
 		return s.ToMyLeft.ledgerChannelAffordsExpectedGuarantees()
-	default: // Invalid
+	case s.MyRole > I1:
```
NiloCK, line 305:
```suggestion
	case s.MyRole >= I1:
```
Or `> Bob:`, but probably the above is better.

NiloCK:
Devil's advocate: are `I1`, ..., `I5` less magic-numbery than the previous implementation? Did their magic numberiness contribute to the highlighted implementation bug?

Similar ambiguities remain: if a programmer now sees a role of 2, they do not know whether that intermediary has a ledger channel with Bob without checking and considering `n`, (the same work to be done previously). It's probably easier to reason incorrectly about relationships between myRole and `n` in the current implementation, since participants are no longer in order.

That said: I really enjoy the readability of constants `Alice` and `Bob` here. I might go as far as to suggest replacing `I(n)` with `Irene`, `Ivan`, ...

protocols/virtual-fund/virtual-fund.go
andrewgordstewart: Is this also indicating a gap in our test suite?

geoknee:
@andrewgordstewart what shall we do with this draft PR? I would like to either close it or get it ready for a final review and merge. 

If the latter, here are my 2 pence: it would be good to update the _specification_ for this protocol prior to (or at the same time as) updating this implementation. For now the specification = the SATP paper.

andrewgordstewart:
I don't currently have the capacity to make progress on this.

> it would be good to update the specification for this protocol prior to (or at the same time as) updating this implementation. For now the specification = the SATP paper.

Great point. I don't recall the SATP paper specifying the order of participants in the joint/virtual channel, but if it does, we should update it in tandem.

If we are considering the SATP paper as a spec that should be kept in sync with go-nitro, let's look for a process to help us keep them in sync.

geoknee:
> Great point. I don't recall the SATP paper specifying the order of participants in the joint/virtual channel, but if it does, we should update it in tandem.
Screenshot 2022-01-28 at 14 34 58

geoknee:
As per discussion in slack:
* we think the use of `enums` or `constants` is a good idea and _somewhat_ independent of the reordering of the roles (not entirely independent, mind)
* we feel it may be too early to tell which ordering makes the most sense or whether a slightly "better' ordering would be worth the effort right now
* we will revisit this if and when it crops up again

PR #126, opened 2022-01-06
rename `AdddressToDestination` to `AddressToDestination`

geoknee:


NiloCK (approved):
I am impressed at how effectively my brain filtered out this spelling mistake. Had to say the letters out loud before I could find it.

PR #127, opened 2022-01-06
unconditionally add a state

geoknee:
Previously, the second condition (which is a function call) would not be evaluated in the case where the first one was false. This meant the function was not behaving as documented, and failing to add a "good" state when accompanied by a "bad" state.

### DIFF
--- a/channel/channel.go
+++ b/channel/channel.go
@@ -196,7 +196,7 @@ func (c *Channel) AddSignedState(s state.State, sig state.Signature) bool {
 func (c Channel) AddSignedStates(mapping map[*state.State]state.Signature) bool {
 	allOk := true
 	for state, sig := range mapping {
-		allOk = allOk && c.AddSignedState(*state, sig)
+		allOk = c.AddSignedState(*state, sig) && allOk
 	}
 	return allOk
 }


channel/channel.go
```diff
@@ -196,7 +196,7 @@ func (c *Channel) AddSignedState(s state.State, sig state.Signature) bool {
 func (c Channel) AddSignedStates(mapping map[*state.State]state.Signature) bool {
 	allOk := true
 	for state, sig := range mapping {
-		allOk = allOk && c.AddSignedState(*state, sig)
+		allOk = c.AddSignedState(*state, sig) && allOk
```
NiloCK, line 199:
```suggestion
		ok := c.AddSignedState(*state, sig)
		if !ok {
			allOk = false
		}
```
I had written this on the prior PR but discarded it after convincing myself that the existing implementation was correct and sufficiently easy to understand.

I am again convinced that the proposed implementation is correct and sufficiently easy to understand, but I am applying a new personal rule: prefer not to call functions inside larger expressions.

NiloCK (changes requested):
A good catch.

channel/channel.go
geoknee: Yep. I think naming the return value makes debugging easier, too.

NiloCK (approved):
:+1:

PR #131, opened 2022-01-07
apply type-assertion once

NiloCK:
Currently, our tests operate on a generic `Objective`-typed value. At each point that we want to make a test assertion, we do a type assertion so that we can access the relevant property.

This refactors the test to narrow the type once, rather than at each test assertion. If there are no objections, I can spread it through the other various Objective tests.

### DIFF
--- a/protocols/virtual-fund/virtual-fund-single-hop_test.go
+++ b/protocols/virtual-fund/virtual-fund-single-hop_test.go
@@ -224,8 +224,7 @@ func TestSingleHopVirtualFund(t *testing.T) {
 			}
 
 			// Approve the objective, so that the rest of the test cases can run.
-			o := s.Approve()
-
+			o := s.Approve().(VirtualFundObjective)
 			// To test the finite state progression, we are going to progressively mutate o
 			// And then crank it to see which "pause point" (WaitingFor) we end up at.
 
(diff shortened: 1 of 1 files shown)


geoknee:
This is nice, especially when you consider that the assertion actually makes a copy.

andrewgordstewart:
@NiloCK would you mind expanding on what this change does? I can't quite understand it.

NiloCK:
Our `Objective` interface is intended to be handled generically by the engine. The engine:
- hears events
- loads relevant objective(s)
- `Update()`s them with event data
- `Crank()`s them
all without knowing which implementation of `Objective` it is dealing with.

Because `Update` and `Crank` must return the parent type `Objective`, test functions which probe these returned values must assert the specific implementations. Previously, we had been doing this in-line more or less each time we needed to "look inside" the objective. This PR reduces the number these assertions by creating new variables which have the "correct" asserted type.

It is a readability / ergonomics improvement. Compare, eg, lines ~178 in `direct-fund_test.go`.

andrewgordstewart:
Here's another alternative: https://github.com/statechannels/go-nitro/compare/ags/type-assertion?expand=1

(taken from https://stackoverflow.com/a/11403616)

NiloCK:
That nicely cleans out the `err` processing step, but applies the same assert-once pattern as above. Also, would require either:

1. passing `t` into `want` in order to error from in there, or
2. beginning (each?) test with a `defer func() { /* catch panic and error w/ t */ }`

(1) makes `want` marginally uglier and so less good at its aim of improving readability of the code. (1) and (2) prevent us from attaching relevant information to the emitted error / test failure, unless we pass _that_ in as well, which further reduces the tidiness of `want`.

This came up in another context recently, but I am on the verge of articulating a tangential **opinion**. Here goes: _maybe it is OK and even good to panic at non-target `error`s in test code_, as long as whatever unexpected panics that occur in the wild end up with test coverage.

andrewgordstewart (approved):

PR #133, opened 2022-01-10
Don't encode the `Outcome` separately when encoding the `State`

geoknee:
Currently we expect a solidity contract to define a struct
```
struct state {
  // ...
  outcome bytes
}
```
This means that to verify signatures of a `state`, we first need to encode its outcome as bytes, plug it into a state struct, and then encode that state as bytes.

The change in this PR allows a solidity contract to define a state which has a property `outcome` that is itself a struct.
```
struct state {
  // ...
  outcome Outcome
}
```
As such, to encode a `state`, we would no longer need to encode `outcome` first.

This is a more elegant and efficient approach which should save some gas: see https://github.com/statechannels/go-nitro/pull/133#issuecomment-1009128748.

### DIFF
--- a/channel/state/state.go
+++ b/channel/state/state.go
@@ -63,7 +63,7 @@ var boolTy, _ = abi.NewType("bool", "bool", nil)
 // TODO: Is this type correct? Shouldn't it be bytes32?
 
 // address type for abi encoding
-var destination, _ = abi.NewType("address", "address", nil)
+var destination, _ = abi.NewType("bytes32", "byte32", nil)
 
 // bytes type for abi encoding
 var bytesTy, _ = abi.NewType("bytes", "bytes", nil)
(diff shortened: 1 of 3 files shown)


andrewgordstewart:
You expect this would save gas because the Outcome would not need to be encoded as bytes on-chain?

andrewgordstewart:
It's unclear to me, as a reviewer, whether this is a refactor or a change in behaviour. Could you use a few words to explain what change you're making?

geoknee:
The outcome would be encoded as bytes, but at the same time as the rest of the members of `State`.

I tried logging out the encoded bytes during the `TestHash` test, on this branch and on the base branch. We get 864 and 928 bytes respectively*. Of course this is only a rough proxy for how gas costs would be, but I think it's safe to assume calling `abi.encode` twice will be more expensive than calling it once. 

*I think there's more self-describing going on with the encode-then-encode-again approach (telling the decoder what the next `x` bytes represent).

channel/state/state.go
```diff
@@ -127,27 +127,25 @@ func (s State) encode() (types.Bytes, error) {
...
+		{Type: outcome.ExitTy}, // outcome
+		{Type: uint256},        // turnNum
+		{Type: boolTy},         // isFinal
 	}.Pack(
 		ChannelId,
 		s.AppDefinition,
 		s.ChallengeDuration,
 		[]byte(s.AppData), // Note: even though s.AppData is types.bytes, which is an alias for []byte], Pack will not accept types.bytes
-		[]byte(outcome),
+		s.Outcome,
```
geoknee, line 148: ... and put the unencoded data straight into the abi encoder. Rather than encoding some data, and mixing it with some more data and encoding it again.

andrewgordstewart:
The PR description has been placed in the commit message of b3eade1dd2c8ed5a29fb0f5e1a38601ddf60e249.

PR #138, opened 2022-01-11
Improve coverage on direct-fund TestNew

NiloCK:
With this PR, `TestNew` now
- hits every error return statement in `direct-fund.New()`
- tests a variety of valid configurations instead of a single one

### DIFF
--- a/protocols/direct-fund/direct-fund_test.go
+++ b/protocols/direct-fund/direct-fund_test.go
@@ -79,6 +79,11 @@ func TestNew(t *testing.T) {
 	if _, err := New(finalState, testState.Participants[0], isTwoPartyLedger, myDestination, theirDestination); err == nil {
 		t.Error("expected an error when constructing with an intial state marked final, but got nil")
 	}
+	if _, err := New(testState, testState.Participants[0], true, myDestination, types.Destination{}); err == nil {
+		t.Error("expected an error when constructing a 2-party-ledger with a blank destination, but got nil")
+	}
+	if _, err := New(testState, testState.Participants[0], true, types.Destination{}, theirDestination); err == nil {
+		t.Error("expected an error when constructing a 2-party-ledger with a blank destination, but got nil")
 	}
 
 }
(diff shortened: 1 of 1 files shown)


protocols/direct-fund/direct-fund_test.go
```diff
@@ -61,19 +61,29 @@ var theirDestination = bob.destination // only needed if isTwoPartyLedger = true
...
 
-	// Assert that constructing with a final state should return an error
 	if _, err := New(finalState, testState.Participants[0], isTwoPartyLedger, myDestination, theirDestination); err == nil {
-		t.Error("Expected an error when constructing with an invalid state, but got nil")
+		t.Error("expected an error when constructing with an intial state marked final, but got nil")
+	}
+	if _, err := New(testState, testState.Participants[0], true, myDestination, types.Destination{}); err == nil {
+		t.Error("expected an error when constructing a 2-party-ledger with a blank destination, but got nil")
+	}
+	if _, err := New(testState, testState.Participants[0], true, types.Destination{}, theirDestination); err == nil {
```
geoknee, line 85: Is the ostensibly inconsistent behaviour justified, here? i.e. a blank destination is not valid for two party ledgers, but it is otherwise?

NiloCK: This inconsistent behaviour is due to the "overloading" of `channel.New` with distinct requirements on its parameters for generic channels vs the case of two party ledger channels. A dangling concern, with or without addressing this inconsistent behaviour, is that `channel.TheirDestination` doesn't make a of lot sense in the general channel case.

This PR aimed at an incremental coverage improvement of the test suite, but your question prompted me to take a shot at a little cleanup of `Channel` in #141.

andrewgordstewart: I believe we don't intend to implement this behaviour, regardless of whether #141 is emerged. If I understand correctly, I'd suggest removing this test case and merging the PR.

andrewgordstewart (approved):

PR #140, opened 2022-01-12
Introduce `NewTestMessageService` constructor function

geoknee:
I think this improves clarity about when to `.Run()` the service, as well as removing some repetition in test files and a nicer dev experience overall.


geoknee:
Edit: this is more crucial than I had previously thought. We cannot construct literal a `TestMessageService` outside of the `messageservice` package due to their being several unexported (private) fields.

NiloCK (approved):
:+1:

PR #141, opened 2022-01-12
Subtype `Channel` with `TwoPartyLedger`

NiloCK:
This refactor was inspired by @geoknee's comment on #138. Motivations for this PR:

- existing `Channel` initializer takes lots of parameters
- existing `Channel` initializer `channel.New()` takes redundant parameters (myDestination, theirDestination).
- existing `Channel` struct has a field that doesn't make sense in the generic case: `TheirDestination`

Solution:
- Reduce the responsibility of `channel.New` by creating a separate initializer for TwoPartyLedger channels
- Move `TwoPartyLedger`-specific implementation details away from `Channel`

Reading this per-commit is recommended. There ought to be green ticks after commit #2, which exists as a more modest refactoring option.

Marked as draft because a little further work would be required: `DirectFundObjective` has a field of type `Channel`. This (and things like it, which want to operate on generic channels) will need to be refactored to expect an interface rather than a concrete implementation.


geoknee:
👏 

One remaining niggle is that the "base class" `Channel` still "knows about" the "derived class" `TwoPartyLedger` in the sense that it has the `isTwoPartyLedger` boolean. Is there a way to fix that as part of your further work? 

If we go down the interface route, I suppose we will have to convert a few property accesses into method calls?

andrewgordstewart:
nit:
> There ought to be green ticks after commit #2, which exists as a more modest refactoring option.

GitHub magically turns #2 into a link. If you paste a few leading characters of the commit hash, it does what you want: 7265900

(force pushing breaks this, but we ought to avoid force pushing on anything that others can see/interact with anyway!)

andrewgordstewart:
This looks like valuable work to continue with. For this project, I would favour specific implementations over abstract, generalized implementations, unless:
1. the abstract implementation is obviously better (eg. leads to more maintainable code)
2. we have a known future need for the abstract implementation

geoknee (approved):

PR #67, opened 2021-11-18
Add `.adr` directory and flowchart/statechart doc

NiloCK:
credit to @geoknee, whose work I have shamelessly rearranged in order to up my commit count.

Adapted from original PR:

> 
> This PR adds an architectural decision record ([preview here!](https://github.com/statechannels/go-nitro/tree/adr-directory/protocols/.adr)).
> 
> The architectural decision record (ADR) is in a hidden folder, and contains much more information about why we went with the current design. It contains discarded code -- but I chose to embed this in markdown to remove any confusion about it forming a functional part of the codebase. It is information only.  See @NiloCK's [pitch](https://www.notion.so/statechannels/Store-Finalized-ADRs-in-their-respective-Repository-ies-b4851f8e1d2a4b629ece64b69daa6a9a).
> 
> The ADR is experimental, and generally a bit rougher as a document. I don't want to spend a lot of time polishing it up, so if it proves controversial we can just drop it for now.

~**TODO**: the readme from the original PR no longer contains a back-link to the adr directory. If this is merged, that link should be re-inserted.~

PR #143, opened 2022-01-13
add adr for signature scheme

### LINKED ISSUE #136: Document security risks present in state signature protocol.
> > this PR may introduce a security vulnerability. Stages 2-3 may be necessary to remove the risk of a hash collision that could be present due to two variable-length attributes, `appData` and `outcome`.
> 
> I don't think this is a problem. I think the potential issue comes not from a hash collision `h(p) = h(p')` but from an encoding collision `p = encode(a,b) = encode'(a',b') = p'`. Put otherwise, the bytes `p` could be decoded in two different ways (perhaps the intended way plus another unintended way), giving rise to distinct variables `a' != a, b' != b`.
> 
> Nitro clients should verify the decoding from the contract where funds are locked. Notwithstanding things like upgradeable contracts, if we use `abi.encode` this decoding is deterministic and therefore totally predictable. (Contrast this with using something like `abi.encodePacked`, where [multiple inputs encode to the same bytes](https://www.youtube.com/watch?v=rxZR3ITZlzE) -- as a consequence there is no decoder for `abi.encodePacked` as the there is no unique pre-image).
> 
> A signature on `h(p)` might get "replayed" to a contract with a different decoder, with unintended effects -- but as long as no funds are ever sent there, there should be no issue. We do need to ensure that the signature is generated by a truly ephemeral key, though -- the **only** place where that key should have meaning is for signing nitro state channel updates.
> 
> The steps 2-3 (hashing the outcome and appdata separately) would make it much harder (effectively impossible) to engineer the malicious decoder contract. But I think this extra protection is unnecessary given the above.

Where does this documentation belong?

_Originally posted by @andrewgordstewart in https://github.com/statechannels/go-nitro/issues/130#issuecomment-1009237494_

geoknee:
Following the pattern set in #67. Closes #136


channel/state/.adr/0000-signature-scheme.md
```diff
@@ -0,0 +1,15 @@
...
+
+We address this issue as follows:
+
+Nitro clients should verify the decoding from the contract where funds are locked. Notwithstanding things like upgradeable contracts, if we use `abi.encode` this decoding is deterministic and therefore totally predictable. (Contrast this with using something like `abi.encodePacked`, where multiple inputs encode to the same bytes -- as a consequence there is no decoder for `abi.encodePacked` as the there is no unique pre-image).
+
+A signature on `h(p)` might get "replayed" to a contract with a different decoder, with unintended effects -- but as long as no funds are ever sent there, there should be no issue. We do need to ensure that the signature is generated by a truly ephemeral key, though -- the only place where that key should have meaning is for signing nitro state channel updates.
```
andrewgordstewart, line 13: > We do need to ensure that the signature is generated by a truly ephemeral key, though -- the only place where that key should have meaning is for signing nitro state channel updates.

It's not totally obvious to me why this is true. I'll take a crack at it:

_If a key `k` is used for signing state channel updates as well as commitments in some other protocol `P`, then there's a risk that the designer of protocol `P` designed the possibility of a hash collision **by means of a specially crafted encoder.**_

Is that about right?

andrewgordstewart:
This is great stuff. I'll point you to concern 2 in https://github.com/statechannels/go-nitro/pull/67#issuecomment-1012453995

channel/state/.adr/0000-signature-scheme.md
geoknee: For each signature made with a given private key, the owner of said key should be checking on the consequences of that signature for each and every contract (on each and every chain) which holds funds or other important state for the owner and can be controlled to some extent by that key. This includes the native assets and accounts of those chains. They do these checks on the data to be signed, and decline to sign if they spot an issue. 

Such contracts/chains might only be deployed _after_ some signatures have been made with that key. So the owner of the key needs to also check all previous signatures that have been made with that key against the new chain/contract before depositing funds there. This is pretty infeasible, too, so the owner should select a new private key for every new protocol that they enter* in order to remain safe.

An example attack would be to take the data which was signed during execution of an existing protocol (for example, state channels) and convince the signer to partake in a new protocol backed by a contract which has the "specially crafted encoder" that you mention. If the signer deposits money it could be at risk by having the existing signatures (which are already in the wild) played onto the new contract. 

---
*this reminds me of the conversation in standup about enter/exit

andrewgordstewart: Can we find existing literature that covers this nuance? If so, we should link to it somehow. If not, it deserves to be written.

PR #145, opened 2022-01-13
change type of TurnNum to uint64

### LINKED ISSUE #95: consider uint64 type for turnNum (instead of big.Int)
In our current solidity code, we have an implicit 48 bit limit on turn numbers https://docs.statechannels.org/docs/protocol-tutorial/understand-adjudicator-status.

This means it is possible to represent using a native 64 bit integer in go, which will make much of our code easier to read and write.

geoknee:
Closes #95 

Pros: 
* we get to use a more ergonomic type, with
    -  `+` instead of `.Add()` , etc
    - no worries about cloning safely
* it mirrors our (anticipated?) on chain code better
* 
Cons:
* it is not safe to use if we had turnnums exceeding 9,223,372,036,854,775,807
* it's an additional type, different to the others we are already using in e.g. `State`.

### DIFF
--- a/channel/channel.go
+++ b/channel/channel.go
@@ -3,7 +3,6 @@ package channel
 import (
 	"bytes"
 	"errors"
-	"math/big"
 	"reflect"
 
 	"github.com/ethereum/go-ethereum/common"
@@ -36,8 +35,8 @@ type Channel struct {
 // New constructs a new Channel from the supplied state.
 func New(s state.State, isTwoPartyLedger bool, myIndex uint, myDestination types.Destination, theirDestination types.Destination) (Channel, error) {
 	c := Channel{}
-	if s.TurnNum.Cmp(big.NewInt(0)) != 0 {
-		return c, errors.New(`objective must be constructed with a turnNum 0 state`)
+	if s.TurnNum != 0 {
+		return c, errors.New(`objective must be constructed with TurnNum=0 state`)
 	}
 	if isTwoPartyLedger && (myDestination == types.Destination{} || theirDestination == types.Destination{}) {
 		return c, errors.New(`two party ledger channels must have non-null myDestination and theirDestination`)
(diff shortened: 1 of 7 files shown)


andrewgordstewart:
I wholeheartedly endorse this change!

channel/state/state.go
```diff
@@ -140,7 +140,7 @@ func (s State) encode() (types.Bytes, error) {
 		ChannelId,
 		[]byte(s.AppData), // Note: even though s.AppData is types.bytes, which is an alias for []byte], Pack will not accept types.bytes
 		s.Outcome,
-		s.TurnNum,
+		big.NewInt(int64(s.TurnNum)),
```
andrewgordstewart, line 143: Why do you need to convert to an `int64` on the way to a `bigInt`?

channel/channel.go
```diff
@@ -36,8 +35,8 @@ type Channel struct {
 // New constructs a new Channel from the supplied state.
 func New(s state.State, isTwoPartyLedger bool, myIndex uint, myDestination types.Destination, theirDestination types.Destination) (Channel, error) {
 	c := Channel{}
-	if s.TurnNum.Cmp(big.NewInt(0)) != 0 {
-		return c, errors.New(`objective must be constructed with a turnNum 0 state`)
+	if s.TurnNum != 0 {
```
andrewgordstewart, line 38: should `0` be an enum `PreFundTurnNum`?

andrewgordstewart (commented):
This change made me wonder: what's responsible for checking for overflows?

channel/state/state.go
geoknee: Because the `big.NewInt` only accepts signed integers. This is different to ethers.js BigNumber.

channel/channel.go
geoknee: Yeah I think that's a good suggestion. Will raise a new PR.

geoknee:
> This change made me wonder: what's responsible for checking for overflows?

I think right now there is nothing checking for overflows. Can you be more specific about what behaviour you think we need?

channel/channel.go
geoknee: #150

PR #147, opened 2022-01-14
add some voucher-related code

andrewgordstewart:
This adds some scaffolding code that we can use to discuss the requirements of a payment provider and receipt provider. The initial prototype aims to outline the core payment flow.

With this prototype, I make the following assumptions:
1. Payments are always unidirectional, and Alice always pays Bob.
2. The virtual-defunder protocol will use a voucher to determine how to conclude a virtually funded channel

Both assumptions are easy to relax, depending on feature requests.

See https://www.notion.so/statechannels/Voucher-channels-f8609ab3f45142eda3e552e58a676bb1 for more detailed discussion.

### DIFF
--- a/payments/payment-provider.go
+++ b/payments/payment-provider.go
@@ -1,0 +1,95 @@
+package payments
+
+import (
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/statechannels/go-nitro/channel/state"
+)
+
+const alice = 0
+
+type (
+	// A PaymentProvider can be used to make a payment for a given channel.
+	PaymentProvider struct {
+		fp             state.FixedPart
+		largestVoucher Voucher
+		pk             []byte
+
+		payments chan Voucher // gochan used to send vouchers to Bob
+	}
+
+	// A receipt provider enables the consumer to listen to incoming
+	// vouchers, validate them, and notify the consumer of a payment
+	// received
+	ReceiptProvider struct {
+		fp             state.FixedPart
+		largestVoucher Voucher
+
+		vouchers chan Voucher // todo: should be unidirectional
+		payments chan uint    // todo: API needs to be clarified
+	}
+)
+
+// implemented just to make use of fp ...
+func NewPaymentProvider(fp state.FixedPart, pk []byte) (PaymentProvider, error) {
+	chanId, err := fp.ChannelId()
+	pp := PaymentProvider{}
+
+	if err != nil {
+		return pp, err
+	}
+
+	pp.largestVoucher = Voucher{
+		channelId: chanId,
+	}
+
+	pp.largestVoucher.sign(pk)
+
+	pp.fp = fp
+
+	return pp, nil
+}
+
+// PayBob will use add amount to the biggest voucher to create
+// a new voucher. This voucher is then signed and sent to Bob.
+func (pp *PaymentProvider) PayBob(amount uint64) error {
+	channelId := pp.largestVoucher.channelId
+	pp.largestVoucher = Voucher{
+		channelId: channelId,
+		amount:    uint(amount),
+	}
+	pp.largestVoucher.sign(pp.pk)
+
+	pp.payments <- pp.largestVoucher
+
+	return nil
+}
+
+func (rp *ReceiptProvider) ValidateVouchers(listener chan uint64) {
+	for voucher := range rp.vouchers {
+
+		signer, err := voucher.recoverSigner()
+
+		if err != nil {
+			// todo: figure out what to do with errors within a goroutine
+			panic(err)
+		}
+
+		if signer != rp.alice() {
+			panic("invalid signature")
+		}
+
+		if rp.largestVoucher.amount >= voucher.amount {
+			panic("received a stale voucher")
+		}
+
+		received := voucher.amount - rp.largestVoucher.amount
+
+		// Clarification needed: What information would a receipt provider need to provide
+		// to its consumer?
+		rp.payments <- received
+	}
+}
+
+func (rp ReceiptProvider) alice() common.Address {
+	return rp.fp.Participants[alice]
+}
--- a/payments/payment-provider.go
+++ b/payments/payment-provider.go
@@ -42,7 +42,9 @@ func NewPaymentProvider(fp state.FixedPart, pk []byte) (PaymentProvider, error) 
 		channelId: chanId,
 	}
 
-	pp.largestVoucher.sign(pk)
+	if err := pp.largestVoucher.sign(pk); err != nil {
+		panic(err)
+	}
 
 	pp.fp = fp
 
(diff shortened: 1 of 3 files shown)


payments/payment-provider.go
```diff
@@ -0,0 +1,99 @@
...
+import (
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/statechannels/go-nitro/channel/state"
+)
+
+const alice = 0
+
+type (
+	// A PaymentProvider can be used to make a payment for a given channel.
+	PaymentProvider struct {
```
geoknee, line 12: Maybe we could use `Manager` or something instead of `Provider`, which is already in use in our lexicon.

payments/payment-provider.go
```diff
@@ -0,0 +1,99 @@
...
+		fp             state.FixedPart
```
geoknee, line 13: Is it possible to use `channelId` instead of `state.FixedPart` ?

payments/payment-provider.go
```diff
@@ -0,0 +1,99 @@
...
+		panic(err)
+	}
+
+	pp.fp = fp
+
+	return pp, nil
+}
+
+// PayBob will use add amount to the biggest voucher to create
+// a new voucher. This voucher is then signed and sent to Bob.
```
geoknee, line 55: typo?

payments/payment-provider.go
```diff
@@ -0,0 +1,99 @@
...
+func (pp *PaymentProvider) PayBob(amount uint64) error {
+	channelId := pp.largestVoucher.channelId
+	pp.largestVoucher = Voucher{
+		channelId: channelId,
+		amount:    uint(amount),
```
geoknee, line 60: I think you need to add the current largest voucher amount to the passed in amount?

andrewgordstewart, line 60: This should be `amount + pp.largestVoucher.amount`.

payments/payment-provider.go
```diff
@@ -0,0 +1,99 @@
...
+
+	pp.fp = fp
+
+	return pp, nil
+}
+
+// PayBob will use add amount to the biggest voucher to create
+// a new voucher. This voucher is then signed and sent to Bob.
+func (pp *PaymentProvider) PayBob(amount uint64) error {
+	channelId := pp.largestVoucher.channelId
```
andrewgordstewart, line 57: Could be `pp.ChannelId`

payments/payment-provider.go
```diff
@@ -0,0 +1,99 @@
...
+	return nil
+}
+
+func (rp *ReceiptProvider) ValidateVouchers(listener chan uint64) {
+	for voucher := range rp.vouchers {
+
+		signer, err := voucher.recoverSigner()
+
+		if err != nil {
+			// todo: figure out what to do with errors within a goroutine
```
andrewgordstewart, line 77: We should ask Hannah/Perun what to do in a situation like this:
- assume `ValidateVouchers` is run as a goroutine
- say the goroutine hits an error. we don't want the error to terminate `ValidateVouchers`

What do we do?

geoknee (commented):
I like the direction of this. Do you have an idea of what kind of API would work well with this approach?

payments/payment-provider.go
andrewgordstewart: 52825bce9952ba352b854e3b35e7cfbf4765c389

payments/payment-provider.go
andrewgordstewart: 4b9e723c4cf28763d05c253be2a3f4e985163044

payments/payment-provider.go
andrewgordstewart: 049fdd0357146d3b91321f91666a06540f3c0afc

payments/payment-provider.go
andrewgordstewart: 7fccd8d8566a43898e02f6a5e78ab3e3bc8a1ba3

payments/payment-provider.go
andrewgordstewart: 049fdd0357146d3b91321f91666a06540f3c0afc

payments/payment-provider.go
andrewgordstewart: 4b9e723c4cf28763d05c253be2a3f4e985163044

payments/payment-provider.go
andrewgordstewart: 52825bce9952ba352b854e3b35e7cfbf4765c389

andrewgordstewart:
> Do you have an idea of what kind of API would work well with this approach?

I've expanded on the idea [here](https://www.notion.so/statechannels/Voucher-channels-f8609ab3f45142eda3e552e58a676bb1). Does [this section](https://www.notion.so/statechannels/Voucher-channels-f8609ab3f45142eda3e552e58a676bb1#552da67a8c2b4699b6ece93a355bd8b1) answer your question?

geoknee:
> > Do you have an idea of what kind of API would work well with this approach?
> 
> I've expanded on the idea [here](https://www.notion.so/statechannels/Voucher-channels-f8609ab3f45142eda3e552e58a676bb1). Does [this section](https://www.notion.so/statechannels/Voucher-channels-f8609ab3f45142eda3e552e58a676bb1#552da67a8c2b4699b6ece93a355bd8b1) answer your question?

It definitely helps, but I am wondering about the API not only of these `Payment/ReceiptManagers` but of the `go-nitro` wallet/client, too. Presumably the overall flow is something like this:

* consuming code asks `go-nitro` for a virtual channel, then 
* constructs a payment manager for that channel (using the id), 
* makes payments, and then 
* supplies the latest voucher back to `go-nitro` to handle defunding of that virtual channel

Is that about right?

andrewgordstewart:
> It definitely helps, but I am wondering about the API not only of these `Payment/ReceiptManagers` but of the `go-nitro` wallet/client, too. Presumably the overall flow is something like this:
> 
> * consuming code asks `go-nitro` for a virtual channel, then
> * constructs a payment manager for that channel (using the id),
> * makes payments, and then
> * supplies the latest voucher back to `go-nitro` to handle defunding of that virtual channel
> 
> Is that about right?

Ah, I didn't quite get that. The document and prototype code in this PR is concerned about the path of making payments _assuming the channel is funded_. I'll clarify this in the doc.

I think I'd prefer the PaymentManager being responsible for constructing a virtual channel. This would encapsulate the complexity of dealing with `go-nitro` and managing channels. The PaymentManager's API is then focused on what its owner cares about, which is making payments.

What do you think?

geoknee:
I am still struggling to see how all the pieces might fit together. Would it be possible to come up with something like this diagram https://docs.statechannels.org/docs/server-app-devs/make-api-calls, showing both `go-nitro` Clients and `PaymentManagers`? From that we could easily see if the `PaymentManager` simply *wraps* the `go-nitro` Client (which is I think what you are driving at?) -- then the application never needs to see the `go-nitro` Client (it becomes an implementation detail of the `PaymentManager`. 

Lately I am liking https://sequencediagram.org/ for making such diagrams.

geoknee:
@andrewgordstewart shall we close this PR if it is no longer being worked on?

lalexgap:
> @andrewgordstewart shall we close this PR if it is no longer being worked on?

I'm going to close this now and seek forgiveness later 😛

PR #149, opened 2022-01-14
Implement a Mock Store (memstore)

NiloCK:
This is a first cut for:
- getting a branch to work from in service of wiring up engine-engine tests
- feedback on approach

It has a few labelled todos:
- locking
- absent tests
- absent interface method implementation

### DIFF
--- a/client/engine/store/mockstore.go
+++ b/client/engine/store/mockstore.go
@@ -1,0 +1,53 @@
+package store
+
+import (
+	"github.com/statechannels/go-nitro/protocols"
+	"github.com/statechannels/go-nitro/types"
+)
+
+type MockStore struct {
+	objectives map[protocols.ObjectiveId]protocols.Objective
+
+	key []byte // the signing key of the store's engine
+}
+
+func NewMockStore(key []byte) Store {
+	ms := MockStore{}
+	ms.key = key
+	ms.objectives = make(map[protocols.ObjectiveId]protocols.Objective)
+
+	return ms
+}
+
+func (ms MockStore) GetChannelSecretKey() *[]byte {
+	return &ms.key
+}
+
+func (ms MockStore) GetObjectiveById(id protocols.ObjectiveId) (obj protocols.Objective, ok bool) {
+	// todo: locking
+	obj, ok = ms.objectives[id]
+	return obj, ok
+}
+
+func (ms MockStore) SetObjective(obj protocols.Objective) error {
+	// todo: locking
+	ms.objectives[obj.Id()] = obj
+	return nil
+}
+
+func (ms MockStore) GetObjectiveByChannelId(channelId types.Destination) (protocols.Objective, bool) {
+	// todo: locking
+	for _, obj := range ms.objectives {
+		for _, ch := range obj.Channels() {
+			if ch == channelId {
+				return obj, true
+			}
+		}
+	}
+
+	return nil, false
+}
+
+func (ms MockStore) UpdateProgressLastMadeAt(protocols.ObjectiveId, protocols.WaitingFor) {
+	// todo
+}
--- a/client/engine/store/mocktore_test.go
+++ b/client/engine/store/mocktore_test.go
@@ -15,9 +15,10 @@ func TestNewMockStore(t *testing.T) {
 
 func TestSetGetObjective(t *testing.T) {
 	ms := NewMockStore([]byte{})
-	got, ok := ms.GetObjectiveById("404")
+        id := "404"
+	got, ok := ms.GetObjectiveById(id)
 	if ok {
-		t.Errorf("expected not to find the 404 objective, but found %v", got)
+		t.Errorf("expected not to find the %s objective, but found %v", id, got)
 	}
 
 	ts := state.TestState
--- a/client/engine/store/mocktore_test.go
+++ b/client/engine/store/mocktore_test.go
@@ -46,7 +46,31 @@ func TestSetGetObjective(t *testing.T) {
 }
 
 func TestGetObjectiveByChannelId(t *testing.T) {
-	// todo
+	ms := NewMockStore([]byte{})
+
+	ts := state.TestState
+	ts.TurnNum = 0
+
+	testObj, _ := directfund.New(
+		ts,
+		ts.Participants[0],
+		false,
+		types.AddressToDestination(ts.Participants[0]),
+		types.AddressToDestination(ts.Participants[1]),
+	)
+
+	if err := ms.SetObjective(testObj); err != nil {
+		t.Errorf("error setting objective %v: %s", testObj, err.Error())
+	}
+
+	got, ok := ms.GetObjectiveByChannelId(testObj.C.Id)
+
+	if !ok {
+		t.Errorf("expected to find the inserted objective, but didn't")
+	}
+	if got.Id() != testObj.Id() {
+		t.Errorf("expected to retrieve same objective Id as was passed in, but didn't")
+	}
 }
 
 // BUG(geoknee)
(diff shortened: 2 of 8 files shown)


NiloCK:
I've left `UpdateProgressLastMadeAt()` as a todo because I'm not sure it should exist. As far as the `Store` is concerned, shouldn't recording this sort of metadata be the responsibility of `store.SetObjective()`? IE, writing a newly cranked objective to the db should internally manage these metadata updates.

andrewgordstewart:
> I've left UpdateProgressLastMadeAt() as a todo because I'm not sure it should exist. As far as the Store is concerned, shouldn't recording this sort of metadata be the responsibility of store.SetObjective()? IE, writing a newly cranked objective to the db should internally manage these metadata updates.

Setting timestamps while setting the objective sounds like the correct approach to me.

client/engine/store/mocktore_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+	testObj, _ := directfund.New(
+		ts,
+		ts.Participants[0],
+		false,
+		types.AddressToDestination(ts.Participants[0]),
+		types.AddressToDestination(ts.Participants[1]),
+	)
+
+	ms.SetObjective(testObj)
```
andrewgordstewart: Screen Shot 2022-01-14 at 4 00 35 PM

This is such a nice feature.

andrewgordstewart: Should this assert on the return value of SetObjective?

client/engine/store/mocktore_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+func TestGetObjectiveByChannelId(t *testing.T) {
+	// todo
+}
+
+// BUG(geoknee)
+func TestGetChannelSecretKey(t *testing.T) {
+	// from state/test-fixtures.go
+	sk := common.Hex2Bytes("caab404f975b4620747174a75f08d98b4e5a7053b691b41bcfc0d839d48b7634")
+	pk := common.HexToAddress("0xF5A1BB5607C9D079E46d1B3Dc33f257d937b43BD")
```
andrewgordstewart, line 80: How come we can't import these fixtures?

client/engine/store/mocktore_test.go
```diff
@@ -0,0 +1,65 @@
...
+	ms := NewMockStore(sk)
+	key := ms.GetChannelSecretKey()
+
+	signedMsg, _ := state.SignEthereumMessage([]byte("asdfasdf"), *key)
+
+	recoveredSigner, _ := state.RecoverEthereumMessageSigner([]byte("asdfasdf"), signedMsg)
+
+	if recoveredSigner != pk {
+		t.Errorf("expected to recover %x, but got %x", pk, recoveredSigner)
+	}
```
andrewgordstewart, line 93: Nice little test tactic!

client/engine/store/mocktore_test.go
```diff
@@ -0,0 +1,94 @@
...
+func TestNewMockStore(t *testing.T) {
+	NewMockStore([]byte{'a', 'b', 'c'})
+}
+
+func TestSetGetObjective(t *testing.T) {
+	ms := NewMockStore([]byte{})
+	got, ok := ms.GetObjectiveById("404")
+	if ok {
+		t.Errorf("expected not to find the 404 objective, but found %v", got)
+	}
```
andrewgordstewart, line 21: nit:
```suggestion
        id := "404"
	got, ok := ms.GetObjectiveById(id)
	if ok {
		t.Errorf("expected not to find the %s objective, but found %v", id, got)
	}
```

client/engine/store/mocktore_test.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/statechannels/go-nitro/types"
+)
+
+func TestNewMockStore(t *testing.T) {
+	NewMockStore([]byte{'a', 'b', 'c'})
+}
+
+func TestSetGetObjective(t *testing.T) {
+	ms := NewMockStore([]byte{})
+	got, ok := ms.GetObjectiveById("404")
```
andrewgordstewart, line 18: Looking at this, I am wondering what benefits a more strictly typed store API would be.

Accepting any old string makes it easier to get garbage input. Defining something like `type ObjectiveId = string` might be a net positive.

andrewgordstewart (approved):
This seems totally reasonable to me, but what do I know!

andrewgordstewart:
One thing that could make this PR easier for the reviewer and posterity is to implement & test one method per commit. This helps a reviewer, for instance, who wouldn't then need to cross-check that each method is tested.

See points 5 & 7 [here](https://mtlynch.io/code-review-love/#5-narrowly-scope-changes).

client/engine/store/mocktore_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+	if !ok {
+		t.Errorf("expected to find the inserted objective, but didn't")
+	}
+	if got.Id() != testObj.Id() {
+		t.Errorf("expected to retrieve same objective Id as was passed in, but didn't")
+	}
+}
+
+// BUG(geoknee)
```
NiloCK, line 76:
```suggestion
```

NiloCK:
> Setting timestamps while setting the objective sounds like the correct approach to me.

moved to #152

client/engine/store/mockstore.go
```diff
@@ -0,0 +1,53 @@
...
+
+	return ms
+}
+
+func (ms MockStore) GetChannelSecretKey() *[]byte {
+	return &ms.key
+}
+
+func (ms MockStore) GetObjectiveById(id protocols.ObjectiveId) (obj protocols.Objective, ok bool) {
+	// todo: locking
```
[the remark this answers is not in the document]
NiloCK: locking is potentially most important with respect to the ledger channel between Irene and Irma. Consider:

```
Alice             Bob 
    \            / 
     Irene - Irma 
    /            \
Andy              Beth
```

If (Alice and Bob) and (Andy and Beth) simultaneously attempt to open virtual channels, this will create multiple funding objectives for each of Irene, Irma, which reference the same I-I ledger channel and can update the state of that channel.

We don't want race conditions here, where changes from one objective overwrite those from the other, or prevent those from the other from taking place.

client/engine/store/store.go
```diff
@@ -10,9 +10,9 @@ import (
 type Store interface {
 	GetChannelSecretKey() *[]byte // Get a pointer to a secret key for signing channel updates
 
-	GetObjectiveById(protocols.ObjectiveId) protocols.Objective // Read an existing objective
-	GetObjectiveByChannelId(types.Bytes32) protocols.Objective  // Get the objective that currently owns the channel with the supplied ChannelId
-	SetObjective(protocols.Objective) error                     // Write an objective
+	GetObjectiveById(protocols.ObjectiveId) (obj protocols.Objective, ok bool)    // Read an existing objective
+	GetObjectiveByChannelId(types.Destination) (obj protocols.Objective, ok bool) // Get the objective that currently owns the channel with the supplied ChannelId
```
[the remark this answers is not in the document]
NiloCK: comma-OK is a common go idiom and (for instance) is what is returned when you read a value from a map:

```
var m map[string]int = { "a": 1 }
x := m["a"] // 1
y := m["b"] // 0 (non-entries return the zero value of the type)
foo, ok := m["a"] // 1, true
bar, ok := m["b"] // 0, false (false allows you to disregard the 0 value)
```

later, the `ok` here will likely need to be replaced with an `error`

protocols/direct-fund/direct-fund.go
```diff
@@ -179,6 +179,12 @@ func (s DirectFundObjective) Crank(secretKey *[]byte) (protocols.Objective, prot
 	return updated, NoSideEffects, WaitingForNothing, nil
 }
 
+func (s DirectFundObjective) Channels() []types.Destination {
+	ret := make([]types.Destination, 0, 1)
+	ret = append(ret, s.C.Id)
```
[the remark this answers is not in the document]
NiloCK: 1. We can append as many values as we like, but the array which backs the slice has a pre-allocated length of 1. If we append more than one value, the array size is dynamically adjusted (doubled, or something, and maybe moved to new physical location in memory etc). Defining `capacity` of `n` is good practice if you know in advance how large your slice will be, or an upper bound on that size, since it precludes these memory reallocation operations.

The rest looks correct up to my understanding.

client/engine/store/mocktore_test.go
NiloCK: This is a good question which I will defer to a follow up PR.

client/engine/store/mocktore_test.go
NiloCK: created #153

NiloCK:
This work was rebased and is to be merged from #154

client/engine/store/mocktore_test.go
NiloCK: ... and closed. `GetObjectiveById(id)` already expects a `protocols.Objective` typed argument. The string literal passes this type check, but string-typed variables passed in do not.

PR #150, opened 2022-01-14
Overhaul turn num constants by introducing an enum

geoknee:
Pros:
* easy to change the meaning of e.g. "post fund turn number" by making a one line change
* better readability

Cons:
* I think we cannot force the use of the enum in Go (as we could in Typescript). This is not really a "con" but just a "not as nice as we might like"

I placed these constants in the `channel` package where I think they belong. The `protocols` package does not need to deal with turn numbers explicitly (only things like "post fund state"). So turn numbers are an implementation detail that is encapsulated within the `channel` package.

### DIFF
--- a/channel/channel.go
+++ b/channel/channel.go
@@ -35,7 +35,7 @@ type Channel struct {
 // New constructs a new Channel from the supplied state.
 func New(s state.State, isTwoPartyLedger bool, myIndex uint, myDestination types.Destination, theirDestination types.Destination) (Channel, error) {
 	c := Channel{}
-	if s.TurnNum != 0 {
+	if s.TurnNum != PREFUNDTURNUM {
 		return c, errors.New(`objective must be constructed with TurnNum=0 state`)
 	}
 	if isTwoPartyLedger && (myDestination == types.Destination{} || theirDestination == types.Destination{}) {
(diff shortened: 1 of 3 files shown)


andrewgordstewart:
> I think we cannot force the use of the enum in Go (as we could in Typescript). This is not really a "con" but just a "not as nice as we might like"

IMO we ought to endorse the use of enums via code review culture.

geoknee:
> Is there a reason why `POSTFUNDTURNNUM` is better than `PostFundTurn`?

I had found some claims online that constants should be in all caps, although not very very authoritative sources. I think that may be a hangover from conventions in other languages like C. I think the idea is to provide a visual hint that this is a constant and cannot / should not be reassigned. It's definitely a thing in Javascript. 

I don't feel strongly about the casing. All caps may not really make sense in Go, since if we want a private (non exported) constant, we would need to lowercase the first letter -- `pOSTURNNUM`, pretty weird.

I don't think shortening by dropping the `Num` is worth it. You save a few characters but you weaken the connection to `TurnNum`. I suppose we may change to `version` at some point in the future?

andrewgordstewart:
> > Is there a reason why `POSTFUNDTURNNUM` is better than `PostFundTurn`?
> 
> I had found some claims online that constants should be in all caps, although not very very authoritative sources. I think that may be a hangover from conventions in other languages like C. I think the idea is to provide a visual hint that this is a constant and cannot / should not be reassigned. It's definitely a thing in Javascript.
> 
> I don't feel strongly about the casing. All caps may not really make sense in Go, since if we want a private (non exported) constant, we would need to lowercase the first letter -- `pOSTURNNUM`, pretty weird.
> 
> I don't think shortening by dropping the `Num` is worth it. You save a few characters but you weaken the connection to `TurnNum`. I suppose we may change to `version` at some point in the future?

- The ["official" code review standards](https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps) suggests breaking the convention from other languages
- This [SO answer](https://stackoverflow.com/questions/22688906/go-naming-conventions-for-const) suggests the standard library uses camel case.
- I picked a [random file](https://github.com/golang/go/blob/5b3ebc8b72f496a5c6892d76fdeee3d9029b55cc/src/net/http/status.go#L9) from the stdlib and found agreement with the SO answer.

I'd prefer `PostFundTurnNum` to `POSTFUNDTURNNUM`.

geoknee:
Thanks for hunting down some better sources. I'll make a PR with your suggestion!

geoknee:
#156

PR #151, opened 2022-01-18
Simplify constructor for `direct-fund-objective`

NiloCK:
This is a follow-up from #141.

### Context

- `direct-fund.New`'s parameter list grew with the introduction of the channel abstraction in conjunction with the virtual fund objective work.
- The splitting of channel constructors in #141 removes the need for these sort of "flag" constructor parameters.

### Changes (per commit)
- the (now-unused) ctor arguments to direct-fund-objective have been removed
- these changes were propagated to other files
- redundant test cases were removed

### DIFF
--- a/client/engine/messageservice/test-messageservice_test.go
+++ b/client/engine/messageservice/test-messageservice_test.go
@@ -12,7 +12,7 @@ import (
 var aliceMS = NewTestMessageService(types.Address{'a'})
 var bobMS = NewTestMessageService(types.Address{'b'})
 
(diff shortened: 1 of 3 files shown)


NiloCK:
An interesting tidbit:

go compiler is strict on unused variable declarations, so one might have expected these unused constructor parameters to be rejected. The reason it doesn't is to allow declarations of methods which satisfy interfaces, even when the specific implementation doesn't need each param of the interface.

geoknee (approved):

PR #155, opened 2022-01-18
Introduce `MockChain` and `SimpleChainService`

### LINKED ISSUE #129: write a TestChainService

geoknee:
The `SimpleChainService` is _per engine_. The `MockChain` is a singleton used for coordination in tests. It is somewhat analogous to ganache or a test rpc network -- although the interface at the moment is essentially the same as the `ChainService`. For this reason the `SimpleChainService` is essentially transparent or "pass through". 

I kept the `SimpleChainService` around because that matches the anticipated real topology of wallet components better.

Closes #129

### DIFF
--- a/client/engine/chainservice/mockchain.go
+++ b/client/engine/chainservice/mockchain.go
@@ -1,0 +1,71 @@
+package chainservice
+
+import (
+	"github.com/statechannels/go-nitro/protocols"
+	"github.com/statechannels/go-nitro/types"
+)
+
+// MockChain provides an interface which simulates a blockchain network. It is designed for use as a central service which multiple
+// ChainServices connect to with Go chans.
+// It keeps a record of of holdings and adjudication status for each channel, accepts transactions and emits events.
+type MockChain struct {
+	out map[types.Address]chan Event // out is a mapping with a chan for each connected ChainService, used to send Events to that service
+	in  chan protocols.Transaction   // in is the chan used to recieve Transactions from multiple ChainServices
+
+	holdings map[types.Destination]types.Funds // holdings tracks funds for each channel
+}
+
+// Out() returns the out chan for a particular ChainService, and narrows the type so that external consumers mays only recieve on it.
+func (mc MockChain) Out(a types.Address) <-chan Event {
+	return mc.out[a]
+}
+
+// In returns the in chan but narrows the type so that external consumers may only send on it.
+func (mc MockChain) In() chan<- protocols.Transaction {
+	return mc.in
+}
+
+// NewMockChain returns a new MockChain with an out chan initialized for each of the addresses passed in.
+func NewMockChain(addresses []types.Address) MockChain {
+
+	mc := MockChain{}
+	mc.out = make(map[types.Address]chan Event)
+	mc.in = make(chan protocols.Transaction)
+	mc.holdings = make(map[types.Destination]types.Funds)
+
+	const BUFFER_SIZE = 10
+	for _, a := range addresses {
+		mc.out[a] = make(chan Event, BUFFER_SIZE)
+	}
+
+	go mc.Run()
+	return mc
+}
+
+// Run starts a listener for transactions on the MockChain's in chan.
+func (mc MockChain) Run() {
+	for tx := range mc.in {
+		mc.handleTx(tx)
+	}
+}
+
+// handleTx responds to the given tx.
+func (mc MockChain) handleTx(tx protocols.Transaction) {
+	if tx.Deposit.IsNonZero() {
+		mc.holdings[tx.ChannelId] = mc.holdings[tx.ChannelId].Add(tx.Deposit)
+	}
+	event := Event{
+		ChannelId:          tx.ChannelId,
+		Holdings:           mc.holdings[tx.ChannelId],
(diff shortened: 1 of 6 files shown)


protocols/interfaces.go
```diff
@@ -15,8 +15,9 @@ type Message struct {
 
 // Transaction is an object to be sent to a blockchain provider.
 type Transaction struct {
```
NiloCK, line 17: Not changed in this PR but:
`Transaction` should probably be `ChainTransaction` or `ChainTx` for disambiguation.

client/engine/chainservice/mockchain.go
```diff
@@ -0,0 +1,71 @@
+package chainservice
+
+import (
+	"github.com/statechannels/go-nitro/protocols"
+	"github.com/statechannels/go-nitro/types"
+)
+
+// MockChain provides an interface which simulates a blockchain network. It is designed for use as a central service which multiple
+// ChainServices connect to with Go chans.
+// It keeps a record of of holdings and adjudication status for each channel, accepts transactions and emits events.
```
NiloCK, line 10:
```suggestion
//
// It keeps a record of of holdings and adjudication status for each channel, accepts transactions and emits events.
```
required if a paragraph break is intended here.

client/engine/chainservice/mockchain.go
```diff
@@ -0,0 +1,71 @@
...
+type MockChain struct {
+	out map[types.Address]chan Event // out is a mapping with a chan for each connected ChainService, used to send Events to that service
+	in  chan protocols.Transaction   // in is the chan used to recieve Transactions from multiple ChainServices
+
+	holdings map[types.Destination]types.Funds // holdings tracks funds for each channel
+}
+
+// Out() returns the out chan for a particular ChainService, and narrows the type so that external consumers mays only recieve on it.
```
NiloCK, line 18:
```suggestion
// Out returns the out chan for a particular ChainService, and narrows the type so that external consumers may only receive on it.
```

client/engine/chainservice/mockchain.go
```diff
@@ -0,0 +1,71 @@
...
+
+// NewMockChain returns a new MockChain with an out chan initialized for each of the addresses passed in.
+func NewMockChain(addresses []types.Address) MockChain {
+
+	mc := MockChain{}
+	mc.out = make(map[types.Address]chan Event)
+	mc.in = make(chan protocols.Transaction)
+	mc.holdings = make(map[types.Destination]types.Funds)
+
+	const BUFFER_SIZE = 10
```
NiloCK, line 36: The test passes with no set buffer. As I read `handleTx`, a non-reading bob will only block further events on his own SimpleMessageService, but not for other engines.

Is this right? Is this buffer required?

NiloCK (approved):
This is great and I think should unblock engine-engine testing.

Questions about buffers can be revisited as-needed.

client/engine/chainservice/mockchain_test.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/statechannels/go-nitro/protocols"
+	"github.com/statechannels/go-nitro/types"
+)
+
+func TestDeposit(t *testing.T) {
+	// The MockChain and SimpleChainService should work together to react to a deposit transaction for a given channel by:
+	// - sending an event with updated holdings for that channel to all SimpleChainServices which are subscribed
+
+	var a = types.Address(common.HexToAddress(`0xF5A1BB5607C9D079E46d1B3Dc33f257d937b43BD`))
+	var b = types.Address(common.HexToAddress(`0xa5A1BB5607C9D079E46d1B3Dc33f257d937b43BD`))
```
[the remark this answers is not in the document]
andrewgordstewart: I like using [vanity addresses](https://vanity-eth.tk) for test addresses.

for instance, here's a private key that starts with `a`, whose address starts with two `a`'s:
- `a24a7ce4163bcf243099184ec3ed29330600803a94c6313668de87bfb1a1f1f9`
- `0xaa6a275f6dF074370Bd50D89E5f9cC03FF725330`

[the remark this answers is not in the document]
geoknee: Yeah good shout on the vanity addresses. In this test we don't need any private key, so I went with simply `"a"` and `"b"` in the end a935fce.

client/engine/chainservice/simplechainservice.go
```diff
@@ -0,0 +1,53 @@
...
+	return mcs.out
+}
+
+// In returns the in chan but narrows the type so that external consumers mays only send on it.
+func (mcs SimpleChainService) In() chan<- protocols.Transaction {
+	return mcs.in
+}
+
+// listenForTransactions pipes transactions to the MockChain
+func (mcs SimpleChainService) listenForTransactions() {
```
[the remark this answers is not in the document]
geoknee: 60d3804. Happily, the VSCode "rename symbol" feature also updates the docstring above the function declaration.

client/engine/chainservice/mockchain.go
geoknee: Yes you're right. I think this was a hangover from a debugging session where I hadn't yet made a goroutine for each send from the `MockChain`.

protocols/interfaces.go
geoknee: #157

PR #156, opened 2022-01-18
change case of constants

geoknee:


channel/constants.go
```diff
@@ -1,7 +1,7 @@
 package channel
 
 const (
-	PREFUNDTURNUM uint64 = iota
-	POSTFUNDTURNNUM
-	MAXTURNNUM = ^uint64(0) // MAXTURNNUM is a reserved value which is taken to mean "there is not yet a supported state"
+	PreFundTurnNum uint64 = iota
+	PostFundTurnNum
+	MaxTurnNum = ^uint64(0) // MAXTURNNUM is a reserved value which is taken to mean "there is not yet a supported state"
```
andrewgordstewart, line 6:
```suggestion
	MaxTurnNum = ^uint64(0) // MaxTurnNum is a reserved value which is taken to mean "there is not yet a supported state"
```

andrewgordstewart (approved):
This looks much nicer to me!

Should we start a short style guide to capture our intended conventions? In this case, we might not need one, since we can defer to the (official ?) [style guide](https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps).

geoknee:
> Should we start a short style guide to capture our intended conventions? In this case, we might not need one, since we can defer to the (official ?) [style guide](https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps).

Let's just add a link to that style guide and https://go.dev/doc/effective_go to the contributing section of the readme? I think for now we are mostly concerned with having _some_ conventions in place and I don't think there is any desire to depart from established ones.

PR #157, opened 2022-01-19
rename `Transaction` -> `ChainTransaction`

geoknee:
Disambiguates from "database transaction" which may appear in the codebase soon.

An alternative which I briefly considered: move `Transaction` into the `chainservice` package. Then, it would be naturally namespaced `chainservice.Transaction`, when imported in the usual manner. I decided against this because the `Transaction` type is designed to be instantiated by a `protocol` , which is the functional core of our design. It seems wrong to have the `protocols` package depend on the `chainservice` package (which it only talks to via the `engine` anyway). We are better off letting the `protocols` define their own types, and converting them if necessary as they reach the more imperative "outer" parts of the architecture.

### DIFF
--- a/client/engine/chainservice/chainservice.go
+++ b/client/engine/chainservice/chainservice.go
@@ -15,5 +15,5 @@ type Event struct {
 
 type ChainService interface {
 	Out() <-chan Event
-	In() chan<- protocols.Transaction
+	In() chan<- protocols.ChainTransaction
 }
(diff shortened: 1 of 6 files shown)


NiloCK (approved):
LGTM. I hadn't thought of moving it inside of `chainservice`, but I think the argument against is fine.

This also disambiguates against, ie, financial transactions.

Issue #158, opened 2022-01-19
run tests with `-count>1`


geoknee:
Motivation 1: Our tests run very fast -- a few seconds at most. I think we could easily live with e.g. doubling that time if it gave us better confidence in our code.

Motivation 2: Running with `-count=2` locally gives me some failures when I run the tests locally. We should get to the bottom of why that is (it is unexpected and probably points to a bad test design). 

There are  other flags which might be useful to add, such as `-shuffle`  and `-race`. See here https://pkg.go.dev/cmd/go/internal/test for more information.

geoknee:
`-count=1` is the idiomatic way of disabling test caching.

PR #159, opened 2022-01-19
Add `TestNew` for `client` package

geoknee:
This minimal test acts as a check for panics in the constructor of our client, as well as enforcing that our test services actually implement the required interfaces. This PR includes a fix for that (they do not currently implement the required interfaces on `main`). 

⚠️ I have noticed that introducing this tests causes a flicker in the `chainservice.TestDeposit` test. We will track that on #158 .


NiloCK (approved):
:+1: but should the mockstore be passed in as well?

geoknee:
> 👍 but should the mockstore be passed in as well?

I suppose it should! Will get on it

PR #160, opened 2022-01-19
make the store an injected dependency

geoknee:
This will make it easy to swap out a mocked / test / memstore for a real one. 

TODO 
- [ ] update readme to reflect this change

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -6,6 +6,7 @@ import (
 	"github.com/statechannels/go-nitro/client/engine"
 	"github.com/statechannels/go-nitro/client/engine/chainservice"
 	"github.com/statechannels/go-nitro/client/engine/messageservice"
+	"github.com/statechannels/go-nitro/client/engine/store"
 )
 
 // Client provides the interface for the consuming application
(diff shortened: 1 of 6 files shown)


client/client.go
```diff
@@ -6,17 +6,18 @@ import (
...
 )
 
 // Client provides the interface for the consuming application
 type Client struct {
 	engine engine.Engine // The core business logic of the client
 }
 
-// New is the constructor for a Client. It accepts a messaging service and a chain service as injected dependencies.
-func New(messageService messageservice.MessageService, chainservice chainservice.ChainService) Client {
+// New is the constructor for a Client. It accepts a messaging service, achain service anda a store as injected dependencies.
```
NiloCK, line 17:
```suggestion
// New is the constructor for a Client. It accepts a messaging service, a chain service, and a a store as injected dependencies.
```

client/client_test.go
```diff
@@ -6,14 +6,18 @@ import (
...
 	"github.com/statechannels/go-nitro/client/engine/chainservice"
 	"github.com/statechannels/go-nitro/client/engine/messageservice"
+	"github.com/statechannels/go-nitro/client/engine/store"
 	"github.com/statechannels/go-nitro/types"
 )
 
 func TestNew(t *testing.T) {
-	a := types.Address(common.HexToAddress(`a`))
+
+	pkA := common.Hex2Bytes("c417e8a75ebe2bfe16fe108e1e04802c324974eef6ea2cc3d55194fa38677b5e")
```
NiloCK, line 15:
```suggestion
	skA := common.Hex2Bytes("c417e8a75ebe2bfe16fe108e1e04802c324974eef6ea2cc3d55194fa38677b5e")
```

NiloCK: `private` / `public` are both p. `secret` / `public` is probably better when abbreviating.

NiloCK (approved):
:+1:

PR #164, opened 2022-01-21
PreApprove objectives when spawned via the API

geoknee:
Currently:
* objectives cannot progress until they are approved
* objectives cannot be approved by an API call (it is unimplemented thus far)

This PR unblocks protocol execution by having objectives automatically approved when they are constructed via an API call (i.e. from the consuming application). The upside is that no second API call is needed to get them approved -- this seems right to me, since it genuinely seems implied by making the API call that the objective is approved. 

The other way objectives will be constructed is by receiving proposals from counterparties. Then, the objective may remain unapproved (or be modified to be unapproved) before some other procedure approves them.


lalexgap (approved):
This also makes sense to me 👍

PR #165, opened 2022-01-21
engine: factor out an `attemptProgress` method

geoknee:
This reduces code repitition and allows the engine to attempt progress immediately when spawning or approving an objective


lalexgap (approved):
Makes sense to me 👍

Issue #167, opened 2022-01-21
Write an integration test for directly funding a channel between two parties


geoknee:
The test will spin up two clients, each with 
* a `MockStore`
* a `TestMessageService` connected to the counterparty
* a `SimpleChainService`

as well as a `MockChain` which each client is connected to. 

The test will call an API method on one of the clients, and assert on some success signal or other in both clients after some time lapse. 

We may need to introduce logging in order to get this test to pass. We may want to assert on the logs as a short term measure only, as this would make the test needlessly brittle. Ideally it probes the public API of each client only.

geoknee:
via #208

PR #169, opened 2022-01-24
Generate secret keys in integration test and pass into store(s) before injecting

geoknee:
Towards #167 

Using injected dependencies in the `go-nitro` client is great for easily swapping out mock dependencies for production dependencies and testing generally. But it poses some issues with respect to having all of those dependencies wired up and sharing information.

In the medium term, we desire a nice developer experience so may want to support constructing a `go-nitro` client *without* needing to inject dependencies. In that case, they can be constructed inside the constructor of the `go-nitro` client, which can also share the secret key and address information between those components. 

In the short term, we need to write integration tests so will continue with the injected dependency approach. All components of the client (message, chain and store) need to know the `address` of the client, and the store needs to know the secret key. The solution of this PR is to construct both in the test script (which plays the role of a consuming application, roughly), and pass into all dependencies prior to injection. The `client` itself can query this information from the store when it needs to.

### DIFF
--- a/crypto/keys.go
+++ b/crypto/keys.go
@@ -1,0 +1,25 @@
+package crypto
+
+import (
+	"crypto/ecdsa"
+	"log"
+
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/statechannels/go-nitro/types"
+)
+
+func GeneratePrivateKeyAndAddress() (types.Bytes, types.Address) {
+	channelSecretKey, err := crypto.GenerateKey()
+	if err != nil {
+		log.Fatal(err)
+	}
+	channelSecretKeyBytes := crypto.FromECDSA(channelSecretKey)
+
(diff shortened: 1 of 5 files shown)


crypto/keys.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"crypto/ecdsa"
+	"log"
+
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/statechannels/go-nitro/types"
+)
+
+func GeneratePrivateKeyAndAddress() (types.Bytes, types.Address) {
```
lalexgap, line 11: This might be overkill but 😆 
```suggestion
// GeneratePrivateKeyAndAddress generates a pseudo-random private key and it's corresponding address
func GeneratePrivateKeyAndAddress() (types.Bytes, types.Address) {
```

lalexgap (approved):
👍

PR #171, opened 2022-01-24
Check that address is a participant when initiating DirectFundObjective

### LINKED ISSUE #73: What happens if we construct a direct funding objective with a non-participant `myAddress`?
There are a few issues with this code:
https://github.com/statechannels/go-nitro/blob/28f23f933c6e80aff3255dcafc18e770c1d6ffaa/protocols/direct-funding.go#L75-L79

Firstly, we do not currently return an error when `myAddress` is not a participant. 

Secondly, the *default* value of this variable has the **incorrect** meaning (it's not "useful out of the box"). If we do not re-set it, we might proceed with `myIndex = 0`, which seems quite dangerous.

lalexgap:
Adds a simple check to make sure that `myAddress` is a participant in `initialState`. Otherwise returns an error.

Fixes #73

### DIFF
--- a/protocols/direct-fund/direct-fund.go
+++ b/protocols/direct-fund/direct-fund.go
@@ -59,10 +59,16 @@ func New(
 	}
 
 	var myIndex uint
+	foundMyAddress := false
 	for i, v := range initialState.Participants {
 		if v == myAddress {
 			myIndex = uint(i)
+			foundMyAddress = true
+			break
 		}
+	}
+	if !foundMyAddress {
+		return DirectFundObjective{}, errors.New("my address not found in participants")
 	}
 
 	init.C = &channel.Channel{}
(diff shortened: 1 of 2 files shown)


geoknee (approved):
Looks good to me! Can you explain the changes to `go.mod` and `go.sum`? I'm not sure they belong here, but I am also not sure if we need to tighten up our CI checks and/or advice to developers about maintaining these files.

lalexgap:
> Looks good to me! Can you explain the changes to `go.mod` and `go.sum`? I'm not sure they belong here, but I am also not sure if we need to tighten up our CI checks and/or advice to developers about maintaining these files.

Hmm I'm not sure how I ended up with a modified `go.mod/go.sum`. I remember seeing a couple changes in `go.mod` then running `go mod tidy` and seeing less. I did have a `go-nitro` repo already so perhaps the `go.mod` changes were from the past?

I reverted the changes to `go.mod/go.sum` and ran `go mod tidy` and I'm now not getting any changes to `go.mod/go.sum` 👍

NiloCK:
@lalexgap I was previously having an issue with modified `go.mod` and `go.sum` files. It was caused by different versions of the `go` compiler existing in different places on my system.

PR #176, opened 2022-01-25
Add basic pre-commit hooks

### LINKED ISSUE #148: Run checks in pre-commit git hook
Currently, [contributing.md](https://github.com/statechannels/go-nitro/blob/main/contributing.md) includes two instructions:
1. format your code using `gofmt -w .`
2. lint your code using `golangci-lint run`

Making these the dev's responsibility
1. adds mental overhead to ones development flow
2. prevents the dev from learning about mistakes until pushed to GH

These checks could run in a pre-commit hook, reducing mental overhead, speeding up time-to-learn-issues and cleaning up the commit history.

(2) leads to a noisier commit history. For instance, see c2b2d64 and 75981e9 from #146 and 272f1b9 from #147.

lalexgap:
Added some basic [pre-commit](https://pre-commit.com/) hooks to run:
- `go-fmt`
- `golangci-lint`
- `go-mod-tidy`

Fixes #148

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,6 @@
+repos:
 - repo: https://github.com/dnephin/pre-commit-golang
-  rev: master
+  rev: v0.4.0
   hooks:
     - id: go-fmt
     - id: golangci-lint
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,6 @@
+- repo: https://github.com/dnephin/pre-commit-golang
+  rev: master
+  hooks:
+    - id: go-fmt
+    - id: golangci-lint
+    - id: go-mod-tidy
--- a/contributing.md
+++ b/contributing.md
@@ -21,6 +21,23 @@ echo $GOPATH
 
 You should see `$HOME/go`.
 
+## Pre-commit Hooks
+
+There are pre-commit hooks to run `gofmt`, `golangci-lint` and `go mod tidy`. To get these to run `pre-commit` must be installed.
+
+### Install pre-commit
+
+```shell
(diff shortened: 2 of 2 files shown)


lalexgap:
I opened this before realizing there was more discussion in #148. Happy to close this if it doesn't quite fit our needs.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,7 @@
+repos:
+- repo: https://github.com/dnephin/pre-commit-golang
+  rev: v0.4.0
+  hooks:
+    - id: go-fmt
+    - id: golangci-lint
```
geoknee, line 6: I think there may be some redundancy here, since `go-fmt` is run as a part of `golangci-lint`. We should make sure our current developer guides reflect this.

geoknee:
This solution seems OK, but:
* it's maybe a bit heavy weight? Asking devs to install an entire program (e.g. via home brew). Maybe  more lightweight: have a shell script running all of our checks (I think there are actually only 2 checks at present, 3 if you count testing).
* I think I prefer the push hook rather than the commit hook.


I suppose we can make the whole thing more-obviously "opt in", which might help.

lalexgap:
> This solution seems OK, but:
> 
> * it's maybe a bit heavy weight? Asking devs to install an entire program (e.g. via home brew). Maybe  more lightweight: have a shell script running all of our checks (I think there are actually only 2 checks at present, 3 if you count testing).
> * I think I prefer the push hook rather than the commit hook.
> 
> I suppose we can make the whole thing more-obviously "opt in", which might help.

I think my preference is to close this for now. I kinda jumped the gun on this as I missed a bunch of discussion about what the team actually wants.

PR #177, opened 2022-01-25
SignedState signature validation and getter functions

### LINKED ISSUE #105: Vet public/private property of Channel and SignedState structs
Similar to above - a container struct like this maybe needs, and almost definitely wants, private data and public setter that vets signatures.

_Originally posted by @NiloCK in https://github.com/statechannels/go-nitro/pull/102#r767693365_

lalexgap:
Fixes #105 

Moves responsibility of validating signatures from `Channel` to `SignedState`.

I'm still getting used to `go` idioms so please point out anything that's not quite right!

# Changes

## Signed State now includes the full state.

Previously we were only storing the `VariablePart` of a state in SignedState. I've switched this to the full state so we can call `s.RecoverSigner`  in `signedstate.go`. This does mean we're now storing FixedPart for every state which is less efficient than only storing it in the channel.

## New addSignature and newSignedState function
I've introduced an `addSignature` function  to `signedState` so signatures can be added to a signedState. This function validates the signature by checking that is signed by a participant. The `newSignedState` is just syntax sugar for initiating a new `signedState` struct.

## Channel no longer checks signatures
Since signatures are now checked in addSignature we no longer need to perform the check in `channel.go` 

## Modified Visibility
Based on the idea that signedState should be read only for external consumers of `channel` I've modified some visibilities:
- signedState is private to the channel package.  External consumers shouldn't be instantiating their own `signedState.
- The state and signatures variables on signedState are now private. These are exposed by the `State()` getter function and the `HasSignature`  helper.
- The new `addSignature` and `newSignedState` function is private as it should only be called by the parent `channel`.

### DIFF
--- a/channel/signedstate.go
+++ b/channel/signedstate.go
@@ -46,8 +46,8 @@ func (ss signedState) State() state.State {
 	return ss.state
 }
 
-// HasSignature returns true if the participant (at participantIndex) has a valid signature.
-func (ss signedState) HasSignature(participantIndex uint) bool {
+// HasSignatureForParticipant returns true if the participant (at participantIndex) has a valid signature.
+func (ss signedState) HasSignatureForParticipant(participantIndex uint) bool {
 	_, found := ss.sigs[uint(participantIndex)]
 	return found
 }
--- a/channel/signedstate.go
+++ b/channel/signedstate.go
@@ -7,7 +7,7 @@ import (
 )
 
 type signedState struct {
-	State state.State
+	state state.State
 	sigs  map[uint]state.Signature // keyed by participant index
 }
 
@@ -20,12 +20,12 @@ func newSignedState(s state.State) signedState {
 // addSignature adds a participant's signature for the state.
 // An error is thrown if the signature is invalid.
 func (ss signedState) addSignature(sig state.Signature) error {
-	signer, err := ss.State.RecoverSigner(sig)
+	signer, err := ss.state.RecoverSigner(sig)
 	if err != nil {
 		return err
 	}
 
-	for i, p := range ss.State.Participants {
+	for i, p := range ss.state.Participants {
 		if p == signer {
 			_, found := ss.sigs[uint(i)]
 			if found {
@@ -41,6 +41,9 @@ func (ss signedState) addSignature(sig state.Signature) error {
 	return errors.New("signature does not match any participant")
 
 }
+func (ss signedState) State() state.State {
+	return ss.state
+}
 
 // HasSignature returns true if the participant (at participantIndex) has a valid signature.
 func (ss signedState) HasSignature(participantIndex uint) bool {
@@ -51,7 +54,7 @@ func (ss signedState) HasSignature(participantIndex uint) bool {
 // HasAllSignatures returns true if every participant has a valid signature.
 func (ss signedState) HasAllSignatures() bool {
 	// Since signatures are validated
-	if len(ss.sigs) == len(ss.State.Participants) {
+	if len(ss.sigs) == len(ss.state.Participants) {
 		return true
 	} else {
 		return false
--- a/channel/signedstate.go
+++ b/channel/signedstate.go
@@ -6,21 +6,20 @@ import (
 	"github.com/statechannels/go-nitro/channel/state"
 )
 
-type SignedState struct {
+type signedState struct {
 	State state.State
 	sigs  map[uint]state.Signature // keyed by participant index
 }
 
-// NewSignedState creates a signed state for the given state with 0 signatures.
-func NewSignedState(s state.State) SignedState {
-	ss := SignedState{s, make(map[uint]state.Signature, len(s.Participants))}
-
-	return ss
+// newSignedState initializes a SignedState struct for the given state.
+// The signedState returned will have no signatures.
+func newSignedState(s state.State) signedState {
+	return signedState{s, make(map[uint]state.Signature, len(s.Participants))}
 }
 
-// AddSignature adds a participant's signature for the state.
+// addSignature adds a participant's signature for the state.
 // An error is thrown if the signature is invalid.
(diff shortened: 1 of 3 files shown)


channel/signedstate.go
```diff
@@ -1,17 +1,60 @@
...
 }
 
-// hasAllSignatures returns true if there are numParticipants distinct signatures on the state and false otherwise.
-func (ss SignedState) hasAllSignatures(numParticipants int) bool {
-	if len(ss.Sigs) == numParticipants {
+// HasAllSignatures returns true if every participant has a valid signature.
+func (ss signedState) HasAllSignatures() bool {
+	// Since signatures are validated
+	if len(ss.sigs) == len(ss.state.Participants) {
 		return true
```
geoknee, line 58: 👍

channel/signedstate.go
```diff
@@ -1,17 +1,60 @@
...
 	"github.com/statechannels/go-nitro/channel/state"
 )
 
-type SignedState struct {
-	State state.VariablePart
-	Sigs  map[uint]state.Signature // keyed by participant index
+type signedState struct {
+	state state.State
+	sigs  map[uint]state.Signature // keyed by participant index
+}
```
geoknee, line 12: I understand the motivation to make these data private, but I also spot that you have made the type a non-exported type -- so it is not available outside of the `channel` package. Was that intentional?

channel/signedstate.go
```diff
@@ -1,17 +1,60 @@
...
+	}
+	return errors.New("signature does not match any participant")
+
+}
+func (ss signedState) State() state.State {
+	return ss.state
+}
+
+// HasSignature returns true if the participant (at participantIndex) has a valid signature.
+func (ss signedState) HasSignature(participantIndex uint) bool {
```
geoknee, line 49: How about naming this `HasSignatureByParticipantWithIndex`, `HasSignatureByParticipant`, or even just`HasSignatureBy` ? Any of these feels a bit better, to me.

channel/signedstate.go
```diff
@@ -1,17 +1,60 @@
...
+type signedState struct {
+	state state.State
+	sigs  map[uint]state.Signature // keyed by participant index
+}
+
+// newSignedState initializes a SignedState struct for the given state.
+// The signedState returned will have no signatures.
+func newSignedState(s state.State) signedState {
+	return signedState{s, make(map[uint]state.Signature, len(s.Participants))}
+}
```
geoknee, line 18: Strongly approve of use of constructor -- I will echo my question above about whether this should be exported or not (capitalised first letter).

channel/signedstate.go
geoknee: I think probably this type wants to live in the `state` package and be exported by that package for use here? Otherwise the public / private distinction doesn't make a heap of sense.

channel/signedstate.go
```diff
@@ -1,17 +1,60 @@
...
+
+// addSignature adds a participant's signature for the state.
+// An error is thrown if the signature is invalid.
+func (ss signedState) addSignature(sig state.Signature) error {
+	signer, err := ss.state.RecoverSigner(sig)
+	if err != nil {
+		return err
```
geoknee, line 25: This is potentially a good candidate for error wrapping, like we do here: 

https://github.com/statechannels/go-nitro/blob/b70f16695f8baf807e5338b132b55b861b5278d9/channel/state/state.go#L101

geoknee: https://go.dev/blog/go1.13-errors

geoknee (approved):
Looks good but I think there are a few tweaks to make.

channel/signedstate.go
lalexgap: > I understand the motivation to make these data private, but I also spot that you have made the type a non-exported type -- so it is not available outside of the channel package. Was that intentional?

Yep although perhaps that's not the correct approach! By making `signedState` a private struct I was trying to prevent it from being instantiated or created outside of the `channel` package. Code external to the `channel` package can still call `State()` or `HasSignatures` but cannot create their own `signedState`
image

My assumption was that `signedState` is a helper structure within `channel`. The `channel` package would be the only one creating `signedState`s and external code would only be reading from the `signedState`, never modifying them.  To follow that pattern maybe there should be a signed state interface that is returned by `SignedStateForTurnNum` ?

It also depends on what the role of `SignedState` is. We could move it to a the `state` package and start accepting it in `func (c *Channel) AddSignedState` (instead of just a state and signature).

channel/signedstate.go
lalexgap: I tried to touch on this above but my goal was that only the `channel` package can create a new `signedState`. Any other packages can only read data from `signedState` (when it's exposed on the channel via `SignedStateForTurnNum`)

channel/signedstate.go
lalexgap: Good idea! Updated in 5992af9

channel/signedstate.go
lalexgap: I've renamed this to `HasSignatureForParticipant` in a1a50ad

channel/signedstate.go
geoknee: I see! Let's leave it as is for now, we can always circle back to it if we want to later on.

geoknee:
> * The state and signatures variables on signedState are now private. These are exposed by the `State()` getter function and the `HasSignature`  helper.

This is well motivated -- but, I think the signed state struct is going to be widely useful, including in places where we need to serialize it (for storage and for transmission). Having private members makes that more difficult, but perhaps not impossible. 

I think one solution is to have the type implement the `Marshaler` interface:
```go
type Marshaler interface {
        MarshalJSON() ([]byte, error)
}
```

and then if the type is found by calling `json.Marshall` on a parent type, it will call into the method above.

PR #178, opened 2022-01-26
Run tests with count=2

### LINKED ISSUE #158: run tests with `-count>1`

lalexgap:
Fixes #158 

Two changes:
- Resolves the `direct-fund_test` failures when running with `count=2`.  The tests were using an objective `s` that would not get reset between test runs. To address this `s` is now instantiated in each test function.
- Updates the CI config to run with `-count=2` and `-shuffle=on`

In this particular case there wasn't much code duplication to instantiate `s` in both tests. However longer term we might want to explore a `beforeEach` pattern?

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -18,7 +18,7 @@ jobs:
         run: go build -v ./...
 
       - name: Test
-        run: go test -v ./... -count=2
+        run: go test -v ./... -count=2 -shuffle=on
 
       - name: Tidy
         run: go mod tidy
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -18,7 +18,7 @@ jobs:
         run: go build -v ./...
 
       - name: Test
-        run: go test -v ./...
+        run: go test -v ./... -count=2
 
       - name: Tidy
         run: go mod tidy
--- a/protocols/direct-fund/direct-fund_test.go
+++ b/protocols/direct-fund/direct-fund_test.go
@@ -145,6 +145,11 @@ func TestUpdate(t *testing.T) {
 
 }
 
+// TestDoubleFailure is an easy way to debug the count=2 case.
+func TestDoubleFailure(t *testing.T) {
+	TestCrank(t)
+	TestCrank(t)
+}
 func TestCrank(t *testing.T) {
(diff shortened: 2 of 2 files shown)


lalexgap:
#158 also references a flicker in `chainservice.TestDeposit` that looks unrelated to `count=2`. I think we should create a seperate issue to track this (if there isn't one already).

geoknee (approved):
👏  nice work tracking this down.

To check my understanding of the underlying issue:

Objective `Updates` and `Cranks` are *pure functions*, so I think we would not expect to see the problem in this test file if we only used those methods. 

However, this test reaches into the `Channel` **pointed to** by the `DirectFundObjective` and calls a method (`.AddSignedState`) which mutates that `Channel`. Because we bypass the `Update/Crank` API (only possible because of a type assertion in this test), we lose the guarantee of a pure function.

Your solution here works because each test has its own `DirectFundObjective`. I do wonder, though, if we can add some annotations of adopt a pattern which makes these subtleties clearer in this test file?

lalexgap:
> Your solution here works because each test has its own `DirectFundObjective`. I do wonder, though, if we can add some annotations of adopt a pattern which makes these subtleties clearer in this test file? After all, one major benefit / motivation for having pure functions is the ease of testing and avoiding problems such as this.

This is a really good point! I think this is also slightly related to this [discussion](https://github.com/statechannels/go-nitro/issues/53#issuecomment-1023129252) on #53. If cranking an objective signs a state and stores a signature then crank will no longer be pure and we'll likely run into this same issue.

geoknee:
> Although maybe that is not an issue as we `clone` the channel before modifying it?

Yes exactly. I think in fact if we clone the objective before modifying it, that should also clone the channel and we'll be back to having a pure function.

PR #181, opened 2022-01-26
Fix TestDeposit Flicker

### LINKED ISSUE #180: chainService.TestDeposit occasionally flickers.
#158 also references a flicker in `chainservice.TestDeposit` that looks unrelated to `count=2`. I think we should create a seperate issue to track this (if there isn't one already).

_Originally posted by @lalexgap in https://github.com/statechannels/go-nitro/issues/178#issuecomment-1022465800_

lalexgap:
Fixes #180 

Caveat: I'm still new to go routines and channels so this might not be the correct solution.

Previously in `MockChain` we were using a go routine to send events to to chain services via a channel. By using go routines we avoided blocking on waiting for the chain service to read from the channel.

However using go routines means that events may not get sent when we expect them, since they're parallel threads of execution. (I'm a little fuzzy on the exact details of how go routines are scheduled/run)

To avoid this I've just switched the `HandleTX` to not use go routines and instead write out to a **buffered channel** instead. Since the channel is buffered `HandleTX` will not block on writing the event to the channel. I've chosen a buffer size of 10 arbitrarily. 

To verify this locally I've been running the command:
```
go test  ./client/engine/chainservice/ -count=10000 -shuffle=on
```

On `main` this fails for me but works on this branch.

### DIFF
--- a/client/engine/chainservice/mockchain.go
+++ b/client/engine/chainservice/mockchain.go
@@ -35,7 +35,8 @@ func NewMockChain(addresses []types.Address) MockChain {
 	mc.holdings = make(map[types.Destination]types.Funds)
 
 	for _, a := range addresses {
-		mc.out[a] = make(chan Event)
+		// Use a buffered channel so we don't have to worry about blocking on writing to the channel.
+		mc.out[a] = make(chan Event, 10)
 	}
 
 	go mc.Run()
@@ -60,7 +61,7 @@ func (mc MockChain) handleTx(tx protocols.ChainTransaction) {
 		AdjudicationStatus: protocols.AdjudicationStatus{TurnNumRecord: 0},
 	}
 	for _, out := range mc.out {
-		go sendEvent(out, event) // we use a goroutine for each send to prevent being blocked by one bad listener
+		sendEvent(out, event)
 	}
 
 }


geoknee (approved):
Our unit tests *require* that these events arrive in order, which as you say having goroutines was playing havoc with. I am wondering if the event should include a block height, so that listeners can receive events in any order and reorder them themselves?

For now your solution looks pretty good -- only I think we may have lost protection against "bad listeners". See #184 for my suggested fix!

lalexgap:
> For now your solution looks pretty good -- only I think we may have lost protection against "bad listeners". See #184 for my suggested fix!

Ooh neat! Since `mockchain` is used for testing I'm wondering if we want to protect against "bad listeners"? If a listener is not reading from the channel like we expect it might be nice to catch that early on?

geoknee:
> Since `mockchain` is used for testing I'm wondering if we want to protect against "bad listeners"? If a listener is not reading from the channel like we expect it might be nice to catch that early on?

Interesting. I'm not really sure at this point. I kind of feel like the `mockchain` ought not to couple the behaviour of several clients together that much? And we could have other checks in place to catch a bad listener.

PR #182, opened 2022-01-27
Add example of channel usage

chumaknadya:


geoknee:
@chumaknadya thanks so much for this contribution! 

I'm really happy to see this example of a directly funded channel between two brokers, along with an interface to our on-chain code from the `nitro-protocol` npm package. 

I can see a few TODOs in this changeset, and I think it would be good for someone on our team to provide some more information and context to help clear those up. 

We plan to schedule an introductory meeting to discuss how this kind of work can fit into our roadmap -- we will be in touch!

PR #184, opened 2022-01-27
use attemptSend function in mockchain

geoknee:
See commit message.

### DIFF
--- a/client/engine/chainservice/mockchain.go
+++ b/client/engine/chainservice/mockchain.go
@@ -61,12 +61,15 @@ func (mc MockChain) handleTx(tx protocols.ChainTransaction) {
 		AdjudicationStatus: protocols.AdjudicationStatus{TurnNumRecord: 0},
 	}
 	for _, out := range mc.out {
-		sendEvent(out, event)
+		attemptSend(out, event)
 	}
 
 }
 
-// sendEvent sends event to the supplied chan
-func sendEvent(out chan Event, event Event) {
-	out <- event
+// attemptSend sends event to the supplied chan, and drops it if the chan is full
+func attemptSend(out chan Event, event Event) {
+	select {
+	case out <- event:
+	default:
+	}
 }


client/engine/chainservice/mockchain.go
```diff
@@ -61,12 +61,15 @@ func (mc MockChain) handleTx(tx protocols.ChainTransaction) {
...
 
-// sendEvent sends event to the supplied chan
-func sendEvent(out chan Event, event Event) {
-	out <- event
+// attemptSend sends event to the supplied chan, and drops it if the chan is full
+func attemptSend(out chan Event, event Event) {
+	select {
+	case out <- event:
+	default:
+	}
```
andrewgordstewart, line 74: Can you explain this for a new gopher?

Is this understanding correct?
1. if `out` receives `event`, do nothing
2. else, do nothing

If `out`'s buffer is full, and it refuses to receive `event`, does `out <- event` evaluate to a "falsy" value or something?

geoknee: Your understanding is pretty much correct. With the go `select` statement, we try each case in turn until one succeeds. So if we can send `event` down the `out` chan, we're done and exit the statement. Otherwise, we revert to the default case, which really does nothing and exits the statement. 

I guess you can think of it as `out <- event` being falsy -- my mental model is that this particular pipe is blocked so you can't flow down it.

geoknee: Edit: "in turn" isn't quite right. 

> A select blocks until one of its cases can run, then it executes that case. It chooses one at random if multiple are ready.

https://go.dev/tour/concurrency/5

client/engine/chainservice/mockchain.go
```diff
@@ -61,12 +61,15 @@ func (mc MockChain) handleTx(tx protocols.ChainTransaction) {
...
 }
 
-// sendEvent sends event to the supplied chan
-func sendEvent(out chan Event, event Event) {
-	out <- event
+// attemptSend sends event to the supplied chan, and drops it if the chan is full
+func attemptSend(out chan Event, event Event) {
+	select {
+	case out <- event:
+	default:
```
NiloCK, line 73: According to my understanding, we hit this `default` case here only when we have badly misconfigured some integration test. Do we want to:

- return an error
- panic

I'm loosely in favour of panic here, since it doesn't complicate the usage of `attemptSend`, and provides the fast feedback that things are wrong in the test setup.

NiloCK (approved):
:+1:

client/engine/chainservice/mockchain.go
geoknee: I anticipate we may do tests in the future where we manually "crash" a client, which would trigger the default case here. So I'm in two minds about whether we should panic or not. 

My feeling is that the code as-is is pretty close to a genuine "broadcast" or "fire and forget" pattern akin to a genuine blockchain, whereas putting in a panic would push us back towards a pattern where "messages being received properly is a _requirement_". To me the latter pattern is characteristic of communicating with `chans`, and not really something that fits the behaviour of what we are trying to mock, here.

Most likely this isn't of huge importance right now and we can amend it in future when we have more context?

NiloCK: That sounds good. In practice a recovering client will have the ability to 'look up' past messages it has missed via a (mock) chain explorer.

PR #188, opened 2022-01-30
Return SideEffects in Direct Funding Protocol

### LINKED ISSUE #53: Return `SideEffects` from `DirectFundObjective.Crank()`

lalexgap:
Fixes #53

The direct funding protocol now sends returns appropriate side effects when Crank is called. There are three scenarios where side effects will be returned:
- If the prefund setup isn't signed by the user then it will be signed, stored and a message containing that signature will be returned as a side effect.
- If it's safe to deposit then a deposit `ChainTransaction` is returned in the side effects
- If it's safe to sign the post fund setup then it will be signed, stored and a message containing that signature will be returned as a side effect.

Tests have been added to check the side effects in each scenario.

## Changes
- [fc8aaf9](https://github.com/statechannels/go-nitro/pull/188/commits/fc8aaf928e0db28295a9e73adee643c5259835b9)  updates `amountToDeposit` to iterate over deposit targets. Otherwise if there are no holdings on chain we would never return a deposit amount.
- [ca44d2f](https://github.com/statechannels/go-nitro/pull/188/commits/ca44d2fe79156f638f82567dc0edf6d1d0a9b604) cleans up some of the logic to send out deposit transactions

### DIFF
--- a/channel/channel.go
+++ b/channel/channel.go
@@ -122,7 +122,7 @@ func (c Channel) PreFundSignedByMe() bool {
 // PostFundSignedByMe() returns true if I have signed the post fund setup state, false otherwise.
 func (c Channel) PostFundSignedByMe() bool {
 	if _, ok := c.SignedStateForTurnNum[PostFundTurnNum]; ok {
-		if c.SignedStateForTurnNum[PreFundTurnNum].HasSignatureForParticipant(c.MyIndex) {
+		if c.SignedStateForTurnNum[PostFundTurnNum].HasSignatureForParticipant(c.MyIndex) {
 			return true
 		}
 	}
--- a/protocols/direct-fund/direct-fund.go
+++ b/protocols/direct-fund/direct-fund.go
@@ -134,6 +134,7 @@ func (s DirectFundObjective) Update(event protocols.ObjectiveEvent) (protocols.O
 func (s DirectFundObjective) Crank(secretKey *[]byte) (protocols.Objective, protocols.SideEffects, protocols.WaitingFor, error) {
 	updated := s.clone()
 
+	se := NoSideEffects
 	// Input validation
 	if updated.Status != protocols.Approved {
 		return updated, NoSideEffects, WaitingForNothing, ErrNotApproved
@@ -141,8 +142,34 @@ func (s DirectFundObjective) Crank(secretKey *[]byte) (protocols.Objective, prot
 
 	// Prefunding
 	if !updated.C.PreFundSignedByMe() {
-		// todo: {SignAndSendPreFundEffect(updated.ChannelId)} as SideEffects{}
-		return updated, NoSideEffects, WaitingForCompletePrefund, nil
+		pf := s.C.PreFundState()
+		sig, err := pf.Sign(*secretKey)
+		if err != nil {
+			return updated, NoSideEffects, WaitingForCompletePrefund, fmt.Errorf("failed to sign pre-fund state: %w", err)
+		}
+		ss := state.NewSignedState(pf)
+		if err := ss.AddSignature(sig); err != nil {
+			return updated, NoSideEffects, WaitingForCompletePrefund, fmt.Errorf("failed to add signature to signed state: %w", err)
+		}
+		ok := updated.C.AddSignedState(ss)
+		if !ok {
+			return updated, NoSideEffects, WaitingForCompletePrefund, fmt.Errorf("failed to add signed state")
+		}
+
+		messages := make([]protocols.Message, 0)
+		for i, participant := range ss.State().Participants {
+
+			// Do not generate a message for ourselves
+			if uint(i) == s.C.MyIndex {
+				continue
+			}
+			message := protocols.Message{To: participant, ObjectiveId: s.Id(), SignedStates: []state.SignedState{ss}, Proposal: nil}
+			messages = append(messages, message)
+		}
+
+		se.MessagesToSend = append(se.MessagesToSend, messages...)
+
+		return updated, se, WaitingForCompletePrefund, nil
 	}
 
 	if !updated.C.PreFundComplete() {
--- a/protocols/direct-fund/direct-fund.go
+++ b/protocols/direct-fund/direct-fund.go
@@ -143,30 +143,11 @@ func (s DirectFundObjective) Crank(secretKey *[]byte) (protocols.Objective, prot
 	// Prefunding
 	if !updated.C.PreFundSignedByMe() {
 		pf := s.C.PreFundState()
-		sig, err := pf.Sign(*secretKey)
+		ss, err := s.signAndStore(pf, secretKey)
 		if err != nil {
-			return updated, NoSideEffects, WaitingForCompletePrefund, fmt.Errorf("failed to sign pre-fund state: %w", err)
+			return updated, NoSideEffects, WaitingForCompletePrefund, fmt.Errorf("could not sign and store state %w", err)
 		}
-		ss := state.NewSignedState(pf)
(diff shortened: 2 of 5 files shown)


protocols/direct-fund/direct-fund_test.go
```diff
@@ -76,6 +77,44 @@ func TestNew(t *testing.T) {
...
+
+	messageSig, err := signMessage.SignedStates[0].GetParticipantSignature(s.C.MyIndex)
+	if err != nil {
+		t.Error(err)
+	}
+	stateSig, err := s.C.SignedStateForTurnNum[0].GetParticipantSignature(s.C.MyIndex)
+	if err != nil {
+		t.Error(err)
+	}
+	if sigMatch := reflect.DeepEqual(messageSig, stateSig); !sigMatch {
```
geoknee, line 112: Here's an alternative pattern using [`go-cmp`](https://github.com/google/go-cmp), which we are already using:

https://github.com/statechannels/go-nitro/blob/3c4925d201fe18bf5fe768b107565e8b3f4a7c47/channel/channel_test.go#L118-L120

lalexgap: Ah that's much cleaner! I've switched over to that pattern in e5ae411

protocols/direct-fund/direct-fund.go
```diff
@@ -243,17 +242,6 @@ func (s DirectFundObjective) safeToDeposit() bool {
 	return true
 }
 
-// amountToDeposit computes the appropriate amount to deposit given the current recorded OnChainHoldings
-func (s DirectFundObjective) amountToDeposit() types.Funds {
```
lalexgap, line 233: My understanding is that `s.C.OnChainFunding` returns the funding per `asset` for a channel. This means that `holdings` could be due to our deposit or another particpant.

I think this means `big.NewInt(0).Sub(s.myDepositTarget[asset], holding)` won't be correct if other participants have already deposited since `holding` will reflect their deposit.

lalexgap: This was incorrect. We were just not iterating over deposit targets.

channel/channel.go
```diff
@@ -122,7 +122,7 @@ func (c Channel) PreFundSignedByMe() bool {
 // PostFundSignedByMe() returns true if I have signed the post fund setup state, false otherwise.
 func (c Channel) PostFundSignedByMe() bool {
 	if _, ok := c.SignedStateForTurnNum[PostFundTurnNum]; ok {
-		if c.SignedStateForTurnNum[PreFundTurnNum].HasSignatureForParticipant(c.MyIndex) {
+		if c.SignedStateForTurnNum[PostFundTurnNum].HasSignatureForParticipant(c.MyIndex) {
```
geoknee, line 125: Yikes, I feel like we made this typo mistake a few times before! :-)

protocols/direct-fund/direct-fund.go
```diff
@@ -87,7 +87,7 @@ func New(
 		types.AddressToDestination(myAddress),
 	)
 	init.myDepositTarget = init.myDepositSafetyThreshold.Add(myAllocatedAmount)
-
+	fmt.Println(init.myDepositTarget)
```
geoknee, line 90: ✂️

protocols/direct-fund/direct-fund_test.go
```diff
@@ -76,6 +77,60 @@ func TestNew(t *testing.T) {
 	}
 }
 
+func TestPreFundSideEffects(t *testing.T) {
+	// Construct various variables for use in TestUpdate
```
geoknee, line 81:
```suggestion
	// Construct various variables for use in the test.
```

protocols/direct-fund/direct-fund_test.go
```diff
@@ -76,6 +77,60 @@ func TestNew(t *testing.T) {
...
+	var s, _ = New(false, testState, testState.Participants[0])
+
+	// Approve the objective, so that the rest of the test cases can run.
+	o := s.Approve()
```
geoknee, line 85: You can save some lines by passing in `preapprove = true` on L82

protocols/direct-fund/direct-fund.go
```diff
@@ -254,6 +270,37 @@ func (s DirectFundObjective) clone() DirectFundObjective {
...
+		if uint(i) == s.C.MyIndex {
+			continue
+		}
+		message := protocols.Message{To: participant, ObjectiveId: s.Id(), SignedStates: []state.SignedState{ss}, Proposal: nil}
+		messages = append(messages, message)
+	}
+	return messages
+}
+
+func (s DirectFundObjective) signAndStore(toSign state.State, secretKey *[]byte) (state.SignedState, error) {
```
geoknee, line 288: Not affected by this PR, but I think we should rename the receiver variable to `o` for objective? `s` makes me think `state`...

protocols/direct-fund/direct-fund.go
```diff
@@ -254,6 +270,37 @@ func (s DirectFundObjective) clone() DirectFundObjective {
...
+
+	sig, err := toSign.Sign(*secretKey)
+	if err != nil {
+		return state.SignedState{}, fmt.Errorf("failed to sign state: %w", err)
+	}
+	ss := state.NewSignedState(toSign)
+	if err := ss.AddSignature(sig); err != nil {
+		return state.SignedState{}, fmt.Errorf("failed to add signature to signed state: %w", err)
+	}
+	ok := s.C.AddSignedState(ss)
```
geoknee, line 298: I wonder if we could save some lines by using 

https://github.com/statechannels/go-nitro/blob/3c4925d201fe18bf5fe768b107565e8b3f4a7c47/channel/channel.go#L170-L171

geoknee (approved):
🌋

protocols/direct-fund/direct-fund.go
geoknee: #209

protocols/direct-fund/direct-fund.go
geoknee: [2bfe0cb](https://github.com/statechannels/go-nitro/pull/188/commits/2bfe0cba11be27316282c4f5e03595aae4e7b11b)

protocols/direct-fund/direct-fund_test.go
geoknee: [3fd478c](https://github.com/statechannels/go-nitro/pull/188/commits/3fd478cc5a30741d7cb9ea130fafa0d41c3ddf56)

protocols/direct-fund/direct-fund_test.go
geoknee: [2e96214](https://github.com/statechannels/go-nitro/pull/188/commits/2e96214ac89ecdd5f9db280ac346227281ef0b5d)

geoknee:
> * If it's safe to deposit then a deposit `ChainTransaction` is returned in the side effects

We should make an issue to ensure that this action is idempotent. We can fire off a transaction, and then crank again before we receive confirmation of it being mined -- which risks sending another transaction. We can either manage idempotency in the protocol itself, or in the chain service.

PR #192, opened 2022-01-31
Serialize messages in `TestMessageService` and change `protocols.Message` type

geoknee:
Work in progress. 

The existing `protocols.Message` type has a `Sigs` property which is a mapping from a pointer to a state, to a signature. That type is needlessly hard to deal with, particularly with respect to serialising for transmission over the wire. 

 Here I'm replacing it with an array of `SignedStates`, and then having the test message service serialize and deserialize messages to simulate a more realistic situation. 

This should also get around problems deferencing pointers that seem to crop up when sharing pointers between go routines -- that approach seems obviously bad but we should document the problem more clearly anyway (TODO).

### DIFF
--- a/channel/channel.go
+++ b/channel/channel.go
@@ -166,9 +166,18 @@ func (c Channel) Affords(
 	return lss.Outcome.Affords(allocationMap, fundingMap)
 }
 
+// AddSignedStateLegacy is a legacy interface, alternative to but calling AddSignedState internally
+func (c *Channel) AddSignedStateLegacy(s state.State, sig state.Signature) bool {
+	ss := state.NewSignedState(s)
+	ss.AddSignature(sig)
+	return c.AddSignedState(ss)
+}
+
 // AddSignedState adds a signed state to the Channel, updating the LatestSupportedState and Support if appropriate.
 // Returns false and does not alter the channel if the state is "stale", belongs to a different channel, or is signed by a non participant.
-func (c *Channel) AddSignedState(s state.State, sig state.Signature) bool {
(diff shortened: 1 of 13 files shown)


channel/channel.go
```diff
@@ -166,9 +166,21 @@ func (c Channel) Affords(
 	return lss.Outcome.Affords(allocationMap, fundingMap)
 }
 
+// AddStateWithSignature constructs a SignedState from the passes state and signature, and calls s.AddSignedState with it.
```
geoknee, line 169:
```suggestion
// AddStateWithSignature constructs a SignedState from the passed state and signature, and calls s.AddSignedState with it.
```

channel/state/signedstate.go
```diff
@@ -0,0 +1,108 @@
...
+	"errors"
+	"fmt"
+)
+
+type SignedState struct {
+	state State
+	sigs  map[uint]Signature // keyed by participant index
+}
+
+// newSignedState initializes a SignedState struct for the given
```
geoknee, line 15:
```suggestion
// NewSignedState initializes a SignedState struct for the given
```

channel/state/signedstate.go
```diff
@@ -0,0 +1,108 @@
...
+// The signedState returned will have no signatures.
+func NewSignedState(s State) SignedState {
+	return SignedState{s, make(map[uint]Signature, len(s.Participants))}
+}
+
+// addSignature adds a participant's signature for the
```
geoknee, line 21:
```suggestion
// AddSignature adds a participant's signature for the
```

PR #196, opened 2022-02-01
Move `SignedState` type to `state` package, and use in `Message` and `ObjectiveEvent` types

geoknee:
The existing `protocols.Message` and `protocols.ObjectiveEvent` types have a `Sigs` property which is a mapping from a pointer to a state, to a signature. That type is needlessly hard to deal with, particularly with respect to serialising for transmission over the wire.

Here I'm replacing it with an array of `SignedStates`, which is a type I have moved from the `channel` package to the `state` package. The constructor of the type is now exported.

Apologies for the relatively large changeset. Going commit-by-commit ought to help!

This effectively rolls back some (but not all) changes from #177:
> signedState is private to the channel package. External consumers shouldn't be instantiating their own `signedState. 

(no longer true, since other packages and tests are likely to instantiate their own)

> The state and signatures variables on signedState are now private. These are exposed by the State() getter function and the HasSignature helper.

still true

> The new addSignature and newSignedState function is private as it should only be called by the parent channel

no longer true, for the same reasons


channel/state/signedstate.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+}
+
+// Merge checks the passed SignedState's state and the reciever's state for equality, andd adds each signature from the former to the latter.
+func (ss SignedState) Merge(ss2 SignedState) error {
+	if !ss.state.Equal(ss2.state) {
+		return errors.New(`cannot merge signed states with distinct state hashes`)
+	}
+	for _, sig := range ss2.sigs {
+		err := ss.AddSignature(sig)
```
lalexgap, line 72: Will this error out if the signature already exists? It might be nice for `Merge` to handle that case and not error.

channel/state/signedstate.go
```diff
@@ -0,0 +1,94 @@
...
+func NewSignedState(s State) SignedState {
+	return SignedState{s, make(map[uint]Signature, len(s.Participants))}
+}
+
+// AddSignature adds a participant's signature for the
+// An error is thrown if the signature is invalid.
+func (ss SignedState) AddSignature(sig Signature) error {
+	signer, err := ss.state.RecoverSigner(sig)
+	if err != nil {
+		return fmt.Errorf("addSignature failed to recover signer %w", err)
```
lalexgap, line 25:
```suggestion
		return fmt.Errorf("AddSignature failed to recover signer %w", err)
```

channel/state/signedstate.go
```diff
@@ -0,0 +1,94 @@
...
+	for _, sig := range ss2.sigs {
+		err := ss.AddSignature(sig)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// Equal returns true if the passed SignedState is deeply equal in value to the reciever.
```
lalexgap, line 80:
```suggestion
// Equal returns true if the passed SignedState is deeply equal in value to the receiver.
```

lalexgap (approved):
🌋

channel/state/signedstate.go
geoknee: This one always gets me 🤦

Issue #189, opened 2022-01-31
`TestMessageService` should serialize messages


geoknee:
Originally posted by @lalexgap:

> Right now Sigs is indexed by pointer to a state. However I don’t think that makes sense if Message is going to be sent over the wire (since we’d be sending a memory address to state). 

https://github.com/statechannels/go-nitro/blob/8e4b2ee63e8e382ee7dd7c428fa3623a367865d9/protocols/interfaces.go#L12

PR #198, opened 2022-02-01
add JSON (un)marshalling support to `SignedState`

geoknee:
~⚠️ stacked on #196 (base not master)~

This in anticipation of serializing messages and towards #189 .

### DIFF
--- a/channel/state/signedstate.go
+++ b/channel/state/signedstate.go
@@ -96,14 +96,12 @@ func (ss SignedState) Equal(ss2 SignedState) bool {
 
 // MarshalJSON marshals the SignedState into JSON, implementing the Marshaler interface.
 func (ss SignedState) MarshalJSON() ([]byte, error) {
-
-	// Similar type but with public members:
-	type SignedState struct {
+	rr := struct {
 		State State
 		Sigs  map[uint]Signature // keyed by participant index
+	}{
+		ss.state, ss.sigs,
 	}
-
-	rr := SignedState{ss.state, ss.sigs}
 
 	return json.Marshal(rr)
 
@@ -112,13 +110,10 @@ func (ss SignedState) MarshalJSON() ([]byte, error) {
 // UnmarshalJSON unmarshals the passed JSON into a SignedState, implementing the Unmarshaler interface.
 func (ss *SignedState) UnmarshalJSON(j []byte) error {
 
-	// Similar type but with public members:
-	type SignedState struct {
+	rr := struct {
 		State State
 		Sigs  map[uint]Signature // keyed by participant index
-	}
-
-	rr := SignedState{}
+	}{}
 
 	err := json.Unmarshal(j, &rr)
 
--- a/channel/state/signedstate.go
+++ b/channel/state/signedstate.go
@@ -2,6 +2,7 @@ package state
 
 import (
 	"bytes"
+	"encoding/json"
 	"errors"
 	"fmt"
 )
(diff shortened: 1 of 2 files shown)


channel/state/signedstate.go
```diff
@@ -92,3 +93,37 @@ func (ss SignedState) Equal(ss2 SignedState) bool {
...
+
+	rr := SignedState{}
+
+	err := json.Unmarshal(j, &rr)
+
+	ss.state = rr.State
+	ss.sigs = rr.Sigs
+	return err
+
+}
```
geoknee, line 129: This allows us to do `json.Marshal` on `SignedState` (and anything that embeds a `SignedState`) *despite* `SignedState` having private / unexported fields.

channel/state/signedstate.go
```diff
@@ -92,3 +93,37 @@ func (ss SignedState) Equal(ss2 SignedState) bool {
 	}
 	return true
 }
+
+// MarshalJSON marshals the SignedState into JSON, implementing the Marshaler interface.
+func (ss SignedState) MarshalJSON() ([]byte, error) {
+
+	// Similar type but with public members:
+	type SignedState struct {
```
lalexgap, line 101: We could use use an anonymous struct so we don't have to declare a new SignedState struct. Not sure if that would be cleaner or not.

channel/state/signedstate_test.go
```diff
@@ -49,3 +50,36 @@ func TestMerge(t *testing.T) {
 	}
 
 }
+
+func TestJSON(t *testing.T) {
+	ss1 := NewSignedState(TestState)
+	sigA, _ := TestState.Sign(common.Hex2Bytes(`caab404f975b4620747174a75f08d98b4e5a7053b691b41bcfc0d839d48b7634`))
+	_ = ss1.AddSignature(sigA)
+
+	msgString := `{"State":{"ChainId":9001,"Participants":["0xf5a1bb5607c9d079e46d1b3dc33f257d937b43bd","0x760bf27cd45036a6c486802d30b5d90cffbe31fe"],"ChannelNonce":37140676580,"AppDefinition":"0x5e29e5ab8ef33f050c7cc10b5a0456d975c5f88d","ChallengeDuration":60,"AppData":"","Outcome":[{"Asset":"0x0000000000000000000000000000000000000000","Metadata":null,"Allocations":[{"Destination":[0,0,0,0,0,0,0,0,0,0,0,0,245,161,187,86,7,201,208,121,228,109,27,61,195,63,37,125,147,123,67,189],"Amount":5,"AllocationType":0,"Metadata":null},{"Destination":[0,0,0,0,0,0,0,0,0,0,0,0,238,24,255,21,117,5,86,145,0,154,162,70,174,96,129,50,197,122,66,44],"Amount":5,"AllocationType":0,"Metadata":null}]}],"TurnNum":5,"IsFinal":false},"Sigs":{"0":{"R":"cEs6/MbnAhAsoa8/c887N/MAfzaMQOi4HKgjpldAoFM=","S":"FAQK1MWY27BVpQQwFCoTUY4TMLedJO7Yb8vf8aepVYk=","V":0}}}`
```
lalexgap, line 59: Will this have to be updated any time the Signed State struct changes? It probably won't change often so probably won't be a big deal.

channel/state/signedstate.go
geoknee: 13593c3

channel/state/signedstate_test.go
geoknee: Yes. An alternative is to just test that `deserialize(serialize(m)) == m`, but I kind of like seeing the raw string that goes over the wire. It's another chance to catch problems like pointers leaking into messages between clients.

PR #199, opened 2022-02-01
Serialize/Deserialize messages that go "across the wire" in `TestMessageService`

### LINKED ISSUE #189: `TestMessageService` should serialize messages

geoknee:
~⚠️ stacked on #198~

Closes #189

### DIFF
--- a/protocols/messages.go
+++ b/protocols/messages.go
@@ -8,7 +8,7 @@ import (
 	"github.com/statechannels/go-nitro/types"
 )
 
-// Message is an object to be sent across the wire. It can contain a proposal and signed state hashes, and is addressed to a counterparty.
+// Message is an object to be sent across the wire. It can contain a proposal and signed states, and is addressed to a counterparty.
 type Message struct {
 	To           types.Address
 	ObjectiveId  ObjectiveId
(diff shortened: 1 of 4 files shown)


protocols/messages.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"bytes"
+	"encoding/json"
+
+	"github.com/statechannels/go-nitro/channel/state"
+	"github.com/statechannels/go-nitro/types"
+)
+
+// Message is an object to be sent across the wire. It can contain a proposal and signed state hashes, and is addressed to a counterparty.
```
lalexgap, line 11:
```suggestion
// Message is an object to be sent across the wire. It can contain a proposal and signed states, and is addressed to a counterparty.
```

protocols/messages.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+// Serialize serializes the message into a string.
+func (m Message) Serialize() (string, error) {
+	bytes, err := json.Marshal(m)
+	return string(bytes), err
+}
+
+// DeserialiseMessage deserializes the passed string into a protocols.Message.
+func DeserialiseMessage(s string) (Message, error) {
```
lalexgap, line 26:
```suggestion
func DeserializeMessage(s string) (Message, error) {
```

lalexgap (approved):

protocols/messages.go
geoknee: f87b36c

protocols/messages.go
geoknee: f87b36c

PR #202, opened 2022-02-02
Overhaul objective proposal types and logic

geoknee:
Changes:
* change `Proposal` field on `Message` to be a `bool`
    * set to `true` when instigator sends the pre fund setup
    * there's enough information on the rest of the message for the recipient to construct their own objective
    * alternative proposal from @lalexgap is for the recipient to simply inspect their store -- create a new objective if necessary
 * fix bigs transcribing events in the engine.
 * use new methods and `getOrCreateObjective` and `constructObjectiveFromProposal` to handle messages appropriately in the engine

This PR allows our integration test to exercise far more code than before:
In `client_test.log`:

```log
0x4d42FF3b758bC31710A0E01b2C5151f9eA95e9D4: 2022/02/02 11:12:05 engine.go:63: Constructed Engine
0x711A0CE1662d092a6b610f575c4f860Ecf6C940D: 2022/02/02 11:12:05 engine.go:63: Constructed Engine
0x4d42FF3b758bC31710A0E01b2C5151f9eA95e9D4: 2022/02/02 11:12:05 engine.go:150: Sending message to 0x711A0CE1662d092a6b610f575c4f860Ecf6C940D
0x4d42FF3b758bC31710A0E01b2C5151f9eA95e9D4: 2022/02/02 11:12:05 engine.go:171: Objective DirectFunding-0xe3ba1c69441fcbb6364db3cf3e926d66f9fca6f5a0fe988a33d657fef8cfb8a5 is WaitingForCompletePrefund
[...]

PR #207, opened 2022-02-02
Remove proposal flag

lalexgap:
⚠️ base not main branch.
Based on https://github.com/statechannels/go-nitro/pull/202

Instead of using a `Proposal` flag on messages this PR just assumes that any objective it hasn't seen is a proposed objective. This simplifies some logic and the Message struct. Since there is nothing special about a proposal message I think it's safe to just treat any new objective as a proposed objective?

### DIFF
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -3,7 +3,6 @@ package engine // import "github.com/statechannels/go-nitro/client/engine"
 
 import (
 	"errors"
-	"fmt"
 	"io"
 	"log"
 	"strings"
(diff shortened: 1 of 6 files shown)


client/engine/engine.go
```diff
@@ -170,24 +169,18 @@ func (e *Engine) attemptProgress(objective protocols.Objective) {
 // getOrCreateObjective creates the objective if the supplied message is a proposal. Otherwise, it attempts to get the objective from the store.
```
geoknee, line 169:
```suggestion
// getOrCreateObjective creates the objective if it doesn't already exist in the store. Otherwise, it returns the objective from the store.
```

geoknee (approved):
This is definitely more terse!

PR #211, opened 2022-02-03
Replace` direct-fund.signAndStore` with `channel.SignAndAddSetupState`

geoknee:
We currently rely on a relatively new method `signAndStore` on the `DirectFundObjective` type. This method has no documentation, has on unduly wide scope, and is not reusable by other protocols (such as `virtual-fund`). 

This PR solves all of these issues by pushing the functionality down into a new, documented method on the `Channel` type, which is already shared by all protocols. It narrows the scope of the function somewhat, to only handling setup states.

A slightly more controversial change is that it also replaces some errors with panics, along code paths that we should never hit (unless there is a bug). Most of those bugs would be considered severe, apart from one: In particular, we will panic if we have already signed the state. I think it is worth keeping this behaviour in place, as we should be careful enough when writing our protocols to not repeat the work of signing a state. 

This solution will be easier to maintain and audit, and will expose bugs more rapidly.

I have not added any unit tests for this method. They surely wouldn't hurt, and I can add them at the request of the reviewer. On the other hand, the replaced function was also not unit tested, and we probably have good coverage "from above" (from integration testing" and "from below" (the method simply wires up other, well unit-tested functions). 

---

### Code quality

### Project management

### DIFF
--- a/channel/channel.go
+++ b/channel/channel.go
@@ -2,6 +2,7 @@ package channel
 
 import (
 	"errors"
+	"fmt"
 	"reflect"
 
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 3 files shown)


channel/channel.go
```diff
@@ -224,3 +225,38 @@ func (c Channel) AddSignedStates(sss []state.SignedState) bool {
...
+}
+
+// SignAndAddPrefund signs and adds the postfund state for the channel, returning a state.SignedState suitable for sending to peers.
+func (c Channel) SignAndAddPostfund(sk *[]byte) (state.SignedState, error) {
+	return c.signAndAddSetupState(false, sk)
+}
+
+// signAndAddPrefund signs and adds a setup state for the channel, returning a state.SignedState suitable for sending to peers.
+// If the preFund is true, the setup state is the Prefund state, otherwise it is the Postfund state.
+func (c Channel) signAndAddSetupState(preFund bool, sk *[]byte) (state.SignedState, error) {
```
lalexgap, line 241: We could change this to `signAndAddState(state.State s, sk *[]byte)`  and then just call `c.signAndAddState(c.Pre/PostFundState(),sk)` in `SignAndAddPre/Postfund`.  This would have us having to use a boolean prefund flag.

channel/channel.go
```diff
@@ -224,3 +225,38 @@ func (c Channel) AddSignedStates(sss []state.SignedState) bool {
...
+		return state.SignedState{}, fmt.Errorf("could not sign prefund %w", err)
+	}
+	ss := state.NewSignedState(setupState)
+	err = ss.AddSignature(sig)
+	if err != nil {
+		panic("could not add own signature")
+	}
+	ok := c.AddSignedState(ss)
+	if !ok {
+		panic("could not add signed prefund to channel")
```
lalexgap, line 259: This could be either a prefund or postfund
```suggestion
		panic("could not add signed state to channel")
```

lalexgap (approved):

channel/channel.go
geoknee: [8dc0308](https://github.com/statechannels/go-nitro/pull/211/commits/8dc03088632a1d94669808ec981bc473c9273356)

Issue #191, opened 2022-01-31
`MockStore` should serialize objectives before storing


geoknee:
And store `Objective` and `Channel` separately (so that a `Channel` can survive the death / completion of an objective)

* `json.Marshall()` might be the easiest / best way
* need to deal with pointers
* should `.Serialize()` be a method on the `Objective` interface?
* will the `Store` interface end up having `Channel` based methods? e.g. `GetChannelById`?

PR #214, opened 2022-02-03
WIP: add JSONMarshaler interface to `Objective`

NiloCK:
This is incremental work toward implementation of a store. See #191 

The CURRENT mockstore places in-memory objects into an in-memory map. This is not ideal because:

- it skips the serialization / deserialization steps that a production store will require (making it a less realistic mock)
- it (potentially) fails to share these in-memory 

The goal is to apply a real world serialization scheme to the data we're interested in storing: `Objective`s and `Channel`s. This is complicated by the following:

- in-memory `Objective`s contain references to `Channel`s, but we wish to store them separately in the store
- `Objective`s and `Channel`s contain a mix of public and private fields, all of which need to be serialized

**How Has This Been Tested?**

The existing test suite via `mockstore_test.go` relies on the correctness of these changes. They are currently failing!

**Current blocker**

https://go.dev/play/p/41si1Khs7xU - `MarshalJSON` implementations which act on structs that embed the receiver type seem to invoke an infinite loop. (stackoverflow error is not reported on the playground, but it's there). By comparison: [the same code](https://go.dev/play/p/mgaCCQLVDjq) with the `MarshalJSON` method removed works fine.

By my estimation, we need to implement a `MarshalJSON` in order to ensure that private fields are persisted, but doing this triggers the stackoverflow problem above.

---

#### Code quality

- [?] This change does not have an unduly wide scope

#### Project management

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -56,7 +56,7 @@ func (c *Client) CreateDirectChannel(counterparty types.Address, appDefinition t
 
 	// Pass in a fresh, dedicated go channel to communicate the response:
 	apiEvent := engine.APIEvent{
-		ObjectiveToSpawn: objective,
+		ObjectiveToSpawn: &objective,
 		Response:         make(chan engine.Response)}
 
 	// Send the event to the engine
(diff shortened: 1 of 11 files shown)


geoknee:
> * it (potentially) fails to share these in-memory

Can you clarify what you mean by this?

PR #219, opened 2022-02-03
sketch out a script for running our CI checks

### LINKED ISSUE #148: Run checks in pre-commit git hook

geoknee:
This is towards #176 .

Fixes #148 

The thing I am tussling with just now is that the CI suite does some "checks" by running code which *may* dirty the git working tree, and then calling `git diff --exit-code` to see if it did. I like this pattern a lot, but it may make less sense to do locally. 

Is it a reasonable expectation that the local script would not make any changes to local files? How about formatting them?

Another issue is that we may lose some of the sweetness of GH actions (e.g. they can inline linting errors in the source code location that triggers them) by wrapping everything in a shell script.


lalexgap:
> The thing I am tussling with just now is that the CI suite does some "checks" by running code which may dirty the git working tree, and then calling git diff --exit-code to see if it did. I like this pattern a lot, but it may make less sense to do locally.
 
I often have other local changes outside of a commit so my local git working tree is often dirty, so I feel like this might be tricky locally.

>Is it a reasonable expectation that the local script would not make any changes to local files? How about formatting them?

My vote would be for the local script to not make any changes to files. We could create a separate `format` script that actually performs formatting.

> Another issue is that we may lose some of the sweetness of GH actions (e.g. they can inline linting errors in the source code location that triggers them) by wrapping everything in a shell script.

This seems like a little bit of a bummer. Is it very important that our CI and local environment share the exact same script? We could just keep the GH actions and use a local script that roughly mimics the behaviour.  If there is a script that I can run that makes me 95% sure the GH actions will pass, that would be good enough for me.

standardize.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+# exit immediately when any command fails
+set -e
+
+go build ./... 
+golangci-lint run # this will fail if the code is formatted incorrectly, see ./.golangci.yml
+go test ./...
+go mod tidy # this may generate some changes...
```
lalexgap, line 9: It looks like a[ check flag ](https://github.com/golang/go/issues/27005)is planned for `go mod tidy` but missed the cut off for golang 1.18 :(

geoknee:
Thanks for your feedback @lalexgap. I agree with what you say. Personally, I feel like the CI script is sufficiently simple that we actually don't need to do anything right now. The contribution guide covers what is expected https://github.com/statechannels/go-nitro/blob/main/contributing.md#pre-pr-checks .

I know others may feel differently, so please chime in if you feel strongly! @andrewgordstewart @NiloCK . If not, let's just close this and the associate issue.

PR #220, opened 2022-02-04
set chain holdings when constructing a Channel

geoknee:
and panic if we try to read one we haven't set.

This is just an alternative suggestion to #215 . I'm not sure if it is more idiomatic or not. 

The underlying bug is a mismatch in two `types.Funds` mappings. One has been populated when the `Channel` was constructed, and one has not. When we come to compare the two mappings, the one which has not been populated will cause a panic unless we "catch" the problem and intervene

The solution here is simply to populate both mappings in the `Channel` constructor. Then, any other method that wants to work with those mappings need not have any special logic inside of it. 

It seems loosely connected to the go proverb about "making the zero value useful" https://go-proverbs.github.io/.  If we were just using `big.Ints`, an uninitialised value wouldn't pose a problem. But `maps` do cause a problem when they are not initialized, so it's hard to see a way to be true to the proverb. 

Solidity is actually "better" at this, I think. An uninitialised mapping would just return a 0 when read.

### DIFF
--- a/protocols/direct-fund/direct-fund.go
+++ b/protocols/direct-fund/direct-fund.go
@@ -224,7 +224,6 @@ func (s DirectFundObjective) fundingComplete() bool {
 
 // safeToDeposit returns true if the recorded OnChainHoldings are greater than or equal to the threshold for safety.
 func (s DirectFundObjective) safeToDeposit() bool {
-	fmt.Println(s.myDepositSafetyThreshold, s.C.OnChainFunding)
 	for asset, safetyThreshold := range s.myDepositSafetyThreshold {
 
 		chainHolding, ok := s.C.OnChainFunding[asset]
--- a/protocols/direct-fund/direct-fund.go
+++ b/protocols/direct-fund/direct-fund.go
@@ -87,6 +87,7 @@ func New(
 		types.AddressToDestination(myAddress),
 	)
 	init.myDepositTarget = init.myDepositSafetyThreshold.Add(myAllocatedAmount)
+
 	return init, nil
 }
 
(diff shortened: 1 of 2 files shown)


protocols/direct-fund/direct-fund.go
```diff
@@ -224,12 +225,13 @@ func (s DirectFundObjective) fundingComplete() bool {
...
 func (s DirectFundObjective) safeToDeposit() bool {
+	fmt.Println(s.myDepositSafetyThreshold, s.C.OnChainFunding)
 	for asset, safetyThreshold := range s.myDepositSafetyThreshold {
+
 		chainHolding, ok := s.C.OnChainFunding[asset]
 
 		if !ok {
-			// If there are no holdings on chain we use 0 for our calculations
-			chainHolding = big.NewInt(0)
+			panic("nil chainHolding for asset in myDepositSafetyThreshold")
```
geoknee, line 234: This whole block is probably unnecessary, since we will get a panic anyway if we don't use the `,ok` pattern on L231.

lalexgap: While strictly not needed, the error is probably helpful for debugging :)

protocols/direct-fund/direct-fund.go
```diff
@@ -224,12 +225,13 @@ func (s DirectFundObjective) fundingComplete() bool {
 
 // safeToDeposit returns true if the recorded OnChainHoldings are greater than or equal to the threshold for safety.
 func (s DirectFundObjective) safeToDeposit() bool {
+	fmt.Println(s.myDepositSafetyThreshold, s.C.OnChainFunding)
```
lalexgap, line 228:
```suggestion
```

lalexgap (approved):
Makes sense to me! I've added [bbed257](https://github.com/statechannels/go-nitro/pull/220/commits/bbed2578c7ef94ff7fd3a3e47483b39506c5c398)so `amountToDeposit` follows the same pattern and rebased this on `main`

PR #224, opened 2022-02-08
WIP: Virtual Funding sends out signed states

lalexgap:
**Title** ☝️ Clear, single line description of the pull request, written as though it was an order.

**Body** A summary of the change and which issue is fixed, including

- what is the problem being solved
- why this is a good approach
- what are the shortcomings of this approach, if any

It may include

- background information, such as "Fixes #"
- benchmark results
- links to design documents
- dependencies required for this change

The summary may omit some of these details if this PR fixes a single ticket that includes these details.

**Changes** [Optional]

Multiple changes are not recommended, so this section should normally be omitted. Unfortunately, they are sometimes unavoidable. If there are multiple logical changes, list them separately.

1. `'foo'` is replaced by `'bar'`
2. `'fizzbuzz'` is optimized for gas

Did you need to run manual tests to verify this change? If so, please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

**⚠️ Does this require multiple approvals?** [Optional]

Please explain which reason, if any, why this requires more than one approval.

- [ ] Is it security related?
- [ ] Is it a significant process change?
- [ ] Is it a significant change to architectural, design?

---

#### Code quality

#### Project management


protocols/virtualfund/virtualfund-single-hop_test.go
```diff
@@ -321,7 +347,7 @@ func TestSingleHopVirtualFund(t *testing.T) {
 			if err != nil {
 				t.Error(err)
 			}
-			if updated.V.PreFundSignedByMe() != true {
+			if updated.V.SignedStateForTurnNum[0].HasSignatureForParticipant(bob.role) != true {
```
lalexgap, line 350: Since the `testUpdate` function and `testCrank` are using the same objective and channels this had to be changed to allow for the state to be signed in `testCrank`.

We should probably be generating data per test instead of sharing it so we don't have to worry about one test influencing another.

protocols/virtualfund/virtualfund.go
```diff
@@ -235,6 +226,7 @@ func (s VirtualFundObjective) Crank(secretKey *[]byte) (protocols.Objective, pro
 
 	if !updated.requestedLedgerUpdates {
 		updated.requestedLedgerUpdates = true
+		// TODO: Should
```
lalexgap, line 229:
```suggestion
```

protocols/virtualfund/virtualfund-single-hop_test.go
geoknee: I can see why sharing the data is a bad pattern here... but can you explain how does this change get around the problem?

geoknee (approved):
Looks good! Nice to see some good code reuse across the protocols. 

Shall we make an issue to get the problems with the tests cleared up as soon as possible?

PR #227, opened 2022-02-08
Add JSON serialization Objectives, Mockstore

### LINKED ISSUE #191: `MockStore` should serialize objectives before storing

NiloCK:
closes #191

## Problem

The existing mockstore held objectives as `go` runtime in-memory objects. This isn't ideal because it:
- removed the (planned) requirement that objective data and channel data be stored independently
- is insufficient as an approximation of a later production store, which will require serde schemes

## Approach

- the `Objective` interface has been modified to adhere to Go's `json.Marshaler` and `json.Unmarshaler` interfaces.

Implementations of `Objective` (directFundObjective, VirtualFundObjective) now have implementations of each and accompanying tests.

- `mockstore` now keeps separate memory maps for Objectives and Channels, indexed by Ids.

## Shortcomings

- :-1:  There is a considerable about of boilerplate in this PR.
  - default behaviour of `json.Marshal` is to _ignore_ private fields. This is not suitable for our needs, so structs which are targeted for serialization have been "copied" with serialization-friendly structs. (eg, `jsonConnection` for `Connection`).
  - :+1:  most (maybe all?) of these boilerplate utility structs are package-private, and do not pollute the overall namespace of `go-nitro`
- :-1:  the implementations of `MarshalJSON` and `UnmarshalJSON` are non-standard in that they discard data.
  - This is by design in order to satisfy our intention to store `channel` and `objective` data separately (since, for example, multiple VFOs can depend on the same ledger channel).
- :-1:  This design requires the `mockstore` to engage in type-assertions, and will require the same of a production store.
  - I don't see any obvious ways around this
  - the assertions are well-contained and seem pretty safe to me

**How Has This Been Tested?**

- new functions are covered by new tests. I've identified some testing //todos which should become new issues if this goes in. 

**⚠️ Does this require multiple approvals?** [Optional]

Maybe. It borders on:

- [x] Is it a significant change to architectural, design?

---

#### Code quality

- [-] This change does not have an unduly wide scope

#### Project management


protocols/directfund/directfund.go
```diff
@@ -101,6 +102,13 @@ func New(
 
 // Public methods on the DirectFundingObjectiveState
 
+// MarshalJSON returns a JSON representation of the DirectFundObjective
+// with channel state replaced by a channel Id
+func (s DirectFundObjective) MarshalJSON() ([]byte, error) {
+	jsonDFO := JSONDirectFundObjective{s, s.C.Id}
```
geoknee: Why not make the `JSON*` struct types private to _this method_ ?

protocols/virtualfund/virtualfund.go
```diff
@@ -104,11 +104,22 @@ type VirtualFundObjective struct {
 	requestedLedgerUpdates bool // records that the ledger update side effects were previously generated (they may not have been executed yet)
 }
 
-// JSONVirtualFundObjective replaces the VirtualFundObjective's channel pointer with the
-// channel's Id, making JSONVirtualFundObjective suitable for serialization
-type JSONVirtualFundObjective struct {
-	VirtualFundObjective
-	V types.Destination
+// jsonVirtualFundObjective is a replaces the VirtualFundObjective's channel pointers
```
geoknee, line 107: typo here

client/engine/store/mockstore.go
```diff
@@ -50,13 +53,48 @@ func (ms MockStore) GetChannelSecretKey() *[]byte {
...
 	// todo: locking
-	ms.objectives[obj.Id()] = obj
+	bytes, err := obj.MarshalJSON()
+	if err != nil {
+		return err
+	}
+	ms.objectives[obj.Id()] = string(bytes)
+
+	id := obj.Id()
+	if strings.HasPrefix(string(id), "DirectFunding-") {
```
geoknee, line 122: This reminds me of a "discriminated union" which is a pattern from Typescript. Elsewhere we have managed to pivot to using the "visitor pattern" instead. Did you consider that option? What would be the pros and cons?

More specifically, could you simply call a new interface method `obj.getChannels` to get an array of channels to pass to `ms.setChannel`? Then, the store does not need to know what kind of objective it is dealing with, only that it can trust it to report which channels need persisting or retrieving.

protocols/directfund/directfund.go
geoknee: I suppose because you can reuse them when un-marshaling. Seems like the right tradeoff, probably.

geoknee:
I didn't take a close look at all the test code yet, but I think the direction of travel here looks pretty good! How do you feel about it, on balance?

I see you still have some conflicts (apologies) and failing checks. We can try to schedule some "quiet time" so that we can get this reviewed and merged.

I would be in favour of multiple reviews on the overall approach of this PR (they wouldn't both need to go into line by line detail though).

NiloCK:
closed in favour of re-implementation on `main`.

Issue #226, opened 2022-02-08
Generate data per virtual funding test


lalexgap:
Since the `testUpdate` function and `testCrank` are using the same objective and channels this had to be changed to allow for the state to be signed in `testCrank`.

We should probably be generating data per test instead of sharing it so we don't have to worry about one test influencing another.

_Originally posted by @lalexgap in https://github.com/statechannels/go-nitro/pull/224#discussion_r801192799_

lalexgap:
This also applies to the current channel tests.

lalexgap:
[Ginkgo](https://github.com/onsi/ginkgo) is a testing framework that would allow us to use familiar patterns from typescript including `BeforeEach`. That might be an easy way to adjust our existing tests to not share data.

PR #229, opened 2022-02-09
Switch channel Sign methods to use pointer receivers

lalexgap:
Previously some of the sign functions on a `channel` were using [value receivers instead of pointer receivers.](https://medium.com/globant/go-method-receiver-pointer-vs-value-ffc5ab7acdb) This meant that we were using a copy of the channel and not modifying the actual channel. 

Since the `SignedStateForTurnNum` map acts like a [pointer](https://go.dev/doc/faq#:~:text=Map%20and%20slice%20values%20behave%20like,not%20the%20data%20it%20points%20to.) it was getting updated correctly with signatures and states. However setting `latestSupportedStateTurnNum` with a value receiver does not work (as it is modifying a copy of the structure) so we end up  with an incorrect `latestSupportedStateTurnNum`.

This PR:
- Implements a simple test for `AddSignedStates` to reproduce the failure.
- Updates all sign methods on channel to use a pointer receiver.
- Adds a Signature.Equal function to make comparing signatures easier.

### DIFF
--- a/channel/channel_test.go
+++ b/channel/channel_test.go
@@ -119,6 +119,93 @@ func TestChannel(t *testing.T) {
 			t.Errorf("TestCrank: side effects mismatch (-want +got):\n%s", diff)
 		}
 	}
+	alicePrivateKey := common.Hex2Bytes(`caab404f975b4620747174a75f08d98b4e5a7053b691b41bcfc0d839d48b7634`)
+	bobPrivateKey := common.Hex2Bytes(`62ecd49c4ccb41a70ad46532aed63cf815de15864bc415c87d507afd6a5e8da2`)
+	testAddSignedState := func(t *testing.T) {
+		myC, _ := New(s, 0)
+		ss := state.NewSignedState(s)
+		sigA, err := ss.State().Sign(alicePrivateKey)
+		if err != nil {
+			t.Error(err)
+		}
+		sigB, err := ss.State().Sign(bobPrivateKey)
+		if err != nil {
+			t.Error(err)
+		}
+		err = ss.AddSignature(sigA)
+		if err != nil {
+			t.Error(err)
+		}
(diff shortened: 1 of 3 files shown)


channel/channel_test.go
```diff
@@ -119,6 +119,93 @@ func TestChannel(t *testing.T) {
...
+				panic(err)
+			}
+			wantSig := expectedSigs[i]
+			if !gotSig.Equal(wantSig) {
+				t.Errorf("Expected to find signature %x at index 0, but got %x", wantSig, gotSig)
+			}
+		}
+	}
+	testAddSignedStates := func(t *testing.T) {
+		myC, _ := New(s, 0)
```
lalexgap, line 166: Most of these tests share the same `channel` which means one test can easily break another. This is similar to #226

channel/channel_test.go
```diff
@@ -119,6 +119,93 @@ func TestChannel(t *testing.T) {
...
+			if err != nil {
+				panic(err)
+			}
+			wantSig := expectedSigs[i]
+			if !gotSig.Equal(wantSig) {
+				t.Errorf("Expected to find signature %x at index 0, but got %x", wantSig, gotSig)
+			}
+		}
+
+	}
```
geoknee, line 208: This test file got quite a bit longer -- do we need it to be this long, do you thnk?

channel/channel_test.go
geoknee: Shall we make an issue for this?

channel/channel_test.go
lalexgap: This could probably be DRYed up with some helper functions.

channel/channel_test.go
lalexgap: I've updated #226 to track this.

PR #235, opened 2022-02-10
Update channel arguments to pointers

### LINKED ISSUE #232: Channels should not be passed by value
Currently some [functions](https://github.com/statechannels/go-nitro/blob/9edaf84cf28610cdd54f299543c9314b2e13e45d/protocols/virtualfund/virtualfund.go#L61) accept a `Channel` struct instead of a pointer to a Channel struct. This means that the function get's a copy of the `Channel` and not access to the original. The copy is only a shallow copy so any `maps` or `slices` or `pointer` will continue to point to their original location.

This makes interacting with a shallow copied `Channel` very confusing, since changes made to any maps or slices like (`SignedStateForTurnNum`) will persist but changes to local variables on the `Channel` will not.

Since a shallow copy of a `Channel` is almost never what we want we should probably pass Channel pointers around instead.

lalexgap:
Fixes #232 

Two main changes in this PR:
## Channel constructor methods now return a pointer
Channel constructor methods now return a pointer to a channel struct instead of a channel struct. Generally we want to pass a channel around by a pointer since passing by value results in a a shallow copied channel. Switching the constructors to return pointers helps reinforce that.

## VirtualFundObjective.New accepts channel pointers instead of a channel struct
  Previously the channels structs were directly being passed into the  `New` function. This meant that `New` would get a shallow copy of the channel. The shallow copy would still reference the same map/slice data  as the original, causing problems.

By accepting a pointer we are now making explicit that we're modifying the same channel as the caller. If we want to create a copy of that we should explicitly call `Clone`

### DIFF
--- a/protocols/virtualfund/virtualfund-single-hop_test.go
+++ b/protocols/virtualfund/virtualfund-single-hop_test.go
@@ -154,7 +154,7 @@ func TestSingleHopVirtualFund(t *testing.T) {
 		my := alice
 
 		// Alice plays role 0 so has no ledger channel on her left
-		var ledgerChannelToMyLeft channel.TwoPartyLedger
+		var ledgerChannelToMyLeft *channel.TwoPartyLedger
 
 		// She has a single ledger channel L_0 connecting her to P_1
 		var ledgerChannelToMyRight, _ = channel.NewTwoPartyLedger(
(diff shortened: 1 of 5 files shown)


protocols/virtualfund/virtualfund-single-hop_test.go
```diff
@@ -200,7 +200,7 @@ func TestSingleHopVirtualFund(t *testing.T) {
 
 		testNew := func(t *testing.T) {
 			// Assert that a valid set of constructor args does not result in an error
-			o, err := New(false, vPreFund, my.address, 2, my.role, ledgerChannelToMyLeft, ledgerChannelToMyRight)
+			o, err := New(false, vPreFund, my.address, 2, my.role, &ledgerChannelToMyLeft, ledgerChannelToMyRight)
```
geoknee, line 203:
```suggestion
			o, err := New(false, vPreFund, my.address, 2, my.role, &ledgerChannelToMyLeft, &ledgerChannelToMyRight)
```

Why isn't this a compiler error??

geoknee (approved):

protocols/virtualfund/virtualfund-single-hop_test.go
lalexgap: `ledgerChannelToMyRight` is set to the value of  `NewTwoPartyLedger` which returns a pointer. `ledgerChannelToMyLeft` is just an empty struct that we set in the test. We should probably update `ledgerChannelToMyLeft` to be a pointer to be consistent.

lalexgap: Updated in https://github.com/statechannels/go-nitro/pull/235/commits/bd2e03c1643bffaebfe52ab371e068f7ee3c9c81 so that `ledgerChannelToMyLeft` is a pointer for consistency.

Issue #217, opened 2022-02-03
Write an integration test for virtually funding a channel between three parties


geoknee:
Can follow the `client_test.go` pattern. 

Depends on #216 , as virtual funder needs to return side effects
Depends on #20, so engine will trigger side effects

# TODO
- [ ] split this ticket into two: a) an MVP version, an b) an integration test of the full protocol

PR #236, opened 2022-02-11
Implement basic Ledger Manager

### LINKED ISSUE #223: Handle Ledger Funding Requests (with mock ledger funding)
We would like basic support for handling ledger requests so we can create a simple MVP of a virtual funding.

We should
- Create an interface for handing ledger requests and update the engine to call this
- Implement a dead simple implementation of the interface that mocks out actual ledger funding and always returns true 

This is split off from #20 .

lalexgap:
Fixes #223 

This PR adds a new `LedgerManager` that handles a ledger request by updating the ledger channel and generating a message with the signed ledger state. The current logic is not a full ledger protocol, but should be enough for now to work towards #217 

The virtual funding tests have been updated to use the `LedgerManager` which has removed a lot of manually created ledger data.

Eventually this `LedgerManager` [will be used in the engine to handle ](https://github.com/statechannels/go-nitro/issues/240)`LedgerRequests` and get us closer to #217.

### DIFF
--- a/protocols/ledger/ledger.go
+++ b/protocols/ledger/ledger.go
@@ -30,6 +30,7 @@ func (l *LedgerCranker) Update(ledger *channel.TwoPartyLedger) {
 	l.ledgers[ledger.Id] = ledger
 }
 
+// CreateLedger creates a new  two party ledger channel based on the provided left and right outcomes.
 func (l *LedgerCranker) CreateLedger(left outcome.Allocation, right outcome.Allocation, secretKey *[]byte, myIndex uint) *channel.TwoPartyLedger {
 
 	leftAddress, _ := left.Destination.ToAddress()
@@ -54,10 +55,13 @@ func (l *LedgerCranker) CreateLedger(left outcome.Allocation, right outcome.Allo
 	}
 
 	l.ledgers[ledger.Id] = ledger
+	// Update the nonce by 1
+	l.nonce = big.NewInt(0).Add(l.nonce, big.NewInt(1))
 	return ledger
 }
 
 // HandleRequest accepts a ledger request and updates the ledger channel based on the request.
+// It returns a signed state message that can be sent to other participants.
 func (l *LedgerCranker) HandleRequest(request protocols.LedgerRequest, oId protocols.ObjectiveId, secretKey *[]byte) (protocols.SideEffects, error) {
 
 	ledger := l.GetLedger(request.LedgerId)
--- a/protocols/ledger/ledger.go
+++ b/protocols/ledger/ledger.go
@@ -53,7 +53,7 @@ func (l *LedgerCranker) CreateLedger(left outcome.Allocation, right outcome.Allo
 
 // HandleRequest accepts a ledger request and updates the ledger channel based on the request.
 // It returns a signed state message that can be sent to other participants.
-func (l *LedgerCranker) HandleRequest(ledger *channel.TwoPartyLedger, request protocols.LedgerRequest, oId protocols.ObjectiveId, secretKey *[]byte) (protocols.SideEffects, error) {
+func (l *LedgerCranker) HandleRequest(ledger *channel.TwoPartyLedger, request protocols.LedgerRequest, secretKey *[]byte) (protocols.SideEffects, error) {
 
 	guarantee, _ := outcome.GuaranteeMetadata{
 		Left:  request.Left,
@@ -109,7 +109,7 @@ func (l *LedgerCranker) HandleRequest(ledger *channel.TwoPartyLedger, request pr
 		return protocols.SideEffects{}, errors.New("Could not add signed state to channel")
 	}
 
-	messages := protocols.CreateSignedStateMessages(oId, ss, ledger.MyIndex)
+	messages := protocols.CreateSignedStateMessages(request.ObjectiveId, ss, ledger.MyIndex)
 	return protocols.SideEffects{MessagesToSend: messages}, nil
 
 }
--- a/protocols/ledger/ledger.go
+++ b/protocols/ledger/ledger.go
@@ -11,11 +11,6 @@ import (
 	"github.com/statechannels/go-nitro/types"
 )
 
-// DirectFundObjective is a cache of data computed by reading from the store. It stores (potentially) infinite data
-type LedgerRequestHandler interface {
-	HandleRequest(request protocols.LedgerRequest, secretKey *[]byte) protocols.SideEffects
-}
-
 type LedgerCranker struct {
 	ledgers map[types.Destination]*channel.TwoPartyLedger
 	nonce   *big.Int
@@ -27,9 +22,13 @@ func NewLedgerCranker() LedgerCranker {
 		nonce:   big.NewInt(0),
 	}
 }
+
+// Update updates the ledger cranker with the given ledger channel
+// Eventually this will be deprecated in favour of using store
 func (l *LedgerCranker) Update(ledger *channel.TwoPartyLedger) {
 	l.ledgers[ledger.Id] = ledger
 }
+
 func (l *LedgerCranker) CreateLedger(left outcome.Allocation, right outcome.Allocation, secretKey *[]byte, myIndex uint) *channel.TwoPartyLedger {
 
 	leftAddress, _ := left.Destination.ToAddress()
@@ -44,7 +43,7 @@ func (l *LedgerCranker) CreateLedger(left outcome.Allocation, right outcome.Allo
 		Outcome: outcome.Exit{outcome.SingleAssetExit{
 			Allocations: outcome.Allocations{left, right},
 		}},
-		TurnNum: 0, // Start at post fund
+		TurnNum: 0,
 		IsFinal: false,
 	}
 
@@ -57,17 +56,23 @@ func (l *LedgerCranker) CreateLedger(left outcome.Allocation, right outcome.Allo
 	return ledger
 }
 
+// HandleRequest accepts a ledger request and updates the ledger channel based on the request.
 func (l *LedgerCranker) HandleRequest(request protocols.LedgerRequest, oId protocols.ObjectiveId, secretKey *[]byte) protocols.SideEffects {
+
 	ledger := l.ledgers[request.LedgerId]
+
 	guarantee, _ := outcome.GuaranteeMetadata{
 		Left:  request.Left,
 		Right: request.Right,
 	}.Encode()
+
 	supported, err := ledger.Channel.LatestSupportedState()
 	if err != nil {
 		panic(err)
 	}
+
 	nextState := supported.Clone()
+	// TODO: We're currently setting the amount to 0 for participants, we should calculate the correct amount
 	nextState.Outcome = outcome.Exit{outcome.SingleAssetExit{
 		Allocations: outcome.Allocations{
 			outcome.Allocation{
@@ -86,7 +91,9 @@ func (l *LedgerCranker) HandleRequest(request protocols.LedgerRequest, oId proto
 			},
 		},
 	}}
+
 	nextState.TurnNum = nextState.TurnNum + 1
+
 	ss := state.NewSignedState(nextState)
 	err = ss.SignAndAdd(secretKey)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


lalexgap:
Moving this back to draft as I want to tweak it slightly.

protocols/ledger/ledger.go
```diff
@@ -0,0 +1,141 @@
...
+	"github.com/statechannels/go-nitro/channel/state/outcome"
+	"github.com/statechannels/go-nitro/protocols"
+	"github.com/statechannels/go-nitro/types"
+)
+
+type LedgerManager struct {
+	nonce *big.Int
+}
+
+func NewLedgerManager() LedgerManager {
```
lalexgap, line 19: The Ledger Manager could accept a `store` and read and write from the `store` itself but that would break with previous patterns where the `engine` is soley responsible for fetching and writing to the store.

protocols/ledger/ledger.go
```diff
@@ -0,0 +1,141 @@
...
+	return LedgerManager{
+		nonce: big.NewInt(0),
```
lalexgap, line 21: Nonce management should probably be handled in the `store`?

protocols/ledger/ledger.go
```diff
@@ -0,0 +1,141 @@
...
+func (l *LedgerManager) HandleRequest(ledger *channel.TwoPartyLedger, request protocols.LedgerRequest, secretKey *[]byte) (protocols.SideEffects, error) {
+
+	guarantee, _ := outcome.GuaranteeMetadata{
+		Left:  request.Left,
+		Right: request.Right,
+	}.Encode()
+
+	supported, err := ledger.Channel.LatestSupportedState()
+	if err != nil {
+		return protocols.SideEffects{}, fmt.Errorf("Could not find a supported state %w", err)
```
NiloCK, line 65:
```suggestion
		return protocols.SideEffects{}, fmt.Errorf("error finding a supported state: %w", err)
```

error messages are non-capitalized & non-punctuated by convention. Changed the text so that the the `: errorText` reads a little better maybe.

protocols/ledger/ledger.go
```diff
@@ -0,0 +1,141 @@
...
+				Destination: request.Left,
+				Amount:      leftAmount,
+			},
+			outcome.Allocation{
+				Destination: request.Right,
+				Amount:      rightAmount,
+			},
+			outcome.Allocation{
+				Destination:    request.Destination,
+				Amount:         request.Amount[types.Address{}],
```
NiloCK, line 94:
```suggestion
				Amount:         request.Amount[asset],
```

protocols/ledger/ledger.go
```diff
@@ -0,0 +1,141 @@
...
+	supported, err := ledger.Channel.LatestSupportedState()
+	if err != nil {
+		return protocols.SideEffects{}, fmt.Errorf("Could not find a supported state %w", err)
+	}
+
+	asset := types.Address{}
+	nextState := supported.Clone()
+
+	// Calculate the amounts
+	amountPerParticipant := big.NewInt(0).Div(request.Amount[asset], big.NewInt(2))
```
NiloCK, line 72: flag for discussion.

alice - irene - bob wishing to fund a new a-b virtual channel with init outcome `{ alice: 5, bob: 0 }` will have to make ledger requests on both ledger channels:
- l(ai) - wants new guarantee { alice:5, irene: 0 }
- l(bi) - wants new guarantee { bob: 0, irene: 5 }

The above calculation doesn't reflect this, and probably the ledgerRequest doesn't pass all of the required information in.

protocols/ledger/ledger.go
```diff
@@ -0,0 +1,141 @@
...
+		Left:  request.Left,
+		Right: request.Right,
+	}.Encode()
+
+	supported, err := ledger.Channel.LatestSupportedState()
+	if err != nil {
+		return protocols.SideEffects{}, fmt.Errorf("Could not find a supported state %w", err)
+	}
+
+	asset := types.Address{}
```
NiloCK, line 68:
```suggestion
	asset := types.Address{} // todo: loop over request.amount's assets
```

NiloCK (commented):
Review still in progress. Bunch of good stuff in here but some tweaks will need to come in eventually.

I'm not sure whether the `LedgerManager` struct adds much - eg, `HandleRequest` could / should probably be a method directly on `TwoPartyLedger`.

protocols/ledger/ledger.go
```diff
@@ -0,0 +1,141 @@
...
+		Outcome: outcome.Exit{outcome.SingleAssetExit{
+			Allocations: outcome.Allocations{left, right},
+		}},
+		TurnNum: 0,
+		IsFinal: false,
+	}
+
+	ledger, lErr := channel.NewTwoPartyLedger(initialState, myIndex)
+	if lErr != nil {
+		panic(lErr)
```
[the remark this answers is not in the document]
lalexgap: Updated to return an error in https://github.com/statechannels/go-nitro/pull/236/commits/57ebaee69559b367433eb9454dc86f965942bf5f

protocols/ledger/ledger_test.go
```diff
@@ -0,0 +1,142 @@
...
+			Allocations: outcome.Allocations{left, right},
+		}},
+		TurnNum: 0,
+		IsFinal: false,
+	}
+
+	gotState := ledger.SignedStateForTurnNum[0].State()
+	if diff := cmp.Diff(expectedState, gotState); diff != "" {
+		t.Errorf("TestCreateLedger: ledger state mismatch (-want +got):\n%s", diff)
+	}
```
NiloCK, line 60:
```suggestion
	if ledger.ChannelNonce.Cmp(big.NewInt(0)) != 0 {
		t.Error("TestCreateLedger: initial ledger channel should use the 0 nonce")
	}
```

the `state` check that's erased wasn't very dry - just checking that the init data is the same in both files, but not clear if that's useful.

The added check compliments the below check on incrementing nonces

protocols/ledger/ledger_test.go
```diff
@@ -0,0 +1,142 @@
...
+	right := outcome.Allocation{Destination: bob.destination, Amount: big.NewInt(2)}
+
+	ledger, _ := ledgerManager.CreateTestLedger(left, right, &alice.privateKey, 0)
+
+	destination := types.AddressToDestination(common.HexToAddress(`0x5e29E5Ab8EF33F050c7cc10B5a0456D975C5F88d`))
+	asset := types.Address{}
+	oId := protocols.ObjectiveId("Test")
+
+	validRequest := protocols.LedgerRequest{ObjectiveId: oId, LedgerId: ledger.Id, Left: left.Destination, Right: right.Destination, Destination: destination, Amount: types.Funds{asset: big.NewInt(4)}}
+	invalidRequest := protocols.LedgerRequest{ObjectiveId: oId, LedgerId: ledger.Id, Left: left.Destination, Right: right.Destination, Destination: destination, Amount: types.Funds{asset: big.NewInt(10)}}
```
NiloCK, line 81:
```suggestion
	validRequest := protocols.LedgerRequest{
		ObjectiveId: oId,
		LedgerId:    ledger.Id,
		Left:        left.Destination,
		Right:       right.Destination,
		Destination: destination,
		Amount:      types.Funds{asset: big.NewInt(4)},
	}
	invalidRequest := protocols.LedgerRequest{
		ObjectiveId: oId,
		LedgerId:    ledger.Id,
		Left:        left.Destination,
		Right:       right.Destination,
		Destination: destination,
		Amount:      types.Funds{asset: big.NewInt(10)},
	}
```

protocols/ledger/ledger_test.go
```diff
@@ -0,0 +1,142 @@
...
+
+	_, err := ledgerManager.HandleRequest(ledger, validRequest, &alice.privateKey)
+	if err == nil {
+		t.Errorf("TestHandleLedgerRequest: expected request to be fail as there is no supported state")
```
NiloCK, line 85:
```suggestion
		t.Errorf("TestHandleLedgerRequest: expected request to fail as there is no supported state")
```

protocols/ledger/ledger_test.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	SignPreAndPostFundingStates(ledger, []*[]byte{&alice.privateKey, &bob.privateKey})
+
+	_, err = ledgerManager.HandleRequest(ledger, invalidRequest, &alice.privateKey)
+	if err == nil {
+		t.Errorf("TestHandleLedgerRequest: expected request to  fail as the ledger does not have enough funds")
```
NiloCK, line 92:
```suggestion
		t.Errorf("TestHandleLedgerRequest: expected request to fail as the ledger does not have enough funds")
```

lalexgap:
> I'm not sure whether the `LedgerManager` struct adds much - eg, `HandleRequest` could / should probably be a method directly on `TwoPartyLedger`.

This is a good point! One reason I separated this out into the `LedgerManager` is that I assumed we'll eventually have a full ledger protocol that should be separate from the `TwoPartyLedger` struct.

protocols/ledger/ledger_test.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	sideEffects, err := ledgerManager.HandleRequest(ledger, validRequest, &alice.privateKey)
```
NiloCK, line 95:
```suggestion
	_, err = ledgerManager.HandleRequest(ledger, invalidRequest, &alice.privateKey)
	if err == nil {
		t.Errorf("TestHandleLedgerRequest: expected request to  fail as the ledger does not have enough funds")
	}

	SignPreAndPostFundingStates(ledger, []*[]byte{&alice.privateKey, &bob.privateKey})
	sideEffects, err := ledgerManager.HandleRequest(ledger, validRequest, &alice.privateKey)
```

moving the 'fix' closer to the test that the fix was applied.

protocols/ledger/ledger.go
lalexgap: Is the missing info from the `LedgerRequest` the amount per participant?

protocols/ledger/ledger_test.go
```diff
@@ -0,0 +1,142 @@
...
+				outcome.Allocation{
+					Destination:    destination,
+					Amount:         big.NewInt(4),
+					AllocationType: outcome.GuaranteeAllocationType,
+					Metadata:       guarantee,
+				},
+			}}},
+		TurnNum: 2,
+		IsFinal: false}
+
```
NiloCK, line 130: Seems to be missing a `gotState, _ := ledger.LatestSupportedState()` + comparison between got and expected.

As far as I can tell, the `NewSignedState` and `SignAndAdd` happen to that state inside of `HandleRequest` and return errors there - the below check may be unnecessary?

NiloCK (approved):
This looks good to me, but we should spin out an issue surrounding the calculation of updated ledger balances (ping @geoknee) and the corresponding testing of those bits via `TestHandleLedgerRequest`.

Caveat: I have not reviewed changes to the virtualfund-single-hop_test.go as carefully as I would have liked, but I'm running out of time!

If those changes can be stashed for another review (and the other pieces still work), that might be a good idea.

protocols/ledger/ledger.go
NiloCK: Yes - in SATP paper these are referred to as `a` and `b`, alice and bob's respective opening balances.

In general (and off-the-cuff - 80% certain), `a` wants to be deducted from the `left` role in the ledger and `b` deducted from the `right`.

protocols/ledger/ledger_test.go
lalexgap: Since this is just a `test` function I agree!

protocols/ledger/ledger_test.go
lalexgap: > Seems to be missing a `gotState, _ := ledger.LatestSupportedState()` + comparison between got and expected.

Good point!

> As far as I can tell, the `NewSignedState` and `SignAndAdd` happen to that state inside of `HandleRequest` and return errors there - the below check may be unnecessary?

While I don't expect this to fail throwing away error responses makes me nervous 😛   My preference would be to always fail a test if a function returns us an error, even if it's a helper function.

PR #238, opened 2022-02-11
Fix overspending `bob` in direct fund objective

NiloCK:
Closes #237. Description in issue.

Issue #237, opened 2022-02-11
Bob is overspending in client_test.go


NiloCK:
`client_test.go` instantiates wallets `alice` and `bob` with the goal of direct-funding a channel with outcome

``` 
// from client_test.go
outcome := outcome.Exit{outcome.SingleAssetExit{
		Allocations: outcome.Allocations{
			outcome.Allocation{
				Destination: types.AddressToDestination(a),
				Amount:      big.NewInt(5),
			},
			outcome.Allocation{
				Destination: types.AddressToDestination(b),
				Amount:      big.NewInt(5),
			},
		},
	}}
```

However: running the test with `e.logger.Printf("handling chain event: %v", chainEvent)` inserted into the `handleChainEvent` method produces

```
...
alice: 2022/02/11 15:31:32 engine.go:182: Sending chain transaction for channel chId
alice: 2022/02/11 15:31:32 engine.go:199: Objective DirectFunding-chId is WaitingForCompleteFunding
bob: 2022/02/11 15:31:32 engine.go:139: handling chain event: {chId {[0x0000000000000000000000000000000000000000,5]} {0}}
bob: 2022/02/11 15:31:32 engine.go:182: Sending chain transaction for channel chId
bob: 2022/02/11 15:31:32 engine.go:199: Objective DirectFunding-chId is WaitingForCompleteFunding
bob: 2022/02/11 15:31:32 engine.go:139: handling chain event: {chId {[0x0000000000000000000000000000000000000000,15]} {0}}
alice: 2022/02/11 15:31:32 engine.go:139: handling chain event: {chId {[0x0000000000000000000000000000000000000000,5]} {0}}
alice: 2022/02/11 15:31:32 engine.go:199: Objective DirectFunding-chId is WaitingForCompleteFunding
alice: 2022/02/11 15:31:32 engine.go:139: handling chain event: {chId {[0x0000000000000000000000000000000000000000,15]} {0}}
bob: 2022/02/11 15:31:32 engine.go:178: Sending message to alice
bob: 2022/02/11 15:31:32 engine.go:199: Objective DirectFunding-chId is WaitingForCompletePostFund
...
```

Note:
- Alice's initial chain event moves the holdings to 5
- Bob's response moves the holdings to 15 (__should be 10__)

This is likely a problem in the `amountToDeposit` function of directFund, or elsewhere in Crank.

PR #239, opened 2022-02-12
DRY out `directfund` unit test, and test all side effects

geoknee:
Our current unit test for this protocol 
- has a TODO for testing side effects
- tests some side effects in a separate test scope to the `TestCrank`
- does not catch the bug found in #237 

This PR pulls the separate tests into `TestCrank`, where much of the setup work can be reused and not replicated, and adds a test which would catch bug #237 . This is good because at present our only protection against regression on this bug is in the `client_test` _integration_ test. 

The shortcomings of the approach are that the test gets a little denser to read, and multiple aspects of cranking are tested at the same time (that's arguably a positive, though). 

**How Has This Been Tested?** 

I manually regressed #238 to reinstate bug #237, and saw this test then FAIL. 

---

#### Code quality

#### Project management

### DIFF
--- a/protocols/directfund/directfund_test.go
+++ b/protocols/directfund/directfund_test.go
@@ -77,55 +77,6 @@ func TestNew(t *testing.T) {
 	}
 }
 
-func TestPreFundSideEffects(t *testing.T) {
-	// Construct various variables for use in the test.
-	var o, _ = New(true, testState, testState.Participants[0])
-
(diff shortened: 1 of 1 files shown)


NiloCK (approved):
:+1: This is a definite improvement in coverage + compactness for this test.

It took me a few minutes to figure out why this was testing against #237, since this test is from an `alice` perspective. (The reason is that the outcome here is prioritized `bob, alice`).

PR #241, opened 2022-02-14
WIP: Update engine to use Ledger Manager

### LINKED ISSUE #240: Update engine to use Ledger Manager
Now that we have a basic way of handling ledger requests we should update the engine to handle the ledger requests it gets back in side effects.

lalexgap:
Fixes #240


lalexgap:
Closing this for now, will reopen when it's no longer stale.

PR #245, opened 2022-02-15
Update Ledger Request to have amount for left and right

### LINKED ISSUE #244: Update Ledger Request to include left and right amounts
Yes - in SATP paper these are referred to as `a` and `b`, alice and bob's respective opening balances.

In general (and off-the-cuff - 80% certain), `a` wants to be deducted from the `left` role in the ledger and `b` deducted from the `right`.

_Originally posted by @NiloCK in https://github.com/statechannels/go-nitro/pull/236#r807199237_

lalexgap:
Fixes #244 

Previously the ledger request only contained a single `Amount`. This meant when handling the request we didn't know how much should be paid by `left` and how much should be paid by `right`.

The ledger request has been updated  to have a `leftAmount` and `rightAmount`.  These get set based on the virtual channel's outcome. IE: `leftAmount=V.Allocations[0]`  `rightAmount=V.Allocations[1]`  

This means that the `LedgerManager` can properly adjust the ledger outcome and construct the guarantee.

### DIFF
--- a/protocols/ledger/ledger.go
+++ b/protocols/ledger/ledger.go
@@ -37,14 +37,13 @@ func (l *LedgerManager) HandleRequest(ledger *channel.TwoPartyLedger, request pr
 	nextState := supported.Clone()
 
 	// Calculate the amounts
-
 	leftAmount := big.NewInt(0).Sub(nextState.Outcome.TotalAllocatedFor(request.Left)[asset], request.LeftAmount[asset])
 	rightAmount := big.NewInt(0).Sub(nextState.Outcome.TotalAllocatedFor(request.Right)[asset], request.RightAmount[asset])
 	total := big.NewInt(0).Add(request.LeftAmount[asset], request.RightAmount[asset])
-	if leftAmount.Cmp(big.NewInt(0)) < 0 {
+	if types.Lt(leftAmount, big.NewInt(0)) {
 		return protocols.SideEffects{}, fmt.Errorf("Allocation for %x cannot afford the amount %d", request.Left, request.LeftAmount[asset])
 	}
-	if rightAmount.Cmp(big.NewInt(0)) < 0 {
+	if types.Lt(rightAmount, big.NewInt(0)) {
 		return protocols.SideEffects{}, fmt.Errorf("Allocation for %x cannot afford the amount %d", request.Right, request.RightAmount[asset])
 	}
 
--- a/protocols/ledger/ledger.go
+++ b/protocols/ledger/ledger.go
@@ -36,16 +36,23 @@ func (l *LedgerManager) HandleRequest(ledger *channel.TwoPartyLedger, request pr
 	asset := types.Address{} // todo: loop over request.amount's assets
 	nextState := supported.Clone()
 
-	// Calculate the amounts
-	leftAmount := big.NewInt(0).Sub(nextState.Outcome.TotalAllocatedFor(request.Left)[asset], request.LeftAmount[asset])
-	rightAmount := big.NewInt(0).Sub(nextState.Outcome.TotalAllocatedFor(request.Right)[asset], request.RightAmount[asset])
-	total := big.NewInt(0).Add(request.LeftAmount[asset], request.RightAmount[asset])
+	// Get the current amounts from the ledger channel
+	currentLeftAmount := nextState.Outcome.TotalAllocatedFor(request.Left)[asset]
+	currentRightAmount := nextState.Outcome.TotalAllocatedFor(request.Right)[asset]
+	// Calculate the new amounts by subtracting the requested amounts from the current amounts
+	leftAmount := big.NewInt(0).Sub(currentLeftAmount, request.LeftAmount[asset])
+	rightAmount := big.NewInt(0).Sub(currentRightAmount, request.RightAmount[asset])
+
+	// If any participant cannot afford the request amount, return an error
 	if types.Lt(leftAmount, big.NewInt(0)) {
 		return protocols.SideEffects{}, fmt.Errorf("Allocation for %x cannot afford the amount %d", request.Left, request.LeftAmount[asset])
 	}
 	if types.Lt(rightAmount, big.NewInt(0)) {
 		return protocols.SideEffects{}, fmt.Errorf("Allocation for %x cannot afford the amount %d", request.Right, request.RightAmount[asset])
 	}
+
+	// Calculate the total amount we need to allocate to the guarantee
+	total := big.NewInt(0).Add(request.LeftAmount[asset], request.RightAmount[asset])
 
 	nextState.Outcome = outcome.Exit{outcome.SingleAssetExit{
 		Allocations: outcome.Allocations{
--- a/protocols/ledger/ledger.go
+++ b/protocols/ledger/ledger.go
@@ -37,14 +37,15 @@ func (l *LedgerManager) HandleRequest(ledger *channel.TwoPartyLedger, request pr
 	nextState := supported.Clone()
 
 	// Calculate the amounts
-	amountPerParticipant := big.NewInt(0).Div(request.Amount[asset], big.NewInt(2))
-	leftAmount := big.NewInt(0).Sub(nextState.Outcome.TotalAllocatedFor(request.Left)[asset], amountPerParticipant)
-	rightAmount := big.NewInt(0).Sub(nextState.Outcome.TotalAllocatedFor(request.Right)[asset], amountPerParticipant)
+
+	leftAmount := big.NewInt(0).Sub(nextState.Outcome.TotalAllocatedFor(request.Left)[asset], request.LeftAmount[asset])
+	rightAmount := big.NewInt(0).Sub(nextState.Outcome.TotalAllocatedFor(request.Right)[asset], request.RightAmount[asset])
+	total := big.NewInt(0).Add(request.LeftAmount[asset], request.RightAmount[asset])
 	if leftAmount.Cmp(big.NewInt(0)) < 0 {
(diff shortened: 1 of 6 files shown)


protocols/ledger/ledger.go
```diff
@@ -32,39 +32,44 @@ func (l *LedgerManager) HandleRequest(ledger *channel.TwoPartyLedger, request pr
...
+	// TODO: We should iterate over assets in RightAmount as well. For now we rely on the assumption that LeftAmount and RightAmount will have the same assets.
+	for asset := range request.LeftAmount {
+
+		currentLeftAmount := supported.Outcome.TotalAllocatedFor(request.Left)[asset]
+		currentRightAmount := supported.Outcome.TotalAllocatedFor(request.Right)[asset]
+		// Calculate the amounts
+		leftAmount := big.NewInt(0).Sub(currentLeftAmount, request.LeftAmount[asset])
+		rightAmount := big.NewInt(0).Sub(currentRightAmount, request.RightAmount[asset])
+		total := big.NewInt(0).Add(request.LeftAmount[asset], request.RightAmount[asset])
+		if leftAmount.Cmp(big.NewInt(0)) < 0 {
```
geoknee, line 46: You could reuse these utils here https://github.com/statechannels/go-nitro/blob/3f8b138550ba2ea7eb3966e2c5f7f98477549ac8/types/bigutils.go#L10

protocols/ledger/ledger.go
```diff
@@ -32,39 +32,44 @@ func (l *LedgerManager) HandleRequest(ledger *channel.TwoPartyLedger, request pr
...
+				},
+				outcome.Allocation{
+					Destination:    request.Destination,
+					Amount:         total,
+					AllocationType: outcome.GuaranteeAllocationType,
+					Metadata:       guarantee,
+				},
+			}}
+
+		nextState.Outcome = append(nextState.Outcome, newOutcome)
```
geoknee, line 71: Don't we also need to make some deductions from existing entries in the outcome?

geoknee (commented):
Looks good to me but one important outstanding question.

protocols/ledger/ledger.go
geoknee: Sorry I read it more carefully and looked at the test and I can see the deductions are there correctly. Variable names like `newLeftAmount` may have helped.

I'm still not sure about the `append`, though? Does that only work currently because there is no supported state on L35?

geoknee:
I note that this PR seems to include the changes from #247 ?

lalexgap:
> I note that this PR seems to include the changes from #247 ?

Oops, I've rebased this to separate out those changes.

protocols/ledger/ledger.go
lalexgap: This change has now been pulled out to #247 but I think you're right. Right now we could lose any exits for assets not in `leftAmount`.

protocols/ledger/ledger.go
lalexgap: Updated in [da0f79c](https://github.com/statechannels/go-nitro/pull/245/commits/da0f79c55e2eaa6a10cc72274441ea77c3cedbb3)

protocols/ledger/ledger.go
```diff
@@ -32,39 +32,44 @@ func (l *LedgerManager) HandleRequest(ledger *channel.TwoPartyLedger, request pr
...
+			return protocols.SideEffects{}, fmt.Errorf("Allocation for %x cannot afford the amount %d", request.Right, request.RightAmount[asset])
+		}
 
-	// Calculate the amounts
-	amountPerParticipant := big.NewInt(0).Div(request.Amount[asset], big.NewInt(2))
-	leftAmount := big.NewInt(0).Sub(nextState.Outcome.TotalAllocatedFor(request.Left)[asset], amountPerParticipant)
-	rightAmount := big.NewInt(0).Sub(nextState.Outcome.TotalAllocatedFor(request.Right)[asset], amountPerParticipant)
-	if leftAmount.Cmp(big.NewInt(0)) < 0 {
-		return protocols.SideEffects{}, fmt.Errorf("Allocation for %x cannot afford the amount %d", request.Left, amountPerParticipant)
+		newOutcome := outcome.SingleAssetExit{
```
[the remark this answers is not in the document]
lalexgap: Since `LedgerRequests` get handled outside of the `virtualfund` protocol I think it's possible that the ledger channel's outcome could change in the time between the request is generated and the request is fulfilled by the `LedgerManager`. I think that means we want to avoid the `LedgerRequest` containing a specific `SingleAssetExit` which could become outdate by the time we go to handle the request?

protocols/ledger/ledger.go
lalexgap: This has been pulled out to #247 so this discussion doesn't exactly apply to the current PR's code. That said in #247 it works by
- [Cloning](https://github.com/statechannels/go-nitro/blob/467a3efa30bfbf40f84b8a4f1d0c8f9a8ada46ad/protocols/ledger/ledger.go#L35) the state into `nextState`
- [Resetting](https://github.com/statechannels/go-nitro/blob/467a3efa30bfbf40f84b8a4f1d0c8f9a8ada46ad/protocols/ledger/ledger.go#L36) the outcome on `nextState`
- Creating new single asset exits based on the previous `supported.Outcome` and `ledgerRequest`

I think this does introduce a problem as we're only iterating through the assets of the ledger request. This means we would lose any `SingleAssetExit`s for assets not in the Ledger Request amounts.

protocols/ledger/ledger_test.go
```diff
@@ -63,15 +63,17 @@ func TestHandleLedgerRequest(t *testing.T) {
...
 	}
 	invalidRequest := protocols.LedgerRequest{
 		ObjectiveId: oId,
 		LedgerId:    ledger.Id,
 		Left:        left.Destination,
 		Right:       right.Destination,
 		Destination: destination,
-		Amount:      types.Funds{asset: big.NewInt(10)},
+		LeftAmount:  types.Funds{asset: big.NewInt(5)},
+		RightAmount: types.Funds{asset: big.NewInt(5)},
```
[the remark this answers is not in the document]
lalexgap: Updated in [4438dde](https://github.com/statechannels/go-nitro/pull/245/commits/4438dde7737e1c31cff617f10fd83d3d9893d83b)

protocols/ledger/ledger.go
```diff
@@ -32,39 +32,44 @@ func (l *LedgerManager) HandleRequest(ledger *channel.TwoPartyLedger, request pr
 	if err != nil {
 		return protocols.SideEffects{}, fmt.Errorf("error finding a supported state: %w", err)
 	}
-
-	asset := types.Address{} // todo: loop over request.amount's assets
 	nextState := supported.Clone()
+	nextState.Outcome = outcome.Exit{}
+	// TODO: We should iterate over assets in RightAmount as well. For now we rely on the assumption that LeftAmount and RightAmount will have the same assets.
```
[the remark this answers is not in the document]
lalexgap: Will update in #247

protocols/ledger/ledger.go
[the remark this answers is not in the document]
geoknee: Yeah you're right. I think this comment was caused by my same confusion as one of my other comments!

protocols/virtualfund/virtualfund.go
```diff
@@ -326,15 +326,24 @@ func (connection *Connection) ledgerChannelAffordsExpectedGuarantees() bool {
 func (s VirtualFundObjective) generateLedgerRequestSideEffects() protocols.SideEffects {
 	sideEffects := protocols.SideEffects{}
 	sideEffects.LedgerRequests = make([]protocols.LedgerRequest, 0)
+
+	leftAmount := s.V.PreFundState().Outcome.TotalAllocatedFor(s.V.MyDestination())
+	// TODO: This is hacky way of getting the second expected outcome.
+	other := s.V.PreFundState().Outcome[0].Allocations[1].Destination
```
geoknee, line 332: I think an elegant longer-term solution to this would be to define a type like so

```go
type VirtualChannel struct {
    Channel 
}
```

as we do with two party ledger channels:

https://github.com/statechannels/go-nitro/blob/c35aba4b207fc7cf6436160e872d140e64c0538f/channel/channel.go#L31-L47
 Then, we can have a constructor for this type which does runtime checks on the number of participants and the structure of the outcome, and write methods for `VirtualChannel` such as `V.GetLeftAmount` and `V.GetRightAmount`, similar to this: 

https://github.com/statechannels/go-nitro/blob/c35aba4b207fc7cf6436160e872d140e64c0538f/channel/channel.go#L93-L96
 
 Pinging @NiloCK for his thoughts.

geoknee (approved):

protocols/ledger/ledger.go
geoknee: It looks like the "resetting outcome" step is missing on #247 right now? (it's on the commit you linked to, which apparently does not belong to a branch).

I agree about the problem of missing assets which are not in the ledger request. But I think there is an additional problem -- even if the asset _is_ in the ledger request, aren't we are going to lose existing allocations that don't allocate to `left` and `right`? For example, allocations which guarantee other virtual channels?

geoknee: We can move the discussion to #247

protocols/virtualfund/virtualfund.go
lalexgap: I would definitely be in favour of something like this!

Another solution is modifying the `Outcome` struct so that you can get `types.Funds` for an index. Similar to [TotalAllocatedFor](https://github.com/statechannels/go-nitro/blob/ag/multiple-asset/channel/state/outcome/exit.go#L90:L90) but it would accept an index. This would be a less elegant but quicker solution to the TODO

PR #247, opened 2022-02-16
Handle multiple assets in handle request

lalexgap:
This updates the `handleRequest` method to iterate through multiple assets and create a `SingleAssetExit` for each asset as opposed to using a hardcoded asset.

The Ledger manager has been updated to call DivertToGuarantee which already handles the logic for updating an allocation with a guarantee.

The test has also been expanded to include handling a second ledger request on the same ledger channel. This makes sure that existing guarantees don't get removed (which @geoknee identified in https://github.com/statechannels/go-nitro/pull/250)

### DIFF
--- a/protocols/ledger/ledger.go
+++ b/protocols/ledger/ledger.go
@@ -23,11 +23,6 @@ func NewLedgerManager() LedgerManager {
 // It returns a signed state message that can be sent to other participants.
 func (l *LedgerManager) HandleRequest(ledger *channel.TwoPartyLedger, request protocols.LedgerRequest, secretKey *[]byte) (protocols.SideEffects, error) {
 
-	guarantee, _ := outcome.GuaranteeMetadata{
-		Left:  request.Left,
-		Right: request.Right,
-	}.Encode()
-
 	supported, err := ledger.Channel.LatestSupportedState()
 	if err != nil {
 		return protocols.SideEffects{}, fmt.Errorf("error finding a supported state: %w", err)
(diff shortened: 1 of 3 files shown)


geoknee:
Following discussion on the other PR, I made this example #250 to highlight the problem of not preserving existing guarantees in the ledger channel. #250 modifies the code on `main`, but I also tried rebasing it on this branch, and got the same test failure.

lalexgap:
> Following discussion on the other PR, I made this example #250 to highlight the problem of not preserving existing guarantees in the ledger channel.

Good catch! I've updated this Ledger Manager to use the DivertToGuarantee which preserves existing guarantees. I've also updated the ledger test to handle a second request, which would catch this issue.

protocols/ledger/ledger.go
```diff
@@ -23,55 +23,42 @@ func NewLedgerManager() LedgerManager {
...
-	currentRightAmount := nextState.Outcome.TotalAllocatedFor(request.Right)[asset]
-	// Calculate the new amounts by subtracting the requested amounts from the current amounts
-	leftAmount := big.NewInt(0).Sub(currentLeftAmount, request.LeftAmount[asset])
-	rightAmount := big.NewInt(0).Sub(currentRightAmount, request.RightAmount[asset])
+	for i, exit := range nextState.Outcome {
+		asset := exit.Asset
+		// If our request doesn't deal with this asset, skip it
+		if types.IsZero(request.LeftAmount[asset]) && types.IsZero(request.RightAmount[asset]) {
+			continue
+		}
```
geoknee, line 37: Is this part unit tested? If the request doesn't deal with an asset in the `nextState.Outcome`, I think this attempted access may result in a panic instead of a zero amount being returned?

protocols/ledger/ledger.go
```diff
@@ -23,55 +23,42 @@ func NewLedgerManager() LedgerManager {
...
+		leftAmount := big.NewInt(0).Sub(currentLeftAmount, request.LeftAmount[asset])
+		rightAmount := big.NewInt(0).Sub(currentRightAmount, request.RightAmount[asset])
+
+		// If any participant cannot afford the request amount, return an error
+		if types.Lt(leftAmount, big.NewInt(0)) {
+			return protocols.SideEffects{}, fmt.Errorf("Allocation for %x cannot afford the amount %d", request.Left, request.LeftAmount[asset])
+		}
+		if types.Lt(rightAmount, big.NewInt(0)) {
+			return protocols.SideEffects{}, fmt.Errorf("Allocation for %x cannot afford the amount %d", request.Right, request.RightAmount[asset])
+		}
```
geoknee, line 51: I think all of these lines are unnecessary, since the `.DivertToGuarantee` method already performs these checks

protocols/ledger/ledger.go
```diff
@@ -23,55 +23,42 @@ func NewLedgerManager() LedgerManager {
...
 
-	// If any participant cannot afford the request amount, return an error
-	if types.Lt(leftAmount, big.NewInt(0)) {
-		return protocols.SideEffects{}, fmt.Errorf("Allocation for %x cannot afford the amount %d", request.Left, request.LeftAmount[asset])
-	}
-	if types.Lt(rightAmount, big.NewInt(0)) {
-		return protocols.SideEffects{}, fmt.Errorf("Allocation for %x cannot afford the amount %d", request.Right, request.RightAmount[asset])
-	}
+		// Get an updated allocation with the guarantee
+		newAlloc, err := nextState.Outcome[i].Allocations.DivertToGuarantee(request.Left, request.Right, request.LeftAmount[asset], request.RightAmount[asset], request.Destination)
```
geoknee, line 54: It might be nice to push this logic into a `DivertToGuarantee` method on the `Outcome` type.

geoknee (approved):

geoknee:
In the interests of making progress faster, I'll merge this and point to a PR where I have made some improvements #262

PR #248, opened 2022-02-16
rename and document `ss.SignAndAdd` => `ss.Sign`

geoknee:
This is a pretty minor rename, which makes a little more sense to me. It might be subjective, though!

The issue with "Sign and add" to me is that it is slightly ambiguous _what_ is being added (is it a state or signature). Coupled with the lack of documentation, we have to read through the source code to understand it.


lalexgap (approved):
Sounds good to me! My only concern is that `State` also has a `Sign` function that behaves slightly differently which may be confusing reading through the code. Probably doesn't matter too much!

PR #250, opened 2022-02-17
add "existing" virtual channel to test ledger outcome

geoknee:
This is just an example of how our unit tests are not opinionated enough about our ledger handler. 

I'm raising this mostly to highlight the issue, but I think we need something like this. 

I'm not sure it's a good approach overall as the (additional) test data is hidden inside the `CreateTestLedger` function, and then replicated in the test body itself.


protocols/ledger/ledger_test.go
```diff
@@ -131,6 +137,7 @@ func TestHandleLedgerRequest(t *testing.T) {
 	expectedMessage := protocols.Message{To: bob.address, ObjectiveId: oId, SignedStates: []state.SignedState{expectedSigned}}
 
 	if diff := cmp.Diff(sideEffects.MessagesToSend[0], expectedMessage); diff != "" {
+		t.Log(cmp.Diff(sideEffects.MessagesToSend[0].SignedStates[0].State().Outcome[0], expectedState.Outcome[0]))
```
geoknee, line 140: This line is triggered and outputs: 


```shell
        - 		Allocations: outcome.Allocations{
        - 			{Destination: s"0x000000000000000000000000f5a1bb"..., Amount: s"1"},
        - 			{Destination: s"0x000000000000000000000000760bf2"..., Amount: s"0"},
        - 			{
        - 				Destination:    s"0x0000000000000000000000005e29e5"...,
        - 				Amount:         s"4",
        - 				AllocationType: 1,
        - 				Metadata:       []uint8{...},
        - 			},
        - 		},
        - 	},
        + 	{
        + 		Allocations: outcome.Allocations{
        + 			{Destination: s"0x000000000000000000000000f5a1bb"..., Amount: s"1"},
        + 			{Destination: s"0x000000000000000000000000760bf2"..., Amount: s"0"},
        + 			{
        + 				Destination:    s"0x000000000000000000000000000000"...,
        + 				Amount:         s"2",
        + 				AllocationType: 1,
        + 				Metadata:       []uint8{...},
        + 			},
        + 			{
        + 				Destination:    s"0x0000000000000000000000005e29e5"...,
        + 				Amount:         s"4",
        + 				AllocationType: 1,
        + 				Metadata:       []uint8{...},
        + 			},
        + 		},
        + 	},

```

lalexgap:
I've updated the test in https://github.com/statechannels/go-nitro/pull/247 to now handle a second ledger request and check the results. This means that we're now checking the case of appending a guarantee to an outcome with existing guarantees similar to this test.

This problem was fixed by using the DivertToGuarantee function on `Allocations` which properly appends the guarantee.

I think with these changes we can probably close this?

PR #251, opened 2022-02-17
check destination field in `LedgerRequest.Equal`

### LINKED ISSUE #243: Check that destinations are equal in `ledgerRequest.Equal`
```suggestion
	return l.LedgerId == m.LedgerId &&
		l.Amount.Equal(m.Amount) &&
		l.Left == m.Left &&
		l.Right == m.Right &&
		l.ObjectiveId == m.ObjectiveId
```

This `Equal` method doesn't check that the `Destionation` fields are equal which seems like an oversight.

_Originally posted by @NiloCK in https://github.com/statechannels/go-nitro/pull/236#discussion_r806918691_

geoknee:
and reorder fields to match type declaration

Closes #243

### DIFF
--- a/protocols/interfaces.go
+++ b/protocols/interfaces.go
@@ -25,12 +25,13 @@ type LedgerRequest struct {
 
 // Equal checks for equality between the receiver and a second LedgerRequest
 func (l LedgerRequest) Equal(m LedgerRequest) bool {
-	return l.LedgerId == m.LedgerId &&
+	return l.ObjectiveId == m.ObjectiveId &&
+		l.LedgerId == m.LedgerId &&
+		l.Destination == m.Destination &&
 		l.LeftAmount.Equal(m.LeftAmount) &&
 		l.RightAmount.Equal(m.RightAmount) &&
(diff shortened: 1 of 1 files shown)


protocols/interfaces.go
```diff
@@ -24,11 +24,12 @@ type LedgerRequest struct {
 
 // Equal checks for equality between the receiver and a second LedgerRequest
 func (l LedgerRequest) Equal(m LedgerRequest) bool {
-	return l.LedgerId == m.LedgerId &&
+	return l.ObjectiveId == m.ObjectiveId &&
```
lalexgap, line 27: Outside the scope of this PR but it'd be nice if we have an easy way to test `Equal` functions that would have caught this.  Maybe using the [cmp](https://pkg.go.dev/github.com/google/go-cmp/cmp) package or reflection?

geoknee: Yes -- I think that's a good idea and doesn't contradict the advice to not used `go-cmp` in production code.

PR #260, opened 2022-02-17
Add a single hop virtual channel

lalexgap:
While working on an `AsIrene` test I ran into bug with how we're calculating the left amount 
```
leftAmount := s.V.PreFundState().Outcome.TotalAllocatedFor(s.V.MyDestination())
```
As Irene there won't be an allocation for `s.V.MyDestination()`, what we actually want is the first allocation(for Alice).

Based on @geoknee [idea](https://github.com/statechannels/go-nitro/pull/245#discussion_r808856150) Ive added a new `SingleHopVirtualChannel` struct. 

`SingleHopVirtualChannel` checks that the channel has the correct amount of participants/allocations. There are also functions `LeftAmount` and `RightAmount` to get the amounts, so the virtual funding objective doesn't have to calculate them.

### DIFF
--- a/channel/channel.go
+++ b/channel/channel.go
@@ -29,6 +29,60 @@ type Channel struct {
 	// Longer term, we should have a more efficient and smart mechanism to store states https://github.com/statechannels/go-nitro/issues/106
 }
 
+type SingleHopVirtualChannel struct {
+	Channel
+}
+
+func NewSingleHopVirtualChannel(s state.State, myIndex uint) (*SingleHopVirtualChannel, error) {
+	if myIndex > 2 {
+		return &SingleHopVirtualChannel{}, errors.New("myIndex in a single hope virtual channel must be 0 1,or 2")
+	}
+	if len(s.Participants) != 3 {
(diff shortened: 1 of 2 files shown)


channel/channel.go
```diff
@@ -29,6 +29,60 @@ type Channel struct {
 	// Longer term, we should have a more efficient and smart mechanism to store states https://github.com/statechannels/go-nitro/issues/106
 }
 
+type SingleHopVirtualChannel struct {
+	Channel
+}
+
+func NewSingleHopVirtualChannel(s state.State, myIndex uint) (*SingleHopVirtualChannel, error) {
+	if myIndex > 2 {
+		return &SingleHopVirtualChannel{}, errors.New("myIndex in a single hope virtual channel must be 0 1,or 2")
```
geoknee, line 38:
```suggestion
		return &SingleHopVirtualChannel{}, errors.New("myIndex in a single hop virtual channel must be 0, 1, or 2")
```

geoknee (approved):
I like this approach. You may not be surprised :-)

I would ask that we add as much Godoc to this code as possible. Its a really good way to declare what the function / type should be doing, which in turn makes reviewing it easier.

PR #255, opened 2022-02-17
Add an `asBob` subtest to `virtualfund-single-hop_test.go`

geoknee:
Currently, we only test the virtual fund protocol over a single hop (3 participants in total), and never as "bob".  Thus there may be bugs in the code which are not surfaced yet. 

This approach adds such a test, and does so by:

- copy-pasting the "as Alice" test 
- making modifications so that it reflects Bob's situation
- ~fixing a bug in the `virtualfund` protocol where we used `n+1` instead of `n`~ this was cherry-picked off to #256 

This is a good approach because it solves the problem in a very verbose way that can be easily checked line-by-line. The downside is that the test is very repetitive and a little tiresome to read through. It is not easy to see the essential differences to the `asAlice` test.  I propose that we tackle refactoring the test (to DRY it out) in a future issue/PR. 

Additionally, I:
- Fixed a bug that confused `myRole` with `myIndex`. Easily done!
- tidied up some test types and functions that polluted the package

Towards #218.

As the change set is quite large, at the request of the reviewer I can split up the additions `TestNew`, `TestUpdate` and `TestCrank`. 

---

#### Code quality

- [x] I have written clear commit messages
- [ ] I have performed a self-review of my own code
- [ ] This change does not have an unduly wide scope
- [x] I have separated logic changes from refactor changes (formatting, renames, etc.)
- [x] I have commented my code wherever necessary (can be 0)
- [ ] I have added tests that prove my fix is effective or that my feature works, if necessary

#### Project management

- [x] I have assigned myself to this PR
- [ ] I have linked the appropriate github issue

Issue #258, opened 2022-02-17
DRY out `virtualfund-single-hop_test.go`


geoknee:
This looks good to me!
 
> I propose that we tackle refactoring the test (to DRY it out) in a future issue/PR.

I'd be in favour of doing this soon! It's already getting quite large and difficult to follow.

_Originally posted by @lalexgap in https://github.com/statechannels/go-nitro/pull/255#pullrequestreview-886581079_

PR #261, opened 2022-02-18
Add AsIrene test

lalexgap:
⚠️ Based on #260 

Adds a test to the virtual funding test for the intermediary role. To keep it consistent with current naming in the file I've called it `AsP1` instead of `AsIrene`. I've followed a similar approach to #255 and just copied an existing test and tailored it to Irene.

The only fix required to get the test working was [a change to how we calculate left/right amounts.](https://github.com/statechannels/go-nitro/commit/ce913d57c466b82e5f440ecdb86c3846aa56bab1)

#### Code quality

#### Project management


geoknee (approved):
Looks good. I pushed some commits which I think should be uncontroversial. It is kind of screaming to be tidied up now, so I'll take a stab #258!

PR #262, opened 2022-02-18
Add `Exit.DivertToGuarantee` method

geoknee:
We currently have an `Allocations.DivertToGuarantee` method which applies to a single asset. In the ledger protocol, we loop over assets and call into this method. It would fit existing patterns better, and lead to greater encapsulation /  abstraction and readability if we had a method on the `Exit` type which was responsible for that looping behaviour. 

**Changes**

- Move logic to `outcome` package
- fix bug where a "missing" asset would result in a panic
- Add a unit test for this

---

#### Code quality

#### Project management

### DIFF
--- a/channel/state/outcome/exit.go
+++ b/channel/state/outcome/exit.go
@@ -192,7 +192,7 @@ func (e Exit) Affords(
 
 }
 
-// DivertToGuarantee returns a new Exit, identical to the reciever but with
+// DivertToGuarantee returns a new Exit, identical to the receiver but with
 // (for each asset of the Exit)
 // the leftDestination's amount reduced by leftFunds[asset],
 // the rightDestination's amount reduced by rightAmount[asset],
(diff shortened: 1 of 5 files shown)


channel/state/outcome/exit.go
```diff
@@ -190,3 +191,42 @@ func (e Exit) Affords(
...
+	rightFunds types.Funds,
+	guaranteeDestination types.Destination,
+) (Exit, error) {
+
+	f := e.Clone()
+
+	for i, sae := range f {
+		asset := sae.Asset
+
+		leftAmount, leftOk := leftFunds.Clone()[asset]
```
lalexgap, line 214: nit but we could probably clone the `leftFunds` and `rightFunds` once instead of cloning it again for every asset.

lalexgap (approved):

channel/state/outcome/exit.go
geoknee: [0ec03ba](https://github.com/statechannels/go-nitro/pull/262/commits/0ec03ba128e5731c4b5a4dba03247e47a6cb5eb8)

PR #267, opened 2022-02-18
Return early and log when objective is not found in `handleChainEvent`.

geoknee:
Currently, an "irrelevant" chain event can cause a panic in a client. It hears the event, tries to pull an objective from the store, ignores the `ok` boolean, and then tries to do stuff with this nil objective. 

Towards #217 .

This approach checks the `ok` boolean and returns early without trying to flog the proverbial dead horse 🐴 . We write to the logger for good measure. I don't think we need to treat this as an error.

---

#### Code quality

#### Project management


lalexgap (approved):
Longer term this might get a bit spammy in the logs but for now it makes sense to me.

PR #269, opened 2022-02-21
Overhaul `Channel.Clone()`

geoknee:
There are several problems with our existing `Channel.Clone()` method, as well as the `SingleHopVirtualChannel.Clone()` and `TwoPartyLedger.Clone()` methods which depend on it. These methods are critically important because they lie at the heart of the "functional core" of `go-nitro`. 

Problem 1: 
`TwoPartyLedger.Clone()` returns a shallow clone. We need a deep clone. 

Problem 2: 
There are no unit tests for `SingleHopVirtualChannel.Clone()` or `TwoPartyLedger.Clone()`.

Problem 3:
The implementation of `Channel.Clone()` uses a value receiver. This was a bit strange, to me, and made it difficult to understand whether the implementation was correct. A pointer receiver seems more natural.

Problem 4:
`Channel.Clone()` returns a `Channel` instead of a `*Channel`. It is more natural to return the pointer, as we mostly work with pointers to channels. 

Problem 5:
If we tried to clone a nil channel, we could get a panic. In the virtual funding protocol, we use nil channels for Alice and Bob (who are at the ends of the chain of participants and have only a single ledger connection, instead of two). When testing this protocol, it is very useful to be able to simply clone a channel and have that work even if it is nil. 

This PR addresses all of the above problems. 

Towards #226 

---

#### Code quality

#### Project management

### DIFF
--- a/channel/channel.go
+++ b/channel/channel.go
@@ -83,8 +83,12 @@ func (v SingleHopVirtualChannel) RightAmount() types.Funds {
 	return v.amountAtIndex(1)
 }
 
-func (v SingleHopVirtualChannel) Clone() SingleHopVirtualChannel {
-	return SingleHopVirtualChannel{v.Channel.Clone()}
+func (v *SingleHopVirtualChannel) Clone() *SingleHopVirtualChannel {
+	if v == nil {
+		return nil
+	}
+	w := SingleHopVirtualChannel{*v.Channel.Clone()}
+	return &w
 }
 
 type TwoPartyLedger struct {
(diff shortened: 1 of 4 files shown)


channel/channel.go
```diff
@@ -154,15 +162,18 @@ func (lc TwoPartyLedger) TheirDestination() types.Destination {
 }
 
 // Clone returns a deep copy of the receiver
-func (c Channel) Clone() Channel {
-	clonedSignedStateForTurnNum := map[uint64]state.SignedState{}
+func (c *Channel) Clone() *Channel {
+	if c == nil {
+		return nil
+	}
+	d, _ := New(c.PreFundState().Clone(), c.MyIndex)
```
lalexgap, line 169: I think we're ignoring the error here because we know `c.PreFundState().Clone()` starts with turnNum=0 and so we'd never get an error back from `New`?  We could add a comment explaining that.

lalexgap (approved):

Issue #259, opened 2022-02-17
`signState` can be removed from `virtualfund-single-hop_test.go` and we can use `ss.Sign` instead.


geoknee:
Not related to this PR but I think `signState` can be removed and we can use `ss.Sign` instead.

_Originally posted by @lalexgap in https://github.com/statechannels/go-nitro/pull/255#discussion_r809483990_

PR #272, opened 2022-02-22
DRY out virtual fund unit tests

### LINKED ISSUE #258: DRY out `virtualfund-single-hop_test.go`

geoknee:
Closes #258 , #226, #259
---

#### Code quality

#### Project management

### DIFF
--- a/protocols/virtualfund/virtualfund-single-hop_test.go
+++ b/protocols/virtualfund/virtualfund-single-hop_test.go
@@ -82,51 +82,52 @@ func TestSingleHopVirtualFund(t *testing.T) {
 
 	TestAs := func(my actor, t *testing.T) {
 
-		// IMPORTANT these are templates. Clone them before using to prevent sharing mutable data between tests
-		var l *channel.TwoPartyLedger
-		var r *channel.TwoPartyLedger
+		// prepareLedgerChannels constructs and return two new ledger channels. Depending on the role, one of them may be nil.
+		prepareLedgerChannels := func(role uint) (*channel.TwoPartyLedger, *channel.TwoPartyLedger) {
+			var l *channel.TwoPartyLedger
+			var r *channel.TwoPartyLedger
+			switch my.role {
+			case 0:
+				{
+					r, _ = ledger.CreateTestLedger(
+						outcome.Allocation{Destination: my.destination, Amount: big.NewInt(5)},
+						outcome.Allocation{Destination: p1.destination, Amount: big.NewInt(5)},
+						&my.privateKey, 0, big.NewInt(0))
+					ledger.SignPreAndPostFundingStates(r, []*[]byte{&alice.privateKey, &p1.privateKey}) // TODO these steps could be absorbed into CreateTestLedger
 
-		switch my.role {
-		case 0:
-			{
-				r, _ = ledger.CreateTestLedger(
-					outcome.Allocation{Destination: my.destination, Amount: big.NewInt(5)},
-					outcome.Allocation{Destination: p1.destination, Amount: big.NewInt(5)},
-					&my.privateKey, 0, big.NewInt(0))
-				ledger.SignPreAndPostFundingStates(r, []*[]byte{&alice.privateKey, &p1.privateKey}) // TODO these steps could be absorbed into CreateTestLedger
+				}
+			case 1:
+				{
+					l, _ = ledger.CreateTestLedger(
+						outcome.Allocation{Destination: alice.destination, Amount: big.NewInt(5)},
+						outcome.Allocation{Destination: my.destination, Amount: big.NewInt(5)},
+						&alice.privateKey, 1, big.NewInt(0))
+					r, _ = ledger.CreateTestLedger(
+						outcome.Allocation{Destination: my.destination, Amount: big.NewInt(5)},
+						outcome.Allocation{Destination: bob.destination, Amount: big.NewInt(5)},
+						&alice.privateKey, 0, big.NewInt(0))
+					ledger.SignPreAndPostFundingStates(l, []*[]byte{&alice.privateKey, &p1.privateKey})
+					ledger.SignPreAndPostFundingStates(r, []*[]byte{&p1.privateKey, &bob.privateKey})
+				}
+			case 2:
+				{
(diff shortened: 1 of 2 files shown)


protocols/virtualfund/virtualfund-single-hop_test.go
```diff
@@ -171,24 +149,58 @@ func TestSingleHopVirtualFund(t *testing.T) {
...
+			mySig, _ := o.V.PreFundState().Sign(my.privateKey)
+			_ = expectedSignedState.AddSignature(mySig)
 
-			forBob := protocols.Message{To: bob.address, ObjectiveId: o.Id(), SignedStates: []state.SignedState{expectedSignedState}}
-			forIrene := protocols.Message{To: p1.address, ObjectiveId: o.Id(), SignedStates: []state.SignedState{expectedSignedState}}
-			want := protocols.SideEffects{MessagesToSend: []protocols.Message{forIrene, forBob}}
+			want := protocols.SideEffects{MessagesToSend: []protocols.Message{}}
+			switch my.role {
+			case 0:
+				{
```
lalexgap, line 158: I think a helper function that checks if a signed state is included in a message would DRY this out a lot.

protocols/virtualfund/virtualfund-single-hop_test.go
```diff
@@ -171,24 +149,58 @@ func TestSingleHopVirtualFund(t *testing.T) {
...
+
 			// Manually progress the extended state by collecting prefund signatures
+			aliceSig, _ := vPreFund.Sign(alice.privateKey)
 			bobSig, _ := vPreFund.Sign(bob.privateKey)
 			p1Sig, _ := vPreFund.Sign(p1.privateKey)
-			o.V.AddStateWithSignature(vPreFund, bobSig)
-			o.V.AddStateWithSignature(vPreFund, p1Sig)
+			switch my.role {
+			case 0:
+				{
```
lalexgap, line 189: This might be another good candidate for a helper function. It could probably be smart enough to figure out which signatures are missing and add them so we don't have to have a case statement.

protocols/virtualfund/virtualfund-single-hop_test.go
```diff
@@ -97,58 +77,56 @@ func TestSingleHopVirtualFund(t *testing.T) {
...
-				Destination:    o.V.Id,
-				Amount:         big.NewInt(0).Set(vPreFund.VariablePart().Outcome[0].TotalAllocated()),
-				AllocationType: outcome.GuaranteeAllocationType,
-				Metadata:       expectedEncodedGuaranteeMetadata,
+	vPostFund := vPreFund.Clone()
+	vPostFund.TurnNum = 1
+
+	TestAs := func(my actor, t *testing.T) {
+
+		// IMPORTANT these are templates. Clone them before using to prevent sharing mutable data between tests
```
lalexgap, line 85: We could move creating these ledger channels into a function and just call that function from each test. This would avoid having to remember to clone the channels at the start of the test.

protocols/virtualfund/virtualfund-single-hop_test.go
```diff
@@ -171,24 +149,58 @@ func TestSingleHopVirtualFund(t *testing.T) {
...
-			o.V.AddStateWithSignature(vPreFund, p1Sig)
+			switch my.role {
+			case 0:
+				{
+					o.V.AddStateWithSignature(vPreFund, bobSig)
+					o.V.AddStateWithSignature(vPreFund, p1Sig)
+				}
+			case 1:
+				{
+					o.V.AddStateWithSignature(vPreFund, aliceSig) // TODO is this necessary?
```
lalexgap, line 195: Is there a reason this wouldn't be necessary?

protocols/virtualfund/virtualfund-single-hop_test.go
```diff
@@ -504,297 +382,121 @@ func TestSingleHopVirtualFund(t *testing.T) {
...
+					}
+				}
 			}
 
 		}
-
-		t.Run(`New`, testNew)
-		t.Run(`Update`, testUpdate)
 		t.Run(`Crank`, testCrank)
+		t.Run(`Update`, testUpdate)
```
lalexgap, line 495: Is there a reason we removed the `testNew` test?

lalexgap (approved):
This looks good to me! I think there's some opportunities to DRY it out even further by creating some helper functions but this is great first step!

protocols/virtualfund/virtualfund-single-hop_test.go
geoknee: 7b98783e540dabd6fc6bee914bbb55909a1708ae

protocols/virtualfund/virtualfund-single-hop_test.go
geoknee: If Alice had proposed V, then P1 and Bob will have the prefund signature *before* their objective is cranked for the first time -- going by our current engine implementation. That's why I was thinking it may not be necessary to add the prefund signature again: of course we are not driving the protocol with the engine in this test -- so the test order doesn't exactly match the order we expect. Thankfully though, we have designed the system to be very insensitive to the order of these things. So I will just remove these comments for now.

protocols/virtualfund/virtualfund-single-hop_test.go
geoknee: 47fc3b77adf2afd44f04ce75eb0c6d696c0b2b14

protocols/virtualfund/virtualfund-single-hop_test.go
geoknee: No reason actually, I just forgot to put it back in! Fixed in [c24e4f8](https://github.com/statechannels/go-nitro/pull/272/commits/c24e4f825b6378f4b9fb0bac035b74cd14221d5a)

Issue #257, opened 2022-02-17
Instead of accepting a `n` value in `New` could we just figure it out based on the participants in `vPreFund` ?


geoknee:
Instead of accepting a `n` value in `New` could we just figure it out based on the participants in `vPreFund` ?

_Originally posted by @lalexgap in https://github.com/statechannels/go-nitro/pull/256#discussion_r809385505_

geoknee:
The same goes for `myRole`.

PR #274, opened 2022-02-23
Get Virtual Funding Integration Test Passing

### LINKED ISSUE #217: Write an integration test for virtually funding a channel between three parties

### LINKED ISSUE #275: Update constructObjectiveFromMessage to support virtual fund objectives
Update [constructObjectiveFromMessage](https://github.com/statechannels/go-nitro/blob/312cc94f96f63c604dbd07f3d2caf7f22701b335/client/engine/engine.go#L227) to handle virtual fund objectives.

lalexgap:
Fixes #217 

This includes the changes from #280 and #281 

This PR introduces a virtual funding integration test and contains a collection of changes to get the test working.

The integration test is now passing on CI 🎉 :
```
.....
0x111A00868581f73AB42FEEF67D235Ca09ca1E8db: 2022/02/25 00:37:40 engine.go:219: Objective VirtualFund-0xe25bb77c1b36b618f553c1fab4b473c53c2a67a1c63591bdc2a72b244e039fd6 is WaitingForNothing
0xBBB676f9cFF8D242e9eaC39D063848807d3D1D94: 2022/02/25 00:37:40 engine.go:219: Objective VirtualFund-0xe25bb77c1b36b618f553c1fab4b473c53c2a67a1c63591bdc2a72b244e039fd6 is WaitingForCompletePostFund
0xBBB676f9cFF8D242e9eaC39D063848807d3D1D94: 2022/02/25 00:37:40 engine.go:121: Handling inbound message {0xBBB676f9cFF8D242e9eaC39D063848807d3D1D94 VirtualFund-0xe25bb77c1b36b618f553c1fab4b473c53c2a67a1c63591bdc2a72b244e039fd6 [{{0xc000271880 [[170 166 98 142 196 74 138 116 41 135 239 58 17 77 223 226 212 247 173 206] [17 26 0 134 133 129 247 58 180 47 238 246 125 35 92 160 156 161 232 219] [187 182 118 249 207 248 210 66 233 234 195 157 6 56 72 128 125 61 29 148]] 0xc0002718a0 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0] 0xc0002718c0 [] [{[0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0] [] [{[0 0 0 0 0 0 0 0 0 0 0 0 170 166 98 142 196 74 138 116 41 135 239 58 17 77 223 226 212 247 173 206] 0xc0002718e0 0 []} {[0 0 0 0 0 0 0 0 0 0 0 0 17 26 0 134 133 129 247 58 180 47 238 246 125 35 92 160 156 161 232 219] 0xc000271900 0 []}]}] 1 false} map[1:{[104 206 53 50 126 77 245 91 175 69 182 194 226 7 171 22 208 134 29 244 98 142 2 102 239 227 219 42 236 92 19 78] [80 128 163 91 255 207 160 38 246 136 106 152 228 28 194 186 160 34 62 148 3 10 87 71 38 90 245 79 0 189 111 142] 1}]}]}
0xAAA6628Ec44A8a742987EF3A114dDFE2D4F7aDCE: 2022/02/25 00:37:40 engine.go:219: Objective VirtualFund-0xe25bb77c1b36b618f553c1fab4b473c53c2a67a1c63591bdc2a72b244e039fd6 is WaitingForNothing
0xBBB676f9cFF8D242e9eaC39D063848807d3D1D94: 2022/02/25 00:37:40 engine.go:219: Objective VirtualFund-0xe25bb77c1b36b618f553c1fab4b473c53c2a67a1c63591bdc2a72b244e039fd6 is WaitingForNothing
```

# Changes

This is a list of various supporting changes made to get the virtual funding integration passing.

- [360e88d](https://github.com/statechannels/go-nitro/pull/274/commits/360e88d14c3e301520e8c4bb9f8ef4c01136e573): Add a basic `CreateVirtualChannel` API function so we can call it from the integration test.
- [b034820](https://github.com/statechannels/go-nitro/pull/274/commits/b034820464177e3823b237d70ac85d04f6f75699): Fix #275 by constructing a virtual funding objective from a message
- [467a27c](https://github.com/statechannels/go-nitro/pull/274/commits/467a27cf1ea7949e1ddd752b598da762c8fd347f): Add a `SetChannel` function so we can update the store after handling ledger requests. 
- [8401cb3](https://github.com/statechannels/go-nitro/pull/274/commits/8401cb316015f616e2f7698b874c47bef63b830d): Make sure we consistently call `SetObjective` whenever we mutate data
- [9b5d134](https://github.com/statechannels/go-nitro/pull/274/commits/9b5d134a0ff2be97ace54f181b7b00c4a19c6808) & [162fdf5](https://github.com/statechannels/go-nitro/pull/274/commits/162fdf5b50483b08e38b3b52a75741bec1c36cee): Make sure that Irene generates the exact same `LedgerRequest`s as Bob and Alice, so all ledger states match exactly. Previously we were using `MyDestination/TheirDestination` for `Left` and `Right` so Irene would generate a different request than Alice/Bob.
- [9e9f7ed](https://github.com/statechannels/go-nitro/pull/274/commits/9e9f7edda71b27334c9247ca245b5a6b1b7c3d4a): Remove `GetChannel`. Right now multiple objectives can have the same channel. However when we call `SetObjective` we only update one objective. This means that the channels on objectives can get out of sync. This should get fixed in https://github.com/statechannels/go-nitro/pull/227 but for now I just removed `GetChannel` and rely on `GetObjectiveById`
- [033bf82](https://github.com/statechannels/go-nitro/pull/274/commits/033bf82c2f0dfe41c065526edac70dcbb85a513a): Call SetChannel in HandleLedgerRequests. Previously we were mutating the ledger channel but not setting it in the store correctly

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -2,9 +2,11 @@
 package client // import "github.com/statechannels/go-nitro/client"
 
 import (
+	"fmt"
 	"io"
 	"math/big"
 
+	"github.com/statechannels/go-nitro/channel"
 	"github.com/statechannels/go-nitro/channel/state"
 	"github.com/statechannels/go-nitro/channel/state/outcome"
 	"github.com/statechannels/go-nitro/client/engine"
@@ -13,6 +15,7 @@ import (
 	"github.com/statechannels/go-nitro/client/engine/store"
 	"github.com/statechannels/go-nitro/protocols"
 	"github.com/statechannels/go-nitro/protocols/directfund"
+	"github.com/statechannels/go-nitro/protocols/virtualfund"
 	"github.com/statechannels/go-nitro/types"
 )
 
@@ -60,6 +63,44 @@ func (c *Client) handleEngineEvents() {
 // CompletedObjectives returns a chan that receives a objective id whenever that objective is completed
 func (c *Client) CompletedObjectives() <-chan protocols.ObjectiveId {
 	return c.completedObjectives
+}
+
+func (c *Client) CreateVirtualChannel(counterParty types.Address, intermediary types.Address, appDefinition types.Address, appData types.Bytes, outcome outcome.Exit, challengeDuration *types.Uint256) protocols.ObjectiveId {
+	right, ok := (*c.store).GetTwoPartyLedger(*c.Address, intermediary)
+
+	if !ok {
+		// TODO: We need to implement proper API error handling
+		panic(fmt.Sprintf("Could not find ledger channel for participants %v,%v", *c.Address, intermediary))
+	}
+
+	var left *channel.TwoPartyLedger
+
+	// Convert the API call into an internal event.
+	objective, _ := virtualfund.New(true,
+		state.State{
+			ChainId:           big.NewInt(0), // TODO
+			Participants:      []types.Address{*c.Address, intermediary, counterParty},
+			ChannelNonce:      big.NewInt(0), // TODO -- how do we get a fresh nonce safely without race conditions? Could we conisder a random nonce?
+			AppDefinition:     appDefinition,
+			ChallengeDuration: challengeDuration,
+			AppData:           appData,
+			Outcome:           outcome,
+			TurnNum:           0,
+			IsFinal:           false,
+		},
+		*c.Address,
+		1, // We only handle a single hop for now
+		0, // We always play the role of alice if we initiate the channel
+		left, right)
+
+	// Pass in a fresh, dedicated go channel to communicate the response:
+	apiEvent := engine.APIEvent{
+		ObjectiveToSpawn: objective,
+	}
+	// Send the event to the engine
+	c.engine.FromAPI <- apiEvent
+
+	return objective.Id()
 }
 
 // CreateDirectChannel creates a directly funded channel with the given counterparty
--- a/client/client.go
+++ b/client/client.go
@@ -21,6 +21,7 @@ type Client struct {
 	engine              engine.Engine // The core business logic of the client
 	Address             *types.Address
 	completedObjectives chan protocols.ObjectiveId
+	store               *store.Store
 }
 
 // New is the constructor for a Client. It accepts a messaging service, a chain service, and a store as injected dependencies.
@@ -29,7 +30,7 @@ func New(messageService messageservice.MessageService, chainservice chainservice
 	c.Address = store.GetAddress()
 	c.engine = engine.New(messageService, chainservice, store, logDestination)
 	c.completedObjectives = make(chan protocols.ObjectiveId, 100)
-
+	c.store = &store
 	// Start the engine in a go routine
 	go c.engine.Run()
 
--- a/client/client.go
+++ b/client/client.go
@@ -93,7 +93,6 @@ func (c *Client) CreateVirtualChannel(counterParty types.Address, intermediary t
 		0, // We always play the role of alice if we initiate the channel
 		left, right)
 
-	// Pass in a fresh, dedicated go channel to communicate the response:
 	apiEvent := engine.APIEvent{
 		ObjectiveToSpawn: objective,
 	}
@@ -121,7 +120,6 @@ func (c *Client) CreateDirectChannel(counterparty types.Address, appDefinition t
 		*c.Address,
 	)
 
-	// Pass in a fresh, dedicated go channel to communicate the response:
 	apiEvent := engine.APIEvent{
 		ObjectiveToSpawn: objective,
 	}
--- a/client/client.go
+++ b/client/client.go
@@ -65,6 +65,7 @@ func (c *Client) CompletedObjectives() <-chan protocols.ObjectiveId {
 	return c.completedObjectives
 }
 
+// CreateVirtualChannel creates a virtual channel with the counterParty using ledger channels with the intermediary.
 func (c *Client) CreateVirtualChannel(counterParty types.Address, intermediary types.Address, appDefinition types.Address, appData types.Bytes, outcome outcome.Exit, challengeDuration *types.Uint256) protocols.ObjectiveId {
 	right, ok := (*c.store).GetTwoPartyLedger(*c.Address, intermediary)
 
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -341,7 +341,7 @@ func (e *Engine) handleLedgerRequests(ledgerRequests []protocols.LedgerRequest, 
 		if err != nil {
 			return se, fmt.Errorf("could not handle ledger request: %w", err)
 		}
-		err = e.store.SetObjective(objective)
+		err = e.store.SetChannel(&ledger.Channel)
 
 		if err != nil {
 			return se, fmt.Errorf("could not set channel: %w", err)
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -130,11 +130,7 @@ func (e *Engine) handleMessage(message protocols.Message) ObjectiveChangeEvent {
 		e.logger.Print(err)
 		return ObjectiveChangeEvent{}
 	}
-	err = e.store.SetObjective(updatedObjective)
-	if err != nil {
-		e.logger.Print(err)
-		return ObjectiveChangeEvent{}
-	}
+
 	return e.attemptProgress(updatedObjective)
 
 }
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -212,7 +212,7 @@ func (e *Engine) attemptProgress(objective protocols.Objective) (outgoing Object
 	crankedObjective, sideEffects, waitingFor, ledgerRequests, _ := objective.Crank(secretKey) // TODO handle error
 	_ = e.store.SetObjective(crankedObjective)                                                 // TODO handle error
 
-	ledgerSideEffects, _ := e.handleLedgerRequests(ledgerRequests) // TODO: handle error
+	ledgerSideEffects, _ := e.handleLedgerRequests(ledgerRequests, crankedObjective) // TODO: handle error
 	sideEffects.Merge(ledgerSideEffects)
 
 	e.executeSideEffects(sideEffects)
@@ -319,22 +319,29 @@ func (e *Engine) constructObjectiveFromMessage(message protocols.Message) (proto
 }
 
 // handleLedgerRequests handles a collection of ledger requests by submitting them to the ledger manager.
-func (e *Engine) handleLedgerRequests(ledgerRequests []protocols.LedgerRequest) (protocols.SideEffects, error) {
+func (e *Engine) handleLedgerRequests(ledgerRequests []protocols.LedgerRequest, objective protocols.Objective) (protocols.SideEffects, error) {
 	sideEffects := protocols.SideEffects{}
 	for _, req := range ledgerRequests {
 		e.logger.Printf("Handling ledger request  %+v", req)
-		ch, ok := e.store.GetChannel(req.LedgerId)
-		if !ok {
+
+		var ledger *channel.TwoPartyLedger
+		found := false
+		for _, ch := range objective.Channels() {
+			if ch.Id == req.LedgerId {
+				ledger = &channel.TwoPartyLedger{Channel: *ch}
+				found = true
+			}
+		}
+		if !found {
 			return protocols.SideEffects{}, fmt.Errorf("Could not find ledger %s", req.LedgerId)
 
 		}
-		ledger := &channel.TwoPartyLedger{Channel: *ch}
 
 		se, err := e.ledgerManager.HandleRequest(ledger, req, e.store.GetChannelSecretKey())
 		if err != nil {
 			return se, fmt.Errorf("could not handle ledger request: %w", err)
 		}
-		err = e.store.SetChannel(ch)
+		err = e.store.SetObjective(objective)
 
 		if err != nil {
 			return se, fmt.Errorf("could not set channel: %w", err)
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -130,6 +130,11 @@ func (e *Engine) handleMessage(message protocols.Message) ObjectiveChangeEvent {
 		e.logger.Print(err)
 		return ObjectiveChangeEvent{}
 	}
+	err = e.store.SetObjective(updatedObjective)
+	if err != nil {
+		e.logger.Print(err)
+		return ObjectiveChangeEvent{}
+	}
 	return e.attemptProgress(updatedObjective)
 
 }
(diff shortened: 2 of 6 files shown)


geoknee:
> * Make sure that Irene generates the exact same `LedgerRequest`s as Bob and Alice, so all ledger states match exactly. Previously we were using `MyDestination/TheirDestination` for `Left` and `Right` so Irene would generate a different request than Alice/Bob.

Can you link to this change? I couldn't seem to find it

client/virtualfund_integration_test.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/statechannels/go-nitro/client/engine/store"
+	"github.com/statechannels/go-nitro/types"
+)
+
+// TestVirtualFundIntegration is a work in progress:
+// It should:
+// [x] spin up three clients with connected test services
+// [x] directly fund a pair of ledger channels
+// [ ] call an API method such as clientA.CreateVirtualChannel
+// [ ] assert on an appropriate objective completing in all clients
```
geoknee, line 22:
```suggestion
// [x] call an API method such as clientA.CreateVirtualChannel
// [x] assert on an appropriate objective completing in all clients
```

lalexgap:
> > * Make sure that Irene generates the exact same `LedgerRequest`s as Bob and Alice, so all ledger states match exactly. Previously we were using `MyDestination/TheirDestination` for `Left` and `Right` so Irene would generate a different request than Alice/Bob.
> 
> Can you link to this change? I couldn't seem to find it

It should be this [80d850c](https://github.com/statechannels/go-nitro/pull/274/commits/80d850c6fdc028de88d101e505c0f3c8fc1cbd18)

client/client.go
```diff
@@ -61,6 +65,44 @@ func (c *Client) CompletedObjectives() <-chan protocols.ObjectiveId {
 	return c.completedObjectives
 }
 
+func (c *Client) CreateVirtualChannel(counterParty types.Address, intermediary types.Address, appDefinition types.Address, appData types.Bytes, outcome outcome.Exit, challengeDuration *types.Uint256) protocols.ObjectiveId {
```
lalexgap, line 68: Add godoc

client/engine/engine.go
```diff
@@ -245,31 +265,102 @@ func (e *Engine) constructObjectiveFromMessage(message protocols.Message) (proto
 			message.SignedStates[0].State(),
 			*e.store.GetAddress(),
 		)
+	case strings.Contains(string(message.ObjectiveId), "Virtual"):
```
lalexgap, line 268: Probably beyond the scope of this PR but this logic should probably live in the protocol package

client/engine/engine.go
```diff
@@ -245,31 +265,102 @@ func (e *Engine) constructObjectiveFromMessage(message protocols.Message) (proto
...
 	default:
-		return directfund.DirectFundObjective{}, errors.New("cannot handle unimplemented objective type")
+		return virtualfund.VirtualFundObjective{}, errors.New("cannot handle unimplemented objective type")
 	}
 
 }
 
 // handleLedgerRequests handles a collection of ledger requests by submitting them to the ledger manager.
-func (e *Engine) handleLedgerRequests(ledgerRequests []protocols.LedgerRequest) (protocols.SideEffects, error) {
+func (e *Engine) handleLedgerRequests(ledgerRequests []protocols.LedgerRequest, objective protocols.Objective) (protocols.SideEffects, error) {
```
lalexgap, line 322: This should probably get moved into the LedgerManager.

client/engine/engine.go
```diff
@@ -245,31 +265,102 @@ func (e *Engine) constructObjectiveFromMessage(message protocols.Message) (proto
...
+		if err != nil {
+			return se, fmt.Errorf("could not set channel: %w", err)
+		}
+
 		sideEffects.Merge(se)
 	}
 	return sideEffects, nil
+}
 
+func getMyRole(myAddress types.Address, participants []types.Address) (uint, error) {
```
lalexgap, line 355: add godoc

client/engine/store/mockstore.go
```diff
@@ -57,12 +60,60 @@ func (ms MockStore) SetObjective(obj protocols.Objective) error {
...
+					return err
+				}
+
+			}
+		}
+	}
+	return nil
+}
+
+func (ms MockStore) GetTwoPartyLedger(firstParty types.Address, secondParty types.Address) (ledger *channel.TwoPartyLedger, ok bool) {
```
lalexgap, line 109: add godoc

client/client.go
```diff
@@ -21,6 +21,7 @@ type Client struct {
 	engine              engine.Engine // The core business logic of the client
 	Address             *types.Address
 	completedObjectives chan protocols.ObjectiveId
+	store               *store.Store
```
geoknee, line 27: What's the thinking here? Currently, the engine owns the store and the client can't access it directly.

client/client.go
```diff
@@ -62,6 +65,44 @@ func (c *Client) CompletedObjectives() <-chan protocols.ObjectiveId {
...
+			Outcome:           outcome,
+			TurnNum:           0,
+			IsFinal:           false,
+		},
+		*c.Address,
+		1, // We only handle a single hop for now
+		0, // We always play the role of alice if we initiate the channel
+		left, right)
+
+	// Pass in a fresh, dedicated go channel to communicate the response:
```
geoknee, line 96: I think this comment (and potentially another one like it below in `CreateDirectChannel`) is stale and related to the "promise"-style pattern we had in the past.

client/virtualfund_integration_test.go
```diff
@@ -95,9 +95,29 @@ func TestVirtualFundIntegration(t *testing.T) {
...
 	directlyFundALedgerChannel(clientI, clientB)
 
+	outcome := outcome.Exit{outcome.SingleAssetExit{
+		Allocations: outcome.Allocations{
+			outcome.Allocation{
+				Destination: types.AddressToDestination(a),
+				Amount:      big.NewInt(5),
+			},
+			outcome.Allocation{
+				Destination: types.AddressToDestination(i),
```
geoknee, line 103: I think this should be an outcome for Bob? Should we be worried that it still works without an outcome for Bob?

client/engine/engine.go
```diff
@@ -245,8 +247,55 @@ func (e *Engine) constructObjectiveFromMessage(message protocols.Message) (proto
 			message.SignedStates[0].State(),
 			*e.store.GetAddress(),
 		)
+	case strings.Contains(string(message.ObjectiveId), "Virtual"):
+		initialState := message.SignedStates[0].State()
+
+		if initialState.TurnNum != 0 {
+			return virtualfund.VirtualFundObjective{}, errors.New("cannot construct virtual fund objective without prefund state")
```
geoknee, line 272: I think the checks on turn number and number of participants are redundant? They are performed already in the constructor for a virtual channel, and the errors should bubble up and arrive at your L288.

client/engine/engine.go
```diff
@@ -245,8 +247,55 @@ func (e *Engine) constructObjectiveFromMessage(message protocols.Message) (proto
...
+		myRole, err := getMyRole(myAddress, participants)
+		if err != nil {
+			return virtualfund.VirtualFundObjective{}, errors.New("could not determine my role: %w")
+		}
+
+		return virtualfund.New(
+			true, // TODO ensure objective in only approved if the application has given permission somehow
+			message.SignedStates[0].State(),
+			*e.store.GetAddress(),
+			1, // Always a single hop virtual channel
```
geoknee, line 310: Is it worth defining this as a constant somewhere?

client/engine/engine.go
```diff
@@ -245,8 +247,55 @@ func (e *Engine) constructObjectiveFromMessage(message protocols.Message) (proto
...
+		}
+
+		myRole, err := getMyRole(myAddress, participants)
+		if err != nil {
+			return virtualfund.VirtualFundObjective{}, errors.New("could not determine my role: %w")
+		}
+
+		return virtualfund.New(
+			true, // TODO ensure objective in only approved if the application has given permission somehow
+			message.SignedStates[0].State(),
```
geoknee, line 308: There's a lot of `message.SignedStates[0].State()` in this file, which could be tidied up using the variable you defined on L251. Also, it seems bad that we just index into the 0th state on the message. Could we refactor this away some how? Do we need a `createObjectiveFromState` function instead?

client/engine/engine.go
geoknee: This should almost certainly be pushed down into the protocols package. We can encapsulate **both** `n` and `myRole` entirely so that the engine doesn't need to know it https://github.com/statechannels/go-nitro/issues/257

client/engine/engine.go
```diff
@@ -130,6 +130,11 @@ func (e *Engine) handleMessage(message protocols.Message) ObjectiveChangeEvent {
 		e.logger.Print(err)
 		return ObjectiveChangeEvent{}
 	}
+	err = e.store.SetObjective(updatedObjective)
```
geoknee, line 133: Do we need to set the objective here? We are going to set it inside `attemptProgress` anyway?

I think its important we have a really well thought out scheme for when we read and write to the store.

protocols/virtualfund/virtualfund.go
```diff
@@ -344,9 +344,9 @@ func (s VirtualFundObjective) generateLedgerRequestSideEffects() []protocols.Led
 				ObjectiveId: s.Id(),
 				LedgerId:    s.ToMyRight.Channel.Id,
 				Destination: s.V.Id,
-				Left:        s.ToMyRight.Channel.MyDestination(),
+				Left:        types.AddressToDestination(s.ToMyRight.Channel.Participants[0]),
 				LeftAmount:  leftAmount,
-				Right:       s.ToMyRight.Channel.TheirDestination(),
+				Right:       types.AddressToDestination(s.ToMyRight.Channel.Participants[1]),
```
geoknee, line 350: I'm not sure this is correct -- I think it should depend on `myRole` ?

geoknee (approved):
🎉  great to hit our milestone!

I left some comments. Overall I think :
* the virtualfund left/right thing looks wrong to me and should be cleared up
* there are a few tweaks / refactors that can be done now (or quickly post-merge)
* still some hard thinking / design around the store and the role of channels/objectives/ledger managers. It's likely the next sprint will have us overhaul this stuff.

protocols/virtualfund/virtualfund.go
geoknee: It works out as correct _if_ we have a convention on the ordering of participants in ledger channels. But if you check the SATP paper, I don't believe this is specified. What _is_ specified is the role of the participants in the virtual fund protocol*.

So I'm nitpicking a bit, but I think it is worth fixing this because it allows ledger channels to be spun up with participants in any order (that might equate to either party -- end user or hub -- proposing the ledger channel). 

Here's my proposed fix https://github.com/statechannels/go-nitro/pull/288. 

* We can take this to be the same as the index of each participant in the participant list in the virtual channel.

andrewgordstewart: This nitpick is very worthwhile. It is not good to couple the participant ordering of the virtual channel `V`, the ledger channels, AND the guarantees.

client/client.go
lalexgap: Currently the `store` is used by the `client` to find the appropriate ledger channels to construct a new VirtualFund objective.

client/virtualfund_integration_test.go
lalexgap: You're right! I'm pretty surprised that things worked as is, we should investigate why that is.

client/engine/engine.go
lalexgap: I think you're right and we can probably just rely on setting it in `attemptProgress`

client/engine/engine.go
lalexgap: I think this gets resolved by https://github.com/statechannels/go-nitro/pull/284

client/virtualfund_integration_test.go
```diff
@@ -0,0 +1,121 @@
...
+	}}
+	id := clientA.CreateVirtualChannel(b, i, types.Address{}, types.Bytes{}, outcome, big.NewInt(0))
+	waitForCompletedObjectiveId(id, &clientA)
+	waitForCompletedObjectiveId(id, &clientB)
+	waitForCompletedObjectiveId(id, &clientI)
+
+}
+
+// waitForCompletedObjectiveId waits for completed objectives and returns when the completed objective id matchs the id waitForCompletedObjectiveId has been given
+func waitForCompletedObjectiveId(id protocols.ObjectiveId, client *Client) {
```
andrewgordstewart, line 116: This integration test asserts that the objective is completed, but does no assertions on the end result. Is this cause for concern?

lalexgap: Unfortunately the current API doesn't expose anything beyond the completed `ObjectiveId`. For no we could probably interrogate the `store` on the client to make sure states look like we expect them to.

client/engine/engine.go
lalexgap: [b1a4096](https://github.com/statechannels/go-nitro/pull/274/commits/b1a40960b99c176f9acd1c5f3209d127ad3b0763)

client/engine/engine.go
lalexgap: This is slightly addressed in [a59c047](https://github.com/statechannels/go-nitro/pull/274/commits/a59c0475342cd93ec949acb4e068812855b853b3) but we should probably extend it to something like `createObjectiveFromState`

client/engine/engine.go
lalexgap: Updated in a59c047

client/engine/store/mockstore.go
lalexgap, line 109:
```suggestion
GetTwoPartyLedger returns a ledger channel between the two parties if it exists.
func (ms MockStore) GetTwoPartyLedger(firstParty types.Address, secondParty types.Address) (ledger *channel.TwoPartyLedger, ok bool) {
```

client/client.go
```diff
@@ -61,6 +65,43 @@ func (c *Client) CompletedObjectives() <-chan protocols.ObjectiveId {
 	return c.completedObjectives
 }
 
+func (c *Client) CreateVirtualChannel(counterParty types.Address, intermediary types.Address, appDefinition types.Address, appData types.Bytes, outcome outcome.Exit, challengeDuration *types.Uint256) protocols.ObjectiveId {
```
lalexgap, line 68:
```suggestion
// CreateVirtualChannel creates a virtual channel with the counterParty using ledger channels with the intermediary.
func (c *Client) CreateVirtualChannel(counterParty types.Address, intermediary types.Address, appDefinition types.Address, appData types.Bytes, outcome outcome.Exit, challengeDuration *types.Uint256) protocols.ObjectiveId {
```

PR #285, opened 2022-02-25
Remove protocol name from Objective struct

geoknee:
...and change `New` to `NewObjective`.

The package namespace already serves the purpose of specifying which protocol we are dealing with.

See https://gist.github.com/adamveld12/c0d9f0d5f0e1fba1e551#package-names and
https://go.dev/blog/package-names?utm_source=dlvr.it&utm_medium=twitter

---

#### Code quality

#### Project management

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -77,7 +77,7 @@ func (c *Client) CreateVirtualChannel(counterParty types.Address, intermediary t
 	var left *channel.TwoPartyLedger
 
 	// Convert the API call into an internal event.
-	objective, _ := virtualfund.New(true,
+	objective, _ := virtualfund.NewObjective(true,
 		state.State{
 			ChainId:           big.NewInt(0), // TODO
 			Participants:      []types.Address{*c.Address, intermediary, counterParty},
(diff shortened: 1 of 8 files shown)


client/client.go
```diff
@@ -64,7 +64,7 @@ func (c *Client) CompletedObjectives() <-chan protocols.ObjectiveId {
 // CreateDirectChannel creates a directly funded channel with the given counterparty
 func (c *Client) CreateDirectChannel(counterparty types.Address, appDefinition types.Address, appData types.Bytes, outcome outcome.Exit, challengeDuration *types.Uint256) protocols.ObjectiveId {
 	// Convert the API call into an internal event.
-	objective, _ := directfund.New(true,
+	objective, _ := directfund.NewObjective(true,
```
geoknee, line 67: Here we consume the constructor from a different package. This is part of the motivation for the renaming done in thiss PR.

client/engine/engine.go
```diff
@@ -238,15 +238,15 @@ func (e *Engine) constructObjectiveFromMessage(message protocols.Message) (proto
...
 		}
-		return directfund.New(
+		return directfund.NewObjective(
 			true, // TODO ensure objective in only approved if the application has given permission somehow
 			message.SignedStates[0].State(),
 			*e.store.GetAddress(),
 		)
 	default:
-		return directfund.DirectFundObjective{}, errors.New("cannot handle unimplemented objective type")
+		return directfund.Objective{}, errors.New("cannot handle unimplemented objective type")
```
geoknee, line 249: ... and here's the other part of the motivation. As is, we end up repeating ourselves unnecessarily.

lalexgap (approved):

PR #286, opened 2022-02-25
Ensure all `.Clone()` methods are documented and unit tested

### LINKED ISSUE #166: Audit all `.Clone()` methods
An example of a well-thought through `Clone` method:

https://github.com/statechannels/go-nitro/blob/b70f16695f8baf807e5338b132b55b861b5278d9/channel/state/state.go#L179-L198

An example of one that probably results in the clone having pointers to the originals data 

https://github.com/statechannels/go-nitro/blob/4d65ed9a6cab409be0e6f86b6cb2a323cdb1b483/channel/channel.go#L48-L50

This is dangerous and can potentially wreak havoc in tests and production. 

We should take another look at all the `Clone` methods (even if they appear correct) and check that modifying the clone leaves the original untouched. Even better would be to enforce this with some tests.

geoknee:
We have 13 such methods (some are private). I audited them all, and have ensured they all have at least a rudimentary unit test using `cmp.Diff`. This approach relies on there being `.Equal()` methods for all involved types, so I have added those were necessary. 

Closes #166 

---

#### Code quality

#### Project management

### DIFF
--- a/channel/channel.go
+++ b/channel/channel.go
@@ -83,6 +83,11 @@ func (v SingleHopVirtualChannel) RightAmount() types.Funds {
 	return v.amountAtIndex(1)
 }
 
+// Equal returns true if the supplied TwoPartyLedger is deeply equal to the receiver, false otherwise.
+func (v *SingleHopVirtualChannel) Equal(w *SingleHopVirtualChannel) bool {
+	return v.Channel.Equal(w.Channel)
+}
+
 // Clone returns a pointer to a new, deep copy of the receiver, or a nil pointer if the receiver is nil.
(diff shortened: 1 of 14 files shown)


channel/channel.go
```diff
@@ -83,6 +83,12 @@ func (v SingleHopVirtualChannel) RightAmount() types.Funds {
 	return v.amountAtIndex(1)
 }
 
+// Equal returns true if the supplied TwoPartyLedger is deeply equal to the receiver, false otherwise.
```
lalexgap, line 86:
```suggestion
// Equal returns true if the supplied SingleHopVirtualChannel is deeply equal to the receiver, false otherwise.
```

lalexgap (approved):

PR #298, opened 2022-02-28
Move objective construction logic into objectives

### LINKED ISSUE #293: Move Objective construction logic to protrocols
Currently the engine is responsible for constructing objectives. However this relies on protocol-specific logic. This should be moved into the protocol package.

See https://github.com/statechannels/go-nitro/blob/9641c39f89714bd1bf207d6311536d5b4237c1af/client/engine/engine.go#L265

lalexgap:
Fixes #293 

Previously the `engine` was responsible for constructing the objectives from messages. However the logic to construct objectives was deeply coupled to the protocol implementations, meaning that the `engine` contained a bunch of logic specific for virtual funding.

This PR moves that logic from the `Engine` into a `ConstructObjectiveFromMessage` in the virtual funding protocol. It also adds a `IsVirtualFundObjective/IsDirectFundObjective` functions so the `engine` doesn't have to directly inspect the objective id string.

To prevent a circular reference to the `Store` package `ConstructObjectiveFromMessage` just accepts a `getTwoPartyLedger` function instead of a whole store. This also has the benefit of preventing `ConstructObjectiveFromMessage` writing to the store accidentally.

### DIFF
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -253,55 +253,18 @@ func (e *Engine) constructObjectiveFromMessage(message protocols.Message) (proto
 	switch {
 	case strings.Contains(string(message.ObjectiveId), `DirectFund`):
 
-		if initialState.TurnNum != 0 {
-			return directfund.Objective{}, errors.New("cannot construct direct fund objective without prefund state")
-		}
 		return directfund.NewObjective(
 			true, // TODO ensure objective in only approved if the application has given permission somehow
 			initialState,
 			*e.store.GetAddress(),
 		)
 	case strings.Contains(string(message.ObjectiveId), "Virtual"):
-		// TODO: This logic belongs in the virtual fund protocol
-		// Once it is moved there it can be cleaned up to use myRole and isAlice/isBob
-		participants := message.SignedStates[0].State().Participants
-		alice := participants[0]
-		intermediary := participants[1]
-		bob := participants[2]
-		myAddress := *e.store.GetAddress()
-
-		var left *channel.TwoPartyLedger
-		var right *channel.TwoPartyLedger
-		var ok bool
-		if alice == myAddress {
-			right, ok = e.store.GetTwoPartyLedger(intermediary, bob)
-			if !ok {
-				return virtualfund.Objective{}, fmt.Errorf("could not find a right ledger channel between %v and %v", intermediary, bob)
-			}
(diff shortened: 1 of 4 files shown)


protocols/directfund/directfund.go
```diff
@@ -273,6 +279,11 @@ func (o Objective) clone() Objective {
 	return clone
 }
 
+// IsDirectFundObjective inspects a objective id and returns true if the objective id is for a direct fund objective.
+func IsDirectFundObjective(id protocols.ObjectiveId) bool {
+	return strings.HasPrefix(string(id), ObjectivePrefix)
```
geoknee, line 284: 👍

protocols/virtualfund/virtualfund.go
```diff
@@ -430,3 +434,60 @@ func (o *Objective) isAlice() bool {
 func (o *Objective) isBob() bool {
 	return o.MyRole == o.n+1
 }
+
+// GetTwoPartyLedgerFunction specifies a function that can be used to retreive ledgers from a store.
+type GetTwoPartyLedgerFunction func(firstParty types.Address, secondParty types.Address) (ledger *channel.TwoPartyLedger, ok bool)
```
geoknee, line 439: 👍

client/engine/engine.go
```diff
@@ -251,57 +250,20 @@ func (e *Engine) constructObjectiveFromMessage(message protocols.Message) (proto
...
+	case directfund.IsDirectFundObjective(message.ObjectiveId):
 
-		if initialState.TurnNum != 0 {
-			return directfund.Objective{}, errors.New("cannot construct direct fund objective without prefund state")
-		}
 		return directfund.NewObjective(
 			true, // TODO ensure objective in only approved if the application has given permission somehow
 			initialState,
 			*e.store.GetAddress(),
 		)
```
geoknee, line 259: Nit: it might be nice for this branch of the switch statement to also call into a `directfund.ConstructObjectiveFromMessage` function?

geoknee (approved):

client/engine/engine.go
lalexgap: Good point! [fc3e6e9](https://github.com/statechannels/go-nitro/pull/298/commits/fc3e6e997e83413faafa358c21b14bd5cef18500)

PR #299, opened 2022-02-28
Add GH action to auto assign to project

lalexgap:
Unfortunately I'm unsure of how to test this besides merging it into `main` and seeing if it works. It also only assigns the `Project` so the `Milestone` would still have to be a manual step.


geoknee (approved):
Looks good! Happy to merge it and see what happens :-) 

Can we get the action to also assign the author to the PR, too?

PR #302, opened 2022-03-01
Panic if encountering an error during `Channel.Clone()`

### LINKED ISSUE #270: Add comment explaining ignored error in Channel.Clone()
I think we're ignoring the error here because we know `c.PreFundState().Clone()` starts with turnNum=0 and so we'd never get an error back from `New`?  We could add a comment explaining that.

_Originally posted by @lalexgap in https://github.com/statechannels/go-nitro/pull/269#discussion_r812242597_

geoknee:
Closes #270 .

We are currently ignoring an error in this function, which is a bad practice and can result in hard-to-debug problems. 

We could considering passing an error back from `Clone`, but there are two downsides to this:
- we would have to bubble up that error passing in a lot of places 
- it seems like the wrong pattern, since there is no reason we would expect `Clone` to ever fail (unless we a programmer makes a mistake)

It is better to panic here, since the problem indicates the presence of a bug which we can't recover from.

---

#### Code quality

#### Project management

### DIFF
--- a/channel/channel.go
+++ b/channel/channel.go
@@ -178,7 +178,12 @@ func (c *Channel) Clone() *Channel {
 	if c == nil {
 		return nil
 	}
-	d, _ := New(c.PreFundState().Clone(), c.MyIndex)
+	d, err := New(c.PreFundState().Clone(), c.MyIndex)
+	if err != nil {
+		// The constructor shouldn't error unless we have made a programming error
+		// (e.g. passed in a turnNum =/= 0 state)
+		panic(err)
+	}
 	d.latestSupportedStateTurnNum = c.latestSupportedStateTurnNum
 	for i, ss := range c.SignedStateForTurnNum {
 		d.SignedStateForTurnNum[i] = ss.Clone()


NiloCK (commented):
I'm for this style of panic (dying on errors that we believe should not be reachable), but I think it should be impossible in general for panics to escape from go-nitro and cause a shutdown of the consuming application.

A quick fix could be something like:

```
// swallowPanic prevents go-nitro panics from escaping to the consuming application
func (e *Engine) swallowPanic() {
	if r := recover(); r != nil {
		e.logger.Printf("Fatal error in go-nitro client: %v", r)
	}
}

// Run kicks of an infinite loop that waits for communications on the supplied channels, and handles them accordingly
func (e *Engine) Run() {
	defer e.swallowPanic()
// ...
```

in `engine.go`

geoknee:
Interesting. I think what you are proposing is a good candidate for either a new issue or even a GH Discussion, and can probably be handled independently of this PR?

NiloCK:
> 



> Interesting. I think what you are proposing is a good candidate for either a new issue or even a GH Discussion, and can probably be handled independently of this PR?

#306

NiloCK (approved):

Issue #296, opened 2022-02-28
Implement a full ledger update protocol


geoknee:
This work depends on discussions and decisions arising from #291.

It should be the major piece of work to get #295 working.

PR #308, opened 2022-03-01
Add a MultiParty Virtual Funding Integration Test

### LINKED ISSUE #295: Write a virtual funding integration test with >3 actors
For example, Alice, Irene, Bob and Brian. Irene is the only intermediary, and Alice should calls `CreateVirtualChannel` _twice_ (once for her connection to Bob, once for Brian).

Otherwise, the test can follow the pattern of the existing script. 

Let's write a new test instead of changing the existing one.

lalexgap:
Fixes #295 
Implements a multi-party virtual funding integration test.

In the new test Alice attempts to create a virtual channel with Bob, and Brian via Irene. This means that ledger channel between Alice and Irene has to be updated twice to fund both virtual channels.

Currently this test fails as one virtual funding objective gets stuck waiting for funding. This is likely due to our current ledger protocol implementation and should get resolved by #296 

#### Code quality

#### Project management


geoknee (approved):
This looks great to me. I've made #309 and #310  to extend on what you did here. #309 includes a change that skips this test. If you are in agreement, I would be in favour of merging this work along with #309 now, closing the associated issue, and replacing it with an issue to unskip the test. We can also have a draft PR hanging around that will close the issue: it will of course fail now but hopefully pass by the end of the sprint.

PR #315, opened 2022-03-03
`Objetive` implements json.Marshaler, json.Unarshaler

NiloCK:
This is incremental work toward:
1. a serialization scheme for the store
2. the mockstore (& store) having a mechanism for splitting and recombining in-memory Objectives (which contain channels) into separate Objective and Channel

See #191, + discussion in #227

This PR splits previous work which had added these methods and integrated them with the existing `mockstore` implementation. `mockstore` will pick these methods up in a follow-up PR after this goes in.

## Shortcomings

- adding the `Unmarshal` interface requires a pointer-receiver method, which forces objectives to be passed around as pointers. (It's unclear if this is a problem, but it is at least a nuisance).
- the test for `virtualfund`'s encode / decode contains a heap of duplicated code from the single-hop virualfund test (init data for a virtualfund Objective)

**Changes**

This PR adds the json.Marshal and json.Unmarshal interface methods to the `Objective` interface, and adds implementations with tests to `directfund` and `virtualfund`.

---

#### Code quality

#### Project management

### DIFF
--- a/protocols/directfund/directfund.go
+++ b/protocols/directfund/directfund.go
@@ -119,20 +119,20 @@ func (o Objective) Approve() protocols.Objective {
 	// todo: consider case of s.Status == Rejected
 	updated.Status = protocols.Approved
 
-	return updated
+	return &updated
 }
 
 func (o Objective) Reject() protocols.Objective {
 	updated := o.clone()
 	updated.Status = protocols.Rejected
-	return updated
+	return &updated
 }
 
 // Update receives an ObjectiveEvent, applies all applicable event data to the DirectFundingObjectiveState,
 // and returns the updated state
 func (o Objective) Update(event protocols.ObjectiveEvent) (protocols.Objective, error) {
 	if o.Id() != event.ObjectiveId {
-		return o, fmt.Errorf("event and objective Ids do not match: %s and %s respectively", string(event.ObjectiveId), string(o.Id()))
+		return &o, fmt.Errorf("event and objective Ids do not match: %s and %s respectively", string(event.ObjectiveId), string(o.Id()))
 	}
 
 	updated := o.clone()
@@ -142,7 +142,7 @@ func (o Objective) Update(event protocols.ObjectiveEvent) (protocols.Objective, 
 		updated.C.OnChainFunding = event.Holdings
 	}
 
-	return updated, nil
+	return &updated, nil
 }
 
 // Crank inspects the extended state and declares a list of Effects to be executed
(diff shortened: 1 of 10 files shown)


protocols/virtualfund/virtualfund.go
```diff
@@ -49,6 +50,64 @@ func (c *Connection) Equal(d *Connection) bool {
...
+	}
+
+	var jsonC jsonConnection
+	err := json.Unmarshal(data, &jsonC)
+
+	if err != nil {
+		return err
+	}
+
+	c.Channel.Id = jsonC.Channel
```
lalexgap, line 95: This could be slightly confusing if you're not aware of how `UnmarshalJSON` works. You would get back connection with a `nil` channel that has the correct `Id`. This is probably fine if it's only the `Store` dealing with marshalling/unmarshalling

protocols/directfund/directfund.go
```diff
@@ -203,6 +215,49 @@ func (o Objective) Channels() []*channel.Channel {
 	return ret
 }
 
+// MarshalJSON returns a JSON representation of the DirectFundObjective
+//
+// NOTE: Marshal -> Unmarshal is a lossy process. All channel data
```
lalexgap, line 220: Having `Marshal -> Unmarshal` being lossy doesn't seem the best.

I wonder if it would make sense to move MarshalJSON/UnMarshalJSON to the `jsonObjective` struct. Instead of an `Objective` marshalling data to json itself we'd just transform it into `jsonObjective`. That way we separate the problem of marshalling data from the problem of constructing an `Objective`  with fully defined channels.

protocols/virtualfund/virtualfund_test.go
```diff
@@ -0,0 +1,151 @@
...
+					Destination: bob.destination,
+					Amount:      big.NewInt(5),
+				},
+			},
+		}},
+		TurnNum: 0,
+		IsFinal: false,
+	}
+
+	// left, _ :=
```
lalexgap, line 70:
```suggestion
```

lalexgap (approved):

protocols/virtualfund/virtualfund.go
```diff
@@ -49,6 +50,64 @@ func (c *Connection) Equal(d *Connection) bool {
...
 }
 
+// jsonConnection is a serialization-friendly struct representation
+// of a Connection
+type jsonConnection struct {
+	Channel            types.Destination
+	ExpectedGuarantees []assetGuarantee
+}
+
+func (c Connection) MarshalJSON() ([]byte, error) {
```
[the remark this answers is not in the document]
NiloCK: 7d1f237

protocols/virtualfund/virtualfund.go
```diff
@@ -49,6 +50,64 @@ func (c *Connection) Equal(d *Connection) bool {
...
+	bytes, err := json.Marshal(jsonC)
+
+	if err != nil {
+		return []byte{}, err
+	}
+
+	return bytes, err
+}
+
+func (c *Connection) UnmarshalJSON(data []byte) error {
```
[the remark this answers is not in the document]
NiloCK: 7d1f237

protocols/directfund/directfund.go
NiloCK: It definitely isn't the best!

The compromise is motivated by us wanting to have channels attached to objectives in memory, as part of the `update` `crank` loop, but we want to persist channels / objectives separately in the store.

Ideally, the marshalling / unmarshalling functionality would live, encapsulated, in the `store`. Issue there is that there are private fields in Objective & Channel which we need to persist, so they aren't visible inside `store`.

protocols/virtualfund/virtualfund.go
NiloCK: Yes, intent here is that this is a utility method specifically for the `store`. (A wart, but the persisting of private fields sort of requires it).

PR #320, opened 2022-03-08
Revert "add GH action to auto assign to project"

lalexgap:
Unfortunately the assign to project action doesn't seem to be working. It seems we're running into an issue similar 
to [this](https://github.com/srggrs/assign-one-project-github-action/issues/80).

At this point I think it's easier to just remove the GH action and rely on manually moving the PR to the project.


NiloCK (approved):
Oh well :/

PR #321, opened 2022-03-09
Get multi party virtual integration test passing

### LINKED ISSUE #313: Enable multi-party integration test
Get the multi-party integration test passing.

lalexgap:
Fixes #313 

This a collection of changes to get the current multi party integration test passing.

This PR **does not implement any locking**. Since this test passes without locking, I've created #322 to implement a failing test and #323 to implement a locking solution to resolve the test.

## Changes
- [7bd6315](https://github.com/statechannels/go-nitro/pull/321/commits/7bd631559a7df821813131ca66384911f920c42b) Fixed an incorrect test assertion that was waiting for Brian on the wrong virtual funding objective
- [455b7b5](https://github.com/statechannels/go-nitro/pull/321/commits/455b7b535b7bce47e4210784217402deae40afb5) lowered the amounts on the virtual channel so the ledger channel can afford multiple virtual channels
- [d84913f](https://github.com/statechannels/go-nitro/pull/321/commits/d84913f5d27e6380a6b518b54b0374e194440103) if a proposal already exists we use the `turnNum` after that instead of always using the `turnNum` after the supported state
- [a3a2c63](https://github.com/statechannels/go-nitro/pull/321/commits/a3a2c639fadb2dac3894c7a9bd3111554ee29778) only create a new proposal if we haven't created proposal already. This allows us to crank multiple times without generating a new proposal every time.
- [a738f11](https://github.com/statechannels/go-nitro/pull/321/commits/a738f11967d42975e25b05d43d14080bd60968e3) This is a workaround for the current store implementation. `SetObjective` only updates the channels on the specific objective, so by manually calling `SetChannel` we know that all objectives will have the latest channel.
- [fbe9bd4](https://github.com/statechannels/go-nitro/pull/321/commits/fbe9bd4207d7e3d371d6ba498684839e5bf54d84) Instead of outputting the whole message to the log just output a very basic summary. The summary is pretty basic at the moment but we can expand it as needed.
- [e7da061](https://github.com/statechannels/go-nitro/pull/321/commits/e7da061b21257b3aebd7b4919445dd498b24c576) Allow the message service channels to buffer incoming and outgoing messages. Previous to this change the engine would get blocked trying to send out messages. I haven't thought too much about this change but it seems to make sense and gets things passing.
---

#### Code quality

#### Project management

### DIFF
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -202,6 +202,17 @@ func (e *Engine) attemptProgress(objective protocols.Objective) (outgoing Object
 	crankedObjective, sideEffects, waitingFor, _ := objective.Crank(secretKey) // TODO handle error
 	_ = e.store.SetObjective(crankedObjective)                                 // TODO handle error
 
+	// TODO: This is hack to get around the fact that currently each objective in the store has it's own set of channels.
+	vfo, isVirtual := crankedObjective.(*virtualfund.Objective)
+	if isVirtual {
+		if vfo.ToMyLeft != nil {
+			_ = e.store.SetChannel(&vfo.ToMyLeft.Channel.Channel)
+		}
+		if vfo.ToMyRight != nil {
+			_ = e.store.SetChannel(&vfo.ToMyRight.Channel.Channel)
+		}
+	}
+
 	e.executeSideEffects(sideEffects)
 	e.logger.Printf("Objective %s is %s", objective.Id(), waitingFor)
 
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -112,7 +112,7 @@ func (e *Engine) Run() {
 // attempts progress.
 func (e *Engine) handleMessage(message protocols.Message) ObjectiveChangeEvent {
 
-	e.logger.Printf("Handling inbound message %v", message)
+	e.logger.Printf("Handling inbound message %+v", summarizeMessage(message))
 	objective, err := e.getOrCreateObjective(message)
 	if err != nil {
 		e.logger.Print(err)
@@ -180,7 +180,7 @@ func (e *Engine) handleAPIEvent(apiEvent APIEvent) ObjectiveChangeEvent {
 // executeSideEffects executes the SideEffects declared by cranking an Objective
 func (e *Engine) executeSideEffects(sideEffects protocols.SideEffects) {
 	for _, message := range sideEffects.MessagesToSend {
-		e.logger.Printf("Sending message to %s", message.To)
+		e.logger.Printf("Sending message %+v", summarizeMessage(message))
 		e.toMsg <- message
 	}
 	for _, tx := range sideEffects.TransactionsToSubmit {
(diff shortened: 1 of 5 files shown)


protocols/virtualfund/virtualfund.go
```diff
@@ -667,8 +667,17 @@ func (o *Objective) proposeLedgerUpdate(connection Connection, sk *[]byte) (prot
 		return protocols.SideEffects{}, fmt.Errorf("error finding a supported state: %w", err)
 	}
 
+	proposed, proposedFound := ledger.Proposed()
+
```
NiloCK, line 671: with respect to ledger syncing and to possible future batch updates to ledgers, this approach may not be correct.

EG: a hub may process N virtualfund objectives at once, update their own local copy of the ledger channel with subsequent proposals that build on top of the existing unanswered proposal, and send only the "latest" on over the wire.

That said, I think this is fine here for now.

client/engine/engine.go
```diff
@@ -202,6 +202,17 @@ func (e *Engine) attemptProgress(objective protocols.Objective) (outgoing Object
 	crankedObjective, sideEffects, waitingFor, _ := objective.Crank(secretKey) // TODO handle error
 	_ = e.store.SetObjective(crankedObjective)                                 // TODO handle error
 
+	// TODO: This is hack to get around the fact that currently each objective in the store has it's own set of channels.
```
NiloCK, line 205: should be fixable post #325

client/engine/engine.go
```diff
@@ -261,3 +272,26 @@ func (e *Engine) constructObjectiveFromMessage(message protocols.Message) (proto
...
+	summary := messageSummary{to: message.To.String(), objectiveId: message.ObjectiveId, signedStates: []signedStateSummary{}}
+	for _, signedState := range message.SignedStates {
+		channelId, err := signedState.State().ChannelId()
+		if err != nil {
+			panic(err)
+		}
+		summary.signedStates = append(summary.signedStates, signedStateSummary{turnNum: signedState.State().TurnNum, channelId: channelId.String()})
+	}
+	return summary
+}
```
NiloCK, line 297: This is great. As a refactor (separate PR is fine), maybe it should live in `protocols/messages.go`?

client/engine/messageservice/test-messageservice.go
```diff
@@ -31,8 +31,8 @@ func NewTestMessageService(address types.Address) TestMessageService {
 	tms := TestMessageService{
 		address: address,
 		toPeers: make(map[types.Address]chan<- string),
-		in:      make(chan protocols.Message),
-		out:     make(chan protocols.Message),
+		in:      make(chan protocols.Message, 5),
+		out:     make(chan protocols.Message, 5),
```
NiloCK, line 35: I've read somewhere or other [citation needed] that buffered channels are generally only a good practice in special circumstances and are a smell when applied in an unconsidered way.

Given the circumstance (a mock communications network) I don't think we need to worry too much about it.

NiloCK (approved):
Great!

PR #325, opened 2022-03-09
Decouple objectives and channels in `mockstore`

NiloCK:
This is incremental toward (re)instating json serialization. See #191 

Previously, `mockstore` stored entire `Objective` objects in memory, which included the channels they referenced. This is not a long term solution, because `channel -> objective` is a `1 -> many` relationship, particularly for ledger channels.

This PR is a minimal separation of `Objective` and `Channel` data.

Existing tests run and pass. This is a strong signal for the correctness of Objective persistence and recovery. No (public) functions were added - just implementation detail.

**Limitations**

Noteworthy: the `mockstore` is currently double-copying channel data. The `Objective` memstore map has the same hold of channel data that it had before, but these channels are overwritten by channel data recovered from the `Channel` memstore map.

This is to be followed with a PR which replaces the memstore map objects with json serde methods from #315.

---

#### Code quality

- [x] ~I have added tests that prove my fix is effective or that my feature works, if necessary~ (existing tests cover all edited code)

#### Project management

### DIFF
--- a/client/engine/store/mockstore.go
+++ b/client/engine/store/mockstore.go
@@ -42,7 +42,7 @@ func (ms MockStore) GetObjectiveById(id protocols.ObjectiveId) (obj protocols.Ob
 	// todo: locking
 	obj, ok = ms.objectives[id]
 
-	// return immediately if no
+	// return immediately if no such objective exists
 	if !ok {
 		return nil, ok
 	}
--- a/client/engine/store/mockstore.go
+++ b/client/engine/store/mockstore.go
@@ -6,6 +6,8 @@ import (
 	"github.com/statechannels/go-nitro/channel"
 	"github.com/statechannels/go-nitro/crypto"
 	"github.com/statechannels/go-nitro/protocols"
+	"github.com/statechannels/go-nitro/protocols/directfund"
+	"github.com/statechannels/go-nitro/protocols/virtualfund"
 	"github.com/statechannels/go-nitro/types"
 )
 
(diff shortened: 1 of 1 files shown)


client/engine/store/mockstore.go
```diff
@@ -39,59 +41,90 @@ func (ms MockStore) GetChannelSecretKey() *[]byte {
...
 	// todo: locking
 	obj, ok = ms.objectives[id]
+
+	// return immediately if no such objective exists
+	if !ok {
+		return nil, ok
+	}
+
+	// populate channel data
+	if directfund.IsDirectFundObjective(id) {
```
lalexgap, line 51: Not very important but instead of using `directfund.IsDirectFundObjective(id)` we could cast `obj` and check the result `dfo, isDirectFund := obj.(*directfund.Objective)` It would save us a line.

lalexgap, line 51: Not very important but instead of using `directfund.IsDirectFundObjective(id)` we could cast `obj` and check the result `dfo, isDirectFund := obj.(*directfund.Objective)` It would save us a line.

client/engine/store/mockstore.go
```diff
@@ -39,59 +41,90 @@ func (ms MockStore) GetChannelSecretKey() *[]byte {
...
+		vfo := obj.(*virtualfund.Objective)
+
+		v, err := ms.getChannelById(vfo.V.Id)
+		if err != nil {
+			return nil, false
+		}
+		vfo.V = &channel.SingleHopVirtualChannel{Channel: v}
+
+		// todo: clean these checks
+		if vfo.ToMyLeft != nil {
```
lalexgap, line 72: I think most other places we assume if `vfo.ToMyLeft != nil` then `vfo.ToMyLeft.Channel != nil` but it's probably safer to check both! Could we compress the two if statements into one?

lalexgap, line 72: I think most other places we assume if `vfo.ToMyLeft != nil` then `vfo.ToMyLeft.Channel != nil` but it's probably safer to check both! Could we compress the two if statements into one?

lalexgap (approved):
👍

PR #327, opened 2022-03-11
Mockstore serialize

### LINKED ISSUE #191: `MockStore` should serialize objectives before storing

NiloCK:
### Problem

- a production store will not store in-memory objects, but instead requires a serialization scheme for persistence

### Approach

- data structures which we are interested in persisting have been given serde methods that comply with the standard library `json.Marshaler` and `json.Unmarshaler` interfaces
- the `mockstore` has been converted to write & read this serialized data

### Shortcomings

Current approach is potentially brittle in the same sense as @kerzhner pointed out with existing `Equal` and `Clone` methods: the addition or subtraction of a field to `Objective`, `Channel`, or any of their child struct types needs to be accounted for in the existing marshaling implementations.

This fragility is partially addressed, and can be more fully addressed, with deeper, automated equality checking (`cmp.Diff`, `reflection.DeepEqual`) in the unit tests for the serialized objects and for the mockstore itself. A follow-up PR should flesh out the test coverage.

### **How Has This Been Tested?**

existing integration test suite succeeds, which is a strong indicator that data is not being lost or corrupted when written & read from the store.

this PR closes #191 

---

#### Code quality

  - could be neater, but I think the entire changeset is probably readable
  - ca

#### Project management

### DIFF
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -230,7 +230,10 @@ func (e *Engine) getOrCreateObjective(message protocols.Message) (protocols.Obje
 	id := message.ObjectiveId
 
 	objective, err := e.store.GetObjectiveById(id)
-	if err != nil { // todo: assert non-existence more specifically ?
+
+	if err == nil {
+		return objective, nil
+	} else if errors.Is(err, store.ErrNoSuchObjective) {
 
 		newObj, err := e.constructObjectiveFromMessage(message)
 		if err != nil {
@@ -244,7 +247,7 @@ func (e *Engine) getOrCreateObjective(message protocols.Message) (protocols.Obje
 		return newObj, nil
 
 	} else {
-		return objective, nil
+		return nil, fmt.Errorf("unexpected error getting/creating objective %s: %w", message.ObjectiveId, err)
 	}
 }
 
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -229,8 +229,8 @@ func (e *Engine) attemptProgress(objective protocols.Objective) (outgoing Object
 func (e *Engine) getOrCreateObjective(message protocols.Message) (protocols.Objective, error) {
 	id := message.ObjectiveId
 
-	objective, ok := e.store.GetObjectiveById(id)
-	if !ok {
+	objective, err := e.store.GetObjectiveById(id)
+	if err != nil { // todo: assert non-existence more specifically ?
 
 		newObj, err := e.constructObjectiveFromMessage(message)
 		if err != nil {
--- a/client/engine/store/mockstore.go
+++ b/client/engine/store/mockstore.go
@@ -1,6 +1,7 @@
 package store
 
 import (
+	"encoding/json"
 	"fmt"
 
 	"github.com/statechannels/go-nitro/channel"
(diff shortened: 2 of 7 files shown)


client/engine/store/mockstore.go
```diff
@@ -136,13 +137,99 @@ func (ms MockStore) GetTwoPartyLedger(firstParty types.Address, secondParty type
...
+		}
+
+		return vfo, nil
+
+	} else {
+		return nil, fmt.Errorf("objective %s did not correctly represent a known Objective type", id)
+	}
+}
+
+// decodeObjective is a helper which encapsulates the deserialization
```
lalexgap, line 219: might be worth mentioning that `decodeObjective` doesn't handle populating channels (done by `populateChannelData`) in the godoc. Although it's a private function so it probably doesn't matter too much.

client/engine/store/mockstore.go
```diff
@@ -136,13 +137,99 @@ func (ms MockStore) GetTwoPartyLedger(firstParty types.Address, secondParty type
...
 			}
 		}
 	}
 
 	return nil, false
 }
+
+// populateChannelData fetches stored Channel data relevent to the given
+// objective, attaches it to the objective, and returns the objective
+func (ms MockStore) populateChannelData(obj protocols.Objective) (protocols.Objective, error) {
```
lalexgap, line 165: Right now `populateChannelData` accepts a pointer to an `Objective` and then modifies the objective. The objective pointer that is returned is the same one that is passed in, which could be confusing.

We might want to either:
- Don't bother returning an `Objective` as a return type. `populateChannelData` can just modify the objective in-place.
- Clone the incoming objective and return a completely new objective that is populated with channel data.

client/engine/engine.go
```diff
@@ -230,8 +230,11 @@ func (e *Engine) attemptProgress(objective protocols.Objective) (outgoing Object
 func (e *Engine) getOrCreateObjective(message protocols.Message) (protocols.Objective, error) {
 	id := message.ObjectiveId
 
-	objective, ok := e.store.GetObjectiveById(id)
-	if !ok {
+	objective, err := e.store.GetObjectiveById(id)
+
+	if err == nil {
+		return objective, nil
+	} else if errors.Is(err, store.ErrNoSuchObjective) {
```
lalexgap, line 237: I prefer `ok` over having to check errors but maybe that's an ok trade off for `GetObjectiveById` being able to return various error types.

lalexgap (approved):

client/engine/store/mockstore.go
NiloCK: ~(As I read it) `populateChannelData` takes an `Objective`, not a pointer, and so it creates a local copy of the passed objective rather than modifying the original.~

(I believe I've just learned that functions with interface typed parameters always receive them by reference, rather than by value.)

Either way, your specific point still applies, because the pieces of the objective being written to (the channels), *are* pointers. `populateChannelData` can (and does) modify the contents of these pointers in-place.

I'll drop the returned objective since:
- internally cloning here would require modifying the API of virtualfund.Objective, directfund.Objective to make them public
- there is no danger here of multiple threads etc overwriting one another's objective data, since the modified objective itself has just been constructed from JSON

NiloCK: addressed in dfb67b6

PR #334, opened 2022-03-14
test: MockChain implements a Subscribe() method

andrewgordstewart:
This enables setupClient to be responsible for "subscribing" to the mock chain service's events, which facilitates more dynamic test setup.

(Also, the logger uses the prefix of an address as the log prefix, which makes the logs easier for humans to read)

### DIFF
--- a/client/engine/chainservice/mockchain.go
+++ b/client/engine/chainservice/mockchain.go
@@ -27,20 +27,20 @@ func (mc MockChain) In() chan<- protocols.ChainTransaction {
 }
 
 // NewMockChain returns a new MockChain with an out chan initialized for each of the addresses passed in.
-func NewMockChain(addresses []types.Address) MockChain {
+func NewMockChain() MockChain {
 
 	mc := MockChain{}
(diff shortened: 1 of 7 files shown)


client/engine/chainservice/mockchain.go
```diff
@@ -27,22 +27,22 @@ func (mc MockChain) In() chan<- protocols.ChainTransaction {
...
-	for _, a := range addresses {
-		// Use a buffered channel so we don't have to worry about blocking on writing to the channel.
-		mc.out[a] = make(chan Event, 10)
-	}
-
 	go mc.Run()
 	return mc
 }
 
+func (mc *MockChain) Subscribe(a types.Address) {
```
lalexgap, line 41: nit but could we add some godoc?

lalexgap (approved):
Makes sense to me!

client/engine/chainservice/mockchain.go
andrewgordstewart: I'll do this in a follow-up PR!

PR #335, opened 2022-03-14
TestMessageService: remove pairwise connections

andrewgordstewart:
Instead of connecting peers in a pairwise fashion, this now defines a Broker struct, which is simply a address -> incoming message channel lookup.

(For some reason, the incoming message channel is called `out` -- why? This is fixed in #336)

This makes it easier to create more dynamic test scenarios: each message service is already configured to receive messages when NewTestMessageService returns.

### DIFF
--- a/client/engine/messageservice/test-messageservice.go
+++ b/client/engine/messageservice/test-messageservice.go
@@ -50,10 +50,6 @@ func (t TestMessageService) In() chan<- protocols.Message {
 	return t.in
 }
 
-func (t TestMessageService) Send(message protocols.Message) {
-	t.in <- message
-}
-
 // Connect creates a gochan for message service to send messages to the given peer.
 func (t TestMessageService) Connect(peer TestMessageService) {
 	toPeer := make(chan string)
--- a/client/engine/messageservice/test-messageservice.go
+++ b/client/engine/messageservice/test-messageservice.go
@@ -3,6 +3,7 @@ package messageservice
 import (
 	"fmt"
 
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/statechannels/go-nitro/protocols"
 	"github.com/statechannels/go-nitro/types"
 )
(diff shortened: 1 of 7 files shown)


client/engine/messageservice/test-messageservice.go
```diff
@@ -18,28 +19,33 @@ import (
...
 
-	// connection to peer message services
-	toPeers map[types.Address]chan<- string
-
 	// connection to Engine:
 	in  chan protocols.Message // for recieving messages from engine
 	out chan protocols.Message // for sending message to engine
 }
 
+type Broker struct {
```
lalexgap, line 27: could we add some godoc?

lalexgap, line 27: could we add some godoc?

client/engine/messageservice/test-messageservice.go
```diff
@@ -50,46 +56,30 @@ func (t TestMessageService) In() chan<- protocols.Message {
...
-
+func (t TestMessageService) Connect(b Broker) {
 	go func() {
-		for msgString := range toPeer {
-			msg, _ := protocols.DeserializeMessage(msgString)
-			peer.out <- msg // send messages directly to peer's engine, using the peer's out chan
+		for message := range t.in {
+			peerChan, ok := b.services[message.To]
+			if ok {
+				msg, err := message.Serialize()
```
lalexgap, line 65: Why does this serialize and then deserialize the message?

andrewgordstewart: Any true message service would have to serialize and deserialize messages!

The honest truth is that I didn't want to change behaviour, and the previous behaviour serializes messages in `forward` and then deserializes them in the anonymous goroutine within `Connect`

lalexgap: That makes sense! Could we a comment that we're doing this for testing purposes?

Could we also rename some of the vars so it's a little clearer? There's `message`,`msg`,and `m` at the moment.

lalexgap (approved):

client/engine/messageservice/test-messageservice.go
andrewgordstewart: > That makes sense! Could we a comment that we're doing this for testing purposes?

Definitely!

> Could we also rename some of the vars so it's a little clearer? There's `message`,`msg`,and `m` at the moment.

Sounds good

andrewgordstewart: That change was not intentional, and can easily be reverted!

PR #336, opened 2022-03-14
refactor: in <--> out

andrewgordstewart:
Note how this is now consistent with "bind to inbound/outbound chans" in engine.go


NiloCK:
A MessageService's `Outbox` is my `Inbox` and vice-verca.

In / Out labels were applied here from the perspective of the services, since that's where the names themselves live and are most "in use". `In` goes into the service, `Out` comes out of it. Note that this change only touches two lines in the consuming `engine` package, and those lines apply an engine-local names to these channels.

andrewgordstewart:
Closing this in favour of improving the documentation on the current naming convention.

PR #337, opened 2022-03-14
document in, out, from and to

andrewgordstewart:
I found these variable names to be confusing, and these docstrings clarify the meaning of the names.

### DIFF
--- a/client/engine/chainservice/chainservice.go
+++ b/client/engine/chainservice/chainservice.go
@@ -14,6 +14,8 @@ type Event struct {
 }
 
 type ChainService interface {
+	// Out is a channel for sending events out of the ChainService
 	Out() <-chan Event
+	// In is a channel for receiving transactions into the ChainService
 	In() chan<- protocols.ChainTransaction
 }
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -59,14 +59,18 @@ func New(msg messageservice.MessageService, chain chainservice.ChainService, sto
 
 	e.store = store
 
-	// bind to inbound chans
+	// FromAPI allows the engine to read API events from the application
 	e.FromAPI = make(chan APIEvent)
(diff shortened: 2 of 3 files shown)


client/engine/chainservice/chainservice.go
```diff
@@ -14,6 +14,8 @@ type Event struct {
 }
 
 type ChainService interface {
+	// Out is a channel for sending events out of the ChainService
 	Out() <-chan Event
```
lalexgap, line 18: This describes `Out` from the perspective of the ChainService, wanting to send messages out.  However this makes the channel directions confusing for any external callers. `ChainService.Out` is a receive only chan but the godoc mentions sending. `ChainService.In` is a send only chan but the godoc mentions receiving. 

Especially since this in an interface I think it makes sense to write godoc from the perspective of an external caller.

andrewgordstewart: So, I personally find the names `In` and `Out` to be confusing, because they are named according to the service's point of view, and don't describe what they do. The docstrings included here are meant to explain that fact, so that it's less confusing.

I would prefer `Out` to be called something like `Events`, and `In` to be called something like `Transactions`:  the consumer of a `ChainService` receives events from `Events` and sends transactions to `Transactions`. I feel like these names would not require any documentation.

But, as discussed in standup, it seemed like there was consensus on names `In` and `Out` for both the `ChainService` and `MessageService`. Hence, the suggestion to explain the names via documentation.

andrewgordstewart: Yes, that would make sense. I think you are right in that the doctoring on `In` should target consumers of the `In` API.

However, because I found `In` and `Out` to be confusingly named, I'm looking for the right way to clarify their meanings. You've pointed out that this isn't a good way to clarify the meaning of these names, so I'll close this PR -- we can add those docstrings separately!

PR #341, opened 2022-03-14
Client now sends Objective Requests to engine instead of Objectives

### LINKED ISSUE #338: ObjectivesToSpawn can become stale
We originally construct the `ObjectiveToSpawn` in the [client](https://github.com/statechannels/go-nitro/blob/9f736c843d6ee097b2ead2f71ca7086500fa1dfb/client/client.go#L80) when the API is called. 

As part of constructing the virtual objective we [call](https://github.com/statechannels/go-nitro/blob/9f736c843d6ee097b2ead2f71ca7086500fa1dfb/protocols/virtualfund/virtualfund.go#L616) `GetTwoPartyLedger` to get the ledgers for the objective.

The problem is that there may be a delay between when we construct `ObjectiveToSpawn` and when the engine handles `ObjectiveToSpawn`.  This means that the ledger channels can become stale by the time the engine goes to handle `ObjectiveToSpawn`

An easy workaround can be seen [e9141f9](https://github.com/statechannels/go-nitro/pull/324/commits/e9141f908d9d0ccbab85870296d021fb8a8b4db3) where we simply refetch the ledger channels when the engine goes to process them.

I think a slightly better solution would move responsibility of constructing the objective to the engine, so it can construct it with the most recent data.

Instead of the Client generating full blown objective for `ObjectiveToSpawn` it could generate an `ObjectiveSpawnRequest` which contains enough information for the engine to generate the actual objective. This would mean that the client doesn't have to call `store.GetTwoPartyLedger`

lalexgap:
The client now sends an `ObjectiveRequest` struct to the engine. The engine is responsible for constructing an `Objective` for this `ObjectiveRequest`.

This removes the need for the `client` to lookup ledger channels in the store and to construct a proper Objective. The `client` now just needs to prepare a simple `ObjectiveRequest` that the engine will handle. The engine can then create the objective using the most recent store data when it's ready to process that request. This fixes #338 as we construct the objective only when the engine is ready to process it.

The protocols now have a `SpawnObjective` function that takes an `ObjectiveRequest` and returns an objective. The engine calls `SpawnObjective` to transform an `ObjectiveRequest` to an `Objective`.

The `ObjectiveRequest` interface has an `Id` function, so the client can easily get the objective Id and return that, without having to construct a full Objective. This also means that the `client` is still responsible for nonce management, since  we need to know the nonce to generate the objective id. For now I think this is ok?

**Note:**  One change is that is easy to miss is the change to use random nonces instead of 0. This allows us to create multiple channels with the same participants.

---

#### Code quality

#### Project management

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -3,9 +3,7 @@ package client // import "github.com/statechannels/go-nitro/client"
 
 import (
 	"io"
-	"math/rand"
 
-	"github.com/statechannels/go-nitro/channel/state/outcome"
 	"github.com/statechannels/go-nitro/client/engine"
 	"github.com/statechannels/go-nitro/client/engine/chainservice"
 	"github.com/statechannels/go-nitro/client/engine/messageservice"
@@ -62,17 +60,8 @@ func (c *Client) CompletedObjectives() <-chan protocols.ObjectiveId {
 }
 
 // CreateVirtualChannel creates a virtual channel with the counterParty using ledger channels with the intermediary.
-func (c *Client) CreateVirtualChannel(counterParty types.Address, intermediary types.Address, appDefinition types.Address, appData types.Bytes, outcome outcome.Exit, challengeDuration *types.Uint256) protocols.ObjectiveId {
-	objectiveRequest := virtualfund.ObjectiveRequest{
-		MyAddress:         *c.Address,
-		CounterParty:      counterParty,
-		Intermediary:      intermediary,
-		AppDefinition:     appDefinition,
-		AppData:           appData,
-		Outcome:           outcome,
-		ChallengeDuration: challengeDuration,
-		Nonce:             rand.Int63(), // TODO: Proper nonce management!
-	}
+func (c *Client) CreateVirtualChannel(objectiveRequest virtualfund.ObjectiveRequest) protocols.ObjectiveId {
+
 	apiEvent := engine.APIEvent{
 		ObjectiveToSpawn: objectiveRequest,
 	}
@@ -83,17 +72,8 @@ func (c *Client) CreateVirtualChannel(counterParty types.Address, intermediary t
 }
 
 // CreateDirectChannel creates a directly funded channel with the given counterparty
-func (c *Client) CreateDirectChannel(counterparty types.Address, appDefinition types.Address, appData types.Bytes, outcome outcome.Exit, challengeDuration *types.Uint256) protocols.ObjectiveId {
-	// Convert the API call into an internal event.
-	objectiveRequest := directfund.ObjectiveRequest{
-		MyAddress:         *c.Address,
-		CounterParty:      counterparty,
-		AppDefinition:     appDefinition,
-		AppData:           appData,
-		Outcome:           outcome,
-		ChallengeDuration: challengeDuration,
-		Nonce:             rand.Int63(), // TODO: Proper nonce management!
-	}
+func (c *Client) CreateDirectChannel(objectiveRequest directfund.ObjectiveRequest) protocols.ObjectiveId {
+
 	apiEvent := engine.APIEvent{
 		ObjectiveToSpawn: objectiveRequest,
 	}
--- a/client/client.go
+++ b/client/client.go
@@ -21,7 +21,6 @@ type Client struct {
 	engine              engine.Engine // The core business logic of the client
 	Address             *types.Address
 	completedObjectives chan protocols.ObjectiveId
-	store               *store.Store
 }
 
 // New is the constructor for a Client. It accepts a messaging service, a chain service, and a store as injected dependencies.
@@ -30,7 +29,7 @@ func New(messageService messageservice.MessageService, chainservice chainservice
 	c.Address = store.GetAddress()
 	c.engine = engine.New(messageService, chainservice, store, logDestination)
 	c.completedObjectives = make(chan protocols.ObjectiveId, 100)
-	c.store = &store
+
 	// Start the engine in a go routine
 	go c.engine.Run()
 
--- a/client/client.go
+++ b/client/client.go
@@ -2,12 +2,9 @@
 package client // import "github.com/statechannels/go-nitro/client"
 
 import (
-	"fmt"
 	"io"
-	"math/big"
+	"math/rand"
 
-	"github.com/statechannels/go-nitro/channel"
-	"github.com/statechannels/go-nitro/channel/state"
 	"github.com/statechannels/go-nitro/channel/state/outcome"
 	"github.com/statechannels/go-nitro/client/engine"
 	"github.com/statechannels/go-nitro/client/engine/chainservice"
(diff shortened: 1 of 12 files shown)


client/client.go
```diff
@@ -67,63 +63,43 @@ func (c *Client) CompletedObjectives() <-chan protocols.ObjectiveId {
...
-		panic(fmt.Sprintf("Could not find ledger channel for participants %v,%v", *c.Address, intermediary))
+	objectiveRequest := virtualfund.ObjectiveRequest{
+		MyAddress:         *c.Address,
+		CounterParty:      counterParty,
+		Intermediary:      intermediary,
+		AppDefinition:     appDefinition,
+		AppData:           appData,
+		Outcome:           outcome,
+		ChallengeDuration: challengeDuration,
+		Nonce:             rand.Int63(), // TODO: Proper nonce management!
```
andrewgordstewart, line 74: An alternative is to expect the API to provide a Nonce. IMO, this is better than leaving a TODO, which in my experience is easy to forget about.

Having the client API require the nonce means:
- if it's a bad design, we will hear complaints about it
- users can actually use the nonce as an idempotency key

lalexgap: That's good point about TODOs being easy to forget!

I've created an [issue](https://github.com/statechannels/go-nitro/issues/348) for this so we can hopefully keep track of it. I think it might make sense to tackle this outside of this PR? Adding nonces to the API seems like a big enough decision that it deserves its own PR/discussion.

client/client.go
```diff
@@ -67,63 +64,43 @@ func (c *Client) CompletedObjectives() <-chan protocols.ObjectiveId {
...
-
+	objectiveRequest := directfund.ObjectiveRequest{
+		MyAddress:         *c.Address,
+		CounterParty:      counterparty,
+		AppDefinition:     appDefinition,
+		AppData:           appData,
+		Outcome:           outcome,
+		ChallengeDuration: challengeDuration,
+		Nonce:             rand.Int63(), // TODO: Proper nonce management!
+	}
```
andrewgordstewart, line 96: I notice that:
- the signature for `CreateDirectChannel` is very long, which is bad for readability and maintainability
- this data structure is almost 1-1 with the data passed to `CreateDirectChannel`

Does it make sense for the signature of `CreateDirectChannel` to simply be `(request direct fund.ObjectiveRequest) protocols.ObjectiveId`?

client/engine/engine.go
```diff
@@ -159,8 +159,29 @@ func (e *Engine) handleChainEvent(chainEvent chainservice.Event) ObjectiveChange
...
 func (e *Engine) handleAPIEvent(apiEvent APIEvent) ObjectiveChangeEvent {
 	if apiEvent.ObjectiveToSpawn != nil {
-		return e.attemptProgress(apiEvent.ObjectiveToSpawn)
+
+		switch request := (apiEvent.ObjectiveToSpawn).(type) {
+
+		case virtualfund.ObjectiveRequest:
+			vfo, err := virtualfund.SpawnObjective(request, e.store.GetTwoPartyLedger)
+			if err != nil {
+				panic(err)
```
andrewgordstewart, line 168: I understand that we are not advised to `panic` in go. Is there a reason it's ok here?

protocols/directfund/directfund.go
```diff
@@ -358,6 +359,51 @@ func ConstructObjectiveFromMessage(m protocols.Message, myAddress types.Address)
...
+		Participants:      []types.Address{r.MyAddress, r.CounterParty},
+		ChannelNonce:      big.NewInt(r.Nonce),
+		ChallengeDuration: r.ChallengeDuration}
+
+	channelId, _ := fixedPart.ChannelId()
+	return protocols.ObjectiveId(ObjectivePrefix + channelId.String())
+}
+
+// SpawnObjective creates a new direct funding objective from a given request.
+func SpawnObjective(request ObjectiveRequest) (Objective, error) {
```
andrewgordstewart, line 385: It's not completely clear to me why we need to have both a `NewObjective` function and a `SpawnObjective` function. It seems like `SpawnObjective` is just another constructor for an `Objective` -- can it be removed?

client/client.go
```diff
@@ -67,63 +63,43 @@ func (c *Client) CompletedObjectives() <-chan protocols.ObjectiveId {
...
-			Outcome:           outcome,
-			TurnNum:           0,
-			IsFinal:           false,
-		},
-		*c.Address,
-		left, right)
-
 	apiEvent := engine.APIEvent{
-		ObjectiveToSpawn: &objective,
+		ObjectiveToSpawn: objectiveRequest,
```
andrewgordstewart, line 77: Should `ObjectiveToSpawn` be renamed, since it is no longer an objective to be spawned?

andrewgordstewart (commented):
It seems like this change is useful to make **if** our goal was to contain `store` API calls within the `engine`.

It's not obvious why this is a good fix to the problem of having a stale ledger channel. Can't there be a delay between when `SpawnObjective` gets the `right` ledger, and when the engine tries to crank the newly spawned objective? What stops the ledger state from going stale after this delay?

lalexgap:
> It's not obvious why this is a good fix to the problem of having a stale ledger channel. Can't there be a delay between when `SpawnObjective` gets the `right` ledger, and when the engine tries to crank the newly spawned objective? What stops the ledger state from going stale after this delay?

Since the engine execution is single threaded this problem is avoided at the moment. Since the engine only has one go-rountine running `Run` we know that nothing will update the "right" ledger before we crank it.

Once we introduce multithreaded engine execution we'll need some more complex locking as the ledger channel could go stale, like you mentioned. This locking would probably resolve the stale ledger issue (by forcing a refetch) but I still think it makes sense to defer the fetching of ledger channels to the engine.

protocols/directfund/directfund.go
lalexgap: `SpawnObjective` and `NewObjective` could probably get merged into one constructor that takes in an `ObjectiveRequest`. We'd probably need to update some tests that reference `NewObjective` but it wouldn't be too much work.

andrewgordstewart:
> Once we introduce multithreaded engine execution we'll need some more complex locking as the ledger channel could go stale, like you mentioned. This locking would probably resolve the stale ledger issue (by forcing a refetch) but I still think it makes sense to defer the fetching of ledger channels to the engine.

Got it. So, this change:
1. fixes problems with the engine on `main`
2. AND, defers fetching of ledger channels to the engine (which we want)

But, we don't expect this to fix the real problem -- we'll address the real problem in the future, once locking is in place.

Is that right?

client/engine/engine.go
lalexgap: This is mostly due to my personal preference 😛  I much prefer panicking over ignoring an error, so it's easy to figure out test failures and debug into them.

Right now we don't really have an error handling framework so we often ignore errors in the engine. This can make debugging/tracing pretty confusing (which motivated this [PR](https://github.com/statechannels/go-nitro/pull/346)) as the errors get swallowed and things don't fail until later.

Perhaps a better fix than panicking in individual engine functions is to just bubble up errors to the root `Run` loop. The `Run` loop can `panic` on error for now (to aid in debugging) and then later can be replaced with something more sophisticated.

andrewgordstewart: That seems to point to incorrect error handling, doesn't it? I feel like we should address that by correctly handling errors, instead of panicking (unless we have a good reason to panic).

lalexgap:
> 
> Got it. So, this change:
> 
> 1. fixes problems with the engine on `main`

Yup, for the single-threaded engine on `main` this prevents stale ledgers coming in from the API.

> 2. AND, defers fetching of ledger channels to the engine (which we want)

Yup!

> But, we don't expect this to fix the real problem -- we'll address the real problem in the future, once locking is in place.

That's correct. The real problem is executing multiple concurrent protocols that may read/write the same ledger channel. This will have to be resolved by locking mechanism so that a protocol can be sure that a ledger channel is safe to read/write.

client/client.go
lalexgap: We could rename it to `ObjectiveSpawnRequests`?

andrewgordstewart: I think that's up to you. I'm just pointing out that `ObjectiveToSpawn` used to contain an objective, which was supposed to be spawned. Now, it doesn't, and so the name seems stale?

andrewgordstewart (approved):
It seems like the use of `panic` should be addressed, and there's a (possibly?) stale name that needs to be updated. Up to you what to do with the rest of the feedback!

client/engine/engine.go
lalexgap: I would argue that error handling is basically not implemented at the moment, which leads to errors being ignored or not handling consistently. We don't have to use `panic`s but I do think unhandled errors in the engine should cause things to fail immediately (unless we can handle that error gracefully) and should fail consistently.

This can be tackled outside of this PR, as I can switch `func (e *Engine) handleAPIEvent(apiEvent APIEvent)` to return an error in this PR and just return an error.

lalexgap: I've removed the panics in [900370c](https://github.com/statechannels/go-nitro/pull/341/commits/900370c27ae4277a9a26271277a8551e80917c03) and created a new issue (#353) to deal with the fact we're ignoring errors in the engine.

protocols/directfund/directfund.go
lalexgap: I've refactored this so `NewObjective` now takes a `ObjectiveRequest`. The previous constructor is now a private function `constructFromInitialState`

client/client.go
lalexgap: Done in https://github.com/statechannels/go-nitro/pull/341/commits/eb74a99310ed81228aa32fa674faebed9a509dd9

client/client.go
lalexgap: I think I'm going to leave this as is for now as it follows the existing pattern of `ObjectiveToApprove` and `ObjectiveToReject`.

PR #324, opened 2022-03-09
WIP: Create test that fails without locking

lalexgap:
Right now this test is failing due to https://github.com/statechannels/go-nitro/issues/338 and https://github.com/statechannels/go-nitro/issues/339

PR #342, opened 2022-03-14
Quick fix to GetTwoPartyLedger

### LINKED ISSUE #339: GetTwoPartyLedger does not use channel store
The current [implementation](https://github.com/statechannels/go-nitro/blob/9f736c843d6ee097b2ead2f71ca7086500fa1dfb/client/engine/store/mockstore.go#L126) of GetTwoPartyLedger iterates through channels off of `ms.Objectives`.

However these could be stale.  It should be looking in `ms.channels` to get the most recent channel.

lalexgap:
Fixes #339 

This fixes GetTwoPartyLedger so it looks for channels in `ms.channel` instead of channels on objectives. This prevents GetTwoPartyLedger returning "stale" channels.

@NiloCK mentioned that this is addresses in his upcoming work, but I figured this small change could go in early (to help unblock #324 )


lalexgap:
> The new code makes more sense to me than the old code anyway: why should we look for channels inside objectives?

This was a holdover from when `ms.channels` didn't exist and there was only `ms.objectives`. It doesn't make much sense now!

PR #344, opened 2022-03-15
TestMessageService: remove pairwise connections

andrewgordstewart:
Instead of connecting peers in a pairwise fashion, this now defines a Broker struct, which is simply a address -> incoming message channel lookup.

(For some reason, the incoming message channel is called out -- why? This is fixed in https://github.com/statechannels/go-nitro/pull/336)

This makes it easier to create more dynamic test scenarios: each message service is already configured to receive messages when NewTestMessageService returns.


andrewgordstewart:
https://github.com/statechannels/go-nitro/pull/335 was merged into `ags/mockchain-subscribe` -- I'm not sure why the branch wasn't updated when #334 was merged.

lalexgap (approved):

PR #346, opened 2022-03-15
Engine panics instead of ignoring errors

lalexgap:
Instead of silently ignoring errors the engine now panics with the error. A more sophisticated error handling system should replace this but for now having the engine `panic` instead of soldering on makes debugging a lot easier.


lalexgap:
I'm closing this for now, as instead of panicking all over the place we can bubble up errors and panic in one place.

PR #347, opened 2022-03-15
Remove workaround for Store

lalexgap:
Now that the `MockStore` properly stores channels when calling `SetObjective` we can remove the workaround that manually calls `SetChannel`.

This is tested by the existing integration test.


lalexgap:
Closing in favour of #327

PR #349, opened 2022-03-15
WIP: Concurrency

lalexgap:
This is work to enable concurrency for the engine and to get the integration tests passing.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -21,15 +21,20 @@ type Client struct {
 	completedObjectives chan protocols.ObjectiveId
 }
 
-// New is the constructor for a Client. It accepts a messaging service, a chain service, and a store as injected dependencies.
-func New(messageService messageservice.MessageService, chainservice chainservice.ChainService, store store.Store, logDestination io.Writer) Client {
+// New is the constructor for a Client.
+// It accepts a messaging service, a chain service, and a store as injected dependencies.
+// It accepts a destination for the log file to write to.
+// It accepts the amount of concurrent run loops that the client should run.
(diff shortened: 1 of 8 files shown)


client/client.go
```diff
@@ -19,6 +19,7 @@ type Client struct {
 	engine              engine.Engine // The core business logic of the client
```
andrewgordstewart, line 19: @lalexgap Can this commit (1a151dd263b63b9449949599acba63f4da258899) be turned into a standalone PR?

lalexgap: Yup, I think it could be. Although I'm not super happy about having logic sprinkled everywhere to refetch the objective.

client/engine/channellocker.go
```diff
@@ -0,0 +1,65 @@
...
+
+// NewChannelLocker returns a new ChannelLocker
+func NewChannelLocker() *ChannelLocker {
+	return &ChannelLocker{
+		channelLocks: sync.Map{},
+	}
+}
+
+// Lock acquires a locks on the given channels.
+func (l *ChannelLocker) Lock(channelIds []types.Destination) {
```
andrewgordstewart, line 25: I thought that `Lock` would _always_ be called with exactly two channels. Does it make sense to reflect that in the signature? This might also simplify the implementation.

lalexgap: It depends on what role the participant is playing in a virtual channel. If the a participant is a hub they will have two ledger channels, otherwise they'll just have one ledger channel.

In our current implementation we also need to lock the virtual channel to prevent concurrently process messages for the same objective at the same time. This means we're locking either 2 or 3 channels when locking for a virtual fund objective.

lalexgap:
Closing this in favour of #374

PR #351, opened 2022-03-15
Chain Event now includes BlockNum

### LINKED ISSUE #350: Chain Events arriving out of order breaks direct funding protocol
Right now the direct funding protocol simply sets `OnChainFunding` based on the [chain event](https://github.com/statechannels/go-nitro/blob/585fe7933bad95aa1fe0cd35a081da8b070edc82/protocols/directfund/directfund.go#L142). This means that if chain events arrive out of order, the direct funding  protocol could override `OnChainFunding` with stale funding information.

I happened to run into this while working on a multi-threaded integration test but I think this it could be an issue in a single threaded case as well, as we're not guaranteed to receive chain events in the correct order.

A simple solution would be attaching a `BlockNumber` to the chain events the chain service sends out. The `directFunding` protocol would then only update the `OnChainFunding` on the channel when there is a new block number.

lalexgap:
Fixes #350 

Currently if chain events arrive out of order the direct funding protocol will overwrite a channels `OnChainFunding` with stale information. I was able to run into this scenario when working on integration tests and it seems slightly dangerous to assume chain events will be received in order.

To address I've added a `BlockNum` to the `ChainEvent`. The direct funding protocol can look at the `BlockNum` and determine if chain event funding is new information or stale.

The `MockChain` has been updated to increase the `BlockNum` for every transaction it handles.

`TestUpdate` has been updated to test that we ignore stale block numbers.

---
#### Code quality

#### Project management

### DIFF
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -146,7 +146,12 @@ func (e *Engine) handleChainEvent(chainEvent chainservice.Event) ObjectiveChange
 		e.logger.Printf("handleChainEvent: No objective in store for channel with id %s", chainEvent.ChannelId)
 		return ObjectiveChangeEvent{}
 	}
-	event := protocols.ObjectiveEvent{Holdings: chainEvent.Holdings, BlockNum: chainEvent.BlockNum, AdjudicationStatus: chainEvent.AdjudicationStatus, ObjectiveId: objective.Id()}
+	event := protocols.ObjectiveEvent{
+		Holdings:           chainEvent.Holdings,
+		BlockNum:           chainEvent.BlockNum,
+		AdjudicationStatus: chainEvent.AdjudicationStatus,
+		ObjectiveId:        objective.Id(),
+	}
 	updatedObjective, err := objective.Update(event)
 	if err != nil {
 		// TODO handle error
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -146,7 +146,7 @@ func (e *Engine) handleChainEvent(chainEvent chainservice.Event) ObjectiveChange
 		e.logger.Printf("handleChainEvent: No objective in store for channel with id %s", chainEvent.ChannelId)
 		return ObjectiveChangeEvent{}
 	}
-	event := protocols.ObjectiveEvent{Holdings: chainEvent.Holdings, AdjudicationStatus: chainEvent.AdjudicationStatus, ObjectiveId: objective.Id()}
+	event := protocols.ObjectiveEvent{Holdings: chainEvent.Holdings, BlockNum: chainEvent.BlockNum, AdjudicationStatus: chainEvent.AdjudicationStatus, ObjectiveId: objective.Id()}
 	updatedObjective, err := objective.Update(event)
 	if err != nil {
 		// TODO handle error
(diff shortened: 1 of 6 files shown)


client/engine/engine.go
```diff
@@ -143,7 +143,7 @@ func (e *Engine) handleChainEvent(chainEvent chainservice.Event) ObjectiveChange
 		e.logger.Printf("handleChainEvent: No objective in store for channel with id %s", chainEvent.ChannelId)
 		return ObjectiveChangeEvent{}
 	}
-	event := protocols.ObjectiveEvent{Holdings: chainEvent.Holdings, AdjudicationStatus: chainEvent.AdjudicationStatus, ObjectiveId: objective.Id()}
+	event := protocols.ObjectiveEvent{Holdings: chainEvent.Holdings, BlockNum: chainEvent.BlockNum, AdjudicationStatus: chainEvent.AdjudicationStatus, ObjectiveId: objective.Id()}
```
NiloCK, line 149:
```suggestion
	event := protocols.ObjectiveEvent{
		Holdings:           chainEvent.Holdings,
		BlockNum:           chainEvent.BlockNum,
		AdjudicationStatus: chainEvent.AdjudicationStatus,
		ObjectiveId:        objective.Id(),
	}
```

protocols/directfund/directfund_test.go
```diff
@@ -119,16 +119,35 @@ func TestUpdate(t *testing.T) {
 
 	// Finally, add some Holdings information to the event
 	// Updating the objective with this event should overwrite the holdings that are stored
-	e.Holdings = types.Funds{}
-	e.Holdings[common.Address{}] = big.NewInt(3)
-	updatedObjective, err = s.Update(e)
+	newFunding := types.Funds{}
+	newFunding[common.Address{}] = big.NewInt(3)
```
NiloCK, line 142: just fyi (I find this more pleasing, but reasonable people might not):

```suggestion
	newFunding := types.Funds{
		common.Address{}: big.NewInt(3),
	}
```

NiloCK, line 142: just fyi (I find this more pleasing, but reasonable people might not):

```suggestion
	newFunding := types.Funds{
		common.Address{}: big.NewInt(3),
	}
```

protocols/directfund/directfund.go
```diff
@@ -39,6 +39,7 @@ type Objective struct {
 	myDepositSafetyThreshold types.Funds // if the on chain holdings are equal to this amount it is safe for me to deposit
 	myDepositTarget          types.Funds // I want to get the on chain holdings up to this much
 	fullyFundedThreshold     types.Funds // if the on chain holdings are equal
+	latestBlockNumber        uint64      //the latest block number we've seen
```
NiloCK, line 43:
```suggestion
	latestBlockNumber        uint64      // the latest block number we've seen
```

NiloCK, line 43:
```suggestion
	latestBlockNumber        uint64      // the latest block number we've seen
```

NiloCK (approved):
:+1: 

Longer term the `chainservice` api may / must be further enriched with things like block difficulty / total difficulty, I guess, as guard against chain reorgs. I peeked at `server-wallet` and saw that it updates based on blocks 5-from-the-top.

See my PR #362, which comes into this branch, for an optional refactor on `mockchain`.

PR #354, opened 2022-03-16
Introduce a rudimentary "benchmarking" test

andrewgordstewart:
This test asserts that creating one virtual channel takes at most 10ms.

If run locally in verbose mode, it will print how long the VirtualFund objective took to complete:

```
go test ./... -run TestBenchmark -v

...
=== RUN   TestBenchmark
VirtualFund-0xe25bb77c1b36b618f553c1fab4b473c53c2a67a1c63591bdc2a72b244e039fd6 took 6.005875ms
...
```

**How Has This Been Tested?** [Optional]

If you add `time.Sleep(time.Millisecond * 10)` in the body of `benchmarkVirtualChannelCreation`, you will see:

```
➜  go-nitro git:(ags/benchmark) ✗ go test ./... -run TestBenchmark -v

...
=== RUN   TestBenchmark
    virtualfund_benchmark_test.go:68: Ran out of time. 0 out of 1 completed
--- FAIL: TestBenchmark (0.01s)
```

This means that the test will fail in the case where `waitTimeForCompletedObjectiveIds` takes more than 10ms to finish.

#### Code quality

- [ ] I have performed a self-review of my own code

#### Project management

- [ ] I have linked the appropriate github issue

PR #364, opened 2022-03-17
Refactor multi-party virtual funding integration test

lalexgap:
This refactor introduces a helper function `createVirtualChannels` that allows us to create virtual channels in bulk. This is handy to easily scale up or down the amount of virtual channels.


andrewgordstewart (approved):
This seems fine. Do note this overlaps a bit with #354!

lalexgap:
I don't think this is needed yet so I'm going to close it.

Issue #370, opened 2022-03-20
test cases are hard to read


andrewgordstewart:
Here is some [sample test data](https://github.com/statechannels/go-nitro/blob/56e0ef08fe21805b6e63409c9b8a76bde6703e6e/channel/state/outcome/exit_test.go#L206-L224). This data is hard for humans to read:
```
e := Exit{
	SingleAssetExit{
		Asset: types.Address{0},
		Allocations: Allocations{
			{
				Destination:    aliceDestination,
				Amount:         big.NewInt(243),
				AllocationType: 0,
				Metadata:       make(types.Bytes, 0),
			},
			{
				Destination:    bobDestination,
				Amount:         big.NewInt(309),
				AllocationType: 0,
				Metadata:       make(types.Bytes, 0),
			},
		},
	},
}
```

Compare with:
```
e := simpleOutcome(
	allocation(alice, aBal),
	allocation(bob, bBal),
)
```

This snippet shows how a helper function can make tests easier to read.

andrewgordstewart:
Context: I am trying to understand to what extend are functions like `DivertToGuarantee` are unit tested, but am having difficulty reading existing test files..

Issue #242, opened 2022-02-15
Consider reorganizing tests into `package_test` packages


NiloCK:
A common Go practice is to create a separate package `foo_test` which tests the public api of package `foo`. Instead of this, we've been writing our tests directly inside their respective packages with the main motivation being:
- **access to package-private fields & methods**: we've been pretty aggressive about encapsulating protocol complexities so that the `engine` and `client` can treat them all generically. We desire access to internal objective state so that tests can probe it.
- **momentum**: we started out with these test locations. They worked well enough, so we kept going.

This has caused a pain point with the creation and sharing of test data. Summary of the problem is:

1. In the case of packages defining data structures (eg, `outcome`, `state`), it's not possible to declare test data that is reusable in other packages (where it is needed!) without making the data part of the public API of the package. 
2. When restricting test data for package `foo` to the package-private level in, say, `foo_test.go`, all variables created at the file scope become part of the global scope in this package, polluting that namespace. (Go does not have 'file scope' - packages can be considered single files for the most part)

The first is a total no-go. The second creates a deterrent - programming inside the package becomes more onerous (and error prone) as more labelled test data is added. And adding more labelled test data isn't even that useful since it is not exported!

This is all the worse because declaring fairly complex literal struct data in Go is burdensome. (Possibly opinion).

## Non-idea

In Typescript, this is a non-issue because circular dependencies are allowed. They are **not** allowed in Go. Preparing exportable test data of types defined in `foo` requires an import of `foo`, and so `foo` cannot re-import that data for its tests.

![image](https://user-images.githubusercontent.com/10780590/153989002-7ce880c2-0b44-4b84-a86f-a53fbbc39a63.png)

## Current idea

@geoknee came up with a couple of incremental improvements to the above statement of things:

1. place test data creation inside of individual test functions rather than at top-of-file as in our TS tests. This hides it from the global scope in the package.
2. amalgamate test functions inside one mega-function and use [`t.Run`](https://github.com/statechannels/go-nitro/blob/ac4b9cd0d5ab408345fe8f55f782efb95211b215/protocols/virtualfund/virtualfund-single-hop_test.go#L379) to call each individually. Now multiple tests have access to single-setup non polluting test data.

This still leaves us unable to declare test data in `foo` and use it in `bar`.

## Proposed Idea

The reusable test data pain goes away if we restrict to testing the public APIs of packages:

[...]

PR #376, opened 2022-03-22
Add a `testdata` package

NiloCK:
This is incremental towards #242, and tangentially #370.

Observation from @andrewgordstewart in #370 is that complex literals in Go are not easy to read. My own experience is that they are tedious to write as well.

This tedium, combined with the namespace problems described in #242, make test authoring some combination of non-DRY / painstaking / error prone.

PR includes a test data package inside an `internal` folder. `internal` has a special interpretation from the compiler - the testdata package defined here cannot be imported by other go modules. (ie, filecoin's lotus package could import `go-nitro/client` package but not `go-nitro/internal/testdata`).

PR includes a defined API for testdata, so that data can be made discoverable & is strongly typed. A test looking for `alice`, `bob`, can access them via `td.Actors.Alice`, etc. This API is possibly an overdesign, but note that it can be optionally used - persons should feel free add one-off data points to the `td` namespace, which can be gathered together under the API in the event that a sensible grouping for them becomes apparent.

PR includes a refactor of the existing `mockstore` test, with an intended follow-up enhancement where `mockstore` is tested against virtualfund & directdefund objectives.

Note: suggesting a slight loosening of inline documentation standards here. EG, the `td.actor` type has a convenience `Destination` function which wraps `types.AddressToDestination`. I can personally live with these types of helpers being undocumented because they are not part of `go-nitro`'s exportable API - only persons writing tests for `go-nitro` ought to see them.

---

#### Code quality

#### Project management

### DIFF
--- a/client/engine/store/mockstore_test.go
+++ b/client/engine/store/mockstore_test.go
@@ -27,7 +27,7 @@ func TestSetGetObjective(t *testing.T) {
 		t.Errorf("expected not to find the %s objective, but found %v", id, got)
 	}
 
-	want := td.Objectives.Directfund.GenericDFO
+	want := td.Objectives.Directfund.GenericDFO()
 
 	if err := ms.SetObjective(&want); err != nil {
 		t.Errorf("error setting objective %v: %s", want, err.Error())
@@ -49,7 +49,7 @@ func TestGetObjectiveByChannelId(t *testing.T) {
 
 	ms := store.NewMockStore(sk)
 
-	want := td.Objectives.Directfund.GenericDFO
+	want := td.Objectives.Directfund.GenericDFO()
 
 	if err := ms.SetObjective(&want); err != nil {
 		t.Errorf("error setting objective %v: %s", want, err.Error())
--- a/client/engine/store/mockstore_test.go
+++ b/client/engine/store/mockstore_test.go
@@ -1,25 +1,25 @@
-package store
+package store_test
 
 import (
 	"testing"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/google/go-cmp/cmp"
-	"github.com/statechannels/go-nitro/channel/state"
+	"github.com/statechannels/go-nitro/client/engine/store"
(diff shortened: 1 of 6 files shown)


internal/testdata/doc.go
```diff
@@ -0,0 +1,3 @@
+// package td (TestData) provides literals and utility functions for generating objects-of-concern
+// (objectives, channels, user-agents, etc) for other test packages.
+package td
```
geoknee, line 3: Why not call the package `testdata` ? It can be aliased to `td` when imported?

client/engine/store/mockstore_test.go
```diff
@@ -1,91 +1,69 @@
...
-		AppData:           ts.AppData,
-		AppDefinition:     ts.AppDefinition,
-		ChallengeDuration: ts.ChallengeDuration,
-		Nonce:             ts.ChannelNonce.Int64(),
-		Outcome:           ts.Outcome,
-	}
-	testObj, _ := directfund.NewObjective(request, false)
+	ms := store.NewMockStore(sk)
+
+	want := td.Objectives.Directfund.GenericDFO
```
geoknee, line 52: As I understand it, this is the result of a constructor call which is only called once (not entirely sure when, perhaps at compile time?). If I write another test and reuse this generic direct funding objective, is there a risk that the two instances have pointers to the same values, which could cause problems if we go about mutating those values during the tests?

Might it be better to expose the constructor function `genericDFO()` itself, and call that function from the test to ensure we get a fresh objective?

NiloCK: This is in fact what I had originally written, but refactored to replace the "complicated" function type with a concrete object. The generator is strictly better for the reason you've described.

fixed in 4033a4f

internal/testdata/doc.go
NiloCK: Felt like brevity was a virtue here because it will probably be repeatedly invoked in lots of places.

More descriptive name + importers discretion is better though.

Fixed in 5cefca3

PR #379, opened 2022-03-22
waitTimeForCompletedObjectiveIds only logs failed ids

lalexgap:
This changes `waitTimeForCompletedObjectiveIds` so when we return a timeout message we only list the ids that we're still waiting for not the complete list of the ids. Examining the logs after a test failure is much easier if we have a list of only the failed objectives.

I've also updated the timeout error message to use the timeout value instead of a hard-coded value.

### DIFF
--- a/client_test/helpers_test.go
+++ b/client_test/helpers_test.go
@@ -61,7 +61,7 @@ func waitTimeForCompletedObjectiveIds(t *testing.T, client *client.Client, timeo
 				incompleteIds = append(incompleteIds, id)
 			}
 		}
-		t.Fatalf("Objective ids %s failed to complete in one second on client %s", incompleteIds, client.Address)
+		t.Fatalf("Objective ids %s failed to complete on client %s within %s", incompleteIds, client.Address, timeout)
 	case <-allDone:
 		return
 	}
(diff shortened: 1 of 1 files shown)


client_test/helpers_test.go
```diff
@@ -23,41 +23,47 @@ const defaultTimeout = time.Second
...
 
-// waitForCompletedObjectiveIds waits for completed objectives and returns when the all objective ids provided have been completed.
-func waitForCompletedObjectiveIds(client *client.Client, ids ...protocols.ObjectiveId) {
-	// Create a map of all objective ids to wait for and set to false
+	allDone := make(chan interface{})
+	// Create a map to keep track of completed objectives
 	completed := make(map[protocols.ObjectiveId]bool)
 	for _, id := range ids {
 		completed[id] = false
 	}
```
geoknee, line 53: Outside scope of this PR but I believe these lines are unnecessary -- the map will return a default (zero) value of `false` for unintialized keys.

geoknee (approved):

client_test/helpers_test.go
lalexgap: Good point, updated in [ed05bab](https://github.com/statechannels/go-nitro/pull/379/commits/ed05babd81a3080773c6665961556b2b5ccddb76)

Issue #366, opened 2022-03-17
Ledger Protocol can stall depending on the order in which states are received


lalexgap:
Right now when creating a new ledger proposal we look for the most recent proposal (or supported state) and use the turn number after [that](https://github.com/statechannels/go-nitro/blob/ccae3d13fc84526692fa5fcfbd1d62b146a71e77/protocols/virtualfund/virtualfund.go#L718).

When a participant goes to accept a ledger proposal they are comparing the proposal state against a supported state to make sure their funds haven't [decreased](https://github.com/statechannels/go-nitro/blob/ccae3d13fc84526692fa5fcfbd1d62b146a71e77/protocols/virtualfund/virtualfund.go#L773).

However if there are multiple proposals we haven't accepted yet this check can fail. That's because the most recent proposal could contain multiple new guarantees compared to `supported`. The virtual funding protocol is only concerned with its guarantee so the funds spent on other guarantees cause `proposedMaintainsOurFunds` to be false.

This issue is dependent on the order in which messages are received/handled. If we receive/handle the proposals in order of turn num we're ok, since `supported` is always one turn behind the proposal we're handling.

geoknee:
This is something we missed when settling on the design for doing ledger updates for this sprint (#291). Although the leader / follower pattern prevents a race between ledger channel participants, it does not prevent a race between message handlers ( ~ multiple objective cranks) on a single participant's machine. 

This seems to underline that, as long as guarantees are added and removed from ledger channels _synchronously_, we will require a module for proposing / accepting those updates that has a view into _multiple_ virtualfund objectives, or at least has the ability to infer the intent of those objectives. This would seem to rule out the idea of making such updates entirely inside a the `virtualfund.Crank()` function. And to push us back in the direction of a queue of ledger requests which is processed separately from objective cranking.

andrewgordstewart:
#373  includes an approach that does not require a ledger request queue, and even supports the design goal of pure objective implementations!

kerzhner:
The leader determines the sequence of ledger updates by setting the turn number (aka version). The follower stores:
- The latest consensus turn number for a ledger channel.
- The proposed update turn number for a ledger channel. This can be stored on the objective.

When a ledger update arrives as part of the virtual fund objective, the follower has enough information to determine whether the update can be applied or is blocked on other updates.

Is the above correct? The goal of this comment is to anchor the conversation to create a common ground for discussing explicit queue vs implicit queue vs no queue.

[...]

PR #380, opened 2022-03-22
Add failing multiparty test

lalexgap:
Previously our virtual funding multi-party integration test was passing even though we knew about an [issue](https://github.com/statechannels/go-nitro/issues/366) that would cause it to fail.

This updates the test to create more virtual channels. This is enough to trigger the [ledger funding issue](https://github.com/statechannels/go-nitro/issues/366) as all it takes is one ledger proposal to arrive out of order. 

The test has been skipped to prevent CI from failing.

### DIFF
--- a/client_test/virtualfund_multiparty_integration_test.go
+++ b/client_test/virtualfund_multiparty_integration_test.go
@@ -5,8 +5,10 @@ import (
 	"math/rand"
 	"testing"
 
+	"github.com/statechannels/go-nitro/client"
 	"github.com/statechannels/go-nitro/client/engine/chainservice"
 	"github.com/statechannels/go-nitro/client/engine/messageservice"
+	"github.com/statechannels/go-nitro/protocols"
 	"github.com/statechannels/go-nitro/protocols/virtualfund"
 	"github.com/statechannels/go-nitro/types"
 )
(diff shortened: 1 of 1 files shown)


geoknee:
Would you expect the current test on `main` to pass 100% of the time? I'm struggling a bit to see why simply adding more virtual channels is the thing to break the test... does it just increase the probability from close to zero to much higher?

lalexgap:
> Would you expect the current test on `main` to pass 100% of the time? I'm struggling a bit to see why simply adding more virtual channels is the thing to break the test... does it just increase the probability from close to zero to much higher?

I think the test on `main` could technically fail, but the chance of that is much lower than when we create multiple virtual channels.

The original multiparty test was creating two virtual channels, one between Alice and Bob and one between Alice and Brian. This means we have three ledger channels : `L_A_I`, `L_I_Bob` and `L_I_Brian`. The ledger issue occurs when we receive a ledger proposal "out of order", which could only happen on `L_A_I` (since the other ledgers only ever receive one proposal). 

With the small amount of messages being sent I suspect it's very unlikely for Irene to receive/process the two proposals for `L_A_I` out of order. By creating multiple virtual channels we create many more opportunities for ledger messages to arrive out of order.

Perhaps a better way of reproducing this issue is [re-ordering/delaying messages](https://github.com/statechannels/go-nitro/pull/381).

lalexgap:
Closing in favour of #381 as I've added a test to reproduce this issue there.

PR #381, opened 2022-03-22
Add message delay to test message service

### LINKED ISSUE #366: Ledger Protocol can stall depending on the order in which states are received

lalexgap:
The test message service now accepts a `maxDelay` that can be used to simulate message latency when testing.

The `TestMessageService` handles the message in a separate go routine after reading it from the `t.in` chan. This allows one message to be delayed without blocking the `TestMessageService` from handling other messages. Otherwise we'd simply slow down the `TestMessageService` but not affect the order in which messages are processed.

A new test has been added that takes advantage of the `maxDelay` to reproduce #366. It has been skipped for now.

### DIFF
--- a/client/engine/messageservice/test-messageservice.go
+++ b/client/engine/messageservice/test-messageservice.go
@@ -63,7 +63,10 @@ func (t TestMessageService) In() chan<- protocols.Message {
 	return t.in
 }
 
-func (t TestMessageService) handleMessage(message protocols.Message, b Broker) {
+// dispatchMessage is responsible for dispatching a message to the appropriate peer message service.
+// If there is a mean delay it will wait a random amount of time(based on meanDelay) before sending the message.
+// It serializes and deserializes a message to mimic a real message service.
+func (t TestMessageService) dispatchMessage(message protocols.Message, b Broker) {
 
 	if t.meanDelay > 0 {
 		randomDelay := time.Duration(rand.Int63n(t.meanDelay.Nanoseconds()))
@@ -95,7 +98,7 @@ func (t TestMessageService) connect(b Broker) {
 	go func() {
 		for message := range t.in {
 
-			go t.handleMessage(message, b)
+			go t.dispatchMessage(message, b)
 		}
 
 	}()
--- a/client/engine/messageservice/test-messageservice.go
+++ b/client/engine/messageservice/test-messageservice.go
@@ -22,9 +22,9 @@ type TestMessageService struct {
 	address types.Address
 
 	// connection to Engine:
-	in        chan protocols.Message // for recieving messages from engine
-	out       chan protocols.Message // for sending message to engine
-	meanDelay time.Duration          // average delay for messages
+	in       chan protocols.Message // for recieving messages from engine
+	out      chan protocols.Message // for sending message to engine
+	maxDelay time.Duration          // the max delay for messages
 }
 
 // A Broker manages a mapping from identifying address to a TestMessageService,
@@ -43,12 +43,14 @@ func NewBroker() Broker {
 }
 
 // NewTestMessageService returns a running TestMessageService
-func NewTestMessageService(address types.Address, broker Broker, meanDelay time.Duration) TestMessageService {
+// It accepts an address, a broker, and a max delay for messages.
+// Messages will be handled with a random delay between 0 and maxDelay
+func NewTestMessageService(address types.Address, broker Broker, maxDelay time.Duration) TestMessageService {
 	tms := TestMessageService{
-		address:   address,
-		in:        make(chan protocols.Message, 5),
-		out:       make(chan protocols.Message, 5),
-		meanDelay: meanDelay,
+		address:  address,
+		in:       make(chan protocols.Message, 5),
+		out:      make(chan protocols.Message, 5),
+		maxDelay: maxDelay,
 	}
 
 	tms.connect(broker)
@@ -67,12 +69,11 @@ func (t TestMessageService) In() chan<- protocols.Message {
 // If there is a mean delay it will wait a random amount of time(based on meanDelay) before sending the message.
 // It serializes and deserializes a message to mimic a real message service.
 func (t TestMessageService) dispatchMessage(message protocols.Message, b Broker) {
-
-	if t.meanDelay > 0 {
-		randomDelay := time.Duration(rand.Int63n(t.meanDelay.Nanoseconds()))
-		delayAmount := t.meanDelay/2 + randomDelay
-		time.Sleep(delayAmount)
+	if t.maxDelay > 0 {
+		randomDelay := time.Duration(rand.Int63n(t.maxDelay.Nanoseconds()))
+		time.Sleep(randomDelay)
 	}
+
 	peerChan, ok := b.services[message.To]
 	if ok {
 		// To mimic a proper message service, we serialize and then
--- a/client/engine/messageservice/test-messageservice.go
+++ b/client/engine/messageservice/test-messageservice.go
@@ -2,6 +2,8 @@ package messageservice
 
 import (
 	"fmt"
+	"math/rand"
+	"time"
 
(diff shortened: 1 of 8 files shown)


client/engine/messageservice/test-messageservice.go
```diff
@@ -20,8 +22,9 @@ type TestMessageService struct {
 	address types.Address
 
 	// connection to Engine:
-	in  chan protocols.Message // for recieving messages from engine
-	out chan protocols.Message // for sending message to engine
+	in        chan protocols.Message // for recieving messages from engine
+	out       chan protocols.Message // for sending message to engine
+	meanDelay time.Duration          // average delay for messages
```
geoknee, line 27: Seeing as you are only letting the calling code specify the mean delay, it might be nice to at least document the distribution of delays that will be used. I think what you end up with is a uniform distribution over `[meanDelay/2, 3*meanDelay/2]`.

client_test/directfund_integration_test.go
```diff
@@ -5,52 +5,68 @@ import (
...
+	return ids
+}
+func directlyFundALedgerChannel(t *testing.T, alpha client.Client, beta client.Client) {
+	ids := createDirectlyFundedChannels(t, alpha, beta, 1)
+	id := ids[0]
 	waitTimeForCompletedObjectiveIds(t, &alpha, defaultTimeout, id)
 	waitTimeForCompletedObjectiveIds(t, &beta, defaultTimeout, id)
 }
+
+const MEAN_MESSAGE_DELAY = 100 * time.Millisecond
```
geoknee, line 57: How did you arrive at this number?

It seems to me as if what we really want is a system for generating random _orderings_ for messages that are to be dispatched. Using random delays is just one proxy for achieving that. Of course, random delays are very realistic and attractive to use in testing for that reason; but if the main symptom of message delays is message reordering, we might be smart to simply shortcut straight to reordering in our tests.

If we do stick with the random delay method, then it seems the condition we want to be in to have reordering (with high probability) is that the _variance_ in the delay is large. But if it gets too large we are needlessly slowing down the test. In your current system, the variance will increase if the mean increases.

Thinking about the distribution of delays again, it might make sense to specify a _maximum_ delay rather than a mean delay -- and that way, include delays of 0 in the distribution. Having a nonzero minimum delay doesn't make a heap of sense to me (if we are just trying to reorder messages). It just adds an unnecessary (albeit negligible, in all likelihood) lag to the test.

geoknee (approved):
I think this will increase confidence in our code, as long as it doesn't introduce flickers :-P. 

Have you done any sanity checks (i.e. checking the logs), to make sure messages are actually being reordered?

Have left some thoughts about some possible improvements.

lalexgap:
> I think this will increase confidence in our code, as long as it doesn't introduce flickers :-P.
> 
> Have you done any sanity checks (i.e. checking the logs), to make sure messages are actually being reordered?
> 
> Have left some thoughts about some possible improvements.

Instead of modifying the direct funding integration test I've now created a new message delay test in https://github.com/statechannels/go-nitro/pull/381/commits/fafff3283790928f1382e4dce616d09059d2518a. This test runs virtual funding with 5 virtual channels and a message delay.

Running that test without the message delay **usually** passes and running the test with the message delay **usually** fails.

I'm also seeing evidence of message re-ordering occurings in the logs:
```
0x111A00: 14:08:33.860012 engine.go:119: Handling inbound message {to:0x111A00868581f73AB42FEEF67D235Ca09ca1E8db objectiveId:VirtualFund-0x099a4f9c56a9108781a2b91239be694d779180f9ca2fbb5817e9e009c6c9d7ef signedStates:[{turnNum:3 channelId:0x9b911093d5dd6104278028be8afa3d59e241513c325eeb9b3d0027011d7516f6}]}
0x111A00: 14:08:33.861779 engine.go:252: Objective VirtualFund-0x099a4f9c56a9108781a2b91239be694d779180f9ca2fbb5817e9e009c6c9d7ef is WaitingForCompleteFunding
0x111A00: 14:08:33.863450 engine.go:119: Handling inbound message {to:0x111A00868581f73AB42FEEF67D235Ca09ca1E8db objectiveId:VirtualFund-0x7bde2749b0478f61281e3e83d5266d6351dbaea65f4c220ce99f0113b9eba7ed signedStates:[{turnNum:4 channelId:0x9b911093d5dd6104278028be8afa3d59e241513c325eeb9b3d0027011d7516f6}]}
0x111A00: 14:08:33.864574 engine.go:252: Objective VirtualFund-0x7bde2749b0478f61281e3e83d5266d6351dbaea65f4c220ce99f0113b9eba7ed is WaitingForCompleteFunding
0x111A00: 14:08:33.876431 engine.go:119: Handling inbound message {to:0x111A00868581f73AB42FEEF67D235Ca09ca1E8db objectiveId:VirtualFund-0xcc2671d924e2d6c2f12914ae81712d8ff1c98ca510a9b0ae984a3362f9eadbbd signedStates:[{turnNum:2 channelId:0x9b911093d5dd6104278028be8afa3d59e241513c325eeb9b3d0027011d7516f6}]}
0x111A00: 14:08:33.878973 engine.go:252: Objective VirtualFund-0xcc2671d924e2d6c2f12914ae81712d8ff1c98ca510a9b0ae984a3362f9eadbbd is WaitingForCompleteFunding
0x111A00: 14:08:33.900511 engine.go:119: Handling inbound message {to:0x111A00868581f73AB42FEEF67D235Ca09ca1E8db objectiveId:VirtualFund-0xc25ff44b307b1ef0c5c2cc0a5a39cd1cb213da2232c753823bd45f2c9e8690e2 signedStates:[{turnNum:6 channelId:0x9b911093d5dd6104278028be8afa3d59e241513c325eeb9b3d0027011d7516f6}]}
0x111A00: 14:08:33.903727 engine.go:252: Objective VirtualFund-0xc25ff44b307b1ef0c5c2cc0a5a39cd1cb213da2232c753823bd45f2c9e8690e2 is WaitingForCompleteFunding
```

client_test/virtual_fund_message_delay_test.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/statechannels/go-nitro/client/engine/messageservice"
+	"github.com/statechannels/go-nitro/protocols"
+	"github.com/statechannels/go-nitro/protocols/virtualfund"
+	"github.com/statechannels/go-nitro/types"
+)
+
+const MAX_MESSAGE_DELAY = time.Millisecond * 100
+
+// Since we are delaying messages we allow for slightly more time to complete the objective
+const OBJECTIVE_TIMEOUT = time.Second * 2
```
geoknee, line 20: Nit 1: I would replace "slightly more time" with "enough time" in the comment (since we don't know what it is slightly more than...)


Nit 2: Consider putting these consts  _inside_ `TestVirtualFundWithMessageDelays`, unless you want to use them in other tests in the `client_test` package.

geoknee (approved):
Thanks for making the improvements 🌋

Issue #318, opened 2022-03-08
Connection should implement ExpectedGuarantees as a function instead of duplicate data


lalexgap:
in that PR, it would be nifty if `Connection` implemented the `ExpectedGuarantees()` method!

another alternative is to modify `ExpectedGuarantees` to include the `LeftAmount` and `RightAmount`. In that way, the actual guarantees (the data structure included in a channel's outcome) are easily computable, by summing `LeftAmount` and `RightAmount`.

_Originally posted by @andrewgordstewart in https://github.com/statechannels/go-nitro/pull/312#r820974364_

PR #382, opened 2022-03-22
Convert expected guarantees to a getter

kerzhner:
Fixes https://github.com/statechannels/go-nitro/issues/318

The changes in this PR are fairly straightforward. The most interesting bit is the introduction of a `panic`. This seems like the lesser evil compared to the changeset required to return errors from `getExpectedGuarantees`. That option is prototyped in https://github.com/statechannels/go-nitro/compare/expected-guarantee?expand=1.

### DIFF
--- a/protocols/virtualfund/serde.go
+++ b/protocols/virtualfund/serde.go
@@ -20,9 +20,8 @@ type assetGuarantee struct {
 // jsonConnection is a serialization-friendly struct representation
 // of a Connection
 type jsonConnection struct {
-	Channel            types.Destination
-	ExpectedGuarantees []assetGuarantee
-	GuarnateeInfo      GuaranteeInfo
+	Channel       types.Destination
+	GuarnateeInfo GuaranteeInfo
 }
 
 // MarshalJSON returns a JSON representation of the Connection
@@ -30,14 +29,7 @@ type jsonConnection struct {
 // NOTE: Marshal -> Unmarshal is a lossy process. All channel data
 //       other than the ID is dropped
 func (c Connection) MarshalJSON() ([]byte, error) {
-	guarantees := []assetGuarantee{}
-	for asset, guarantee := range c.ExpectedGuarantees {
-		guarantees = append(guarantees, assetGuarantee{
-			asset,
-			guarantee,
-		})
-	}
-	jsonC := jsonConnection{c.Channel.Id, guarantees, c.GuaranteeInfo}
+	jsonC := jsonConnection{c.Channel.Id, c.GuaranteeInfo}
 	bytes, err := json.Marshal(jsonC)
 
 	if err != nil {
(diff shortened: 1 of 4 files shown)


protocols/virtualfund/virtualfund.go
```diff
@@ -344,39 +340,56 @@ func (o Objective) Channels() []*channel.Channel {
...
+	}
+	_, err := metadata.Encode()
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
 // insertExpectedGuaranteesForLedgerChannel mutates the reciever Connection struct.
```
geoknee, line 367: I think this godoc is stale.

protocols/virtualfund/serde.go
```diff
@@ -5,39 +5,23 @@ import (
...
-}
-
 // jsonConnection is a serialization-friendly struct representation
 // of a Connection
 type jsonConnection struct {
-	Channel            types.Destination
-	ExpectedGuarantees []assetGuarantee
-	GuarnateeInfo      GuaranteeInfo
+	Channel       types.Destination
+	GuarnateeInfo GuaranteeInfo
```
geoknee, line 16:
```suggestion
	GuaranteeInfo GuaranteeInfo
```

protocols/virtualfund/virtualfund.go
```diff
@@ -344,39 +340,56 @@ func (o Objective) Channels() []*channel.Channel {
...
-		Right: right,
+		Left:  connection.GuaranteeInfo.Left,
+		Right: connection.GuaranteeInfo.Right,
 	}
 	encodedGuarantee, err := metadata.Encode()
+	// This error is unexpected. insertGuaranteeInfo checks that the guarantee metadata can be encoded.
+	// If this panic is triggered, GuranteeInfo has been modified after creation
 	if err != nil {
-		return err
+		panic(err)
```
geoknee, line 378: I suppose we can't be 100% sure that this is the reason, if we get a panic. What if the metadata cannot be encoded, but the first time we try, the error is swallowed somewhere in the call stack? 

The check performed in `insertGuaranteeInfo` seems quite far away from this panic, which makes me a bit uneasy about using that check as justification for the panic.

That said, I think this approach is fine. We don't really expect any genuine errors encoding guarantee metadata, so panicking actually seems like the right thing to do in _any case_, because it probably indicates a programming error / bug.

geoknee (approved):

protocols/virtualfund/serde.go
kerzhner: Good catch. Addressed in 776bd37

protocols/virtualfund/virtualfund.go
kerzhner: Updated in 81cf404

protocols/virtualfund/virtualfund.go
kerzhner: I agree with your thoughts. One interesting question is why metadata encoding is here at all. `getExpectedGuarantees` is only used to check whether a channel affords the guarantees. It seems that encoding is not necessary for that operation. However, the `Affords` api accepts an `Allocation` with encoded metadata: https://github.com/statechannels/go-nitro/blob/81cf404220048c83e6fe3e1b153cc2130340ae05/channel/state/outcome/exit.go#L177-L179

My larger takeaways are the following:
1. If a low-level operation errors, error handling becomes cumbersome. Every function that uses the operation must now return an error.
2. One strategy is to ensure low-level operations do not error.

In this particular case, encoding of metadata can be isolated to logic that need to encode metadata. Encoding must take place for data authentication (signatures on state) and submission to chain.

Issue #328, opened 2022-03-11
Establish a struct comparison pattern


kerzhner:
The goal of this issue is to standardize struct comparison in the project. In other words, how should comparison be done when `==` cannot be used. Some of the options are: 
- [`DeepEqual` from `reflect`](https://pkg.go.dev/reflect#DeepEqual). Known caveats are:
  -  NaN float values can result in unexpected inequality.
  - circular references can lead to issues.
  - function fields can lead to issues.
- [go-cmp](https://github.com/google/go-cmp). The package can use only public fields to compare structs or all fields.
- Custom `Equal` methods. These are fragile. When a new field is added to a struct, existing unit tests will not necessarily flag that `Equal` method must be updated.

More notes on the discussion are [here](https://www.notion.so/statechannels/March-11-standup-2b4db98163f549c4b8b3bd6652ae30c7).

andrewgordstewart:
Meeting notes are here: https://www.notion.so/statechannels/March-11-standup-2b4db98163f549c4b8b3bd6652ae30c7

geoknee:
One option I like is to have custom `Equal` methods, and ensure they are unit tested by using `go-cmp`, which  is intended to be

>  a more powerful and safer alternative to reflect.DeepEqual for comparing whether two values are semantically equal. 

 but also

> it's propensity towards panicking means that it's unsuitable for production environments where a spurious panic may be fatal.

andrewgordstewart:
> > it's propensity towards panicking means that it's unsuitable for production environments where a spurious panic may be fatal.

Is this a quote from somewhere? Can you link to the source?

geoknee:
Yes, https://pkg.go.dev/github.com/google/go-cmp/cmp

kerzhner:
Is there a strong argument against using go-cmp instead of custom `Equal` methods?

andrewgordstewart:
> I think the fact that the `go-cmp` documentation warns against use in production environments is quite a strong argument?

This is a strong enough argument to me :)

andrewgordstewart:
Some recap from today's PL sync:

- George said that many of the `Equal` implementations were originally introduced for test purposes at the beginning of the repo

- Hannah pointed out two things:
1. in tests, `reflect.DeepEqual` should be totally fine. [aside: `go-cmp` is also fine (that's its intended purpose!) but shouldn't be used in production code]
[...]

PR #383, opened 2022-03-22
remove most Equal implementations

andrewgordstewart:
Most `Equal` implementations are used only in tests. These `Equal` implementations are removed. Other `Equal` implementations are used in application logic, and should be removed in a different PR.

Previously `cmp.Diff` depended on `Equal` being implemented. Now, `cmp.Diff` receives a `AllowUnexported` option, allowing `cmp.Diff` to reflect on specified, unexported options.

### DIFF
--- a/channel/channel_test.go
+++ b/channel/channel_test.go
@@ -3,6 +3,7 @@ package channel
 import (
 	"errors"
 	"math/big"
+	"reflect"
 	"testing"
 
 	"github.com/ethereum/go-ethereum/common"
@@ -34,7 +35,7 @@ func TestChannel(t *testing.T) {
 		}
 
 		r.latestSupportedStateTurnNum++
-		if r.Equal(*c) {
+		if reflect.DeepEqual(r, *c) {
 			t.Error("Clone: modifying the clone should not modify the original")
 		}
 
@@ -361,7 +362,7 @@ func TestTwoPartyLedger(t *testing.T) {
 		}
 
 		r.latestSupportedStateTurnNum++
-		if r.Channel.Equal(c.Channel) {
+		if reflect.DeepEqual(r.Channel, c.Channel) {
 			t.Error("Clone: modifying the clone should not modify the original")
 		}
 
@@ -396,7 +397,7 @@ func TestSingleHopVirtualChannel(t *testing.T) {
 		}
 
 		r.latestSupportedStateTurnNum++
-		if r.Channel.Equal(c.Channel) {
+		if reflect.DeepEqual(r.Channel, c.Channel) {
 			t.Error("Clone: modifying the clone should not modify the original")
 		}
 
--- a/channel/channel_test.go
+++ b/channel/channel_test.go
@@ -13,6 +13,14 @@ import (
 )
 
 func TestChannel(t *testing.T) {
+	compareChannels := func(a, b *Channel) string {
+		return cmp.Diff(*a, *b, cmp.AllowUnexported(*a, big.Int{}, state.SignedState{}))
+	}
+
+	compareStates := func(a, b state.SignedState) string {
+		return cmp.Diff(a, b, cmp.AllowUnexported(a, big.Int{}))
+	}
+
 	s := state.TestState.Clone()
 
 	_, err1 := New(s, 0)
@@ -30,8 +38,9 @@ func TestChannel(t *testing.T) {
 
 	testClone := func(t *testing.T) {
 		r := c.Clone()
-		if diff := cmp.Diff(*r, *c, cmp.Comparer(types.Equal)); diff != "" {
-			t.Fatalf("Clone: mismatch (-want +got):\n%s", diff)
+
+		if diff := compareChannels(c, r); diff != "" {
+			t.Errorf("Clone: mismatch (-want +got):\n%s", diff)
 		}
 
 		r.latestSupportedStateTurnNum++
@@ -287,8 +296,8 @@ func TestChannel(t *testing.T) {
 		if err != nil {
 			t.Error(err)
 		}
-		if diff := cmp.Diff(expectedSignedState, latestSignedState, cmp.Comparer(types.Equal)); diff != "" {
-			t.Fatalf("LatestSignedState: mismatch (-want +got):\n%s", diff)
+		if diff := cmp.Diff(expectedSignedState, latestSignedState, cmp.AllowUnexported(expectedSignedState)); diff != "" {
+			t.Errorf("LatestSignedState: mismatch (-want +got):\n%s", diff)
 		}
 
 		got2 := c.SignedStateForTurnNum[1]
@@ -325,8 +334,9 @@ func TestChannel(t *testing.T) {
 		if err != nil {
 			t.Error(err)
 		}
-		if diff := cmp.Diff(latestSignedState, expectedSignedState, cmp.Comparer(types.Equal)); diff != "" {
-			t.Fatalf("LatestSignedState: mismatch (-want +got):\n%s", diff)
+
+		if diff := compareStates(latestSignedState, expectedSignedState); diff != "" {
+			t.Errorf("LatestSignedState: mismatch (-want +got):\n%s", diff)
 		}
 
 	}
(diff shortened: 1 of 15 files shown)


andrewgordstewart:
This does a lot to address #328, but doesn't fully address it.

channel/channel_test.go
```diff
@@ -29,12 +31,12 @@ func TestChannel(t *testing.T) {
 
 	testClone := func(t *testing.T) {
 		r := c.Clone()
-		if diff := cmp.Diff(*r, *c, cmp.Comparer(types.Equal)); diff != "" {
```
geoknee, line 32: I ran a manual mutation test on `Channel.Clone()` -- I broke it on purpose 241e081695590d468d59afa02efbd97bcd95e92a. Strangely, the code on master successfully detected the problem, whereas `deep.Equal` doesn't seem to?

On this branch / the commit above, if I run `go  test ./channel` 
```
ok  	github.com/statechannels/go-nitro/channel	(cached)
```

```
--- FAIL: TestChannel (0.00s)
    --- FAIL: TestChannel/TestClone (0.00s)
        channel_test.go:33: Clone: mismatch (-want +got):
              channel.Channel(
            - 	{
            - 		Id:             s"0x252e8d12b641d8dd3cde74d299047963c99ce52bded2fbd38afa8e07b76a9306",
            - 		OnChainFunding: s"{[0x0000000000000000000000000000000000000000,0]}",
            - 		FixedPart: state.FixedPart{
            - 			ChainId:           s"9001",
            - 			Participants:      []common.Address{s"0xF5A1BB5607C9D079E46d1B3Dc33f25"..., s"0x760bf27cd45036a6C486802D30B5D9"...},
            - 			ChannelNonce:      s"37140676580",
            - 			AppDefinition:     s"0x5e29E5Ab8EF33F050c7cc10B5a0456D975C5F88d",
            - 			ChallengeDuration: s"60",
            - 		},
            - 		latestSupportedStateTurnNum: 999999999,
            - 		SignedStateForTurnNum: map[uint64]state.SignedState{
            - 			0: {state: state.State{...}, sigs: map[uint]crypto.Signature{}},
            - 			1: {state: state.State{...}, sigs: map[uint]crypto.Signature{}},
            - 		},
            - 	},
            + 	{
            + 		Id:             s"0x252e8d12b641d8dd3cde74d299047963c99ce52bded2fbd38afa8e07b76a9306",
            + 		OnChainFunding: s"{[0x0000000000000000000000000000000000000000,0]}",
            + 		FixedPart: state.FixedPart{
            + 			ChainId:           s"9001",
            + 			Participants:      []common.Address{s"0xF5A1BB5607C9D079E46d1B3Dc33f25"..., s"0x760bf27cd45036a6C486802D30B5D9"...},
            + 			ChannelNonce:      s"37140676580",
            + 			AppDefinition:     s"0x5e29E5Ab8EF33F050c7cc10B5a0456D975C5F88d",
            + 			ChallengeDuration: s"60",
            + 		},
            + 		latestSupportedStateTurnNum: 18446744073709551615,
            + 		SignedStateForTurnNum: map[uint64]state.SignedState{
            + 			0: {state: state.State{...}, sigs: map[uint]crypto.Signature{}},
            + 			1: {state: state.State{...}, sigs: map[uint]crypto.Signature{}},
            + 		},
            + 	},
              )

```

channel/state/signedstate_test.go
```diff
@@ -2,10 +2,11 @@ package state
...
 	"encoding/json"
+	"reflect"
 	"testing"
 
 	"github.com/ethereum/go-ethereum/common"
-	"github.com/google/go-cmp/cmp"
+	"github.com/go-test/deep"
 )
 
 func TestSignedStateEqual(t *testing.T) {
```
geoknee, line 12: I don't think we need this test at all, if the function it is testing is being removed.

geoknee (dismissed):
It looks like `go-test/deep` has bugs where `go-cmp` does not. I think the thrust of this PR is good, but swapping out `go-cmp` for `go-test/deep` is not necessary to achieve the desired aim. Do you think we could revert to `go-cmp`? Happy to approve once that is done (or the issue is otherwise resolved).

Also, can you comment on why you didn't remove _all_ Equal implementations?

andrewgordstewart:
> It looks like `go-test/deep` has bugs where `go-cmp` does not. I think the thrust of this PR is good, but swapping out `go-cmp` for `go-test/deep` is not necessary to achieve the desired aim. Do you think we could revert to `go-cmp`? Happy to approve once that is done (or the issue is otherwise resolved).

Good catch!

We cannot easily revert to `go-cmp`. `go-cmp` either:
1. panics when it hits an unexported field
2. requires us to pass in `AllowUnexported`
3. relies on `Equal` being implemented for the structs passed in

I did (2) in 751640b2180d864c6ac05917a32f9b41d15cc15c. It is a little tedious, but seems fine to me. It would be helpful to collect comparer functions in a test-utils package (CC @NiloCK)

> Also, can you comment on why you didn't remove _all_ Equal implementations?

I updated the PR description to say "Most Equal implementations are used only in tests. These Equal implementations are removed. Other Equal implementations are used in application logic, and should be removed in a different PR."

NiloCK:
> I did (2) in [751640b](https://github.com/statechannels/go-nitro/commit/751640b2180d864c6ac05917a32f9b41d15cc15c). It is a little tedious, but seems fine to me. It would be helpful to collect comparer functions in a test-utils package (CC @NiloCK)

Agreed. It's not super clear to me whether either of 

- separate `internal/testdata`, `internal/testutils`, `internal/testservices`* 
- one `test___` package containing all of the above

is **substantially** better than the other. I think I prefer the first, just for the sake of smaller, better labeled packages whose contents should be more obvious.



_*_ (`mockstore` and `mockchain` are test-only utilities, so should probably be quarantined off as consumable packages from go-nitro similarly to the testdata)

kerzhner:
It seems that perhaps we should reconsider using `reflect.DeepEqual`. We can also choose to adopt a decision tree like the following for comparisons in tests:
1. Use `reflect.DeepEqual`.
2. If floats, circular references, or function fields are involved, use `go-cmp`.

I don't have strong feeling around this, but it did take a while to wrap my head around [AllowUnexported](https://pkg.go.dev/github.com/google/go-cmp/cmp#AllowUnexported)

geoknee:
I think one major advantage of `go-cmp` is that it reports a diff, not just a bool, when making comparisons. That seems very useful for debugging purposes. I do agree that having to manage `AllowUnexported` is a bit of a chore -- and actually seems to push us back (slightly) towards the `.Equal` methods that this PR is removing (in the sense that we need to specify some struct-specific information to the comparer function on a case by case basis). 

I also do not have massively strong feelings either way, so I'm going to approve this so @andrewgordstewart can decide what to do. I imagine refactoring everything back to `reflect.DeepEqual` will be slightly painful, since we will have to rip out the error handlers that log the diff.

geoknee (approved):

andrewgordstewart:
> I think one major advantage of go-cmp is that it reports a diff, not just a bool, when making comparisons. That seems very useful for debugging purposes.

An alternative technique is, given a failing comparison, compare the fields one by one in the failing test case. This is tedious, but should work reliably.

> I do agree that having to manage AllowUnexported is a bit of a chore -- and actually seems to push us back (slightly) towards the .Equal methods that this PR is removing (in the sense that we need to specify some struct-specific information to the comparer function on a case by case basis).

`AllowUnexported` seems much less error-prone to me than using `Equal`:
- If we allow an unexported struct that doesn't exist, that amounts to an extra argument that does not harm.
- If we update a struct and forget to add an `AllowUnexported` option, `cmp.Diff` panics and we learn about it while running tests. This is nicer than having depending on `Equal`, where we can update a struct and forget to update `Equal`, leading to false positives.
- Another (very) big winner is "At least that code can live with our tests and not our production code."

> I imagine refactoring everything back to reflect.DeepEqual will be slightly painful, since we will have to rip out the error handlers that log the diff.

I don't think the refactoring is very difficult.

Nevertheless, I believe that using `AllowUnexported` is semantically equivalent to using `reflect.DeepEqual` -- both have the (stated) negative that our comparison can depend on implementation details. We happen to be ok with that, since our `Equal` implementations are looking for equality by explicitly comparing private fields.

channel/channel_test.go
lalexgap: I still think it makes sense to go with `go-cmp` as it seems like the more mature library.

PR #385, opened 2022-03-23
fix waitTimeForCompletedObjectiveIds

lalexgap:
A small [change](https://github.com/statechannels/go-nitro/pull/379#discussion_r833569672) in my PR accidentally caused  `waitTimeForCompletedObjectiveIds` to always return an empty slice of objective ids. Unfortunately I didn't catch this before merging in the PR!

The problem was that we were relying on iterating through the keys of `completed` to construct the objective id slice. By removing the for loop that was setting `completed[i]=false`  `completed` no longer contains keys for all the objective ids.

 This PR fixes this by iterating through the `ids` instead of the `completed` map.

### DIFF
--- a/client_test/helpers_test.go
+++ b/client_test/helpers_test.go
@@ -54,7 +54,8 @@ func waitTimeForCompletedObjectiveIds(t *testing.T, client *client.Client, timeo
 	select {
 	case <-time.After(timeout):
 		incompleteIds := make([]protocols.ObjectiveId, 0)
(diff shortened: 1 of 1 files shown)


geoknee (approved):
Apologies for making the suggestion that broke this!

lalexgap:
> Apologies for making the suggestion that broke this!

I think your suggestion was valid, I just didn't implement it right🙂

PR #402, opened 2022-03-25
Add error handling to `engine.go`

### LINKED ISSUE #353: Errors should not be ignored in the engine
This is mostly due to my personal preference 😛  I much prefer panicking over ignoring an error, so it's easy to figure out test failures and debug into them.

Right now we don't really have an error handling framework so we often ignore errors in the engine. This can make debugging/tracing pretty confusing (which motivated this [PR](https://github.com/statechannels/go-nitro/pull/346)) as the errors get swallowed and things don't fail until later.

Perhaps a better fix than panicking in individual engine functions is to just bubble up errors to the root `Run` loop. The `Run` loop can `panic` on error for now (to aid in debugging) and then later can be replaced with something more sophisticated.

_Originally posted by @lalexgap in https://github.com/statechannels/go-nitro/pull/341#discussion_r828412682_

Here's a list of known cases where errors need to be handled:
- [x] https://github.com/statechannels/go-nitro/issues/355

### LINKED ISSUE #355: handleAPIEvent should properly handle errors
`handleAPIEvent` [currently](https://github.com/statechannels/go-nitro/blob/6ed2c14afc3c315c4ee74c57d4779be5bee8a7ac/client/engine/engine.go#L166-L187) prints an error statement when it detects an error. This leads to silently failing software.

Instead, `handleAPIEvent` should either handle the error locally, or return an error.

geoknee:
Closes #353.
Closes #355.

Error handling in the engine is next-to non-existent. This PR ensures no errors are swallowed. Instead, they are all passed back up to the engine run loop where they trigger a panic. There is one exception: the chain event handler continues to swallow store read errors when trying to lookup the objective for a chain event. This is because the current mock chain and chain service is quite dumb -- *all* information is broadcast to *all* clients. I have added comments to make this clearer.

Triggering a panic is not the long term solution, since this is irresponsible for library code to do. It has the great advantage, however, of surfacing errors quickly so we can know about them when developing / testing / debugging. 

An advantage of the approach here is that there is a single place where (most) errors will end up. We should make a new issue for having those errors dealt with properly, which should include a mechanism to communicate the errors to the consuming application (see #388 for a sketch).  

---

#### Code quality

#### Project management

### DIFF
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -111,7 +111,6 @@ func (e *Engine) Run() {
 			panic(err)
 			// TODO do not panic if in production.
 			// TODO report errors back to the consuming application
-
 		}
 
 	}
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -86,16 +86,17 @@ func (e *Engine) ToApi() <-chan ObjectiveChangeEvent {
 func (e *Engine) Run() {
 	for {
 		var res ObjectiveChangeEvent
+		var err error
 		select {
 		case apiEvent := <-e.FromAPI:
-			res = e.handleAPIEvent(apiEvent)
+			res, err = e.handleAPIEvent(apiEvent)
 
 		case chainEvent := <-e.fromChain:
 
-			res = e.handleChainEvent(chainEvent)
+			res, err = e.handleChainEvent(chainEvent)
 
 		case message := <-e.fromMsg:
-			res = e.handleMessage(message)
+			res, err = e.handleMessage(message)
 
 		}
 		// Only send out an event if there are changes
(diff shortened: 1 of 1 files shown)


geoknee:
See https://github.com/statechannels/go-nitro/discussions/306

client/engine/engine.go
```diff
@@ -86,25 +86,34 @@ func (e *Engine) ToApi() <-chan ObjectiveChangeEvent {
...
+			res, err = e.handleChainEvent(chainEvent)
 
 		case message := <-e.fromMsg:
-			res = e.handleMessage(message)
+			res, err = e.handleMessage(message)
+		}
 
+		// Handle errors
+		if err != nil {
+			panic(err)
```
lalexgap, line 104: We could use `e.logger.Panic()` here so the error also get's written out to the log.

lalexgap (approved):

lalexgap:
I think this change will make debugging a much friendlier experience!

client/engine/engine.go
geoknee: [d72d90c](https://github.com/statechannels/go-nitro/pull/402/commits/d72d90cf0617e3f06b6c9801013ab08f997736c1)

PR #406, opened 2022-03-25
Vars.Add explicitly mutates its receiver

### LINKED ISSUE #400: Vars.Add uses a value receiver but modifies the guarantees map
Currently Vars.Add uses a [value receiver](https://github.com/statechannels/go-nitro/blob/ag/receive/channel/consensus_channel/consensus_channel.go#L186:L186), which mean the receiver is a shallow clone. However `Add` calls `DivertToGuarantee` which modifies the `guarantees` map on the `Outcome`. Since maps are reference type the shallow copy still points to the same map as the original.

This can be dangerous as you might not be aware you're modifying the underlying map.

```	currentVars := Vars{TurnNum: 5, Outcome: currentOutcome}
	current := SignedVars{Vars: currentVars, Signatures: [2]*state.Signature{nil, nil}}
	proposal := add(current.TurnNum+1, 5, channelId, alice.Destination(), bob.Destination())
	fmt.Printf("current BEFORE add is called: turn num %d, guarantees length: %d\n", currentVars.TurnNum, len(currentVars.Outcome.guarantees))
	_, _ = currentVars.Add(proposal)
	// Current still has the previous turn number but now it includes a guarantee
	fmt.Printf("current AFTER add is called: turn num %d, guarantees length: %d\n", currentVars.TurnNum, len(currentVars.Outcome.guarantees))
```
Outputs:
```
current BEFORE add is called: turn num 5, guarantees length: 0
current AFTER add is called: turn num 5, guarantees length: 1
```

andrewgordstewart:
Fixes #400 by:
- making `Vars.Add` a pointer receiver that mutates its value
- implementing `latestProposedVars` which uses a new `clone` method to clone the current vars

I decided to get rid of `DivertToGuarantee` (note: it is only used in one place) since this collects the logic for incorporating an `Add` proposal into `Vars`.

`Vars.Add` mutating its receiver may deviate from some other patterns found in `go-nitro`, but this solution seems to be the simplest solution: it avoids having to implement multiple `clone` methods, and avoids creating many unused copies of almost-the-same-data as the ProposalQueue is processed.

### DIFF
--- a/channel/consensus_channel/consensus_channel.go
+++ b/channel/consensus_channel/consensus_channel.go
@@ -143,32 +143,6 @@ var ErrInvalidDeposit = fmt.Errorf("unable to divert to guarantee: invalid depos
 var ErrInsufficientFunds = fmt.Errorf("unable to divert to guarantee: insufficient funds")
 var ErrDuplicateGuarantee = fmt.Errorf("duplicate guarantee detected")
 
-// DivertToGuarantee deducts g.amount from o.left's balance, and
-// adds g to o.guarantees
-func (o LedgerOutcome) DivertToGuarantee(p Add) (LedgerOutcome, error) {
-	g := p.Guarantee
-
-	if types.Gt(&p.LeftDeposit, &g.amount) {
-		return LedgerOutcome{}, ErrInvalidDeposit
-	}
-
-	if types.Gt(&g.amount, &o.left.amount) {
-		return LedgerOutcome{}, ErrInsufficientFunds
-	}
-
-	o.left.amount.Sub(&o.left.amount, &p.LeftDeposit)
-	rightDeposit := p.RightDeposit()
-	o.right.amount.Sub(&o.right.amount, &rightDeposit)
-
-	_, found := o.guarantees[g.target]
-	if found {
-		return LedgerOutcome{}, ErrDuplicateGuarantee
-	}
-	o.guarantees[g.target] = g
-
-	return o, nil
-}
-
 // Vars stores the turn number and outcome for a state in a consensus channel
 type Vars struct {
 	TurnNum uint64
@@ -218,22 +192,42 @@ func (a Add) RightDeposit() big.Int {
 
 var ErrIncorrectTurnNum = fmt.Errorf("incorrect turn number")
 
-// Add updates Vars by including a guarantee, updating balances accordingly
-func (vars Vars) Add(p Add) (Vars, error) {
+// Add mutates Vars by
+// - increasing the turn number by 1
+// - including the guarantee
+// - adjusting balances accordingly
+func (vars *Vars) Add(p Add) error {
 	if p.turnNum != vars.TurnNum+1 {
-		return Vars{}, ErrIncorrectTurnNum
+		return ErrIncorrectTurnNum
 	}
 
+	// Increase the turn number
 	vars.TurnNum += 1
 
-	o, err := vars.Outcome.DivertToGuarantee(p)
+	o := vars.Outcome
+	g := p.Guarantee
 
-	if err != nil {
-		return Vars{}, err
+	// Include the guarantee
+	_, found := o.guarantees[g.target]
+	if found {
+		return ErrDuplicateGuarantee
+	}
+	o.guarantees[g.target] = g
+
+	// Adjust balances
+	if types.Gt(&p.LeftDeposit, &g.amount) {
+		return ErrInvalidDeposit
 	}
 
-	vars.Outcome = o
-	return vars, nil
+	if types.Gt(&g.amount, &o.left.amount) {
+		return ErrInsufficientFunds
+	}
+
+	o.left.amount.Sub(&o.left.amount, &p.LeftDeposit)
+	rightDeposit := p.RightDeposit()
+	o.right.amount.Sub(&o.right.amount, &rightDeposit)
+
+	return nil
 }
 
 // Propose receives a proposal to add a guarantee, and generates and stores a SignedProposal in
@@ -257,7 +251,7 @@ func (c *ConsensusChannel) Propose(add Add, sk []byte) (SignedProposal, error) {
 
 	add.turnNum = latestProposal.turnNum + 1
 
-	vars, err = vars.Add(add)
+	err = vars.Add(add)
 	if err != nil {
 		return SignedProposal{}, fmt.Errorf("propose could not add new state vars: %w", err)
 	}
@@ -280,7 +274,7 @@ func (c *ConsensusChannel) latestProposedVars() (Vars, error) {
 
 	var err error
 	for _, p := range c.proposalQueue {
-		vars, err = vars.Add(p.Proposal.(Add))
+		err = vars.Add(p.Proposal.(Add))
 		if err != nil {
 			return Vars{}, err
 		}
--- a/channel/consensus_channel/consensus_channel.go
+++ b/channel/consensus_channel/consensus_channel.go
@@ -175,6 +175,20 @@ type Vars struct {
 	Outcome LedgerOutcome
 }
 
+// clone returns a deep clone of v
+func (v Vars) clone() Vars {
+	v.Outcome.left.amount = *new(big.Int).Set(&v.Outcome.left.amount)
+	v.Outcome.right.amount = *new(big.Int).Set(&v.Outcome.right.amount)
+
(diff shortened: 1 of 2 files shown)


channel/consensus_channel/consensus_channel.go
```diff
@@ -143,38 +143,26 @@ var ErrInvalidDeposit = fmt.Errorf("unable to divert to guarantee: invalid depos
...
-}
-
 // Vars stores the turn number and outcome for a state in a consensus channel
 type Vars struct {
 	TurnNum uint64
 	Outcome LedgerOutcome
 }
 
+// clone returns a deep clone of v
+func (v Vars) clone() Vars {
```
kerzhner, line 153: From [Code Review Comments](https://github.com/golang/go/wiki/CodeReviewComments#receiver-type):
"Don't mix receiver types. Choose either pointers or struct types for all available methods."

Later in this file, we have https://github.com/statechannels/go-nitro/blob/c3e14a8775b04d5bd54cf02d52537e1707f9937b/channel/consensus_channel/consensus_channel.go#L199

channel/consensus_channel/consensus_channel.go
```diff
@@ -143,38 +143,26 @@ var ErrInvalidDeposit = fmt.Errorf("unable to divert to guarantee: invalid depos
...
+	v.Outcome.left.amount = *new(big.Int).Set(&v.Outcome.left.amount)
+	v.Outcome.right.amount = *new(big.Int).Set(&v.Outcome.right.amount)
+
+	guarantees := make(map[types.Destination]Guarantee)
+	for d, g := range v.Outcome.guarantees {
+		guarantees[d] = g
```
kerzhner, line 159: `Guarantee` struct has an `amount` field that is a `big.Int`. It seems that this `big.Int` is treated differently than the `Outcome.left.amount` and `Outcome.right.amount` at the beginning of the function. On this line, we use assignment. For outcome amounts, we allocate a new `big.Int` struct.

kerzhner: It seems that the first 2 lines of this function might be unnecessary?

channel/consensus_channel/consensus_channel.go
```diff
@@ -204,22 +192,42 @@ func (a Add) RightDeposit() big.Int {
...
-		return Vars{}, ErrIncorrectTurnNum
+		return ErrIncorrectTurnNum
 	}
 
+	// Increase the turn number
 	vars.TurnNum += 1
 
-	o, err := vars.Outcome.DivertToGuarantee(p)
+	o := vars.Outcome
+	g := p.Guarantee
```
kerzhner, line 208: This doesn't seem necessary. Fields of the embedded struct can be accessed directly on `p` like `p.target`.

channel/consensus_channel/consensus_channel.go
```diff
@@ -204,22 +192,42 @@ func (a Add) RightDeposit() big.Int {
...
 
-	if err != nil {
-		return Vars{}, err
+	// Include the guarantee
+	_, found := o.guarantees[g.target]
+	if found {
+		return ErrDuplicateGuarantee
 	}
+	o.guarantees[g.target] = g
```
kerzhner, line 215: It might be better to delay setting this key until all of the error checking is completed. Otherwise, to recover from an error, the value for the key will need to be removed.

channel/consensus_channel/consensus_channel.go
```diff
@@ -264,6 +267,22 @@ func (c *ConsensusChannel) Propose(add Add, sk []byte) (SignedProposal, error) {
...
 }
 
+// latestProposedVars returns the latest proposed vars in a consensus channel
+// by cloning its current vars and applying each proposal in the queue
+func (c *ConsensusChannel) latestProposedVars() (Vars, error) {
+	vars := c.current.Vars.clone()
+
+	var err error
+	for _, p := range c.proposalQueue {
+		err = vars.Add(p.Proposal.(Add))
```
kerzhner, line 277: If `p.Proposal` is not of type `Add`, is panic the desired behavior?

channel/consensus_channel/consensus_channel.go
```diff
@@ -143,38 +143,26 @@ var ErrInvalidDeposit = fmt.Errorf("unable to divert to guarantee: invalid depos
...
-
 // Vars stores the turn number and outcome for a state in a consensus channel
 type Vars struct {
 	TurnNum uint64
 	Outcome LedgerOutcome
 }
 
+// clone returns a deep clone of v
+func (v Vars) clone() Vars {
+	v.Outcome.left.amount = *new(big.Int).Set(&v.Outcome.left.amount)
```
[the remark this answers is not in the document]
andrewgordstewart: I was actually following the implementation of [this](https://github.com/statechannels/go-nitro/blob/c0fef878771ec094697d6af00ccd5f0f871d5a6e/channel/state/state.go#L180-L188), which uses `new(big.Int)`.

Per [the docs](https://go.dev/doc/effective_go#allocation_new),
>The expressions new(File) and &File{} are equivalent
so your suggestion is equivalent, so I think it comes down to preference about readability.

The codebase currently leans towards
```suggestion
	v.Outcome.left.amount = *big.NewInt(0).Set(&v.Outcome.left.amount)
```
which is equivalent to `new(big.Int).SetInt64(0).set(&v.Outcome.left.amount)`.

Looking at
```
x = *new(big.Int).set(100)
x = *big.newInt(0).set(100)
x = *(&big.Int{}).Set(100)
```

I feel `new(big.Int)` is the clearest, but it doesn't really matter much at all. What do you think?

channel/consensus_channel/consensus_channel_test.go
```diff
@@ -67,16 +67,53 @@ func TestConsensusChannel(t *testing.T) {
...
+	if fingerprint(clone1) != f1 {
+		t.Fatal("vars incorrectly cloned: ", f1, fingerprint(clone1))
+	}
+
+	clone1.Outcome.guarantees[targetChannel] = guarantee(999, bob, bob, bob)
+	if f1 != fingerprint(vars) {
+		t.Fatal("vars shares data with clone")
+	}
+
+	clone2 := vars.clone()
```
andrewgordstewart, line 94: This test is ineffective -- see https://github.com/statechannels/go-nitro/pull/411/commits/4612ae86e3935197c15f500a708dab0d42a612a1

channel/consensus_channel/consensus_channel.go
andrewgordstewart: You're definitely right, and they should not be treated differently.

> It seems that the first 2 lines of this function might be unnecessary?

I learned that they are actually necessary! See https://github.com/statechannels/go-nitro/pull/411/commits/4612ae86e3935197c15f500a708dab0d42a612a1

This is fixed in cc57d4e3c6a793ab5a045b5596bae908a613ce61

channel/consensus_channel/consensus_channel.go
andrewgordstewart: 7f43928c9d44a0634248726f6857b4b3ff5a9867 naturally addresses this

channel/consensus_channel/consensus_channel.go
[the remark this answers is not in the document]
andrewgordstewart: FYI I use the existing convention here: c3fcd40085705d2cd44e9d7aa081dade8b189494

channel/consensus_channel/consensus_channel.go
andrewgordstewart: a44421f4f76e551b0192e4d82d6e806256c23347

channel/consensus_channel/consensus_channel.go
andrewgordstewart: b56cf79ed6a8919f46a4205fff1b99d193a250dd

channel/consensus_channel/consensus_channel.go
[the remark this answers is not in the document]
kerzhner: No strong preference. I was just initially thrown off by the `new` keyword.

kerzhner (approved):

channel/consensus_channel/consensus_channel.go
kerzhner: Good to know that a `big.Int` struct contains a slice!

PR #414, opened 2022-03-28
Address feedback on UpdateConsensus

andrewgordstewart:
This addresses the feedback on #410, which was merged quickly to let other folks branch off of it.

### DIFF
--- a/channel/consensus_channel/consensus_channel.go
+++ b/channel/consensus_channel/consensus_channel.go
@@ -114,7 +114,7 @@ type LedgerOutcome struct {
 	guarantees   map[types.Destination]Guarantee
 }
 
-// includes returns true when o includes g in its list of guarantees.
+// includes returns true when the receiver includes g in its list of guarantees.
 func (o *LedgerOutcome) includes(g Guarantee) bool {
 	existing, found := o.guarantees[g.target]
 	if !found {
--- a/channel/consensus_channel/consensus_channel.go
+++ b/channel/consensus_channel/consensus_channel.go
@@ -109,6 +109,7 @@ type LedgerOutcome struct {
 	guarantees   map[types.Destination]Guarantee
 }
 
+// includes returns true when o includes g in its list of guarantees.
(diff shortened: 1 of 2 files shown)


channel/consensus_channel/consensus_channel.go
```diff
@@ -109,6 +114,7 @@ type LedgerOutcome struct {
 	guarantees   map[types.Destination]Guarantee
 }
 
+// includes returns true when o includes g in its list of guarantees.
```
geoknee, line 117: Nit: I don't believe `o` makes it into the generated documentation / hover modal in VSCode. So I prefer referencing "the receiver" instead of using the named variable.
Screenshot 2022-03-28 at 20 02 56

channel/consensus_channel/leader_channel.go
```diff
@@ -70,26 +65,31 @@ func (c *LeaderChannel) Propose(add Add, sk []byte) (SignedProposal, error) {
...
-// If their proposal is stale (ie. theirP.TurnNum <= c.current.TurnNum) then
-// their proposal is ignored
-func (c *LeaderChannel) UpdateConsensus(theirP SignedProposal) error {
-	vars := Vars{
+// If the countersupplied is stale (ie. proposal.TurnNum <= c.current.TurnNum) then
+// their proposal is ignored.
+//
+// An error is returned if:
+// - the countersupplied proposal is not found
+// - or if it is found but not correctly by the Follower
```
geoknee, line 78: Another Nit: Bullet points like this don't really work :-/ They end up getting in-lined in the documentation https://stackoverflow.com/questions/52744468/godoc-documentation-not-outputting-lists

geoknee (approved):

channel/consensus_channel/consensus_channel.go
andrewgordstewart: 5c3c1dbac1106b009eeec84d29194d67db66b57f

PR #419, opened 2022-03-28
WIP: Add IsShallowCopy to detect when clones are shallow copies

lalexgap:
This is a first stab at using reflection to ensure that cloning does not generate a shallow copy in any way. It seems to work for the basic use case of a shallow slice, but I haven't tested much further than that.

This is definitely a viable approach but I think it requires a bit more time to think through and write tests for. I'm going to pause work on this for now.

### DIFF
--- a/channel/channel_test.go
+++ b/channel/channel_test.go
@@ -8,6 +8,7 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/google/go-cmp/cmp"
 	"github.com/statechannels/go-nitro/channel/state"
+	"github.com/statechannels/go-nitro/helpers"
 	"github.com/statechannels/go-nitro/types"
 )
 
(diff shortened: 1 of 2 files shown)


lalexgap:
As I mentioned in the [discussion](https://github.com/statechannels/go-nitro/discussions/413) it probably makes sense to not use this work and just use serde instead. I'll leave this up for now until we're fully decided.

geoknee:
I may be nitpicking, but I think it is a little confusing to name this as a shallow clone detector. That's because I think sometimes a deep clone is the same thing as a shallow clone (depending on the type). We want deep clones, so it can't be right that a shallow clone is "wrong". Do you think there is a better name for what we are trying to detect with this function?

andrewgordstewart:
Would this be more clear?
- `HasShallowCopy` (sic) is renamed to `IsDeepCopy(a,b)`
- returns false if `reflect.DeepEqual(a, b)` is false
- returns `false` when `a` is not a deep copy of `b` (and therefore must be a shallow copy)

lalexgap:
Closing this as we don't plan on implementing this now.

PR #423, opened 2022-03-28
rename testdata.go -> exports_test.go

andrewgordstewart:
Defining (exported) global variables in testadata.go pollutes the package. Defining (imported) global variables in testdata.go does not, but these global variables could unintentially be used in client code.

Naming the file `exports_test.go` means that **non-test** code cannot use these global variables. It also means that any exported global variables from testdata.go do not make it into the package API.

(In this case, these variables don't appear to be used in client code, since this rename does not cause `go build ./...` to fail.)

As an example, in [this branch](https://github.com/statechannels/go-nitro/compare/ags/testdata-example?expand=1) I added an Exported variable:
Screen Shot 2022-03-28 at 3 38 47 PM

[After the rename](https://github.com/statechannels/go-nitro/commit/2020071a3234687fb8434df5c9729cd9ae7fc46f), this variable is not in the package's godoc:
Screen Shot 2022-03-28 at 3 39 32 PM


andrewgordstewart:
See https://pkg.go.dev/cmd/go@master#hdr-Compile_packages_and_dependencies

> When compiling packages, build ignores files that end in '_test.go'.

kerzhner (commented):
Why is the changed name `exports_test.go`? It doesn't look like any of the file contents are exported outside of the `outcome` package.

andrewgordstewart:
The file contents are _currently_ not exported outside of the `outcome` package. But it is defining global variables in the package's scope.

The fear here is that client code could use eg. the variable `e` defined [here](https://github.com/statechannels/go-nitro/blob/604dc555eba35a569ec0ba103062572ad85ad4d1/channel/state/outcome/exports_test.go#L40). Given some believable circumstances, a developer may not realize that `e` is not the variable they believe it to be. eg.

```
package outcome 

if someCond {
  e := 10 // this is now a shadowed variable, since there is a global variable `e`
}}

e // this is the global variable defined in `testdata.go`, of type `Exit`
```
Screen Shot 2022-03-28 at 4 42 40 PM

geoknee (approved):

Issue #391, opened 2022-03-24
Implement a `ConsensusChannel`


andrewgordstewart:
A `ConsensusChannel` managing the _running state_ of a two-party ledger channel that runs a "consensus app" (ie. full consensus is required). One of the participants is designated the leader, and the other is designated the follower. 

State updates are proposed by the leader and signed by the follower. This allows the leader to create a consistent set of concurrent updates without having to wait for a countersignature.

# Expected API
Roughly, 
```
type ConsensusChannel interface {
	Propose(update Update, pk []byte) (T, error)
	SignNextProposal(expectedUpdate Update, pk []byte) (T, error)
	Receive(p SignedProposal) (T, error)
	UpdateConsensusState(p SignedProposal) (T, error)
}
```

# Implementation

The `ConsensusChannel` stores the following state:
1. a `consensusState`, a single state signed by both parties
2. a `proposalQueue`, a FIFO queue of signed, sequential, proposed updates

Some code (mostly data structures) is sketched out [here](https://github.com/statechannels/go-nitro/pull/373/files).

- [x] Propose: https://github.com/statechannels/go-nitro/issues/392
- [x] Receive: https://github.com/statechannels/go-nitro/issues/393
- [x] SignNextProposal: https://github.com/statechannels/go-nitro/issues/394
- [x] UpdateConsensusState: https://github.com/statechannels/go-nitro/issues/395  

## Comment

Objectives receive events via `Update` and take actions via `Crank`. If objectives received events and took actions via `Crank` (like a state machine does) then the follower would not need to store a `proposalQueue`, and would not need to receive-then-accept proposals. It could instead call `L.Accept(p SignedProposal, pk []byte)`, which would need to:
1. construct the next state `s` by applying `p` to `consensusState`
2. check that the leader signed `s`. if not, return an error
3. replace `consensusState` with `s`

# Integration 
- [x] #412
- [x] #422
- [x] #420

PR #425, opened 2022-03-29
Disable Flickering Test

### LINKED ISSUE #366: Ledger Protocol can stall depending on the order in which states are received

lalexgap:
Due to the [ledger funding issue](https://github.com/statechannels/go-nitro/issues/366) the multi party integration test can fail.The multi party test creates multiple virtual channels using the same ledger channel, which means ledger proposals can arrive out of order and cause things to fail. Due to the small amount of virtual channels it happens rarely, but it still [occurs](https://github.com/statechannels/go-nitro/runs/5729776796?check_suite_focus=true).

This disables the test until we have completed the work in #391 with a new ledger protocol that doesn't suffer the problem.

### DIFF
--- a/client_test/virtualfund_multiparty_integration_test.go
+++ b/client_test/virtualfund_multiparty_integration_test.go
@@ -14,7 +14,9 @@ import (
 
 // TestMultiPartyVirtualFundIntegration tests the scenario where Alice creates virtual channels with Bob and Brian using Irene as the intermediary.
 func TestMultiPartyVirtualFundIntegration(t *testing.T) {
-
+	// TODO: This test can occasionally fail due to https://github.com/statechannels/go-nitro/issues/366
+	// It should be re-enabled when this issue is fixed.
+	t.Skip()
 	logFile := "virtualfund_multiparty_client_test.log"
 	truncateLog(logFile)
 


geoknee:
Fine -- I would just like it to be clear what we should do to test when we've finished upgrading the ledger protocol. Do I unskip a test, add a command line flag in the github action somewhere, or increase a parameter in a test file somewhere?

My preference would be for simply unskipping a test without needing any other changes?

geoknee (approved):

PR #426, opened 2022-03-29
Allow `ConsensusChannels` to be persisted to the `MockStore`

### LINKED ISSUE #412: The `store` should be able to store and return a ConsensusChannel

geoknee:
Closes #412 .

**Changes** 
* Adds json marshalling / unmarshalling to consensus channels. 
* Adds a getter and setter to the store interface
* Implements the expanded interface in the mock store (by following existing patterns for persisting channels). 
* Exports the following from `consensus_channel` package
    - `ConsensusChannel` itself (necessary to properly type the methods on the store
    - `Vars.asState` helper. Not strictly necessary, but useful for constructing test data
    - a new `Id` field
    - a `Participants()` getter
   
Exporting those symbols leads to a large-ish diff here. I recommend reviewing commit-by-commit to reduce the noise (I did each renaming change in its own commit). 

**Shortcomings**
I did ~little~ nothing to address the issue of storing `LeaderChannels` or `FollowerChannels` (which are sub types of `ConsensusChannel`). For now, to store such types calling code will need to extract a `ConsensusChannel` from them. And when retrieving, a type assertion will be necessary somewhere to enable the channels to be used as intended. 

**How Has This Been Tested?**

A new test has been written, mirroring some of the existing tests in the `mockstore` package. However, I believe the current tests to be a bit broken, and I generally had a miserable time testing for deep equality between a test `ConsensusChannel`, and the result of storing and retrieving it.  So an incomplete set of checks is used, for now. 

https://github.com/statechannels/go-nitro/issues/428

 

---

#### Code quality

#### Project management

### DIFF
--- a/channel/consensus_channel/consensus_channel.go
+++ b/channel/consensus_channel/consensus_channel.go
@@ -19,8 +19,8 @@ const (
 	follower ledgerIndex = 1
 )
 
-// consensusChannel is used to manage states in a running ledger channel
-type consensusChannel struct {
+// ConsensusChannel is used to manage states in a running ledger channel
+type ConsensusChannel struct {
 	// constants
 	myIndex ledgerIndex
 	fp      state.FixedPart
@@ -37,23 +37,23 @@ func newConsensusChannel(
 	initialTurnNum uint64,
 	outcome LedgerOutcome,
 	signatures [2]state.Signature,
-) (consensusChannel, error) {
+) (ConsensusChannel, error) {
 	vars := Vars{TurnNum: initialTurnNum, Outcome: outcome.clone()}
 
 	leaderAddr, err := vars.asState(fp).RecoverSigner(signatures[leader])
 	if err != nil {
-		return consensusChannel{}, fmt.Errorf("could not verify sig: %w", err)
+		return ConsensusChannel{}, fmt.Errorf("could not verify sig: %w", err)
 	}
 	if leaderAddr != fp.Participants[leader] {
-		return consensusChannel{}, fmt.Errorf("leader did not sign initial state: %v, %v", leaderAddr, fp.Participants[leader])
+		return ConsensusChannel{}, fmt.Errorf("leader did not sign initial state: %v, %v", leaderAddr, fp.Participants[leader])
 	}
 
 	followerAddr, err := vars.asState(fp).RecoverSigner(signatures[follower])
 	if err != nil {
-		return consensusChannel{}, fmt.Errorf("could not verify sig: %w", err)
+		return ConsensusChannel{}, fmt.Errorf("could not verify sig: %w", err)
 	}
 	if followerAddr != fp.Participants[follower] {
-		return consensusChannel{}, fmt.Errorf("leader did not sign initial state: %v, %v", followerAddr, fp.Participants[leader])
+		return ConsensusChannel{}, fmt.Errorf("leader did not sign initial state: %v, %v", followerAddr, fp.Participants[leader])
 	}
 
 	current := SignedVars{
@@ -61,7 +61,7 @@ func newConsensusChannel(
 		signatures,
 	}
 
-	return consensusChannel{
+	return ConsensusChannel{
 		fp:            fp,
 		myIndex:       myIndex,
 		proposalQueue: make([]SignedProposal, 0),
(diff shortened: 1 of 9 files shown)


client/engine/store/mockstore.go
```diff
@@ -184,6 +199,34 @@ func (ms *MockStore) GetTwoPartyLedger(firstParty types.Address, secondParty typ
 	return ledger, ok
 }
 
+// GetTwoPartyLedger returns a ledger channel between the two parties if it exists.
```
geoknee, line 202:
```suggestion
// GetConsensusChannel returns a ConsensusChannel between the two parties if it exists.
```

client/engine/store/mockstore.go
```diff
@@ -184,6 +199,34 @@ func (ms *MockStore) GetTwoPartyLedger(firstParty types.Address, secondParty typ
 	return ledger, ok
 }
 
+// GetConsensusChannel returns a ConsensusChannel between the two parties if it exists.
+func (ms *MockStore) GetConsensusChannel(firstParty types.Address, secondParty types.Address) (channel *consensus_channel.ConsensusChannel, ok bool) {
```
andrewgordstewart, line 203:
```suggestion
func (ms *MockStore) GetConsensusChannel(firstParty, secondParty types.Address) (channel *consensus_channel.ConsensusChannel, ok bool) {
```

client/engine/store/mockstore.go
```diff
@@ -184,6 +199,34 @@ func (ms *MockStore) GetTwoPartyLedger(firstParty types.Address, secondParty typ
...
+		var ch consensus_channel.ConsensusChannel
+		err := json.Unmarshal(chJSON, &ch)
+
+		if err != nil {
+			return true // channel not found, continue looking
+		}
+
+		participants := ch.Participants()
+		if len(participants) == 2 {
+			// TODO: Should order matter?
```
andrewgordstewart, line 216: Can we resolve this question before merging?

client/engine/store/mockstore.go
lalexgap: We could rename `firstParty` to `leader` and `secondParty` to `follower`.

client/engine/store/mockstore.go
lalexgap: If `GetConsensusChannel` accepts a leader and follower address then participant order does matter since the leader is always the first participant and the follower is the second, so we could remove this TODO.

client/engine/store/mockstore.go
andrewgordstewart: That is also a nice change!

channel/consensus_channel/consensus_channel.go
```diff
@@ -36,32 +39,39 @@ func newConsensusChannel(
 	myIndex ledgerIndex,
 	outcome LedgerOutcome,
 	signatures [2]state.Signature,
-) (consensusChannel, error) {
+) (ConsensusChannel, error) {
+
+	cId, err := fp.ChannelId()
```
lalexgap, line 44: Outside the scope of this PR but its annoying that calculating a channelId returns an error.

lalexgap (approved):

channel/consensus_channel/consensus_channel.go
andrewgordstewart: Is there a way around that?

lalexgap:
@geoknee I rebased this on `main` to resolve some downstream conflicts.

channel/consensus_channel/consensus_channel.go
lalexgap: We could just `panic` instead of returning an error 😉

client/engine/store/mockstore.go
geoknee: [a4568d5](https://github.com/statechannels/go-nitro/pull/426/commits/a4568d5c4aeece2aef34990afb414d95102706fe)

channel/consensus_channel/consensus_channel.go
geoknee: I actually think returning an error is the right thing to do here, because we need to validate the `FixedPart` and check for an (admittedly almost impossible) hash collision. 

The approach I took here is to do have those checks done when constructing the `consensusChannel`, and then (if successful) cache the channelId. Then, it can be looked up later on (for example inside `MarshalJSON`) without needing to handle an error. 

All of this machinery is there to prevent unexpected `panics`, especially when they could be triggered by the application providing "bad" data.

PR #427, opened 2022-03-29
Contracts SATP compatability

nksazonov:
**Contracts changes:**
- [x] appDefinition and challengeDuration are now used in channelId computations (which includes chainID, participants, channelNonce, appDefinition, challengeDuration)
- [x] State struct representation has changed: bytes appPart and bytes outcome instead of bytes32 appPartHash and bytes32 outcomeHash
- [x] in ForceMove State is encoded in the same way as in go-nitro - not as a struct, but as a set of State's fields. Also order of arguments has changed.
- [x] function signatures, EmbeddedApplication.sol were updated to support changes above
- [x] additional layer of decoding added to CountingApp to correspond to go-nitro
- [x] tests updated to support changes above. Also some ts types fixed and imports updates to support newer packages versions

**TODO**:
 - [ ] change validTransition to latestSupportedState in ForceMoveApp.sol
 - [ ] update function signatures and tests to support that update


nksazonov:
Trying to clarify what is need to be changed related to `latestSupportedState` implementation. As I understand from SATP paper, all state transition checks should be done in it in ForceMoveApp. If so, should `latestSupportedState` replace [`_requireStateSupportedBy`](https://github.com/statechannels/statechannels/blob/master/packages/nitro-protocol/contracts/ForceMove.sol#L488) logic currently implemented in ForceMove?

andrewgordstewart:
> As I understand from SATP paper, all state transition checks should be done in it in ForceMoveApp. If so, should latestSupportedState replace [_requireStateSupportedBy](https://github.com/statechannels/statechannels/blob/master/packages/nitro-protocol/contracts/ForceMove.sol#L488) logic currently implemented in ForceMove?

Yes, that's correct.

In the current `nitro-protocol`, the adjudicator holds a strong opinion about what makes a state "supported". "Supported" should mean: the adjudicator is willing

SATP suggests delegating that responsibility `latestSupportedState`. This removes a guard rail built into the protocol, but is a lot more flexible. We can extract the current code into a library that different `latestSupportedState` implementations can use, if that's useful.

As an, imagine a 3-party channel between Alice, Irene and Bob. The current rules say that at any point in time, either Alice can force an update on chain or Bob can; it's not possible to allow both of them to force an update (which is useful in a channel when Alice and Bob are trading between themselves, as opposed to Alice always giving assets to Bob).

Here's [an example](https://github.com/statechannels/statechannels/blob/e3f604b05d21f7caffca83afa3a1efb656d234cd/packages/nitro-protocol/src/contract/examples/embedded-app.ts#L40) of how these kinds of rules might be written (replace `validSupport`  with `latestSupportedState`). Is this example clear? (You can leave comments [here](https://github.com/statechannels/statechannels/pull/3654/files#diff-cd4e5eabc0c00a3c2c8cefef051fe35fe80b33360ec810af3ec6668f12ae9951R1) if you like!)

nitro-protocol/contracts/libraries/ECRecovery.sol
```diff
@@ -0,0 +1,28 @@
...
+
+    // If the version is correct return the signer address
+    if (v != 27 && v != 28) {
+      return (address(0));
+    } else {
+      return ecrecover(hash, v, r, s);
+    }
+  }
+
+}
```
geoknee, line 28: This file is new, but doesn't appear to be used anywhere? 

It looks to me like a wrapper for `ecrecover` which may add some extra safety. Is that right? Is there something equivalent in the openzeppelin library that we could use https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/ECDSA.sol ?

nitro-protocol/contracts/interfaces/IForceMove.sol
```diff
@@ -99,17 +93,17 @@ interface IForceMove {
      * @dev Finalizes a channel by providing a finalization proof.
      * @param largestTurnNum The largest turn number of the submitted states; will overwrite the stored value of `turnNumRecord`.
      * @param fixedPart Data describing properties of the state channel that do not change with state updates.
-     * @param appPartHash The keccak256 of the abi.encode of `(challengeDuration, appDefinition, appData)`. Applies to all states in the finalization proof.
-     * @param outcomeHash The keccak256 of the abi.encode of the `outcome`. Applies to all stats in the finalization proof.
+     * @param appData The keccak256 of the abi.encode of `(challengeDuration, appDefinition, appData)`. Applies to all states in the finalization proof.
```
geoknee, line 96: I think this comment needs updating -- `appData` is not the keccak256 of anything?

geoknee (commented):
Thanks for this @nksazonov! I'm still working through the changes but thought it was a good idea to share a couple of points early on :-)

nitro-protocol/contracts/libraries/ECRecovery.sol
nksazonov: It is a wrapper for `ecrecover` indeed. The reason this file was introduced is inconsistency of returned `v` signature value in `go-ethereum` (as described [here](https://github.com/ethereum/go-ethereum/issues/19751)) and its difference with what pure `ecrecover` solidity function [expects](https://bitcoin.stackexchange.com/a/112489) (i.e. `v` to be either 27 or 28).

We thought of using fault-proof OpenZeppelin analogy, but it has a lot of unnecessary tweaks and, more importantly, [requires modification](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/ECDSA.sol#L159-L160) to support what we need. Thus it was simpler and clearer to make a library with the same modifications that should have been done to OpenZeppelin contract.

nitro-protocol/contracts/interfaces/IForceMove.sol
nksazonov: You are definitely right. I changed `@param` name, but overlooked the description.

nitro-protocol/contracts/libraries/ECRecovery.sol
geoknee: I see. I wonder if it is better to fix this problem in the `go-nitro` client code, rather than on chain? That seems to be the advice from the `go-ethereum` team. What do you think?

nksazonov: I agree. We will try to fix it from `go-nitro` client code ASAP.

nitro-protocol/test/test-helpers.test.ts
```diff
@@ -25,7 +25,7 @@ const multiAssetReplaced = {
   '0xTOK': {'0xANOTHERCHANNEL': BigNumber.from(4)},
 };
 
-describe('replaceAddressesAndBigNumberify', () => {
```
geoknee: Is this `.only` necessary?

nitro-protocol/tsconfig.json
```diff
@@ -1,31 +1,13 @@
...
-    "outDir": "lib" /* Redirect output structure to the directory. */,
-    "module": "commonjs" /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */,
-    "target": "es5",
-    "strict": false /* Enable all strict type-checking options. */,
-    "removeComments": false,
-    "esModuleInterop": true /* Enables emit interoperability between CommonJS and ES Modules via creation of namespace objects for all imports. Implies 'allowSyntheticDefaultImports'. */,
-    "types": ["node", "jest"],
+    "target": "es2018",
+    "module": "commonjs",
+    "strict": true,
```
geoknee, line 5: Good to see `"strict": true`. Does `tsc .` work, or are there errors?

nitro-protocol/test/contracts/ForceMove/challenge.test.ts
```diff
@@ -255,14 +257,17 @@ describe('challenge', () => {
...
-    await (await ForceMove.setStatus(getChannelId(twoPartyChannel), HashZero)).wait();
+    await (await ForceMove.setStatus(getChannelId(twoPartyFixedPart), HashZero)).wait();
   });
+  // FIX: even if dropping channel status before each test, turn nums from prev tests are saved and can cause reverts
   it.each`
     description                                     | appData   | outcome                            | turnNums  | challenger
     ${'challenge(0,1) accepted'}                    | ${[0, 0]} | ${[]}                              | ${[0, 1]} | ${1}
     ${'challenge(1,2) accepted'}                    | ${[0, 0]} | ${[]}                              | ${[1, 2]} | ${0}
-    ${'challenge(1,2) accepted, MAX_OUTCOME_ITEMS'} | ${[0, 0]} | ${largeOutcome(MAX_OUTCOME_ITEMS)} | ${[1, 2]} | ${0}
+    ${'challenge(3,4) accepted, MAX_OUTCOME_ITEMS'} | ${[0, 1]} | ${largeOutcome(MAX_OUTCOME_ITEMS)} | ${[3, 4]} | ${0}
```
geoknee, line 270: I'd like to understand this a little better. What's going wrong?

nitro-protocol/src/transactions.ts
```diff
@@ -8,7 +8,7 @@ import {getStateSignerAddress, SignedState} from './signatures';
 // CONSTANTS
 export const MAGIC_ADDRESS_INDICATING_ETH = constants.AddressZero;
 export const MAX_TX_DATA_SIZE = 128 * 1024; // (bytes) https://github.com/ethereum/go-ethereum/blob/f59ed3565d18c1fa676fd34f4fd41ecccad707e8/core/tx_pool.go#L51
-export const NITRO_MAX_GAS = 6_000_000; // should be below the block gas limit, which can change over time and is generally different on mainnet, testnet and ganache.
+export const NITRO_MAX_GAS = 16_000_000; // should be below the block gas limit, which can change over time and is generally different on mainnet, testnet and ganache.
```
geoknee, line 11: Could you explain this increase from 6M to 16M?

nitro-protocol/test/test-helpers.test.ts
nksazonov: Definitely not.

nitro-protocol/tsconfig.json
nksazonov: `tsc --project ./` from root folder produces
```bash
Errors  Files
     7  node_modules/@types/jest/index.d.ts:35
     2  scripts/postdeploy.ts:6
```

geoknee: Maybe outside the scope of this work, but it might be nice to get to zero errors. Also, we want to gitignore the resulting `dist` folder.

nitro-protocol/src/transactions.ts
nksazonov: I've seen block gas limit has increased on all chains, but ganache. This change should be undone.

Also why have you put gas limit as 6 mil, not 6_721_975 as hardcoded in ganache?

nitro-protocol/contracts/interfaces/IForceMoveApp.sol
```diff
@@ -0,0 +1,29 @@
...
+pragma experimental ABIEncoderV2;
+
+/**
+ * @dev The IForceMoveApp interface calls for its children to implement an application-specific validTransition function, defining the state machine of a ForceMove state channel DApp.
+ */
+interface IForceMoveApp {
+    struct VariablePart {
+        bytes outcome;
+        bytes appData;
+    }
```
geoknee, line 12: @nksazonov This struct needs updating, too, I think?

Take a look at 744dbbef0ef22ba8085d60b8b0a8b468f038bc16 and 3b4bfbae06a4e5942ca2d88d80c2af1cea15740c

nitro-protocol/test/contracts/ForceMove/challenge.test.ts
nksazonov: The last test fails due to transaction revert with reason `turnNumRecord not increased`, which corresponds to `CountingApp.sol`. Both changing `turnNums` to `[3, 4]` and removing middle test help to avoid this revert. Unfortunately, I was not able to debug `CountingApp.sol` (via `console.log`) behavior with this tests because `ganache` is created dynamically and even setting env args to enable logging does not help.

nitro-protocol/contracts/interfaces/IForceMoveApp.sol
nksazonov: Yeah, I've also realized it needs updating to support `latestSupportedState`.

geoknee:
@nksazonov thanks a lot for all these contributions! We would like to propose the following workflow:

1. We will merge this PR as is very soon. 
2. We will make issues with the tag "on-chain" to manage any immediate issues arising from this PR (i.e. the comments above). 
3. We will add you and any other openware devs as external collaborators so you can make branches and raise pull requests in this repo. 
4. We can assign issues to prevent duplicating work.
5. We will setup continuous integration (see #487) as soon as possible. One of the outstanding pieces of work will be to get the tests passing 100% (I think we are very close). 
6. We may add some other CI checks such as linting, prettier and so on. 

How does this sound to you? Please let me know your thoughts.

geoknee:
There is also another thing I would like to resolve before merging!

0. We need to resolve an issue with the readme files. Most of our team use Mac OSX, which has a case insensitive file system. When I check out this branch on a Mac, I immediately see some unstaged changes to `./nitro-protocol/README.md`. These changes are reverting the readme back to the old `nitro-protocol` readme from the statechannels monorepo. If I restore the file (reject the changes), I immediately get some unstaged changes to `./nitro-protocol/readme.md`, changing that file to your new readme. It's hard to tell on my system, but I guess there are two files there which differ only in the casing? In that case, deleting one of them ought to solve the problem?

nksazonov:
> How does this sound to you? Please let me know your thoughts.

@geoknee sounds like an amazing collaboration start!
Before merging I propose we resolve all minor problems we have discussed (e.g. [comments](https://github.com/statechannels/go-nitro/pull/427#discussion_r841974776) and [gas limit](https://github.com/statechannels/go-nitro/pull/427#discussion_r842618730)).

geoknee:
> Before merging I propose we resolve all minor problems we have discussed (e.g. [comments](https://github.com/statechannels/go-nitro/pull/427#discussion_r841974776) and [gas limit](https://github.com/statechannels/go-nitro/pull/427#discussion_r842618730)).

Sounds good to me!

nksazonov:
> It's hard to tell on my system, but I guess there are two files there which differ only in the casing?

Yes, this is the reason. Deleting ours should solve the problem.

nitro-protocol/src/transactions.ts
geoknee: > Also why have you put gas limit as 6 mil, not 6_721_975 as hardcoded in ganache?

I struggle to remember exactly :-P. I think the idea is that we want to assert that transactions in nitro/satp do not consume excessive gas, where excessive means "getting near the block gas limit". As I recall, getting near the limit is enough to make the transaction unattractive to miners. So we picked a value well below the limit on most test nets (ganache is a bit of an exception because there is no gas auction to compete in). 

Anyways, probably not super important -- I suggest we just leave this value as is (unless it causes problems). We can revisit if needs be.

geoknee (approved):

PR #433, opened 2022-03-29
Consensus Channel constructors now accept a turn number

lalexgap:
This modifies the constructors for consensus channels to accept a turn number instead of assuming a turn number of 0.

This allows the consensus channel to be agnostic to what turn we're starting at. This means it is unaffected by decisions to sign post fund setups or not.

### DIFF
--- a/channel/consensus_channel/consensus_channel.go
+++ b/channel/consensus_channel/consensus_channel.go
@@ -34,11 +34,11 @@ type consensusChannel struct {
 func newConsensusChannel(
 	fp state.FixedPart,
 	myIndex ledgerIndex,
-	turnNum uint64,
+	initialTurnNum uint64,
 	outcome LedgerOutcome,
 	signatures [2]state.Signature,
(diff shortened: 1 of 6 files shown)


channel/consensus_channel/consensus_channel.go
```diff
@@ -34,10 +34,11 @@ type consensusChannel struct {
 func newConsensusChannel(
 	fp state.FixedPart,
 	myIndex ledgerIndex,
+	turnNum uint64,
```
andrewgordstewart, line 37: Would `initialTurnNum` be more clear?

andrewgordstewart (approved):

channel/consensus_channel/consensus_channel.go
lalexgap: Updated in [253b246](https://github.com/statechannels/go-nitro/pull/433/commits/253b246559ca0980e34344e75a4a4c928a4f332b)

Issue #435, opened 2022-03-29
The `store` doesn't properly store/retrieve the outcome


lalexgap:
Currently it looks like the `LedgerOutcome` that comes back has a zeroed out destinations.

lalexgap:
I think this is due to `LedgerOutcome` only containing private fields. `The raw json string seems to not include any private fields:
```{"Id":[43,102,29,60,92,242,199,232,69,165,154,35,101,113,70,211,0,173,136,123,102,86,197,26,145,39,100,224,102,56,245,60],"MyIndex":0,"FP":{"ChainId":0,"Participants":["0xaaa6628ec44a8a742987ef3a114ddfe2d4f7adce","0xbbb676f9cff8d242e9eac39d063848807d3d1d94"],"ChannelNonce":37140676580,"AppDefinition":"0x5e29e5ab8ef33f050c7cc10b5a0456d975c5f88d","ChallengeDuration":60},"Current":{"TurnNum":0,"Outcome":{},"Signatures":[{"R":"ixqul6ghVVjVQ1HHfpLyXhz6F7qZ/tzjW1m6Koiq1Uo=","S":"XYJDKQ67hVpr8v/CIa9HlBd6FZLQLwJPy10fxNg1TJI=","V":0},{"R":"rKC5Yz2hL4oiHTUb+kuKQBDCp7QYad6+KNmXt9KmeyQ=","S":"CqPy+y0SchbHyDWf9wC554KVyD0X6NNcoZ6EUj8t9qA=","V":0}]},"ProposalQueue":[]}
```

If I change the LedgerOutcome fields to be public I start seeing them show up in the generated JSON.

geoknee:
I've started looking into this. The private fields are an issue, but not the only issue! I'll start by writing some explicit serde tests for a consensus channel.

PR #437, opened 2022-03-29
Create Consensus Channel in the Store when Direct Funding Objective Completes

lalexgap:
⚠️ Based on #449 

Whenever a direct funding objective completes the engine now constructs a consensus channel and stores it in the `store`.

## Testing

To this test this new functionality the direct funding integration test now inspects the `store` for a consensus channel after the objective has been completed. To support this the `setupClient` now returns the `store` it generates and passes into the client constructor.

~The test currently doesn't inspect the details of the Consensus Channel as that would fail due to #435~
The test checks that the consensus outcome and turn num are what we expect 

#### Code quality

#### Project management

### DIFF
--- a/channel/consensus_channel/consensus_channel.go
+++ b/channel/consensus_channel/consensus_channel.go
@@ -9,6 +9,8 @@ import (
 	"github.com/statechannels/go-nitro/channel/state"
 	"github.com/statechannels/go-nitro/channel/state/outcome"
 	"github.com/statechannels/go-nitro/crypto"
+	"github.com/statechannels/go-nitro/protocols/directfund"
+
 	"github.com/statechannels/go-nitro/types"
 )
 
@@ -228,15 +230,15 @@ func (o *LedgerOutcome) includes(g Guarantee) bool {
 // - The first alloction entry is for left
 // - The second alloction entry is for right
 // - We ignore guarantee metadata and just assume that it is [left,right]
-func FromExit(sae outcome.SingleAssetExit) *LedgerOutcome {
+func FromExit(sae outcome.SingleAssetExit) LedgerOutcome {
 
-	left := Balance{destination: sae.Allocations[0].Destination, amount: *sae.Allocations[0].Amount}
-	right := Balance{destination: sae.Allocations[1].Destination, amount: *sae.Allocations[1].Amount}
+	left := Balance{destination: sae.Allocations[0].Destination, amount: sae.Allocations[0].Amount}
+	right := Balance{destination: sae.Allocations[1].Destination, amount: sae.Allocations[1].Amount}
 	guarantees := make(map[types.Destination]Guarantee)
 	for _, a := range sae.Allocations {
 
 		if a.AllocationType == outcome.GuaranteeAllocationType {
-			g := Guarantee{amount: *a.Amount,
+			g := Guarantee{amount: a.Amount,
 				target: a.Destination,
 				// Instead of decoding the metadata we make an assumption that the metadata has the left/right we expect
 				left:  left.destination,
@@ -245,7 +247,7 @@ func FromExit(sae outcome.SingleAssetExit) *LedgerOutcome {
 		}
 
 	}
-	return &LedgerOutcome{left: left, right: right, guarantees: guarantees, assetAddress: sae.Asset}
+	return LedgerOutcome{left: left, right: right, guarantees: guarantees, assetAddress: sae.Asset}
 
 }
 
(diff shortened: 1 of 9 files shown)


geoknee:
@lalexgap I rebased this branch.

channel/consensus_channel/consensus_channel.go
```diff
@@ -410,3 +438,47 @@ func (c *ConsensusChannel) UnmarshalJSON(data []byte) error {
 
 	return nil
 }
+
+// CreateFromDirectFundingObjective accepts an objective and generates a new consensus channel from it.
+// It assumes that EVERY DirectFundingObjective is for a ledger channel.
+func CreateFromDirectFundingObjective(dfo directfund.Objective) (*ConsensusChannel, error) {
```
lalexgap, line 444: Originally I thought this could live in the `directFund` package but I ran into circular dependencies.

lalexgap: That may be resolved by https://github.com/statechannels/go-nitro/pull/450

client/engine/engine.go
```diff
@@ -264,6 +265,18 @@ func (e *Engine) attemptProgress(objective protocols.Objective) (outgoing Object
...
+		if dfo, isDfo := crankedObjective.(*directfund.Objective); isDfo {
+			c, err := consensus_channel.CreateFromDirectFundingObjective(*dfo)
+			if err != nil {
+				return ObjectiveChangeEvent{}, fmt.Errorf("could not create consensus channel for objective %s: %w", objective.Id(), err)
+			}
+			err = e.store.SetConsensusChannel(c)
+			if err != nil {
+				return ObjectiveChangeEvent{}, fmt.Errorf("could not store consensus channel for objective %s: %w", objective.Id(), err)
+			}
+		}
```
geoknee, line 279: Could this be factored into a function?

channel/consensus_channel/consensus_channel.go
geoknee: I think the `directfund` package needs to know about (i.e. import `ConsensusChannel`), but `consensus_channel` should _not_ know about `directfund`.

channel/consensus_channel/consensus_channel.go
```diff
@@ -223,6 +231,32 @@ func (o *LedgerOutcome) includes(g Guarantee) bool {
...
+
+	left := Balance{destination: sae.Allocations[0].Destination, amount: sae.Allocations[0].Amount}
+	right := Balance{destination: sae.Allocations[1].Destination, amount: sae.Allocations[1].Amount}
+	guarantees := make(map[types.Destination]Guarantee)
+	for _, a := range sae.Allocations {
+
+		if a.AllocationType == outcome.GuaranteeAllocationType {
+			g := Guarantee{amount: a.Amount,
+				target: a.Destination,
+				// Instead of decoding the metadata we make an assumption that the metadata has the left/right we expect
```
geoknee, line 249: Is it worth encoding our expectation and panicking if it doesn't match?

client/engine/engine.go
geoknee: And should we do this _before_ we do `e.executeSideEffects(sideEffects)`? I think this would help to reduce the chance of a race between an incoming message  from a counterparty which is triggered by `e.executeSideEffects`, and the insertion of the `ConsensusChannel` into the store.

client_test/directfund_integration_test.go
```diff
@@ -41,9 +44,29 @@ func TestDirectFundIntegration(t *testing.T) {
...
+		got := vars.Outcome.AsOutcome()
+
+		if diff := cmp.Diff(want, got); diff != "" {
+			t.Fatalf("expected outcome to be %v, got %v:\n %v", want, got, diff)
+		}
+		if vars.TurnNum != 1 {
+			t.Fatal("expected consensus turn number to be the post fund setup 1, received #$v", vars.TurnNum)
+		}
+
+	}
```
geoknee, line 70: This could also be factored into a helper.

channel/consensus_channel/consensus_channel.go
```diff
@@ -428,3 +462,47 @@ func (v Vars) AsState(fp state.FixedPart) state.State {
 func (c *ConsensusChannel) Participants() []types.Address {
 	return c.fp.Participants
 }
+
+// CreateFromDirectFundingObjective accepts an objective and generates a new consensus channel from it.
+// It assumes that EVERY DirectFundingObjective is for a ledger channel.
+func CreateFromDirectFundingObjective(dfo directfund.Objective) (*ConsensusChannel, error) {
+	// The current assumption is that ANY direct funding objective is for a ledger channel
```
geoknee, line 469: You've stated this assumption twice. But I would argue the assumption applies to the code in the engine, not here? The engine could choose not to call this function for certain direct fund objectives.

channel/consensus_channel/consensus_channel.go
```diff
@@ -428,3 +462,47 @@ func (v Vars) AsState(fp state.FixedPart) state.State {
...
+
+	if ledger.MyIndex == uint(leader) {
+		con, err := NewLeaderChannel(ledger.FixedPart, turnNum, outcome, signatures)
+		if err != nil {
+			return nil, fmt.Errorf("could not create consensus channel as leader: %w", err)
+		}
+		return &con.ConsensusChannel, nil
+
+	} else {
+		con, err := NewLeaderChannel(ledger.FixedPart, turnNum, outcome, signatures)
```
geoknee, line 501:
```suggestion
		con, err := NewFollowerChannel(ledger.FixedPart, turnNum, outcome, signatures)
```

client_test/directfund_integration_test.go
```diff
@@ -41,9 +44,29 @@ func TestDirectFundIntegration(t *testing.T) {
...
+	clientA, storeA := setupClient(alice.PrivateKey, chain, broker, logDestination, 0)
+	clientB, storeB := setupClient(bob.PrivateKey, chain, broker, logDestination, 0)
 
 	directlyFundALedgerChannel(t, clientA, clientB)
 
+	want := testdata.Outcomes.Create(*clientA.Address, *clientB.Address, 5, 5)
+	// Ensure that we create a consensus channel in the store
+	for _, store := range []store.Store{storeA, storeB} {
+		con, ok := store.GetConsensusChannel(*clientA.Address, *clientB.Address)
+
```
geoknee, line 56: Could you assert on the type of `con` to catch the leader/follower typo you made above?

Although, given #462 those types may not exist for long.

channel/consensus_channel/consensus_channel.go
geoknee: Why didn't this cause a failing test?

geoknee (approved):
🌋 

Left a few comments, but I'm inclined to merge this and work on improvements in a follow up PR, given that @lalexgap is out today.

channel/consensus_channel/consensus_channel.go
geoknee: A: the **only** difference in the resulting channel is that `myIndex` will be incorrect. The **only** place where that is actually used is when signing vars, which doesn't happen in the integration test.

PR #440, opened 2022-03-30
make types.Destination an encoding.TextMarshaler 

geoknee:
Towards #435

This allows types.Destination to be used as the key in a mapping such as this one: 

https://github.com/statechannels/go-nitro/blob/909aa58c22bb2bc98680720ebc03c5e3f5a74996/channel/consensus_channel/consensus_channel.go#L180

which will be json.Marshaled. This is because JSON objects must have strings as keys (so a big.Int being a struct, won't do). 

It has the nice side effect of making the data much easier to read (bye bye bytes arrays!). 

TODO
- [x] unit test to make sure the marshaler and unmarshaler are inverses of each other
- [x] fixing remaining unit tests which expect byte arrays (and now need to expect hex strings)


lalexgap:
This includes c2dbe4a that has its own [pull request](https://github.com/statechannels/go-nitro/pull/443).

types/serde_test.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"testing"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestTextMarshalling(t *testing.T) {
+	const destString = "0x000000000000000000000000f5a1bb5607c9d079e46d1b3dc33f257d937b43bd"
+	dest := Destination(common.HexToHash("0x000000000000000000000000f5a1bb5607c9d079e46d1b3dc33f257d937b43bd"))
```
andrewgordstewart, line 12: I'm confused about this syntax -- is `Destination` a function? Or is it a different way of creating a `Destination` value?

andrewgordstewart: ah, it is type conversion...

andrewgordstewart (approved):

Issue #420, opened 2022-03-28
The virtualfund objective should use a ConsensusChannel in its Connection struct


lalexgap:

andrewgordstewart:
Myself & @NiloCK discussed a strategy for doing this work iteratively:

## 1. Define the interface that a `virtualfund` objective would use.
Note that `LeaderChannel` and `FollowerChannel` both implement
  - a `(sp SignedProposal, privateKey []byte) -> (SideEffects, error)` method that is used to sign some proposal
  - a `(sp SignedProposa) -> (SideEffects, error)` method that is used to sign some proposal
So, we can define a `LedgerChannel` interface like this:
```
type interface LedgerChannel {
  IncludeGuarantee(p Proposal, sk []byte) (SideEffects, error)
  ReceiveProposal(p SignedProposal) (SideEffects, error)
}
```
and stick it on a LedgerConnection. Keep the `TwoPartyLedgerChannel` in place!

## 2. Add proposals to `Message`
Keeping the existing message format in place, add data to messages to allow the transfer of Leader/Follower SignedProposal.

## 3. Use the interface throughout `virtualfund`

Where the TwoPartyLedger is used, we can start using a LedgerChannel. This can be done iteratively, across multiple PRs, eg, these might be multiple PRs:
1. peers use it to include a guarantee [here](https://github.com/statechannels/go-nitro/blob/c9ed4bd109cbae3b81eb4cc7ac486482ed07dc4b/protocols/virtualfund/virtualfund.go#L643)
2. then, they use it to receive a proposal [here](https://github.com/statechannels/go-nitro/blob/c9ed4bd109cbae3b81eb4cc7ac486482ed07dc4b/protocols/virtualfund/virtualfund.go#L589)
3. then, they use it to countersign a proposal [here](https://github.com/statechannels/go-nitro/blob/c9ed4bd109cbae3b81eb4cc7ac486482ed07dc4b/protocols/virtualfund/virtualfund.go#L650)

## 4. (maybe) Assert against existing behaviour

By iteratively adding in `consensus_channel` functionality, we can think about asserting that it has the expected behaviour in the client code. (This is temporary!)

## 5. Remove the old code
Once the new code is fully in place, we can remove the old code.

This strategy can allow for work to be done in parallel -- eg. 1 & 2 can be done in parallel, as well as many parts of 3. It also allows for the gradual inclusion of the new code.

NiloCK:
[...]

PR #442, opened 2022-03-30
WIP: issue #420

NiloCK:
This can be a working branch for #420. PR as stands includes a number of labelled todos of the form `// todo: #420`, and stub inserts of ConsensusChannels into the `Connection` struct.

PR also includes a refactor of `consensus_channel`'s tests to avoid importing `testdata` and the cyclic import issue observed in #434.

### DIFF
--- a/protocols/virtualfund/virtualfund.go
+++ b/protocols/virtualfund/virtualfund.go
@@ -63,6 +63,66 @@ func (c *Connection) Equal(d *Connection) bool {
 
 }
 
+// ledgerChannelAffordsExpectedGuarantees returns true if, for the channel inside the connection, and for each asset keying the input variables, the channel can afford the allocation given the funding.
+// The decision is made based on the latest supported state of the channel.
+//
+// Both arguments are maps keyed by the same asset.
+func (c *Connection) ledgerChannelAffordsExpectedGuarantees() bool {
+	return c.Channel.Affords(c.getExpectedGuarantees(), c.Channel.OnChainFunding) // todo: #420 deprecate
+}
+
+// insertGuaranteeInfo mutates the reciever Connection struct.
+func (c *Connection) insertGuaranteeInfo(a0 types.Funds, b0 types.Funds, vId types.Destination, left types.Destination, right types.Destination) error {
+
+	c.GuaranteeInfo = GuaranteeInfo{
+		Left:                 left,
+		Right:                right,
+		LeftAmount:           a0,
+		RightAmount:          b0,
+		GuaranteeDestination: vId,
+	}
+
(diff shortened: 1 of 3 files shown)


protocols/virtualfund/virtualfund.go
```diff
@@ -38,8 +39,9 @@ type GuaranteeInfo struct {
 	GuaranteeDestination types.Destination
 }
 type Connection struct {
-	Channel       *channel.TwoPartyLedger // todo: #420 deprecate in favor of consensus_channel.ConsensusChannel
-	GuaranteeInfo GuaranteeInfo
+	Channel          *channel.TwoPartyLedger // todo: #420 deprecate in favor of
+	ConsensusChannel *consensus_channel.ConsensusChannel
```
andrewgordstewart, line 43: Does it make sense to hold a `ConsensusChannel` here? Would holding something that implement this interface make more sense?
```
type LedgerChannel interface {
  IncludeGuarantee(p Proposal, sk []byte) (SideEffects, error)
  ReceiveProposal(p SignedProposal) (SideEffects, error)
  Funds(g Guarantee) bool
  Proposed(g Guarantee) bool
  // alternatively
  Status(g Guarantee) InclusionStatus
}
```

My understanding is that a `ConsensusChannel` value would not provide the functionality that `virtualfund` would need.

On the other hand, both a `LeaderChannel` and a `FollowerChannel` can implement the `LedgerChannel` interface. In this case, the `virtualfund` objective can:
- receive proposals in `Update` by calling `ReceiveProposal(p)` when a message contains `p`
- in `Crank`, check the state of ledger funding using `Funds`/`Proposed`/`Status`, and use `IncludeGuarantee` when appropriate.

NiloCK: Possibly!

My intention had been to make `Connection` a working boundary between `virtualfund` and `consensus_channel`, give it (some approximation of) the above API methods, and have it internally manage delegation to the concrete types.

andrewgordstewart: Ok! I think I understand your approach, which could also work (I don't understand yet how that would work, but that's fine!)

andrewgordstewart:
I've cherry-picked c9d17c572f2704e2405a97a2db68753da8c1457c into its own branch, and created a PR for it, since it can go in on its own:  https://github.com/statechannels/go-nitro/pull/450

protocols/virtualfund/virtualfund.go
kerzhner: Colin, I would be really interested to see your approach in more detail. I played around with the `LedgerChannel` interface, and that approach does seem to encapsulate the leader/follower logic nicely outside of the virtual fund protocol logic.

PR #445, opened 2022-03-30
generate allocations deterministically in createOutcomeFromMap

### LINKED ISSUE #444: Flickering test
Rarely the mock store tests will fail due to a panic:
```
panic: error constructing genericVFO: error creating objective: Allocation in slot 0 does not correspond to participant 0 [recovered]
	panic: error constructing genericVFO: error creating objective: Allocation in slot 0 does not correspond to participant 0
```

Here is an example of the [flicker](https://github.com/statechannels/go-nitro/runs/5759099344?check_suite_focus=true#step:6:319) on a feature branch.

Locally I've been able to reproduce this on `main` but it takes a while.

lalexgap:
Fixes #444 

Updates createOutcomeFromMap to deterministically create allocations by always sorting the addresses instead of relying on the order we happen to get by iterating through the map.

### DIFF
--- a/internal/testdata/states.go
+++ b/internal/testdata/states.go
@@ -1,7 +1,9 @@
 package testdata
 
 import (
+	"bytes"
 	"math/big"
+	"sort"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/statechannels/go-nitro/channel/state"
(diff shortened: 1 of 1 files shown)


andrewgordstewart (commented):
This resolves the issue, but it doesn't feel very ergonomic to use `CreateOutcomeFromMap`:
- if you're using it to create a 2-item outcome only, where Alice's item goes first, it works as expected
- if you're using it to create a more-than-2 item outcome, or if you want Bob's item to go first, you're probably out of luck

`CreateOutcomeFromMap` is only used in one place, and creates an Alice-first outcome, so that's not a problem right now.

Would it be better for the function's signature to be something like `CreateOutcome(items ...SomeType)`? Then, both of these return the expected outcome
```
CreateOutcome(
  Item{Alice, 6},
  Item{Bob, 10},
}

CreateOutcome(
  Item{Bob, 10},
  Item{Alice, 6},
}
```

lalexgap:
> This resolves the issue, but it doesn't feel very ergonomic to use `CreateOutcomeFromMap`:
> 
> * if you're using it to create a 2-item outcome only, where Alice's item goes first, it works as expected
> * if you're using it to create a more-than-2 item outcome, or if you want Bob's item to go first, you're probably out of luck
> 
> `CreateOutcomeFromMap` is only used in one place, and creates an Alice-first outcome, so that's not a problem right now.
> 
> Would it be better for the function's signature to be something like `CreateOutcome(items ...SomeType)`? Then, both of these return the expected outcome
> 
> ```
> CreateOutcome(
>   Item{Alice, 6},
>   Item{Bob, 10},
> }
> 
> CreateOutcome(
>   Item{Bob, 10},
>   Item{Alice, 6},
> }
> ```

That sounds good! Perhaps that could be tackled separately under it's own issue?

andrewgordstewart (approved):

PR #447, opened 2022-03-30
Switch to pointers to big.int in ConsensusChannel

lalexgap:
According to the [documentation](https://pkg.go.dev/math/big#Int) shallow copies of `big.Ints` can lead to errors. However we have a couple of structs (like `Balance` and `Guarantee`) that contain `big.Int` instead of a pointer. This means that when these structs get passed around by value (like [here](https://github.com/statechannels/go-nitro/blob/ag/store-bug/channel/consensus_channel/leader_channel.go#L35:L35)) we're creating shallow copies of `big.Int`.

It seems safer to use pointers to `big.Int` in consensus channel so that:
- we follow what's recommended by the documentation
-  we keep consistent with how we deal with big.ints in the rest of the [code](https://github.com/statechannels/go-nitro/blob/ag/store-bug/channel/state/outcome/exit.go#L40:L40)
- BONUS: Using pointers means that big.ints will properly marshal to json

Once I made the change to pointers in [b3bd3ed](https://github.com/statechannels/go-nitro/pull/447/commits/b3bd3eddc796ec4a412001bf64371881739f6b50) I had to update the store test in [da6d998](https://github.com/statechannels/go-nitro/pull/447/commits/da6d9983cb430abb75d9f5c1f9cf46d892576168). This was because the store test was using a nil `LedgerOutcome` which now would contain `nil` big.int pointers and would not serialize out properly. To address this I've added some simple constructor functions and constructed a proper outcome for the store test.

### DIFF
--- a/channel/consensus_channel/consensus_channel.go
+++ b/channel/consensus_channel/consensus_channel.go
@@ -132,6 +132,16 @@ func (c *ConsensusChannel) latestProposedVars() (Vars, error) {
 	return vars, nil
 }
 
+// NewBalance returns a new Balance struct with the given amount and destination
+func NewBalance(destination types.Destination, amount *big.Int) Balance {
+	balanceAmount := big.NewInt(0).Set(amount)
+	return Balance{
+		destination: destination,
+		amount:      balanceAmount,
+	}
+
+}
+
 // Balance represents an Allocation of type 0, ie. a simple allocation.
 type Balance struct {
 	destination types.Destination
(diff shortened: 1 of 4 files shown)


channel/consensus_channel/consensus_channel.go
```diff
@@ -132,37 +132,49 @@ func (c *ConsensusChannel) latestProposedVars() (Vars, error) {
 	return vars, nil
 }
 
+// NewBalance returns a new Balance struct with the given amount and destination
+func NewBalance(destination types.Destination, amount *big.Int) Balance {
```
andrewgordstewart, line 136: I hope this isn't a trademark violation.

andrewgordstewart (approved):

PR #448, opened 2022-03-30
fix issue 446

### LINKED ISSUE #446: Update CreateOutcomeFromMap to be more ergonomic
This resolves the issue, but it doesn't feel very ergonomic to use `CreateOutcomeFromMap`:
- if you're using it to create a 2-item outcome only, where Alice's item goes first, it works as expected
- if you're using it to create a more-than-2 item outcome, or if you want Bob's item to go first, you're probably out of luck

`CreateOutcomeFromMap` is only used in one place, and creates an Alice-first outcome, so that's not a problem right now.

Would it be better for the function's signature to be something like `CreateOutcome(items ...SomeType)`? Then, both of these return the expected outcome
```
CreateOutcome(
  Item{Alice, 6},
  Item{Bob, 10},
}

CreateOutcome(
  Item{Bob, 10},
  Item{Alice, 6},
}
```

_Originally posted by @andrewgordstewart in https://github.com/statechannels/go-nitro/pull/445#pullrequestreview-926638058_

andrewgordstewart:
Fixes #446

### DIFF
--- a/internal/testdata/states.go
+++ b/internal/testdata/states.go
@@ -9,9 +9,8 @@ import (
 	"github.com/statechannels/go-nitro/types"
 )
 
-// a simpleItem is an ergonomic way to create a simple allocation item, without having to create a big.Int
-// This is more ergonomic for creating test data
-type simpleItem struct {
+// SimpleItem is an ergonomic way to create a simple allocation item, without having to create a big.Int
+type SimpleItem struct {
 	Dest   types.Destination
 	Amount int64
 }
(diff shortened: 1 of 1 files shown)


internal/testdata/states.go
```diff
@@ -108,28 +111,15 @@ func createOutcome(first types.Address, second types.Address, x, y uint) outcome
...
 }
 
-func createOutcomeFromMap(amounts map[types.Address]uint) outcome.Exit {
-	var allocations []outcome.Allocation
-
-	// Generate a list of addresses from the map
-	addresses := make([]types.Address, 0, len(amounts))
-	for address, _ := range amounts {
-		addresses = append(addresses, address)
+func createLongOutcome(items ...simpleOutcome) outcome.Exit {
```
lalexgap, line 114: What does `Long` mean? More than one entry?

lalexgap (approved):
👍

internal/testdata/states.go
andrewgordstewart: yes -- I can clarify this in a doctoring?

andrewgordstewart: https://github.com/statechannels/go-nitro/pull/448/commits/8086fb6bba3e05a3e86cbf60c8bc276336dbf4b6

PR #450, opened 2022-03-30
Remove testdata dependency from consensus_channel

andrewgordstewart:
Having `consensus_channel` depend on `testdata` would create a circular dependency if `consensus_channel` is imported into `virtual_channel`. This problem is resolved by copying the required functionality into `consensus_channel`

### DIFF
--- a/channel/consensus_channel/consensus_channel_test.go
+++ b/channel/consensus_channel/consensus_channel_test.go
@@ -7,10 +7,48 @@ import (
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/google/go-cmp/cmp"
+
 	"github.com/statechannels/go-nitro/channel/state"
-	"github.com/statechannels/go-nitro/internal/testdata"
 	"github.com/statechannels/go-nitro/types"
 )
+
+type actor struct {
+	Address    types.Address
+	PrivateKey []byte
+}
+
+func (a actor) Destination() types.Destination {
+	return types.AddressToDestination(a.Address)
+}
+
+// actors namespaces the actors exported for test consumption
+type actors struct {
+	Alice actor
+	Bob   actor
+	Brian actor
+	Irene actor
+}
+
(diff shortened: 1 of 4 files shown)


channel/consensus_channel/consensus_channel_test.go
```diff
@@ -7,11 +7,49 @@ import (
...
+type actors struct {
+	Alice actor
+	Bob   actor
+	Brian actor
+	Irene actor
+}
+
+// Actors is the endpoint for tests to consume constructed statechannel
+// network participants (public-key secret-key pairs)
+var Actors actors = actors{
```
kerzhner, line 34: I don't Actors need to be exported.
```suggestion
var actors actors = actors{
```

channel/consensus_channel/consensus_channel_test.go
```diff
@@ -7,11 +7,49 @@ import (
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/google/go-cmp/cmp"
+
 	"github.com/statechannels/go-nitro/channel/state"
-	"github.com/statechannels/go-nitro/internal/testdata"
 	"github.com/statechannels/go-nitro/types"
 )
 
+type actor struct {
```
kerzhner, line 15: These definitions should probably live in `helpers_test.go`

kerzhner (approved):

channel/consensus_channel/consensus_channel_test.go
andrewgordstewart: Done in another PR: db3eeec081d65f24cbce3106108692186b3dd8fc

channel/consensus_channel/consensus_channel_test.go
andrewgordstewart: It's removed in https://github.com/statechannels/go-nitro/commit/c3bcfa7c70e994eb3fb1e3b68958f6f632323488 (in another PR)

PR #453, opened 2022-03-31
Add tests for unmarshaling a `ConsensusChannel`

geoknee:
* refactors existing serde test into a table
* adds unmarshaling subtest
* ~surfaces an error, which I think is caused by trying to unmarshal into an `interface`? See #454~
* hits problems with serializing an `Add` (which causes everything else to break).


lalexgap:
Closing in favour of #459

PR #454, opened 2022-03-31
Change type of `Proposal` from `interface{}` to intersection type.

geoknee:
Towards #435

>[ interface{} says nothing](https://go-proverbs.github.io/)

Currently, we use a `Proposal interface{}` property nested within a `ConsensusChannel`. Using the `interface{}` type has some drawbacks:
1. A lack of type safety.  Any type could be used, here. And furthermore, we need to perform type assertions on the interface in our production code:
https://github.com/statechannels/go-nitro/blob/2b43476b53c451ea80325cdc2705627dd7b7be9c/channel/consensus_channel/consensus_channel.go#L125
2. Difficulties unmarshaling from JSON. The JSON will typically contain a concrete type of `Add` (or `Remove` which does not yet exist) -- unless of course the lack of type safety has let something else sneak in there. 

The immediate concern is 2, because this blocks #435. There are a few different ways to solve this: 

1. We could to write a custom unmarshaler which inspects the json, figures out which concrete type it should be unmarshaled into, and then calls into a custom unmarshaler for the appropriate concrete type.
2. There could be a solution involving generics, but I haven't thought much about that.
3. We take the approach adopted here (and elsewhere in the codebase) to avoid having a sort of "poor man's union type", in favour of something more like an intersection type. That is, a `Proposal` contains _both_ an `Add` and a `Remove` -- and we aim to have only one of them as non-zero. If we encounter both, we can choose what to do (in this PR, return an error). 

**Changes**

See comment below for the main type modification. The other changes are minor, aside from error handling when both Add and Remove are found. 

**How Has This Been Tested?** 
Existing test assertions have been updated to match the new type. 

---

#### Code quality

#### Project management


kerzhner (approved):
This seems like a reasonable approach! It would be nice to spend a bit of time exploring generics to understand if there is any benefit to using generics.

NiloCK:
I haven't read this changeset (only the highlighted rub) but I had an alternative thought on this yesterday. It's fairly different but I don't have much immediate opinion on better-or-worse.

How about:

```
type LedgerChannelUpdater interface {
  Update(vars *Vars) error // could update-in-place or return a new Vars
}

func (a Add) Update(vars *Vars) error {
  // perform mutation currently done in Vars.Add(a Add)
}

type SignedProposal {
  state.Signature
  Proposal LedgerChannelUpdater
}
```

geoknee:
@NiloCK your thought there certainly reads like idiomatic Go. I feel like the fact that Go interfaces tell us about what something _does_  -- rather than what something _is_ -- makes them pretty awkward when it comes to serializing data.

andrewgordstewart:
> @NiloCK your thought there certainly reads like idiomatic Go. I feel like the fact that Go interfaces tell us about what something _does_ -- rather than what something _is_ -- makes them pretty awkward when it comes to serializing data.

If the `SignedProposal` interface defines `MarshalJSON` and `UnmarshalJSON`, it would be the responsibility of each type implementing `SignedProposal` to serialize itself, right? Would that relieve the awkwardness?

```
type SignedProposal {
  type 'add' | 'remove'
  Proposal Proposal
  Signature state.Signature
}

type Proposal interface {
  MarshalJSON()...
  UnmarshalJSON() ...
}

func (a *Add) MarshalJSON() {
  // a knows what it is, and can be marshalled without type assertion
}

func (r *Remove) MarshalJSON() {
  // likewise
}
```

PR #464, opened 2022-03-31
Enhance checks for the ConsensusChannel Store test

### LINKED ISSUE #460: Update Consensus Store tests to check to fully inspect the fetched value
We should update the test to properly check the `got` [value](https://github.com/statechannels/go-nitro/blob/da6d9983cb430abb75d9f5c1f9cf46d892576168/client/engine/store/mockstore_test.go#L161).

lalexgap:
Fixes #460 

This PR:
- Introduces a deep comparison of the fetched value against the expected value
- Updates the mock store test to use a richer consensus channel (that includes a proposal and guarantee)
- Adds some new constructor functions to aid in constructing a richer consensus channel

~This currently fails due to `Add` serialization that is resolved in #459.~ 
Now that #459 is merged into `main` this passes 🎉

### DIFF
--- a/channel/consensus_channel/consensus_channel.go
+++ b/channel/consensus_channel/consensus_channel.go
@@ -161,6 +161,11 @@ type Guarantee struct {
 	right  types.Destination
 }
 
+// NewGuarantee constructs a new guarantee
+func NewGuarantee(amount *big.Int, target types.Destination, left types.Destination, right types.Destination) Guarantee {
+	return Guarantee{amount, target, left, right}
+}
+
 func (g Guarantee) equal(g2 Guarantee) bool {
 	if !types.Equal(g.amount, g2.amount) {
 		return false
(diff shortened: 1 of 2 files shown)


client/engine/store/mockstore_test.go
```diff
@@ -145,7 +148,17 @@ func TestConsensusChannelStore(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	if err := ms.SetConsensusChannel(&want.ConsensusChannel); err != nil {
+	// Generate a new proposal so we test that the proposal queue is being fetched properly
+	proposedGuarantee := cc.NewGuarantee(big.NewInt(1), types.Destination{2}, left.AsAllocation().Destination, right.AsAllocation().Destination)
+	proposal := cc.NewAdd(2, proposedGuarantee, big.NewInt(1))
+	_, err = leader.Propose(proposal, td.Actors.Alice.PrivateKey)
```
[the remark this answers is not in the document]
andrewgordstewart: You can actually pass in any `LedgerOutcome` to `NewConsensusChannel` (which may not be great in production, but can be used here instead of this hack?)

channel/consensus_channel/consensus_channel.go
```diff
@@ -161,6 +161,11 @@ type Guarantee struct {
 	right  types.Destination
 }
 
+// NewGuarantee constructs a new guarantee
+func NewGuarantee(amount *big.Int, target types.Destination, left types.Destination, right types.Destination) Guarantee {
+	return Guarantee{amount, target, left, right}
+}
```
geoknee, line 167: My first thought was: why? Calling this function isn't any easier than constructing a literal `Guarantee`. But then, thinking about it: using the constructor forces you into creating a "rich" `Guarantee` (you can't miss out a field). And I guess that is your intention, here?

geoknee: It could be a pattern that we want to spread around the codebase.

Tempting to go even further and make `Guarantee` => `guarantee` to prevent creating literal types. I wonder if that would have other adverse consequences, though?

geoknee (approved):

Issue #417, opened 2022-03-28
Table tests for leader channel


andrewgordstewart:
General comment: I found this test pretty hard to read. I think that is because test data, helper functions, API calls  and assertions aren't organised very well. 

I'm wondering if a table test or something would help? Something which looks a little more like this when I scan it:

```
0[ ] + Propose(1) = 0[1]
0[1] + Propose(2) = 0[1,2]
0[1,2] + Propose(3) = 0[1,2,3]
0[1,2,3] + UpdateConsensus(2) = 2[3]
2[3] + UpdateConsensus(3) = 3[ ]
```

Here I'm using a notation like `consensusTurnNum.ProposalQueue` (I think you get the idea).

_Originally posted by @geoknee in https://github.com/statechannels/go-nitro/pull/410#discussion_r836384672_

geoknee:
Here's an example of an existing table test https://github.com/statechannels/go-nitro/blob/main/channel/consensus_channel/serde_test.go

PR #465, opened 2022-03-31
Unite leader & follower ConsensusChannels

### LINKED ISSUE #452: Is it a problem that`*big.Int` gets marshalled to a plain old number?
Interesting that `*big.Int` gets marshalled to a plain old number. We should think about whether that may cause us problems when the numbers are big! It may only be an issue when wanting to use the JSON in (say) javascript.

_Originally posted by @geoknee in https://github.com/statechannels/go-nitro/pull/449#discussion_r839319962_

### LINKED ISSUE #462: Merge `LeaderChannel` and `FollowerChannel`
Having these two structs presents problems:
### during serde
when a ConsensusChannel is fetched from a mock store, how do you decide which type to assign it to?

### when being used by virtualfund

should a `virtualfund.Connection` hold a `LeaderChannel` _and_ a `FollowerChannel`?
```
type Connection struct {
  Leader LeaderChannel
  Follower FollowerChannel
  GuaranteeInfo GuaranteeInfo
}
```

In this case, exactly one of `Leader` and `Follower` should be not-nil -- ugly!

# Solution
In this case, it is sensible to merge Leader/Follower channel functionality into a single struct, removing these complications.

Both of these options make sense:
- a ConsensusChannel implements `Receive(sp SignedProposal)`, and `Receive` has conditional logic based on whether it's a leader or a follower channel
- a ConsensusChannel implements `UpdateConsensus` and `Receive` -- `UpdateConsensus` should return an error if it's not a leader channel, and likewise for `Receive`

## Question

If it didn't make sense to merge the functionality into a single struct, how could we resolve this puzzle?

NiloCK:
closes #462

This PR is slightly minimalist: it retains the separate files AND constructors for `leader` and `follower` channels, which are semantically useful.

It adds checks on each function call, plus a test assertion that each incorrect call fails with the expected `ErrNot(correctcaller)` error.

---

#### Code quality

#### Project management

### DIFF
--- a/channel/consensus_channel/consensus_channel.go
+++ b/channel/consensus_channel/consensus_channel.go
@@ -495,14 +495,14 @@ func CreateFromDirectFundingObjective(dfo directfund.Objective) (*ConsensusChann
 		if err != nil {
 			return nil, fmt.Errorf("could not create consensus channel as leader: %w", err)
 		}
-		return &con.ConsensusChannel, nil
+		return &con, nil
 
 	} else {
 		con, err := NewLeaderChannel(ledger.FixedPart, turnNum, outcome, signatures)
 		if err != nil {
 			return nil, fmt.Errorf("could not create consensus channel as follower: %w", err)
 		}
-		return &con.ConsensusChannel, nil
+		return &con, nil
 	}
 
 }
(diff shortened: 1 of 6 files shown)


geoknee:
> closes #452

Do you mean a different issue?

NiloCK:
> Do you mean a different issue?

Absolutely I do! Edited.

NiloCK:
> 



> Sweet!
> 
> What's great about this PR is:
> 
>     1. it's simple, and makes very little change
> 
>     2. Alice can now make progress on #420
> 
>     3. Simultaneously, Bob can refactor ConsensusChannel (eg. work on #417) without bothering Alice.
> 
> 
> The one niggle is that in (3), Bob might change the function signatures, which would bother Alice. Is there good way to guard against that?

No obvious way to me :/

PR #466, opened 2022-04-01
`cc.CreateFromDirectFundingObjective` => `dfo.CreateConsensusChannel`

geoknee:
See https://github.com/statechannels/go-nitro/pull/437#discussion_r840436118 and https://github.com/statechannels/go-nitro/pull/437#discussion_r840460237

PR #467, opened 2022-04-01
Execute side effects after storing ConsensusChannel and introduce SpawnConsensusChannelIfDirectFundObjective method

geoknee:
⚠️ stacked on #466 

See https://github.com/statechannels/go-nitro/pull/437#discussion_r840447242 and https://github.com/statechannels/go-nitro/pull/437#discussion_r840458644


NiloCK (commented):
:+1:

NiloCK (approved):

PR #469, opened 2022-04-01
Fix Leader/Follower typo

geoknee:
⚠️ stacked on #467 

See https://github.com/statechannels/go-nitro/pull/437#discussion_r840565692

Ideally I would have included a failing test to exhibit the bug. But I couldn't find an easy way to do that (it would have surfaced eventually in a virtual fund integration test).


NiloCK (approved):
:+1:

PR #475, opened 2022-04-02
Add SignedProposals to Message 

kerzhner:
This PR works toward https://github.com/statechannels/go-nitro/issues/420.

`SignedProposals` field in the `Message` struct might not be the long term solution but perhaps good enough for now. Only a certain type of message needs the `SignedProposal` field populated (messages that are related to a consensus channel). But all messages will contain the field.

### DIFF
--- a/protocols/messages.go
+++ b/protocols/messages.go
@@ -35,7 +35,7 @@ func CreateSignedStateMessages(id ObjectiveId, ss state.SignedState, myIndex uin
 	return createMessages(id, &ss, nil, myIndex)
 }
 
-// CreateSignedProposalMessages creates an list of messages containing the signed proposal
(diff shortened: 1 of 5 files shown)


protocols/messages.go
```diff
@@ -72,3 +95,16 @@ func (se *SideEffects) Merge(other SideEffects) {
 	se.TransactionsToSubmit = append(se.TransactionsToSubmit, other.TransactionsToSubmit...)
 
 }
+
+type comparable[T any] interface {
```
lalexgap, line 99: I just merged in https://github.com/statechannels/go-nitro/pull/383 so I think most of the `Equal` work in here can be removed 🎉

lalexgap (approved):

Issue #479, opened 2022-04-04
Write Integration test for virtual defunding


lalexgap:

lalexgap:
This has been completed in https://github.com/statechannels/go-nitro/pull/618

PR #484, opened 2022-04-04
Refactor leader_channel tests

### LINKED ISSUE #417: Table tests for leader channel

andrewgordstewart:
Resolves #417.

It doesn't use table tests: there are a few different ways to set up and call expectations on the outcomes. So, the tests are generally cleaned up and made more readable.

This refactor was done with #479 in mind, though I expect the tests might still change a little bit.

### DIFF
--- a/channel/consensus_channel/leader_channel_test.go
+++ b/channel/consensus_channel/leader_channel_test.go
@@ -311,7 +311,7 @@ func TestLeaderChannel(t *testing.T) {
 			if !errors.Is(err, expectedErr) {
 				t.Fatalf("expected error %v, got %v", expectedErr, err)
 			}
-			
+
 			if currentTurnNum != channel.ConsensusTurnNum() {
 				t.Fatalf("consensus changed in error case")
 			}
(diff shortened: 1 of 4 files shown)


channel/consensus_channel/leader_channel_test.go
```diff
@@ -2,182 +2,367 @@ package consensus_channel
...
-	outcomeSigned := makeOutcome(
-		allocation(alice, aBal-amountAdded),
-		allocation(bob, bBal),
-		guarantee(vAmount, existingChannel, alice, bob),
-		guarantee(amountAdded, targetChannel, alice, bob),
-	)
-	stateSigned := Vars{TurnNum: 1, Outcome: outcomeSigned}
-	sig, _ := stateSigned.AsState(fp()).Sign(alice.PrivateKey)
-	expected := SignedProposal{Proposal: p, Signature: sig}
+	// aliceSignedProposal generates a proposal given the vars & proposed change
```
lalexgap, line 44:
```suggestion
	// createSignedProposal generates a proposal given the vars & proposed change
```

lalexgap (approved):

Issue #485, opened 2022-04-04
Run nitro-protocol tests in CI


andrewgordstewart:
On https://github.com/statechannels/go-nitro/pull/427, tests do not run in CI. We should add a GitHub action to build `nitro-protocol` and run its tests.

PR #487, opened 2022-04-05
Onchain.gha

geoknee:
Trying out a github action for #485


geoknee:
It would be nice to skip this new GH action in the case where `nitro-protocol` was not changed.

NiloCK:
> It would be nice to skip this new GH action in the case where `nitro-protocol` was not changed.

An example. In our case `paths` would be `nitro-protocol` (if my first glance is correct), and the workflow would run only when there are changes in that path.

```
on:
  push:
    branches:
      - deploy
    paths:
      - 'packages/express/**'
```

.github/workflows/node.yml
```diff
@@ -0,0 +1,20 @@
+name: Node
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+
```
NiloCK, line 7: (untested)

```suggestion
on:
  push:
    branches: [main]
    paths: "nitro-protocol/**"
  pull_request:
    paths: "nitro-protocol/**"

```

geoknee: I test this manually, it seems to work [5328ee2](https://github.com/statechannels/go-nitro/pull/487/commits/5328ee20430c13101b5693fd6dff08368dd09ea6) !

geoknee:
> Could we skip the failing test temporarily, so that we can merge this with a ✅ ?
> 
> e: n/m, just saw [#495 (review)](https://github.com/statechannels/go-nitro/pull/495#pullrequestreview-935293255)

@andrewgordstewart I merged #495 and rebased this. So now we have our ✅ .

PR #493, opened 2022-04-06
Add consensus_channel logic to `virtualfund.Objective.Update`

NiloCK:
incremental toward #420

This PR introduces some processing of `consensus_channel.SignedProposal` messages in the virtualfund `Update()` method.

The PR adds:
- a range loop in virtualfund.Objective.Update which handles incoming signed proposals
- a `handleProposal` function to the `Connection` struct (VFO's interface with ledger channels)
- a forwarding `GetTwoPartyConsensusLedgerFunction` to virtualfund.NewObjective(...)
- identity helper functions `IsLeader`, `IsFollower`, `Follower` to ConsensusChannel
- `testdata` mocking function for finding ConsensusChannels

2dece03ade9df212325a21f5c842f09620a6df7f is unrelated to the PR. A drive-by change of a godoc.

## :warning: broken tests

test breakage is expected here because consensus channels are not yet serializing / deserializing in mockstore. Because of this, a deserialized `vfo` ends up with `nil` ConsensusChannels where it expects non-nil ones.

---

#### Code quality

  - testing blocked by storing / retrieving updated consensus channels

#### Project management

### DIFF
--- a/protocols/virtualfund/virtualfund.go
+++ b/protocols/virtualfund/virtualfund.go
@@ -75,6 +75,28 @@ func (c *Connection) insertGuaranteeInfo(a0 types.Funds, b0 types.Funds, vId typ
 	return nil
 }
 
+// handleProposal recieves a signed proposal and acts according to the leader / follower
+// status of the Connection's ConsensusChannel
+func (c *Connection) handleProposal(sp consensus_channel.SignedProposal) error {
+	if c == nil {
+		return fmt.Errorf("nil connection should not handle proposals")
+	}
+
+	if sp.Proposal.ChannelID != c.ConsensusChannel.Id {
+		return consensus_channel.ErrIncorrectChannelID
+	}
+
+	if c.ConsensusChannel.IsFollower() {
+		return c.ConsensusChannel.Receive(sp)
+	}
+
+	if c.ConsensusChannel.IsLeader() {
+		return c.ConsensusChannel.UpdateConsensus(sp)
+	}
+
+	return nil
+}
+
 // getExpectedGuarantees returns a map of asset addresses to guarantees for a Connection.
 func (c *Connection) getExpectedGuarantees() map[types.Address]outcome.Allocation {
 	expectedGuaranteesForLedgerChannel := make(map[types.Address]outcome.Allocation)
--- a/protocols/virtualfund/virtualfund.go
+++ b/protocols/virtualfund/virtualfund.go
@@ -304,17 +304,35 @@ func (o Objective) Update(event protocols.ObjectiveEvent) (protocols.Objective, 
 
 	var toMyLeftId types.Destination
 	var toMyRightId types.Destination
+	var ccLeftID types.Destination
+	var ccRightID types.Destination
 
 	if !o.isAlice() {
 		toMyLeftId = o.ToMyLeft.Channel.Id // Avoid this if it is nil // todo: #420 deprecate
-		// toMyLeftId = o.ToMyLeft.ConsensusChannel.Id
+		ccLeftID = o.ToMyLeft.ConsensusChannel.Id
 	}
 	if !o.isBob() {
 		toMyRightId = o.ToMyRight.Channel.Id // Avoid this if it is nil // todo: #420 deprecate
-		// toMyRightId = o.ToMyRight.ConsensusChannel.Id
+		ccRightID = o.ToMyRight.ConsensusChannel.Id
 	}
 
-	// todo: #420 range over event.Proposals (or similar)
+	for _, sp := range event.SignedProposals {
+		var err error
+		switch sp.Proposal.ChannelID {
+		case types.Destination{}:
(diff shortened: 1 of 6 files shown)


protocols/virtualfund/virtualfund.go
```diff
@@ -75,6 +75,28 @@ func (c *Connection) insertGuaranteeInfo(a0 types.Funds, b0 types.Funds, vId typ
...
+func (c *Connection) handleProposal(sp consensus_channel.SignedProposal) error {
+	if c == nil {
+		return fmt.Errorf("nil connection should not handle proposals")
+	}
+
+	if sp.Proposal.ChannelID != c.ConsensusChannel.Id {
+		return consensus_channel.ErrIncorrectChannelID
+	}
+
+	if c.ConsensusChannel.IsFollower() {
```
kerzhner, line 89: Seems like `consensus_channel` has enough information internally to determine how to react to a signed proposal. Would it be better to encapsulate this `isFollower`/`isLeader` logic inside the `consensus_channel` struct logic (instead of exposing this logic to the virtual fund protocol)?

kerzhner (approved):

andrewgordstewart:
Let's recall our strategy:
1. incorporate ConsensusChannel logic in virtualfund, asserting its correctness against the behaviour of TwoPartyLedger channels
2. THEN, remove TwoPartyLedger channels

This PR aims to do (1), but mixes in some of (2).

I would suggest removing:
- https://github.com/statechannels/go-nitro/pull/493/commits/4775ad308278c3dbeb83b08881f31956902dcaeb
- https://github.com/statechannels/go-nitro/pull/493/commits/282aea7b4ac30811455eec6f0d3e025912afa7cb
- https://github.com/statechannels/go-nitro/pull/493/commits/cf2899dbbd1dfe71fc48e9c4af805deda1cacdfd
which are all related to the _engine_, and impact integration tests.

After merging https://github.com/statechannels/go-nitro/pull/494, one person can continue to make progress on part (1).

If we want to, these three commits can be added independently by another developer in preparation for (2).

NiloCK:
> 



> Let's recall our strategy:
> 
>     1. incorporate ConsensusChannel logic in virtualfund, asserting its correctness against the behaviour of TwoPartyLedger channels
> 
>     2. THEN, remove TwoPartyLedger channels
> 
> 
> This PR aims to do (1), but mixes in some of (2).
> 
> I would suggest removing:
> 
>     * [4775ad3](https://github.com/statechannels/go-nitro/commit/4775ad308278c3dbeb83b08881f31956902dcaeb)
> 
>     * [282aea7](https://github.com/statechannels/go-nitro/commit/282aea7b4ac30811455eec6f0d3e025912afa7cb)
> 
>     * [cf2899d](https://github.com/statechannels/go-nitro/commit/cf2899dbbd1dfe71fc48e9c4af805deda1cacdfd)
> 
> 
> which are all related to the _engine_, and impact integration tests.
> 
> After merging #494, one person can continue to make progress on part (1).
> 
> If we want to, these three commits can be added independently by another developer in preparation for (2).

Having trouble understanding this!
- I don't see any removal of TwoPartyLedger channels in the highlighted commits
- How can side-by-side implementations be compared without including ConsensusChannel in the integration tests?

andrewgordstewart:
> Having trouble understanding this!
> 
> * I don't see any removal of TwoPartyLedger channels in the highlighted commits (or on this PR)

My expectation is that TwoPartyLedger channels would be removed after ConsensusChannel is fully integrated and fully working. That would be a few PRs down the road!

> incorporate ConsensusChannel logic in virtualfund, asserting its correctness against the behaviour of TwoPartyLedger channels

What this means to me is, we update `Crank` and `Update` to utilize a `ConsensusChannel` alongside a `TwoPartyLedger`. We can run [assertions](https://github.com/statechannels/go-nitro/compare/cmk.vfo-consensus-update...ags/vfo-assertions?expand=1) that they maintain the same outcome & turn number on their latest state and consensus state.

> * How can side-by-side implementations be compared without including ConsensusChannel in the integration tests?

We can use unit tests! In #494, changes are made to allow the `virtual fund-single-hop-test` to run (and pass) with a ConsensusChannel. When an integration test runs, the `ConsensusChannel` on each connection is `nil`, and the code is bypassed. (If we were doing this work on code running in production, we would have to use a feature flag here to bypass the still-in-progress change in production.)

Unfortunately, I am not sure that the existing unit tests are very amenable to this strategy:
- The `testUpdate` function sends an [arbitrary signed state](https://github.com/statechannels/go-nitro/blob/3956d5ac214bbaf6345d01268428e762457856a1/protocols/virtualfund/virtualfund-single-hop_test.go#L570-L596) destined for the` TwoPartyLedger`, which the `Update` function accepts (**is this a bug?**).
- Then, it checks that this arbitrary state was received
- Because `LatestSignedState` doesn't actually return the latest signed state (it should return this arbitrary state), [this sanity check passes](https://github.com/statechannels/go-nitro/compare/cmk.vfo-consensus-update...ags/vfo-assertions?expand=1#diff-e78ba83d3714e679fa001e0cc48267b733ce93be39eb4ab1a37a777d25f6ca57R387) even though it should fail

Issue #499, opened 2022-04-07
Update VariablePart to match off-chain code


geoknee:
@nksazonov This struct needs updating, too, I think?

Take a look at 744dbbef0ef22ba8085d60b8b0a8b468f038bc16 and 3b4bfbae06a4e5942ca2d88d80c2af1cea15740c

_Originally posted by @geoknee in https://github.com/statechannels/go-nitro/pull/427#discussion_r843108788_

nksazonov:
@geoknee why don't also add `turnNum` and `isFinal` fields to `VariablePart` as was done in [`744dbbe`](https://github.com/statechannels/go-nitro/commit/744dbbef0ef22ba8085d60b8b0a8b468f038bc16) you have sent?

This would make the code significantly clearer and function signatures would resemble ones from SATP paper more.

nksazonov:
I am thinking about adding `exitsEqual` method to `ExitFormat` to enclose outcome comparisons (as in [`3b4bfba`](https://github.com/statechannels/go-nitro/commit/3b4bfbae06a4e5942ca2d88d80c2af1cea15740c#diff-0f9a53252feac4825a0d3108507d0fd4aad893bd91361bb492e9cdf74fb5fb77R46)).
It would be a bit more gas efficient and a clearer way (without invoking [5 functions in one line](https://github.com/statechannels/go-nitro/commit/3b4bfbae06a4e5942ca2d88d80c2af1cea15740c#diff-0f9a53252feac4825a0d3108507d0fd4aad893bd91361bb492e9cdf74fb5fb77R46)).

@geoknee what do you think?

geoknee:
> I am thinking about adding `exitsEqual` method to `ExitFormat`

Yes I think this is a good idea. I suppose we can split your suggestion into two components:
1. Refactor the "encode and hash" equality check into a helper that lives in `ExitFormat`.
2. Change the implementation to be more gas efficient.

nksazonov:
> I did something like 1 (but kept it in this repo) on a branch I have been working on: https://github.com/statechannels/go-nitro/compare/go-solidity-signature-test?expand=1#diff-483aad04b1f90d19c78f0fd86f7ebb7869c6c194797c3d0774ce9e80dff5a2b8R632-R647

This link leads to a PR, unfortunately not a difference comparison as it might suggest. 
Could you specify the commit, file and lines, please?

geoknee:
Yes, sorry, I updated the link to something more useful!

[...]

Issue #496, opened 2022-04-07
Migrate from `validTransition` to `latestSupportedState`


geoknee:
Trying to clarify what is need to be changed related to `latestSupportedState` implementation. As I understand from SATP paper, all state transition checks should be done in it in ForceMoveApp. If so, should `latestSupportedState` replace [`_requireStateSupportedBy`](https://github.com/statechannels/statechannels/blob/master/packages/nitro-protocol/contracts/ForceMove.sol#L488) logic currently implemented in ForceMove?

_Originally posted by @nksazonov in https://github.com/statechannels/go-nitro/issues/427#issuecomment-1085609335_

- [x] #557
- [x] #558
- [x] #559

geoknee:
I'm splitting this issue down into more manageable chunks. 

- [ ] modifying the `ForceMoveApp` interface and integrating into `ForceMove.sol`
- [ ] rewriting our example applications to implement the new interface & updating unit tests
- [ ] writing a VirtualChannel application plus unit tests

PR #508, opened 2022-04-08
Update VariablePart to match off-chain code

nksazonov:
Relates to #499

### Changes

Update `VariablePart` struct to include:
- [x] `outcome` as `SingleAssetExit[]`
- [x] `turnNum` and `isFinal` flag 

Update function logic (majorly in `ForceMove.sol`) to support aforementioned changes:
- [x] support changing `outcome` as `SingleAssetExit[]`
- [x] support adding `turnNum` and `isFinal` flag 

---

#### Code quality

#### Project management

### DIFF
--- a/nitro-protocol/contracts/ForceMove.sol
+++ b/nitro-protocol/contracts/ForceMove.sol
@@ -105,7 +105,7 @@ contract ForceMove is IForceMove, StatusManager {
                 largestTurnNum,
                 uint48(block.timestamp) + fixedPart.challengeDuration, //solhint-disable-line not-rely-on-time
                 supportedStateHash,
-                keccak256(variableParts[variableParts.length - 1].outcome)
+                keccak256(Outcome.encodeExit(variableParts[variableParts.length - 1].outcome))
             )
         );
     }
@@ -134,7 +134,7 @@ contract ForceMove is IForceMove, StatusManager {
         bytes32 challengeStateHash = _hashState(
             channelId,
             variablePartAB[0].appData,
-            variablePartAB[0].outcome,
+            Outcome.encodeExit(variablePartAB[0].outcome),
             turnNumRecord,
             isFinalAB[0]
         );
@@ -142,14 +142,14 @@ contract ForceMove is IForceMove, StatusManager {
         bytes32 responseStateHash = _hashState(
             channelId,
             variablePartAB[1].appData,
-            variablePartAB[1].outcome,
+            Outcome.encodeExit(variablePartAB[1].outcome),
             turnNumRecord + 1,
             isFinalAB[1]
         );
 
         // checks
 
-        bytes32 challengeOutcomeHash = keccak256(variablePartAB[0].outcome);
+        bytes32 challengeOutcomeHash = keccak256(Outcome.encodeExit(variablePartAB[0].outcome));
 
         _requireSpecificChallenge(
             ChannelData(turnNumRecord, finalizesAt, challengeStateHash, challengeOutcomeHash),
@@ -542,7 +542,7 @@ contract ForceMove is IForceMove, StatusManager {
             stateHashes[i] = _hashState(
                 channelId,
                 variableParts[i].appData,
-                variableParts[i].outcome,
+                Outcome.encodeExit(variableParts[i].outcome),
                 turnNum,
                 turnNum >= firstFinalTurnNum
             );
@@ -581,11 +581,11 @@ contract ForceMove is IForceMove, StatusManager {
         // chainId, participants, channelNonce, appDefinition, challengeDuration
         // and that the b.turnNum = a.turnNum + 1
         if (isFinalAB[1]) {
-            require(_bytesEqual(ab[1].outcome, ab[0].outcome), 'Outcome change verboten');
+            require(Outcome.exitsEqual(ab[1].outcome, ab[0].outcome), 'Outcome change verboten');
         } else {
             require(!isFinalAB[0], 'isFinal retrograde');
             if (turnNumB < 2 * nParticipants) {
-                require(_bytesEqual(ab[1].outcome, ab[0].outcome), 'Outcome change forbidden');
+                require(Outcome.exitsEqual(ab[1].outcome, ab[0].outcome), 'Outcome change forbidden');
                 require(_bytesEqual(ab[1].appData, ab[0].appData), 'appData change forbidden');
             } else {
                 return IsValidTransition.NeedToCheckApp;
--- a/nitro-protocol/contracts/ForceMove.sol
+++ b/nitro-protocol/contracts/ForceMove.sol
@@ -48,9 +48,9 @@ contract ForceMove is IForceMove, StatusManager {
     function challenge(
         FixedPart memory fixedPart,
         IForceMoveApp.VariablePart[] memory variableParts,
-        Signature[] memory sigs,
-        uint8[] memory whoSignedWhat,
-        Signature memory challengerSig
+        Signature[] calldata sigs,
+        uint8[] calldata whoSignedWhat,
+        Signature calldata challengerSig
     ) external override {
         // input type validation
         requireValidInput(
@@ -117,7 +117,7 @@ contract ForceMove is IForceMove, StatusManager {
         IForceMoveApp.VariablePart[2] memory variablePartAB,
         // variablePartAB[0] = challengeVariablePart
         // variablePartAB[1] = responseVariablePart
-        Signature memory sig
+        Signature calldata sig
     ) external override {
         // No need to validate fixedPart.participants.length here, as that validation would have happened during challenge
 
@@ -176,8 +176,8 @@ contract ForceMove is IForceMove, StatusManager {
     function checkpoint(
         FixedPart memory fixedPart,
         IForceMoveApp.VariablePart[] memory variableParts,
-        Signature[] memory sigs,
-        uint8[] memory whoSignedWhat
+        Signature[] calldata sigs,
+        uint8[] calldata whoSignedWhat
     ) external override {
         // input type validation
         requireValidInput(
@@ -218,8 +218,8 @@ contract ForceMove is IForceMove, StatusManager {
         FixedPart memory fixedPart,
         IForceMoveApp.VariablePart memory latestVariablePart,
         uint8 numStates,
-        uint8[] memory whoSignedWhat,
-        Signature[] memory sigs
+        uint8[] calldata whoSignedWhat,
+        Signature[] calldata sigs
     ) external override {
         _conclude(
             fixedPart,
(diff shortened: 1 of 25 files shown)


nksazonov:
Have a few doubts about the following:
- should a function for returning the last element of `VariablePart[]` be added? (e.g. for usage [here](https://github.com/statechannels/go-nitro/pull/508/files#diff-483aad04b1f90d19c78f0fd86f7ebb7869c6c194797c3d0774ce9e80dff5a2b8R91) and [here](https://github.com/statechannels/go-nitro/pull/508/files#diff-483aad04b1f90d19c78f0fd86f7ebb7869c6c194797c3d0774ce9e80dff5a2b8R191)) . The benefit from it is clearer code (repetitions are in a separate function), but the drawback is slightly increased gas usage (by 100 gas).

nitro-protocol/contracts/NitroAdjudicator.sol
```diff
@@ -13,34 +13,28 @@ contract NitroAdjudicator is ForceMove, MultiAssetHolder {
...
-            appData,
-            outcomeBytes,
+            latestVariablePart,
             numStates,
             whoSignedWhat,
             sigs
         );
 
-        transferAllAssets(channelId, outcomeBytes, bytes32(0));
+        transferAllAssets(channelId, Outcome.encodeExit(latestVariablePart.outcome), bytes32(0));
```
geoknee, line 37: There are different options here, but to me it seems slightly nicer to:

* pass the `latestVariablePart.outcome`
* have `transferAllAssets` encode and hash it

(rather than encoding it here, passing the bytes, and having `transferAllAssets` hash it and decode it).

geoknee:
> * update `_hashState` to be called with `SingleAssetExit[] outcome`

We should tackle this in a follow-up PR. When this one is merged, we can make a new issue to track that. 

> * should a function for returning the last element of `VariablePart[]` be added? (e.g. for usage [here](https://github.com/statechannels/go-nitro/pull/508/files#diff-483aad04b1f90d19c78f0fd86f7ebb7869c6c194797c3d0774ce9e80dff5a2b8R91) and [here](https://github.com/statechannels/go-nitro/pull/508/files#diff-483aad04b1f90d19c78f0fd86f7ebb7869c6c194797c3d0774ce9e80dff5a2b8R191)) . The benefit from it is clearer code (repetitions are in a separate function), but the drawback is slightly increased gas usage (by 100 gas).

This could also be tackled in a follow-up PR. The links you gave don't seem to point to a particular line, so I'm not sure exactly where you propose to use such a function. In general I would say it is worth a little bit of gas to improve the quality of the code. 

This is reminding me that we should reintroduce some tracking of gas costs in CI, to make these decisions easier to make.

geoknee:
@nksazonov this looks great to me! We need to do a bit more work before merging, however:

1. We need to get the CI tests to run on this branch. If you're comfortable, I would suggest rebasing this branch on `main` (which now includes #487). Then we should get a second github action running. 
2. I expect that to fail since this PR changes the contract API and we didn't update the tests yet. I'm afraid it will be a bit laborious :-/ . Happy to help out with that work, though.

nitro-protocol/contracts/interfaces/IForceMove.sol
```diff
@@ -35,78 +35,64 @@ interface IForceMove {
...
+     * @param variableParts An ordered array of structs, each decribing the properties of the state channel that may change with each state update. Length is from 1 to the number of participants (inclusive).
+     * @param sigs An array of signatures that support the state with the `largestTurnNum`. There must be one for each participant, e.g.: [sig-from-p0, sig-from-p1, ...]
      * @param whoSignedWhat An array denoting which participant has signed which state: `participant[i]` signed the state with index `whoSignedWhat[i]`.
      * @param challengerSig The signature of a participant on the keccak256 of the abi.encode of (supportedStateHash, 'forceMove').
      */
     function challenge(
-        FixedPart calldata fixedPart,
-        uint48 largestTurnNum,
-        IForceMoveApp.VariablePart[] calldata variableParts,
-        uint8 isFinalCount, // how many of the states are final
```
geoknee, line 49: Glad to see this go!

nitro-protocol/contracts/interfaces/IForceMove.sol
```diff
@@ -35,78 +35,64 @@ interface IForceMove {
...
     function conclude(
-        uint48 largestTurnNum,
-        FixedPart calldata fixedPart,
-        bytes memory appData,
-        bytes memory outcome,
+        FixedPart memory fixedPart,
+        IForceMoveApp.VariablePart memory latestVariablePart,
         uint8 numStates,
-        uint8[] calldata whoSignedWhat,
-        Signature[] calldata sigs
```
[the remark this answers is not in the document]
nksazonov: Sure

nksazonov:
> > update _hashState to be called with SingleAssetExit[] outcome

> We should tackle this in a follow-up PR. When this one is merged, we can make a new issue to track that.

Has not noticed it on time. Let me revert these changes and force-push.

nitro-protocol/contracts/NitroAdjudicator.sol
nksazonov: This way looks better to me too.

nitro-protocol/contracts/interfaces/IForceMove.sol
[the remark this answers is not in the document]
nksazonov: Adding `calldata` to `FixedPart` or `IForceMoveApp.VariablePart` and compilation results in `UnimplementedFeatureError:`. This is pretty strange, because `calldata` is present in current implementation of on-chain statechannels part in monorepo and compiling it finishes successfully.

nksazonov:
> 2. I expect that to fail since this PR changes the contract API and we didn't update the tests yet. I'm afraid it will be a bit laborious :-/ . Happy to help out with that work, though.

I will start working on changing tests today then.

nitro-protocol/contracts/interfaces/IForceMove.sol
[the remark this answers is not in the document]
geoknee: I can't remember, possibly! Happy to just go with `memory` for now if it unblocks us. We can make an issue to circle back to it.

nksazonov:
The only thing left now is to update `EmbeddedApp.sol` to have the same `VariablePart` struct and test related to it.

geoknee:
> The only thing left now is to update `EmbeddedApp.sol` to have the same `VariablePart` struct and tests related to it.

I would be in favour of simply removing or disabling `EmbeddedApp.sol` and tests, for now, because:

* We can reinstate that code as a part of #496 
  * The `EmbeddedApp` could become the `VirtualChannelApp`
  * It might be easier to write from scratch, possibly following Andrew's example https://github.com/statechannels/go-nitro/pull/427#issuecomment-1086025068
* It would be good to get the changes on this PR merged quickly

nksazonov:
> I would be in favour of simply removing or disabling EmbeddedApp.sol and tests, for now

Okay, then I will just tune the tests for `EmbeddedApp.sol`.

nksazonov:
@geoknee all tests are now passing.

nitro-protocol/test/contracts/ForceMove/challenge.test.ts
```diff
@@ -169,6 +170,9 @@ describe('challenge', () => {
       const variableParts = states.map(state => getVariablePart(state));
       const fixedPart = getFixedPart(states[0]);
       const channelId = getChannelId(fixedPart);
+      console.log("variableParts");
+      console.log(variableParts);
```
geoknee, line 174: ✂️

nitro-protocol/contracts/examples/HashLockedSwap.sol
```diff
@@ -17,11 +17,10 @@ contract HashLockedSwap is IForceMoveApp {
     function validTransition(
         VariablePart memory a,
         VariablePart memory b,
-        uint48 turnNumB,
         uint256
     ) public override pure returns (bool) {
         // is this the first and only swap?
-        require(turnNumB == 4, 'turnNumB != 4');
+        require(b.turnNum == 4, 'turnNumB != 4');
```
geoknee, line 23: Nit: 
```suggestion
        require(b.turnNum == 4, 'b.turnNum != 4');
```

nitro-protocol/src/contract/state.ts
```diff
@@ -53,7 +55,12 @@ export interface VariablePart {
  * @returns the VariablePart, which usually changes during state channel updates
  */
 export function getVariablePart(state: State): VariablePart {
-  return {outcome: encodeOutcome(state.outcome), appData: encodeAppData(state.appData)};
+  return {
+    outcome: state.outcome,
+    appData: encodeAppData(state.appData),
```
geoknee, line 60: Outside the scope of this PR -- This seems a little strange, to me. I wouldn't expect we need to re-encode `appData` when extracting it from the state?

nitro-protocol/test/contracts/ForceMove/challenge.test.ts
```diff
@@ -44,7 +45,7 @@ const chainId = process.env.CHAIN_NETWORK_ID;
 const participants = ['', '', ''];
 const wallets = new Array(3);
 const challengeDuration = 86400; // 1 day
-const outcome: Outcome = [{allocations: [], asset: Wallet.createRandom().address, metadata: '0x'}];
+const outcome: Outcome = [{allocations: [{destination: "0x00000000000000000000000096f7123e3a80c9813ef50213aded0e4511cb820f", amount: "0x01", allocationType: 1, metadata: "0x"}], asset: Wallet.createRandom().address, metadata: '0x'}];
```
geoknee, line 48: Was this change necessary?

nitro-protocol/contracts/ForceMove.sol
```diff
@@ -134,22 +127,22 @@ contract ForceMove is IForceMove, StatusManager {
...
+            Outcome.encodeExit(variablePartAB[1].outcome),
             turnNumRecord + 1,
-            isFinalAB[1]
+            variablePartAB[1].isFinal
         );
 
         // checks
 
-        bytes32 challengeOutcomeHash = keccak256(variablePartAB[0].outcome);
+        bytes32 challengeOutcomeHash = keccak256(Outcome.encodeExit(variablePartAB[0].outcome));
```
geoknee, line 145: Outside the scope of this PR: might be nice to have a `hashOutcome` helper?

geoknee (approved):
👏

nitro-protocol/test/contracts/ForceMove/challenge.test.ts
nksazonov: It was done for testing purposes. Removed in [`0fea82b`](https://github.com/statechannels/go-nitro/pull/508/commits/0fea82b1634e81ba892dad33ad304ae4d36db885)

nitro-protocol/contracts/examples/HashLockedSwap.sol
nksazonov: Implemented in [`894e94d`](https://github.com/statechannels/go-nitro/pull/508/commits/894e94db7ea9d87ae2ce9118feb144561abe920c)

nitro-protocol/test/contracts/ForceMove/challenge.test.ts
nksazonov: Removed in [`0fea82b`](https://github.com/statechannels/go-nitro/pull/508/commits/0fea82b1634e81ba892dad33ad304ae4d36db885)

nitro-protocol/contracts/ForceMove.sol
nksazonov: Good idea.

nitro-protocol/src/contract/state.ts
nksazonov: If I remember correctly, this was added to be correspond to `go-nitro` state encoding.

PR #512, opened 2022-04-08
Remove `channel.SingleHopVirtualChannel`  from Virtual Defunding 

### LINKED ISSUE #511: Update VirtualDefund to Store Channel info on Objective
Since we're writing a new objective here, how would you feel about trying out NOT embedding `V`?

Here's the rough requirements:
1. Alice (could be Bob) calls `closeVirtualChannel(cId, paid: big.Int})`
2. Alice's engine needs to create a virtual-defund-objective (VDO) that knows how to construct a final state
3. Alice's engine sends some message to Bob, Irene indicating that they want to close `V`. (What should this message look like?)
4. The VDO brokers the exchange of signatures on the final state
5. Then moves through the ledger-defunding process.

(4) is easy to do without needing to store `V`, since the final state is a function of `V`'s starting outcome + `paid` + fixed part. So, the only thing we'd need to do in step (2) is store `V`'s fixed part + initial outcome, and the only thing we need to store in step 4 is some signatures. Something like:
```
type Objective struct {
 	Status protocols.ObjectiveStatus

 	VFixedPart state.FixedPart
        Deposits struct{Alice, Bob big.Int}
        AmountPaid big.int
        FinalStateSignatures [3]state.Signature

       ToMyLeft  *consensus_channel.ConsensusChannel
 	ToMyRight *consensus_channel.ConsensusChannel

 	MyRole uint
}
```

Doing it this way has some benefits:
1. unit tests for the VDO are easier to write (you don't need to construct a valid `V`, and you have one fewer dependency to worry about)
2. there's one fewer relation, which means: one less entry in `Related()`, cloned, json marshalled, diffed, etc. these things add up over time!
3. (theoretical) there would be queries required against a store during the defunding process, if we were actually querying against a store
4. code is more explicit: rather than asking "is the latest supported state in `V` final, and does it have the correct outcome and turn number?" the VDO would ask "do I have three (valid) signatures on the final state?"

To expand on (4), I'll look at the implications of using `V` in the virtualfund objective for comparison.

Currently, `V` could contain arbitrary states when the VDO starts:
A. the virtualfund objective objective stores [all incoming states](https://github.com/statechannels/go-nitro/blob/c1e82e07fcd599581a77a1edcf767202159a6bc8/protocols/virtualfund/virtualfund.go#L363-L364) for `V`
B. it completes successfully when `V` has a [full set of signatures on the post fund state](https://github.com/statechannels/go-nitro/blob/c1e82e07fcd599581a77a1edcf767202159a6bc8/channel/channel.go#L182). This check 

(A) means it is easy for Eve to slip arbitrary states into Alice's store. (B) means the state of `V` could be fairly arbitrary. Hopefully Alice didn't **sign** any of the malicious states added in (A):
- if she did, then her funds are at risk
- if she didn't, then we have to be careful in the VDO to **only sign the correct final state**. This means we have to be careful with malicious `TurnNum == 3` states that might have been stored by `V` in (A).

I feel that a lot of this care could be avoided by using explicit data structures that store exactly the data we expect, and reject any other data. Since we're writing a new objective here, it seems we could more easily try out this pattern and make a judgement call part way through.

What do you think?

_Originally posted by @andrewgordstewart in https://github.com/statechannels/go-nitro/pull/503#discussion_r846321196_

lalexgap:
Fixes #511 

Updates the Virtual Defunding Objective to store the information on the objective instead of relying on a `channel.SingleHopVirtualChannel`

We make some simplifying assumptions:
- Every virtual channel will be single hop virtual channel
- Payments are always going from `alice` to `bob`

### DIFF
--- a/protocols/virtualdefund/virtualdefund.go
+++ b/protocols/virtualdefund/virtualdefund.go
@@ -115,7 +115,7 @@ func (o *Objective) clone() Objective {
 
 	clone.Signatures = [3]state.Signature{}
 	for i, s := range o.Signatures {
-		clone.Signatures[i] = s
+		clone.Signatures[i] = state.CloneSignature(s)
 	}
 	clone.MyRole = o.MyRole
 
--- a/protocols/virtualdefund/virtualdefund.go
+++ b/protocols/virtualdefund/virtualdefund.go
@@ -3,12 +3,12 @@ package virtualdefund
 import (
 	"errors"
 	"fmt"
+	"math/big"
 
 	"github.com/statechannels/go-nitro/channel/consensus_channel"
 	"github.com/statechannels/go-nitro/channel/state"
 	"github.com/statechannels/go-nitro/channel/state/outcome"
 	"github.com/statechannels/go-nitro/protocols"
-	"github.com/statechannels/go-nitro/types"
 )
 
 const (
@@ -22,10 +22,10 @@ type Objective struct {
 	Status protocols.ObjectiveStatus
 
 	// InitialOutcome is the initial outcome of the virtual channel
-	InitialOutcome outcome.Exit
+	InitialOutcome outcome.SingleAssetExit
 
 	// PaidToBob is the amount that should be paid from Alice (participant 0) to Bob (participant 2)
-	PaidToBob types.Funds
+	PaidToBob *big.Int
 
 	// VFixed is the fixed channel information for the virtual channel
 	VFixed state.FixedPart
@@ -49,18 +49,16 @@ const ObjectivePrefix = "VirtualDefund-"
 
 // finalState returns the final state for the virtual channel
 func (o Objective) finalState() state.State {
-	vp := state.VariablePart{Outcome: o.finalOutcome(), TurnNum: 3, IsFinal: true}
+	vp := state.VariablePart{Outcome: outcome.Exit{o.finalOutcome()}, TurnNum: 3, IsFinal: true}
 	return state.StateFromFixedAndVariablePart(o.VFixed, vp)
 }
 
 // finalOutcome returns the outcome for the final state calculated from the InitialOutcome and PaidToBob
-func (o Objective) finalOutcome() outcome.Exit {
+func (o Objective) finalOutcome() outcome.SingleAssetExit {
 	finalOutcome := o.InitialOutcome.Clone()
-	for _, exit := range finalOutcome {
-		paid := o.PaidToBob[exit.Asset]
-		exit.Allocations[0].Amount.Sub(exit.Allocations[0].Amount, paid)
-		exit.Allocations[1].Amount.Add(exit.Allocations[1].Amount, paid)
-	}
+
+	finalOutcome.Allocations[0].Amount.Sub(finalOutcome.Allocations[0].Amount, o.PaidToBob)
+	finalOutcome.Allocations[1].Amount.Add(finalOutcome.Allocations[1].Amount, o.PaidToBob)
 
 	return finalOutcome
 }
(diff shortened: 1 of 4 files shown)


protocols/virtualdefund/virtualdefund.go
```diff
@@ -16,28 +17,56 @@ const (
...
 // Objective contains relevent information for the defund objective
 type Objective struct {
 	Status protocols.ObjectiveStatus
-	V      *channel.SingleHopVirtualChannel
+
+	// InitialOutcome is the initial outcome of the virtual channel
+	InitialOutcome outcome.SingleAssetExit
+
+	// PaidToBob is the amount that should be paid from Alice (participant 0) to Bob (participant 2)
+	PaidToBob *big.Int
```
lalexgap, line 28: This assumes payments are one-way. This is easy to change if we want.

protocols/virtualdefund/virtualdefund.go
```diff
@@ -16,28 +17,56 @@ const (
...
+	// 1 is Irene
+	// 2 is Bob
+	MyRole uint
 }
 
 const ObjectivePrefix = "VirtualDefund-"
 
+// finalState returns the final state for the virtual channel
+func (o Objective) finalState() state.State {
+	vp := state.VariablePart{Outcome: outcome.Exit{o.finalOutcome()}, TurnNum: 3, IsFinal: true}
```
andrewgordstewart, line 52: While `TurnNum: 3` is a reasonable final turn number, why not `2`?

Does it make sense to define a `FinalTurnNum` constant?

protocols/virtualdefund/virtualdefund.go
```diff
@@ -74,10 +103,15 @@ func (o *Objective) Related() []protocols.Storable {
...
-	clone.V = vClone
 
-	// todo: #420 consider cloning for consensusChannels
+	clone.VFixed = o.VFixed.Clone()
+	clone.InitialOutcome = o.InitialOutcome.Clone()
+	clone.PaidToBob = big.NewInt(0).Set(o.PaidToBob)
+
+	clone.Signatures = [3]state.Signature{}
+	for i, s := range o.Signatures {
+		clone.Signatures[i] = s
```
andrewgordstewart, line 113: Does this clone the signatures?

(I'm not sure that's super important, TBH.)

andrewgordstewart (approved):
The change looks good.

How do you feel about this approach? What's better or worse about it? (Maybe that's hard to say without seeing the objective implemented?)

protocols/virtualdefund/virtualdefund.go
lalexgap: I believe this does! Since `state.Signature` is a struct (and not a pointer) when we assign it (on line 113) it makes a copy of the struct. I think!

andrewgordstewart: It does, but the `Signature` struct contains two slices; a copy of `s` would copy these slices, which would copy the _pointers_ that manage the slice's values.

I believe if you do:
```
s := state.Signature{...}
s_copy := s
s_copy.R = append(s_copy.R, 1)
```
then you would also be modifying the value of `s.R`?

lalexgap:
> How do you feel about this approach? What's better or worse about it? (Maybe that's hard to say without seeing the objective implemented?)

Overall I like this approach. It simplifies the objective code to not have to worry about dealing with a `channel`.  My only concern is that we're introducing another way a `channel` is stored, but that's probably not too big an issue right now.

protocols/virtualdefund/virtualdefund.go
lalexgap: Done in https://github.com/statechannels/go-nitro/pull/512/commits/a732771c053ddf5d1879765ff2ed2428032c5bc2

protocols/virtualdefund/virtualdefund.go
lalexgap: Good point! I forgot that the signature has slices of bytes.

I've implemented a `CloneSignature` function in [0bb26e6](https://github.com/statechannels/go-nitro/pull/512/commits/0bb26e6df92d586c427a451b686c439358f864cb) to properly clone the signature.

It probably doesn't matter at the moment but it seems like a good thing to fix before it bites us :)

PR #513, opened 2022-04-08
Vision question part 1: write some unit tests

andrewgordstewart:
We were struggling to make progress towards #420, and we identified that there was a key reason we were struggling:
> the virtualfund unit tests take advantage of the fact that `Channel` receives arbitrary states, and test the `Crank` method by sending an arbitrary state for the ledger channel

This has a few implications when switching to ConsensusChannels
- ConsensusChannels are strict about what data they receive: an invalid or unexpected proposal is rejected.
- Our old strategy was to assert on equal behaviour between a ConsensusChannel and a TwoPartyLedger channel. But, because the test took advantage of arbitrary TwoPartyLedger behaviour, we needed to be really careful about setting up ledger channels correctly!
- testCrank and testUpdate are written to apply to Alice, Bob and P1. But, unless they are set up with ledger channels where they have the same role, applying the same test function to all actors is really awkward

This PR is attempting a different strategy towards migrating to ConsensusChannel:
1. remove TwoPartyLedger from a `connection`. This causes the build to break!
2. fix the build errors, and review the failing tests
3. skip the failing virtual fund-related tests
4. write more rigorous virtualfund unit tests

In step 4, I chose to copy the old unit test file (`virtualfund-single-hop_test.go`) to a new file (`virtualfund_single_hop_test.go`). The difference in filename is the old file had hyphens (which I believe are not kosher?)

In the new unit test file, I unwrap some of the nesting from the old file, and rewrite the intent of `testNew`, `testClone`, and (some of) `testCrank` (from Alice's perspective). If you look at the two files side by side, you should hopefully see that they have the same intent.

When writing the new test file, I aimed to make the tests easy to understand. So, I took inspiration from https://github.com/benbjohnson/testing and wrote some assertions that help developers pinpoint test failures

Further work:
- finishing testCrank from Alice's perspective
- write testCrank from p1 & Bob's perspective
- write testUpdate
- delete the old unit test file
- unskip failing integration tests

#### Code quality

#### Project management

### DIFF
--- a/channel/consensus_channel/leader_channel.go
+++ b/channel/consensus_channel/leader_channel.go
@@ -38,6 +38,11 @@ func (c *ConsensusChannel) Propose(proposal Proposal, sk []byte) (SignedProposal
 		return SignedProposal{}, ErrNotLeader
 	}
 
+	// TODO: the Propose API should be less confusing!
+	// Currently, the TurnNum is ignored, and Propose could easily
+	// return the same ChannelId that it's been passed
+	proposal.ChannelID = c.Id
+
 	vars, err := c.latestProposedVars()
 	if err != nil {
 		return SignedProposal{}, fmt.Errorf("unable to construct latest proposed vars: %w", err)
--- a/channel/consensus_channel/leader_channel_test.go
+++ b/channel/consensus_channel/leader_channel_test.go
@@ -227,8 +227,7 @@ func TestLeaderChannel(t *testing.T) {
 
 		c := testChannel(startingOutcome, startingQueue)
 
-		newAdd := Proposal{ToAdd: add(2, amountAdded, types.Destination{3}, alice, bob)}
-		newAdd.ChannelID = p1.ChannelID
+		newAdd := Proposal{ChannelID: p1.ChannelID, ToAdd: add(2, amountAdded, types.Destination{3}, alice, bob)}
 
 		currentlyProposed, _ := c.latestProposedVars()
 		expectedSp := aliceSignedProposal(currentlyProposed, newAdd).SignedProposal
--- a/channel/consensus_channel/leader_channel_test.go
+++ b/channel/consensus_channel/leader_channel_test.go
@@ -228,6 +228,7 @@ func TestLeaderChannel(t *testing.T) {
 		c := testChannel(startingOutcome, startingQueue)
 
 		newAdd := Proposal{ToAdd: add(2, amountAdded, types.Destination{3}, alice, bob)}
+		newAdd.ChannelID = p1.ChannelID
 
 		currentlyProposed, _ := c.latestProposedVars()
 		expectedSp := aliceSignedProposal(currentlyProposed, newAdd).SignedProposal
--- a/protocols/virtualfund/helpers_test.go
+++ b/protocols/virtualfund/helpers_test.go
@@ -41,6 +41,8 @@ var bob = actor{
 	role:        2,
 }
 
+var allActors = []actor{alice, p1, bob}
+
 func prepareConsensusChannel(role uint, left, right actor) *consensus_channel.ConsensusChannel {
 	fp := state.FixedPart{
 		ChainId:           big.NewInt(9001),
--- a/protocols/virtualfund/helpers_test.go
+++ b/protocols/virtualfund/helpers_test.go
@@ -14,6 +14,7 @@ type actor struct {
 	destination types.Destination
 	privateKey  []byte
 	role        uint
+	name        string
 }
 
 ////////////
@@ -25,6 +26,7 @@ var alice = actor{
 	destination: types.AddressToDestination(common.HexToAddress(`0xD9995BAE12FEe327256FFec1e3184d492bD94C31`)),
 	privateKey:  common.Hex2Bytes(`7ab741b57e8d94dd7e1a29055646bafde7010f38a900f55bbd7647880faa6ee8`),
 	role:        0,
+	name:        "alice",
 }
 
 var p1 = actor{ // Aliases: The Hub, Irene
@@ -32,6 +34,7 @@ var p1 = actor{ // Aliases: The Hub, Irene
 	destination: types.AddressToDestination(common.HexToAddress(`0xd4Fa489Eacc52BA59438993f37Be9fcC20090E39`)),
 	privateKey:  common.Hex2Bytes(`2030b463177db2da82908ef90fa55ddfcef56e8183caf60db464bc398e736e6f`),
 	role:        1,
+	name:        "p1",
 }
 
 var bob = actor{
@@ -39,6 +42,7 @@ var bob = actor{
 	destination: types.AddressToDestination(common.HexToAddress(`0x760bf27cd45036a6C486802D30B5D90CfFBE31FE`)),
 	privateKey:  common.Hex2Bytes(`62ecd49c4ccb41a70ad46532aed63cf815de15864bc415c87d507afd6a5e8da2`),
 	role:        2,
+	name:        "bob",
 }
 
 var allActors = []actor{alice, p1, bob}
--- a/protocols/virtualfund/virtualfund-single-hop_test.go
+++ b/protocols/virtualfund/virtualfund-single-hop_test.go
@@ -52,7 +52,6 @@ func compareGuarantees(a, b consensus_channel.Guarantee) string {
 // signLatest is a test utility function which applies signatures from
 // multiple participants to the latest recorded state
 // func signLatest(ledger *consensus_channel.ConsensusChannel, secretKeys [][]byte) {
-// 	panic("whoops")
 
 // Find the largest turn num and therefore the latest state
 // turnNum := uint64(0)
--- a/protocols/virtualfund/virtualfund.go
+++ b/protocols/virtualfund/virtualfund.go
@@ -625,14 +625,19 @@ func (o *Objective) updateLedgerWithGuarantee(ledgerConnection Connection, sk *[
 	ledger := ledgerConnection.Channel // todo: #420 deprecate
 
 	var sideEffects protocols.SideEffects
+	g := ledgerConnection.getExpectedGuarantee()
+	proposed, err := ledger.IsProposed(g)
+
 	if ledger.IsLeader() { // If the user is the proposer craft a new proposal
+		if proposed {
+			return protocols.SideEffects{}, nil
+		}
 		se, err := o.proposeLedgerUpdate(ledgerConnection, sk)
 		if err != nil {
 			return protocols.SideEffects{}, fmt.Errorf("error proposing ledger update: %w", err)
 		}
 		sideEffects = se
 	} else {
-		proposed, err := ledger.IsProposed(ledgerConnection.getExpectedGuarantee())
 		if err != nil {
 			return protocols.SideEffects{}, err
 		}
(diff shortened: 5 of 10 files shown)


protocols/virtualfund/virtualfund-single-hop_test.go
```diff
@@ -33,142 +31,76 @@ func compareObjectives(a, b Objective) string {
 	)
 }
 
+func compareGuarantees(a, b consensus_channel.Guarantee) string {
```
andrewgordstewart, line 34: This file is slated for deletion, so you can ignore the diffs!

protocols/virtualfund/helpers_test.go
```diff
@@ -0,0 +1,87 @@
...
+	address:     common.HexToAddress(`0x760bf27cd45036a6C486802D30B5D90CfFBE31FE`),
+	destination: types.AddressToDestination(common.HexToAddress(`0x760bf27cd45036a6C486802D30B5D90CfFBE31FE`)),
+	privateKey:  common.Hex2Bytes(`62ecd49c4ccb41a70ad46532aed63cf815de15864bc415c87d507afd6a5e8da2`),
+	role:        2,
+	name:        "bob",
+}
+
+var allActors = []actor{alice, p1, bob}
+
+func prepareConsensusChannel(role uint, left, right actor) *consensus_channel.ConsensusChannel {
```
andrewgordstewart, line 50: This is copy/pasted from `virtualfund-single-hop_test.go`

protocols/virtualfund/helpers_test.go
```diff
@@ -0,0 +1,87 @@
+package virtualfund
```
andrewgordstewart, line 1: Changes:

This file is copy-past code from virtualfund-single-hop_test.go.

`actor` now has a `name` field, to help print friendly test messages.

protocols/virtualfund/virtualfund.go
```diff
@@ -38,17 +38,8 @@ type GuaranteeInfo struct {
 	GuaranteeDestination types.Destination
```
andrewgordstewart, line 38: There are two parts of the diffs in this file:
- the old ledger channel is removed from the Connection struct: https://github.com/statechannels/go-nitro/pull/513/commits/d0e34c5759bd0dfb4f7ff8d4178fbe4a1a669c05. Other changes made in that commit are more or less forced.
- subsequent changes, which are more reasonable: https://github.com/statechannels/go-nitro/compare/d0e34c5759bd0dfb4f7ff8d4178fbe4a1a669c05...528b1cda71736ccf32afd825307d94eb70c5f9f6#diff-e78ba83d3714e679fa001e0cc48267b733ce93be39eb4ab1a37a777d25f6ca57

lalexgap:
> Further work:
> 
> * finishing testCrank from Alice's perspective
> * write testCrank from p1 & Bob's perspective
> * write testUpdate
> * delete the old unit test file
> * unskip failing integration tests

Could we create some issues for these tasks?

channel/consensus_channel/leader_channel.go
```diff
@@ -38,6 +38,11 @@ func (c *ConsensusChannel) Propose(proposal Proposal, sk []byte) (SignedProposal
 		return SignedProposal{}, ErrNotLeader
 	}
 
+	// TODO: the Propose API should be less confusing!
```
lalexgap, line 41: We could create an issue for that!

protocols/virtualfund/virtualfund.go
```diff
@@ -98,6 +99,11 @@ func (c *Connection) Funded() bool {
 // getExpectedGuarantee returns a map of asset addresses to guarantees for a Connection.
 func (c *Connection) getExpectedGuarantee() consensus_channel.Guarantee {
 	amountFunds := c.GuaranteeInfo.LeftAmount.Add(c.GuaranteeInfo.RightAmount)
+
+	//HACK: GuaranteeInfo stores amounts as types.Funds.
```
lalexgap, line 103: This is beyond the scope of this PR but I think we should decide how we want to treat multi-asset outcomes throughout our codebase.  Right now some code supports multi-asset (but is not fully tested) and some does not.  It would be nice if we could be consistent in our approach.

protocols/virtualfund/virtualfund_single_hop_test.go
```diff
@@ -0,0 +1,304 @@
...
+	equals(t, waitingFor, WaitingForCompleteFunding)
+}
+
+// Copied from https://github.com/benbjohnson/testing
+
+// assert fails the test if the condition is false.
+func assert(tb testing.TB, condition bool, msg string, v ...interface{}) {
+	if !condition {
+		_, file, line, _ := runtime.Caller(1)
+		fmt.Printf("\033[31m%s:%d: "+msg+"\033[39m\n\n", append([]interface{}{filepath.Base(file), line}, v...)...)
```
lalexgap, line 240: Could we create a function to wrap the `msg` in these magic strings so we don't have to copy and paste?

protocols/virtualfund/virtualfund_single_hop_test.go
```diff
@@ -0,0 +1,304 @@
...
+	o = oObj.(*Objective)
+	ok(t, err)
+	equals(t, got, emptySideEffects)
+	equals(t, waitingFor, WaitingForCompleteFunding)
+}
+
+// Copied from https://github.com/benbjohnson/testing
+
+// assert fails the test if the condition is false.
+func assert(tb testing.TB, condition bool, msg string, v ...interface{}) {
```
lalexgap, line 237: Is there a reason we use a `testing.TB` instead of just a `testing.T`

protocols/virtualfund/virtualfund_single_hop_test.go
```diff
@@ -0,0 +1,304 @@
...
+	oObj, got, waitingFor, err = o.Crank(&my.privateKey)
+	o = oObj.(*Objective)
+	ok(t, err)
+	equals(t, got, emptySideEffects)
+	equals(t, waitingFor, WaitingForCompleteFunding)
+}
+
+// Copied from https://github.com/benbjohnson/testing
+
+// assert fails the test if the condition is false.
```
lalexgap, line 236: Do we expect to use these in other tests? Should we move them somewhere central and add some better godoc?

protocols/virtualfund/virtualfund_single_hop_test.go
```diff
@@ -0,0 +1,304 @@
...
+		_, file, line, _ := runtime.Caller(1)
+		fmt.Printf("\033[31m%s:%d:\n\n\texp: %#v\n\n\tgot: %#v\033[39m\n\n", filepath.Base(file), line, exp, act)
+		tb.FailNow()
+	}
+}
+
+// The following assertions are inspired by the ok, assert and equals above
+
+// assertSideEffectsContainsMessageWith fails the test instantly if the supplied side effects does not contain a message for the supplied actor with the supplied expected signed state.
+func assertProposalSent(t *testing.T, ses protocols.SideEffects, sp consensus_channel.SignedProposal, to actor) {
```
lalexgap, line 266: It would be nice if formatting could be handled separately so `assertProposalSent` would be simpler to read.

channel/consensus_channel/leader_channel_test.go
```diff
@@ -228,6 +228,7 @@ func TestLeaderChannel(t *testing.T) {
 		c := testChannel(startingOutcome, startingQueue)
 
 		newAdd := Proposal{ToAdd: add(2, amountAdded, types.Destination{3}, alice, bob)}
+		newAdd.ChannelID = p1.ChannelID
```
lalexgap, line 231: nit but could we set `ChannelID` when initialing `Proposal` in the line above?

protocols/virtualfund/virtualfund_single_hop_test.go
```diff
@@ -0,0 +1,304 @@
...
+}
+
+func TestNew(t *testing.T) {
+	for _, a := range allActors {
+		msg := fmt.Sprintf("Testing new as %v", a.name)
+		t.Run(msg, testNew(a))
+	}
+}
+
+func testClone(my actor) Tester {
```
lalexgap, line 134: Could we rename this so it's distinct from `TestClone`? Maybe `testCloneAs`?

protocols/virtualfund/virtualfund.go
```diff
@@ -600,10 +601,21 @@ func (o *Objective) acceptLedgerUpdate(c Connection, sk *[]byte) (protocols.Side
...
 
 	sideEffects := protocols.SideEffects{}
-	messages := protocols.CreateSignedProposalMessages(o.Id(), signedProposal, uint(ledger.MyIndex))
-	sideEffects.MessagesToSend = append(sideEffects.MessagesToSend, messages...)
+	message := o.createSignedProposalMessage(signedProposal, c.Channel)
+	sideEffects.MessagesToSend = append(sideEffects.MessagesToSend, message)
 	return sideEffects, nil
+}
 
+func (o *Objective) createSignedProposalMessage(sp consensus_channel.SignedProposal, ledger *consensus_channel.ConsensusChannel) protocols.Message {
```
lalexgap, line 609: nit but could we add godoc to this

lalexgap (approved):

channel/consensus_channel/leader_channel_test.go
andrewgordstewart: 6dd1662f2ce6c90b483addd120f458590a42cd14

protocols/virtualfund/virtualfund_single_hop_test.go
andrewgordstewart: c3b855de620e1f235ae2718acd5adcba29b33b97

protocols/virtualfund/virtualfund.go
andrewgordstewart: 216c4c4a720bfbadff170a5792e44ac57c3ab49f

protocols/virtualfund/virtualfund_single_hop_test.go
andrewgordstewart: Only because it's copied from the given link :P

protocols/virtualfund/virtualfund.go
andrewgordstewart: https://github.com/statechannels/go-nitro/issues/516

protocols/virtualfund/virtualfund_single_hop_test.go
andrewgordstewart: I think that a function like that would be a bit annoying. It would accept a formatted string and wrap it on either end. This means we'd have to do something like
```
printRed(fmt.Sprintf("%s:%d: "+msg, append([]interface{}{filepath.Base(file), line}, v...)...))
```

Given that concatenating strings is so common, I might prefer
```
fmt.Printf(makeRed+"%s:%d: "+msg+makeBlack, append([]interface{}{filepath.Base(file), line}, v...)...)
```

What do you think about 9e626058a6aac0e3ef2dbdae0243a115f4af0510?

protocols/virtualfund/virtualfund_single_hop_test.go
andrewgordstewart: That is definitely a good idea!

I've created https://github.com/statechannels/go-nitro/issues/517 -- it could be useful to use these assertions in some other tests to make them more readable.

protocols/virtualfund/virtualfund_single_hop_test.go
andrewgordstewart: IMO, it's not very important to understand the formatting in `assertProposalSent`. It's only important that it works well :P

9e626058a6aac0e3ef2dbdae0243a115f4af0510 should make it simpler to read?

PR #504, opened 2022-04-07
refactor: simplify api for receiving proposals...

NiloCK:
closes #501 

ConsensusChannel.Receive(signedProposal) is now the consuming endpoint
for both leader and follower channels. This relieves calling code from
having to inspect the channel and decide which method to use.

`Receive(sp)` routes the signed proposal to the appropriate, private
leader or follower receiving functions.

(this is not technically a refactor, because it alters the API of a struct. It is a refactor in spirit.)

---

#### Code quality

- [x] I have written clear commit messages
- [x] I have performed a self-review of my own code
- [x] This change does not have an unduly wide scope
- [x] I have separated logic changes from refactor changes (formatting, renames, etc.)
- [x] I have commented my code wherever necessary (can be 0)
- [x] I have added tests that prove my fix is effective or that my feature works, if necessary
- [x] I have added comprehensive [godoc](https://go.dev/blog/godoc) comments 

#### Project management

- [x] I have assigned myself to this PR
- [x] I have linked the appropriate github issue
- [x] I have assigned this PR to the appropriate GitHub project
- [x] I have assigned this PR to the appropriate GitHub Milestone

PR #519, opened 2022-04-11
refactor: simplify api for receiving proposals...

NiloCK:
This is the same content as the approved #504 PR, cleaned into a single commit, after botching the in browser merge.


lalexgap (approved):
👍

PR #522, opened 2022-04-11
Implement round 1 of virtual defunding

### LINKED ISSUE #478: Create a virtual defunding protocol that signs/exchanges `V`'s final state
The goal of this task is create a virtual defunding protocol that completes Round 1 of the defunding protocol. The virtual defunding protocol should exchange signed final states for `V`.

We should add a suite of unit tests for this functionality.

lalexgap:
Fixes #478

This implements the first round of virtual defunding, which is exchanging the final state for `V`.  Tests have been added for `Update` and `Crank` to make sure they properly handle the exchange of final states.

### DIFF
--- a/protocols/virtualdefund/virtualdefund.go
+++ b/protocols/virtualdefund/virtualdefund.go
@@ -228,9 +228,9 @@ func (o Objective) signedByMe() bool {
 
 }
 
-// ValidateSignature returns whether the given signature is valid for the given participant
+// validateSignature returns whether the given signature is valid for the given participant
 // If a signature is invalid an error will be returned containing the reason
-func (o Objective) ValidateSignature(sig state.Signature, participantIndex uint) (bool, error) {
+func (o Objective) validateSignature(sig state.Signature, participantIndex uint) (bool, error) {
 	if participantIndex > 2 {
 		return false, fmt.Errorf("participant index %d is out of bounds", participantIndex)
 	}
@@ -280,7 +280,7 @@ func (o Objective) Update(event protocols.ObjectiveEvent) (protocols.Objective, 
 					}
 				}
 				// Otherwise we validate the incoming signature and update our signatures
-				isValid, err := updated.ValidateSignature(incomingSig, i)
+				isValid, err := updated.validateSignature(incomingSig, i)
 				if isValid {
 					// Update the signature
 					updated.Signatures[i] = incomingSig
--- a/protocols/virtualdefund/virtualdefund.go
+++ b/protocols/virtualdefund/virtualdefund.go
@@ -51,14 +51,32 @@ type Objective struct {
 
 const ObjectivePrefix = "VirtualDefund-"
 
-//nolint:unused // not used yet
+// newObjective constructs a new virtual defund objective
+func newObjective(preApprove bool, vFixed state.FixedPart, initialOutcome outcome.SingleAssetExit, paidToBob *big.Int, myRole uint) Objective {
+	var status protocols.ObjectiveStatus
+
+	if preApprove {
+		status = protocols.Approved
+	} else {
+		status = protocols.Unapproved
+	}
+
+	return Objective{
+		Status:         status,
+		InitialOutcome: initialOutcome,
+		PaidToBob:      paidToBob,
+		VFixed:         vFixed,
+		Signatures:     [3]state.Signature{},
+		MyRole:         myRole,
+	}
+}
+
 // finalState returns the final state for the virtual channel
 func (o Objective) finalState() state.State {
 	vp := state.VariablePart{Outcome: outcome.Exit{o.finalOutcome()}, TurnNum: FinalTurnNum, IsFinal: true}
 	return state.StateFromFixedAndVariablePart(o.VFixed, vp)
 }
 
-//nolint:unused // not used yet
 // finalOutcome returns the outcome for the final state calculated from the InitialOutcome and PaidToBob
 func (o Objective) finalOutcome() outcome.SingleAssetExit {
 	finalOutcome := o.InitialOutcome.Clone()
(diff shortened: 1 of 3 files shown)


protocols/virtualdefund/virtualdefund.go
```diff
@@ -145,6 +252,77 @@ func (o Objective) Update(event protocols.ObjectiveEvent) (protocols.Objective,
...
 
+	updated := o.clone()
+
+	for _, ss := range event.SignedStates {
+		incomingChannelId, _ := ss.State().ChannelId() // TODO handle error
+		vChannelId, _ := updated.VFixed.ChannelId()    // TODO handle error
+
+		if incomingChannelId != vChannelId {
+			return &o, errors.New("event channelId out of scope of objective")
+		} else {
```
kerzhner, line 264: It would be cleaner not to nest this logic in an else statement.

protocols/virtualdefund/virtualdefund.go
```diff
@@ -145,6 +252,77 @@ func (o Objective) Update(event protocols.ObjectiveEvent) (protocols.Objective,
...
+			incomingSignatures := ss.Signatures()
+			for i := uint(0); i < 3; i++ {
+				existingSig := o.Signatures[i]
+				incomingSig := incomingSignatures[i]
+
+				// If the incoming signature is zeroed we ignore it
+				if isZero(incomingSig) {
+					continue
+				}
+				// If the existing signature is not zeroed we check that it matches the incoming signature
```
kerzhner, line 274: Why is this logic needed? If the objective contains a signature for the final state of a virtual channel for a participant, we can ignore any new signatures for that participant.

A rogue, new signature indicates an out-of-protocol behaviour by a participant. But this behaviour does not prevent the objective from moving forward/succeeding.

protocols/virtualdefund/virtualdefund.go
```diff
@@ -145,6 +252,77 @@ func (o Objective) Update(event protocols.ObjectiveEvent) (protocols.Objective,
...
+	}
+	if o.ToMyRight != nil {
+		toMyRightId = o.ToMyRight.Id
+	}
+
+	for _, sp := range event.SignedProposals {
+		var err error
+		switch sp.Proposal.ChannelID {
+		case types.Destination{}:
+			return &o, fmt.Errorf("signed proposal is for a zero-addressed ledger channel") // catch this case to avoid unspecified behaviour -- because if Alice or Bob we allow a null channel.
```
kerzhner, line 307:
```suggestion
			return &o, fmt.Errorf("signed proposal is for a zero-addressed ledger channel") // catch this case to avoid unspecified behaviour -- because of Alice or Bob we allow a null channel.
```

protocols/virtualdefund/virtualdefund_test.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	if my.Role != 2 {
+		_ = signedState.Sign(&bob.PrivateKey)
+	}
+	return signedState
+}
+
+// assertStateSentToEveryone asserts that ses contains a message for every participant but from
+func assertStateSentToEveryone(t *testing.T, ses protocols.SideEffects, expected state.SignedState, from testactors.Actor) {
```
kerzhner, line 83: I find this function with 5 levels of nesting hard to interpret. Any way to flatten out this function?

protocols/virtualdefund/virtualdefund_test.go
```diff
@@ -0,0 +1,200 @@
...
+		}
+
+		testhelpers.Equals(t, waitingFor, WaitingForCompleteFinal)
+		signedByMe := state.NewSignedState(data.vFinal)
+		_ = signedByMe.Sign(&my.PrivateKey)
+		assertStateSentToEveryone(t, se, signedByMe, my)
+
+		// Update the signatures on the objective so the final state is fully signed
+		signedByOthers := signByOthers(my, state.NewSignedState(data.vFinal))
+		for i, sig := range signedByOthers.Signatures() {
```
kerzhner, line 178: Should we just use `Update` here?

kerzhner (approved):

protocols/virtualdefund/virtualdefund.go
```diff
@@ -136,7 +171,79 @@ func (o *Objective) clone() Objective {
 
 // Crank inspects the extended state and declares a list of Effects to be executed
 func (o Objective) Crank(secretKey *[]byte) (protocols.Objective, protocols.SideEffects, protocols.WaitingFor, error) {
-	return &o, protocols.SideEffects{}, WaitingForCompleteFinal, errors.New("TODO: UNIMPLEMENTED")
+	updated := o.clone()
```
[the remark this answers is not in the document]
lalexgap: Using a value receiver means the receiver will be a shallow clone of the original. However since slices are reference types the shallow clone of `o` would contain slices that point to the same underlying elements. This means modifying a slice in `Crank` would modify the slice for all copies of `o`.

protocols/virtualdefund/virtualdefund.go
```diff
@@ -136,7 +171,79 @@ func (o *Objective) clone() Objective {
...
+
+// signedByMe returns whether the current participant has signed the final state
+func (o Objective) signedByMe() bool {
+	return o.signedBy(o.MyRole)
+
+}
+
+// ValidateSignature returns whether the given signature is valid for the given participant
+// If a signature is invalid an error will be returned conaining the reason
+func (o Objective) ValidateSignature(sig state.Signature, participantIndex uint) (bool, error) {
```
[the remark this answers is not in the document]
lalexgap: Nope! I think it can be private to the package.

protocols/virtualdefund/virtualdefund.go
lalexgap: I think you're right and this logic is not strictly needed. However I think it might be helpful when debugging; if we receive an incorrect signature (after receiving a valid one) something has probably gone wrong at some point.

Happy to change this if you feel strongly about it though!

protocols/virtualdefund/virtualdefund.go
```diff
@@ -145,6 +252,77 @@ func (o Objective) Update(event protocols.ObjectiveEvent) (protocols.Objective,
 	if o.Id() != event.ObjectiveId {
 		return &o, fmt.Errorf("event and objective Ids do not match: %s and %s respectively", string(event.ObjectiveId), string(o.Id()))
 	}
-	return &o, errors.New("TODO: UNIMPLEMENTED")
 
+	updated := o.clone()
```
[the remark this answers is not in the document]
lalexgap: Same reason as this [comment](https://github.com/statechannels/go-nitro/pull/522/files/d79d0426b9a05644bc8bd211c29a4fcfc0cd8203#r854331336).

protocols/virtualdefund/virtualdefund.go
```diff
@@ -51,14 +54,46 @@ type Objective struct {
 
 const ObjectivePrefix = "VirtualDefund-"
 
-//nolint:unused // not used yet
+// newObjective constructs a new virtual defund objective
+func newObjective(preApprove bool, vFixed state.FixedPart, initialOutcome outcome.SingleAssetExit, paidToBob *big.Int, myRole uint) Objective {
```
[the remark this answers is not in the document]
lalexgap: Once we hook up virtual defunding into the engine (#564) I think we'll need to write a function like `ConstructObjectiveFromMessage` that would call `newObjective`

protocols/virtualdefund/virtualdefund_test.go
lalexgap: While it'd probably be fine I figured it'd be safer for the `crank` test to not rely on the functionality of `Update`.

protocols/virtualdefund/virtualdefund_test.go
lalexgap: I've flattened this a bit in https://github.com/statechannels/go-nitro/pull/522/commits/a079b5aad7e99c791a04c97c585359202f87af90 by breaking it into two functions.

protocols/virtualdefund/virtualdefund.go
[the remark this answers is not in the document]
lalexgap: https://github.com/statechannels/go-nitro/pull/522/commits/cdb230e8e234f424f35a9f13d8018484ab84faf2

PR #525, opened 2022-04-12
Add eslint (prettier), fix errors and warnings, run in CI

### LINKED ISSUE #500: Add prettier formatting for `./nitro-protocol` folder
This should be enforced in CI. See statechannels monorepo for configuration. 

This is high priority to prevent formatting noise on changesets.

geoknee:
Towards #500 

This PR
- adds linting configuration, which includes prettier via plugins
- runs a lint check in CI
- fixes (or ignores) all lint warnings and errors

The large changes are housed in a single commit running `eslint`. So reviewing commit-by-commit (and ignoring that one) is a reasonable review strategy.


nitro-protocol/.eslintrc.js
```diff
@@ -0,0 +1,43 @@
+/* global module */
+module.exports = {
+  parser: '@typescript-eslint/parser',
+  plugins: [
+    // All packages in this monorepo use TypeScript
+    '@typescript-eslint',
+    // All packages in this monorepo use Prettier
```
geoknee, line 7:
```suggestion
    '@typescript-eslint',
```

nitro-protocol/.prettierrc.js
```diff
@@ -0,0 +1,16 @@
+module.exports = {
+  trailingComma: "none",
```
geoknee, line 2: We could change this rule to generate a smaller diff when "fixing" the prettier warnings. 

https://prettier.io/docs/en/options.html#trailing-commas

The actual rule is not that important -- having _some_ rule is what we are really interested in.

nksazonov:
Why don't you use `arrowParens: "always"` [inhere](https://github.com/statechannels/go-nitro/pull/525/files#diff-1963db219cd24c80461b6d4fa4e384b4b5e29a393a4b53252b26a4d27d5a285cR3)? 

https://prettier.io/docs/en/options.html#arrow-function-parentheses

Suggestion to use them from `Prettier` docs sounds reasonable.

geoknee:
> 

Yes it seems reasonable. However, I'd rather that was tackled in a separate PR. This one is about getting the boilerplate code for linting and formatting in `main`. Tweaks to the actual configuration can be done separately.

nksazonov (approved):

PR #529, opened 2022-04-12
Move test helper functions from virtualfund to their own packages

### LINKED ISSUE #517: create standard assertion helpers
Do we expect to use these in other tests? Should we move them somewhere central and add some better godoc?

_Originally posted by @lalexgap in https://github.com/statechannels/go-nitro/pull/513#discussion_r846491888_

lalexgap:
Fixes #517

This moves the test helper functions used by the virtual funding test into their own package. That way they can be shared by other tests.

I took a quick look at formatting using `lipgloss` but ran into some problems and decided it wasn't worth the effort at the moment.


protocols/virtualfund/virtualfund_single_hop_test.go
```diff
@@ -4,15 +4,14 @@ import (
...
-	"path/filepath"
 	"reflect"
-	"runtime"
 	"testing"
 
 	"github.com/statechannels/go-nitro/channel"
 	"github.com/statechannels/go-nitro/channel/consensus_channel"
 	"github.com/statechannels/go-nitro/channel/state"
 	"github.com/statechannels/go-nitro/channel/state/outcome"
+	helpers "github.com/statechannels/go-nitro/internal/testhelpers"
```
andrewgordstewart, line 14: Is this a reasonable practice here?

```suggestion
	. "github.com/statechannels/go-nitro/internal/testhelpers"
```

protocols/virtualfund/virtualfund_single_hop_test.go
```diff
@@ -219,80 +218,32 @@ func TestCrankAsAlice(t *testing.T) {
...
-	if len(ses.MessagesToSend) != 1 {
-		fmt.Printf(makeRed+"%s:%d:\n\n\texpected one message"+makeBlack, filepath.Base(file), line)
-		t.FailNow()
-	}
-	if len(ses.MessagesToSend[0].SignedProposals) != 1 {
-		fmt.Printf(makeRed+"%s:%d:\n\n\texpected one signed proposal"+makeBlack, filepath.Base(file), line)
-		t.FailNow()
-	}
+
+	helpers.Assert(t, len(ses.MessagesToSend) == 1, "expected one message")
```
andrewgordstewart, line 236: There's a (relatively minor) issue here: one of the benefits of `helpers.Assert` is it points out the line number of the failed assertion. This function would currently print a less useful error to the console: `virtualfund_single_hop_test.go:236: ...`.  This is because `helpers.Assert` looks at one function up the call stack.

andrewgordstewart (approved):

andrewgordstewart:
Note that this conflicts with https://github.com/statechannels/go-nitro/pull/521

protocols/virtualfund/virtualfund_single_hop_test.go
lalexgap: That makes sense to me! I've applied that in https://github.com/statechannels/go-nitro/pull/529/commits/eceefe6ee3fd6c794c263028437af8bf7492d0bf

protocols/virtualfund/virtualfund_single_hop_test.go
lalexgap: That's a good point! I think for now I'm going to just ignore it. However we could always look at creating an assert that takes in a call stack depth.

lalexgap:
> Note that this conflicts with #521

I've now rebased on `main` so the conflicts should be resolved

Issue #490, opened 2022-04-05
`GetObjectiveByChannelId` is not fully specified


andrewgordstewart:
During the lifecycle of any given channel `c`, we expect **at least two** objectives to "point to" `c`.

For instance, a ledger channel is
1. directly funded, managed by some `DFO` (direct-fund objective)
2. directly defunded, managed by some `DDO`

Currently, when [`Engine.handleChainEvent` is called](https://github.com/statechannels/go-nitro/blob/2615ed67b6cb7aebd84553400b5ab635979cb109/client/engine/engine.go#L153),
1. an objective `o` is fetched via `GetObjectiveByChannelId`
2. an "objective event" `e` is created
3. `o.update(e)` is called
4. `o.crank()` is called

What should happen if `DFO` and `DDO` both existed in the store? Should that be impossible?

(inserted by @NiloCK) Related **Todo:** `GetObjectiveByChannelId` has no explicit tests.

andrewgordstewart:
Furthermore, what should happen if a chain event is triggered for a ledger channel?

In this case, `GetObjectiveByChannelId` may return:
- an objective running for the ledger channel
- any virtual fund/defund objective which involves the ledger channel

geoknee:
Here are some options: 

1. Maintain the following principle: A channel may be owned by exactly one approved Objective at any given time. 



A DFO and a DDO should not both be approved and crankable at the same time. When a DDO exists and is approved, any DFO for that channel should have been terminated. Hence, GetObjectiveByChannelId mostly stays the same (perhaps a rename), and we solve the problem by carefully managing Objectives and their relation to Channels. 


```
event --> objective-bar --> channel
```
                  objective-baz
event --> objective-bar --> channel

3.  The store could have an API like `GetDirectFundObjectiveByChannelID`, `GetDirectDefundObjectiveByChannelID`, etc.  The engine inspects chain events to infer their type. Deposit events are routed to a DirectFundObjective (if it exists), and AssetTransferred events are routed to a DirectDefundObjective (if it exists). Challenge events spawn a new objective. If the objective does not exist, either it is spawned or the event is dropped.

4. Instead of routing events to objectives, we route them directly to channels. I.e. channels could have an `update()` method. In this model, the engine looks up which objectives are interested in a channel that was just updated, and cranks all of them. 




But the problem is broader than that

kerzhner:
[...]

PR #538, opened 2022-04-13
Separate chainservice.Event from ObjectiveEvent 

kerzhner:
Fixes https://github.com/statechannels/go-nitro/issues/490.

This PR separates the logic for handling chain events. When the engine processes a chain event:
1. There must be an active objective for the channel targeted by the chain event.
2. The objective must be able to handle chain events.

Every objective decides whether it handles chain events. If an objective handles chain events but receives an unexpected chain event type, the objective returns an error.

### DIFF
--- a/client/engine/chainservice/chainservice.go
+++ b/client/engine/chainservice/chainservice.go
@@ -9,7 +9,6 @@ import (
 // Event dictates which methods all chain events must implement
 type Event interface {
 	ChannelID() types.Destination
-	GetBlockNum() uint64
 }
 
 // CommonEvent declares fields shared by all chain events
@@ -29,10 +28,6 @@ func (de DepositedEvent) ChannelID() types.Destination {
 	return de.channelID
 }
 
-func (de DepositedEvent) GetBlockNum() uint64 {
-	return de.BlockNum
-}
-
 // AllocationUpdated is an internal representation of the AllocatonUpdated blockchain event
 type AllocationUpdatedEvent struct {
 	CommonEvent
@@ -41,10 +36,6 @@ type AllocationUpdatedEvent struct {
 
 func (de AllocationUpdatedEvent) ChannelID() types.Destination {
 	return de.channelID
-}
-
-func (de AllocationUpdatedEvent) GetBlockNum() uint64 {
-	return de.BlockNum
 }
 
 // todo implement other event types
--- a/client/engine/chainservice/chainservice.go
+++ b/client/engine/chainservice/chainservice.go
@@ -6,13 +6,27 @@ import (
 	"github.com/statechannels/go-nitro/types"
 )
 
-// ChainEvent is an internal representation of a blockchain event
-type Event struct {
-	ChannelId          types.Destination
+type Event interface {
+	ChannelId() types.Destination
+}
+
+// DepositedEvent is an internal representation of the deposited blockchain event
+type DepositedEvent struct {
+	IntoChannel        types.Destination
 	Holdings           types.Funds // indexed by asset
 	AdjudicationStatus protocols.AdjudicationStatus
 	BlockNum           uint64
 }
+
+func (de DepositedEvent) ChannelId() types.Destination {
+	return de.IntoChannel
+}
+
+// todo implement other event types
+// AllocationUpdated
+// Concluded
+// ChallengeRegistered
+// ChallengeCleared
 
 type ChainService interface {
 	Out() <-chan Event
--- a/client/engine/chainservice/chainservice.go
+++ b/client/engine/chainservice/chainservice.go
@@ -28,6 +28,10 @@ func (de DepositedEvent) ChannelId() types.Destination {
 // ChallengeRegistered
 // ChallengeCleared
 
+type ChainEventHandler interface {
+	UpdateWithChainEvent(event Event) (protocols.Objective, error)
+}
+
 type ChainService interface {
 	Out() <-chan Event
 	In() chan<- protocols.ChainTransaction
--- a/client/engine/chainservice/chainservice.go
+++ b/client/engine/chainservice/chainservice.go
@@ -11,12 +11,16 @@ type Event interface {
 	GetBlockNum() uint64
 }
 
-// DepositedEvent is an internal representation of the deposited blockchain event
-type DepositedEvent struct {
+type CommonEvent struct {
 	ChannelId          types.Destination
-	Holdings           types.Funds // indexed by asset
 	AdjudicationStatus protocols.AdjudicationStatus
 	BlockNum           uint64
+}
+
+// DepositedEvent is an internal representation of the deposited blockchain event
+type DepositedEvent struct {
+	CommonEvent
+	Holdings types.Funds // indexed by asset
 }
 
 func (de DepositedEvent) GetChannelId() types.Destination {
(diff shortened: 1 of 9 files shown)


client/engine/engine.go
```diff
@@ -15,6 +15,16 @@ import (
 	"github.com/statechannels/go-nitro/protocols/virtualfund"
 )
 
+// UnhandledChainEvent is an engine error when the the engine cannot process a chain event
+type UnhandledChainEvent struct {
```
NiloCK, line 19:
```suggestion
type ErrUnhandledChainEvent struct {
```

client/engine/engine.go
```diff
@@ -15,6 +15,16 @@ import (
...
+	event     chainservice.Event
+	objective protocols.Objective
+}
+
+func (uce *UnhandledChainEvent) Error() string {
+	return fmt.Sprintf("chain event could not be handled %#v by objective %#v", uce.event, uce.objective)
```
NiloCK, line 25:
```suggestion
	return fmt.Sprintf("chain event %#v could not be handled by objective %#v", uce.event, uce.objective)
```

client/engine/engine.go
```diff
@@ -152,26 +162,20 @@ func (e *Engine) handleMessage(message protocols.Message) (ObjectiveChangeEvent,
...
 	e.logger.Printf("handling chain event %v", chainEvent)
-	objective, ok := e.store.GetObjectiveByChannelId(chainEvent.ChannelId)
+	objective, ok := e.store.GetObjectiveByChannelId(chainEvent.GetChannelId())
 	if !ok {
-		// Because the MockChain and SimpleChainService broadcast all events to all subscribers,
-		// it is likely that a client receives an irrelevant event. So we log and continue without
-		// returning an error.
-		e.logger.Printf("handleChainEvent: No objective in store for channel with id %s", chainEvent.ChannelId)
-		return ObjectiveChangeEvent{}, nil
+		return ObjectiveChangeEvent{}, &UnhandledChainEvent{event: chainEvent, objective: objective}
```
NiloCK, line 167: This returns something like "chain event `x` could not be handled by objective nil", which is possibly less descriptive than we would like.

NiloCK: maybe:

```
type UnhandledChainEvent struct {
	event     chainservice.Event
	objective protocols.Objective
	reason string
}

func (uce *UnhandledChainEvent) Error() string {
	return fmt.Sprintf("chain event could not be handled %#v by objective %#v: %s", uce.event, uce.objective, uce.reason)
}
```
which could be supplied things like "objective not found", "objective not a chainEventHandler", etc.

(ignores other suggestions)

client/engine/chainservice/chainservice.go
```diff
@@ -6,14 +6,57 @@ import (
 	"github.com/statechannels/go-nitro/types"
 )
 
-// ChainEvent is an internal representation of a blockchain event
-type Event struct {
+// Event dictates which methods all chain events must implement
+type Event interface {
+	GetChannelId() types.Destination
```
NiloCK, line 11: nit, but a nit from [on high](https://go.dev/doc/effective_go) (ctrl+f getters):

> Go doesn't provide automatic support for getters and setters. There's nothing wrong with providing getters and setters yourself, and it's often appropriate to do so, but it's neither idiomatic nor necessary to put Get into the getter's name. If you have a field called owner (lower case, unexported), the getter method should be called Owner (upper case, exported), not GetOwner

```suggestion
	ChannelID() types.Destination
	BlockNum() uint64
```

NiloCK: (I've lost the documentation, but `ID` is considered an initialism by picky Go persons, and should be denoted either as `id` or `ID`)

protocols/directfund/directfund.go
```diff
@@ -203,12 +204,24 @@ func (o Objective) Update(event protocols.ObjectiveEvent) (protocols.Objective,
...
+}
+
+func (o Objective) UpdateWithChainEvent(event chainservice.Event) (protocols.Objective, error) {
+	updated := o.clone()
+
+	de, ok := event.(chainservice.DepositedEvent)
+	if !ok {
+		return &updated, fmt.Errorf("objective %+v cannot handle event %+v", updated, event)
+	}
+	if de.Holdings != nil && de.GetBlockNum() > updated.latestBlockNumber {
```
NiloCK, line 218: how reorg safe is this check?

I'm not well informed on the specifics, but is a check against total difficulty a better approach than blocknum?

Do we intend to rely on the chain service to apply these checks, or should it be in-protocol?

protocols/directdefund/directdefund.go
```diff
@@ -97,60 +99,86 @@ func (o Objective) Id() protocols.ObjectiveId {
...
+	return &updated, nil
+}
+
+func (o Objective) UpdateWithChainEvent(event chainservice.Event) (protocols.Objective, error) {
+	updated := o.clone()
+	de, ok := event.(chainservice.AllocationUpdatedEvent)
+	if !ok {
+		return &updated, fmt.Errorf("objective %+v cannot handle event %+v", updated, event)
+	}
+	// todo: check block number
```
NiloCK, line 160: can this be included now?

NiloCK (approved):
This is great!

client/engine/engine.go
kerzhner: Addressed in d23f4fd.

protocols/directdefund/directdefund.go
```diff
@@ -97,60 +99,86 @@ func (o Objective) Id() protocols.ObjectiveId {
...
 
 	updated := o.clone()
 	updated.C.AddSignedStates(event.SignedStates)
-	if event.Holdings != nil {
-		updated.C.OnChainFunding = event.Holdings
+
+	return &updated, nil
+}
+
+func (o Objective) UpdateWithChainEvent(event chainservice.Event) (protocols.Objective, error) {
```
NiloCK, line 154:
```suggestion
// UpdateWithChainEvent updates the objective with observed on-chain data.
//
// Only Channel Deposit events are currently handled.
func (o Objective) UpdateWithChainEvent(event chainservice.Event) (protocols.Objective, error) {
```

client/engine/engine.go
kerzhner: Good call! Done in 72b1b02

client/engine/chainservice/chainservice.go
kerzhner: Makes sense to me. Addressed in 1e939c8.

kerzhner: Also, `GetBlockNum` has been removed in 1ee011a.

protocols/directfund/directfund.go
kerzhner: These are great questions! We should reevaluate our chain reorg strategy. That work is outside the scope of this PR, so I will file a new issue.

protocols/directdefund/directdefund.go
kerzhner: This PR keeps the existing logic for reorgs in place. I added this `todo` as I noticed that the direct defunder did not address reorgs. As per above comment, this is important work that should be done at a later point.

protocols/directfund/directfund.go
kerzhner: Filed https://github.com/statechannels/go-nitro/issues/549

PR #540, opened 2022-04-13
Round 2 of Virtual Defunding

### LINKED ISSUE #480: Update Virtual Defunding Protocol to defund Ledger Channel
This continues on the work of #478 and should update the virtual defunding protocol to use the `consensus_channel` to remove the guarantee from the ledger channel.

lalexgap:
Fixes #480.

This PR extends the virtual defunding protocol to handle defunding the ledger channels. 

To help with organization test helpers have been moved to a separate file.

#### Code quality

#### Project management

### DIFF
--- a/channel/consensus_channel/consensus_channel.go
+++ b/channel/consensus_channel/consensus_channel.go
@@ -310,6 +310,14 @@ func (lo *LedgerOutcome) Clone() LedgerOutcome {
 	}
 }
 
+func (l LedgerOutcome) Left() Balance {
+	return l.left
+}
+
+func (l LedgerOutcome) Right() Balance {
+	return l.right
+}
+
 // NewLedgerOutcome creates a new ledger outcome with the given asset address and balances and guarantees
 func NewLedgerOutcome(assetAddress types.Address, left, right Balance, guarantees []Guarantee) *LedgerOutcome {
 	guaranteeMap := make(map[types.Destination]Guarantee, len(guarantees))
--- a/channel/consensus_channel/consensus_channel.go
+++ b/channel/consensus_channel/consensus_channel.go
@@ -83,6 +83,11 @@ func newConsensusChannel(
 
 }
 
+// FixedPart returns the fixed part of the channel
+func (c *ConsensusChannel) FixedPart() state.FixedPart {
+	return c.fp
+}
+
 // Receive accepts a proposal signed by the ConsensusChannel counterparty,
 // validates its signature, and performs updates the proposal queue and
 // consensus state
--- a/channel/consensus_channel/consensus_channel.go
+++ b/channel/consensus_channel/consensus_channel.go
@@ -107,6 +107,24 @@ func (c *ConsensusChannel) Includes(g Guarantee) bool {
 	return c.current.Outcome.includes(g)
 }
 
+// IncludesTarget returns whether or not the consensus state includes a guarantee targeting the given channel
+func (c *ConsensusChannel) IncludesTarget(target types.Destination) bool {
+	return c.current.Outcome.includesTarget(target)
+}
+
+// RemoveProposedFor returns whether or not a proposal exists to remove the guaranatee for the target
+func (c *ConsensusChannel) RemoveProposedFor(target types.Destination) bool {
+	for _, p := range c.proposalQueue {
+		if p.Proposal.Type() == RemoveProposal {
+			remove := p.Proposal.ToRemove
+			if remove.Target == target {
+				return true
+			}
+		}
+	}
+	return false
+}
+
 // IsLeader returns true if the calling client is the leader of the channel,
 // and false otherwise
 func (c *ConsensusChannel) IsLeader() bool {
@@ -304,6 +322,12 @@ func NewLedgerOutcome(assetAddress types.Address, left, right Balance, guarantee
 		right:        right,
 		guarantees:   guaranteeMap,
 	}
+}
+
+// includesTarget returns true when the receiver includes a guarantee that targets the given destination
+func (o *LedgerOutcome) includesTarget(target types.Destination) bool {
+	_, found := o.guarantees[target]
+	return found
 }
 
 // includes returns true when the receiver includes g in its list of guarantees.
--- a/channel/consensus_channel/consensus_channel.go
+++ b/channel/consensus_channel/consensus_channel.go
@@ -118,8 +118,8 @@ func (c *ConsensusChannel) IncludesTarget(target types.Destination) bool {
 	return c.current.Outcome.includesTarget(target)
 }
 
-// RemoveProposedFor returns whether or not a proposal exists to remove the guaranatee for the target
-func (c *ConsensusChannel) RemoveProposedFor(target types.Destination) bool {
+// HasRemovalBeenProposedFor returns whether or not a proposal exists to remove the guaranatee for the target
+func (c *ConsensusChannel) HasRemovalBeenProposedFor(target types.Destination) bool {
 	for _, p := range c.proposalQueue {
 		if p.Proposal.Type() == RemoveProposal {
 			remove := p.Proposal.ToRemove
--- a/channel/consensus_channel/consensus_channel.go
+++ b/channel/consensus_channel/consensus_channel.go
@@ -2,6 +2,7 @@
 package consensus_channel
 
 import (
+	"bytes"
 	"fmt"
 	"math/big"
 	"sort"
@@ -540,7 +541,7 @@ func (p *Proposal) TurnNum() uint64 {
 
 // equal returns true if the supplied Proposal is deeply equal to the receiver, false otherwise.
 func (p *Proposal) equal(q *Proposal) bool {
-	return p.ToAdd.equal(q.ToAdd) && p.ToRemove == q.ToRemove
+	return p.ToAdd.equal(q.ToAdd) && p.ToRemove.equal(q.ToRemove)
 }
 
 // SignedProposal is a Proposal with a signature on it
@@ -614,6 +615,23 @@ func (a Add) equal(a2 Add) bool {
 		return false
 	}
 	return types.Equal(a.LeftDeposit, a2.LeftDeposit)
+}
+
+func (r Remove) equal(r2 Remove) bool {
+	if r.turnNum != r2.turnNum {
+		return false
+	}
+	if !bytes.Equal(r.Target.Bytes(), r2.Target.Bytes()) {
+
+		return false
+	}
+	if !types.Equal(r.LeftAmount, r2.LeftAmount) {
+		return false
+	}
+	if !types.Equal(r.RightAmount, r2.RightAmount) {
+		return false
+	}
+	return true
 }
 
 var ErrIncorrectTurnNum = fmt.Errorf("incorrect turn number")
--- a/channel/consensus_channel/consensus_channel.go
+++ b/channel/consensus_channel/consensus_channel.go
@@ -316,14 +316,6 @@ func (lo *LedgerOutcome) Clone() LedgerOutcome {
 	}
 }
 
-func (l LedgerOutcome) Left() Balance {
-	return l.left
-}
-
-func (l LedgerOutcome) Right() Balance {
-	return l.right
-}
-
 // NewLedgerOutcome creates a new ledger outcome with the given asset address and balances and guarantees
 func NewLedgerOutcome(assetAddress types.Address, left, right Balance, guarantees []Guarantee) *LedgerOutcome {
 	guaranteeMap := make(map[types.Destination]Guarantee, len(guarantees))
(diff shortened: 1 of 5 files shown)


protocols/virtualdefund/virtualdefund.go
```diff
@@ -166,6 +169,14 @@ func (o *Objective) clone() Objective {
 	}
 	clone.MyRole = o.MyRole
 
+	// TODO: Properly clone the consensus channels
+	if o.ToMyLeft != nil {
+		clone.ToMyLeft = o.ToMyLeft
+	}
+	if o.ToMyRight != nil {
+		clone.ToMyRight = o.ToMyRight
```
lalexgap, line 177: We should probably create an issue for this.

channel/consensus_channel/consensus_channel.go
```diff
@@ -107,6 +113,24 @@ func (c *ConsensusChannel) Includes(g Guarantee) bool {
 	return c.current.Outcome.includes(g)
 }
 
+// IncludesTarget returns whether or not the consensus state includes a guarantee targeting the given channel
+func (c *ConsensusChannel) IncludesTarget(target types.Destination) bool {
+	return c.current.Outcome.includesTarget(target)
+}
```
geoknee, line 119: Is this check thorough enough? Is there an attack where a guarantee is added with (say) a zero amount?

channel/consensus_channel/consensus_channel.go
```diff
@@ -107,6 +113,24 @@ func (c *ConsensusChannel) Includes(g Guarantee) bool {
...
+
+// RemoveProposedFor returns whether or not a proposal exists to remove the guaranatee for the target
+func (c *ConsensusChannel) RemoveProposedFor(target types.Destination) bool {
```
geoknee, line 122: Nit: `RemoveProposedFor` sounds like _imperative_ code. I would prefer something like `HasRemovalBeenProposedFor`.

protocols/virtualdefund/helpers_test.go
```diff
@@ -0,0 +1,319 @@
...
+	ta "github.com/statechannels/go-nitro/internal/testactors"
+	"github.com/statechannels/go-nitro/internal/testhelpers"
+	. "github.com/statechannels/go-nitro/internal/testhelpers"
+	"github.com/statechannels/go-nitro/protocols"
+	"github.com/statechannels/go-nitro/types"
+)
+
+// assertSideEffectsContainsMessageWith fails the test instantly if the supplied side effects does not contain a message for the supplied actor with the supplied expected signed state.
+// TODO: This is copied from https://github.com/statechannels/go-nitro/blob/0722a1127241583944f32efa0638012f64b96bf0/protocols/virtualfund/virtualfund_single_hop_test.go#L409
+// We should probably establish some shared helpers
```
geoknee, line 22: I think something is off with the godoc here (mismatch in function name).

Is it possible to avoid adding all 300 lines of test helpers by reusing anything in `internal/`?

protocols/virtualdefund/virtualdefund.go
```diff
@@ -55,7 +55,7 @@ type Objective struct {
 const ObjectivePrefix = "VirtualDefund-"
 
 // newObjective constructs a new virtual defund objective
-func newObjective(preApprove bool, vFixed state.FixedPart, initialOutcome outcome.SingleAssetExit, paidToBob *big.Int, myRole uint) Objective {
+func newObjective(preApprove bool, vFixed state.FixedPart, initialOutcome outcome.SingleAssetExit, paidToBob *big.Int, toMyLeft, toMyRight *consensus_channel.ConsensusChannel, myRole uint) Objective {
```
geoknee, line 58: Shouldn't we match the pattern here

https://github.com/statechannels/go-nitro/blob/ca82de1b3ddbf963b0fdb53ddad10633d29b613c/protocols/virtualfund/virtualfund.go#L131-L132

and pass a _getter_ instead of the actual ConsensusChannels?

protocols/virtualdefund/virtualdefund.go
```diff
@@ -203,8 +214,28 @@ func (o Objective) Crank(secretKey *[]byte) (protocols.Objective, protocols.Side
...
+			return &o, protocols.SideEffects{}, WaitingForNothing, fmt.Errorf("error updating ledger funding: %w", err)
+		}
+		sideEffects.Merge(ledgerSideEffects)
+	}
+
+	if fullyDefunded := updated.isLeftDefunded() && updated.isRightDefunded(); !fullyDefunded {
+		return &updated, sideEffects, WaitingForCompleteLedgerDefunding, nil
+	} else {
+		return &updated, sideEffects, WaitingForNothing, nil
+	}
```
geoknee, line 237: I think you don't need the `else` block here.

protocols/virtualdefund/virtualdefund.go
```diff
@@ -217,6 +248,71 @@ func (o Objective) fullySigned() bool {
...
 
+// isAlice returns true if the receiver represents participant 0 in the virtualdefund protocol
+func (o Objective) isAlice() bool {
+	return o.MyRole == 0
+}
+
+// isBob returns true if the receiver represents participant 2 in the virtualdefund protocol
+func (o Objective) isBob() bool {
+	return o.MyRole == 2
+}
```
geoknee, line 259: Occurs to me that there is scope for sharing production code between virtualfund and virtualdefund (as well as test code).

protocols/virtualdefund/virtualdefund.go
```diff
@@ -217,6 +248,71 @@ func (o Objective) fullySigned() bool {
...
+// ledgerProposal generates a ledger proposal to remove the guarantee for V for ledger
+func (o Objective) ledgerProposal(ledger *consensus_channel.ConsensusChannel) consensus_channel.Proposal {
+	left := o.finalOutcome().Allocations[0].Amount
+	right := o.finalOutcome().Allocations[1].Amount
+
+	return consensus_channel.NewRemoveProposal(ledger.Id, FinalTurnNum, o.VId(), left, right)
+}
+
+// defundLedger updates the ledger channel to remove the guarantee that funds V.
+func (o *Objective) defundLedger(ledger *consensus_channel.ConsensusChannel, sk *[]byte) (protocols.SideEffects, error) {
```
geoknee, line 270: Nit: this sounds like imperative code again. Would it be better to name this (or at least describe this in the godoc) as "attempting to defund the ledger by generating side effects "?

protocols/virtualdefund/virtualdefund.go
```diff
@@ -217,6 +248,71 @@ func (o Objective) fullySigned() bool {
...
+			if err != nil {
+				return protocols.SideEffects{}, fmt.Errorf("could not sign proposal: %w", err)
+			}
+
+			message := o.createSignedProposalMessage(sp, ledger)
+			sideEffects.MessagesToSend = append(sideEffects.MessagesToSend, message)
+
+		} else {
+
+		}
```
geoknee, line 303: You definitely don't need this else block :-)

protocols/virtualdefund/virtualdefund.go
geoknee: Also, the name is ambiguous. We aren't trying to defund the ledger here, we are trying to defund the virtual channel.

Something like `generateProposalToRemoveFromLedger`?

protocols/virtualdefund/virtualdefund.go
```diff
@@ -217,6 +248,71 @@ func (o Objective) fullySigned() bool {
...
+	}
+
+	return sideEffects, nil
+}
+
+// VId returns the channel id of the virtual channel.
+func (o Objective) VId() types.Destination {
```
geoknee, line 310: I think we have a guideline that _all_ methods should have a pointer receiver unless there is a really good reason to have a value receiver.

protocols/virtualdefund/virtualdefund.go
```diff
@@ -228,6 +324,41 @@ func (o Objective) signedByMe() bool {
...
+	if o.ToMyLeft == nil {
+		return true
+	}
+
+	included := o.ToMyLeft.IncludesTarget(o.VId())
+	return !included
+}
+
+// createSignedProposalMessage returns a signed proposal message addressed to the counterparty in the given ledger
+func (o *Objective) createSignedProposalMessage(sp consensus_channel.SignedProposal, ledger *consensus_channel.ConsensusChannel) protocols.Message {
```
geoknee, line 350: This code ought to be shared. Do we need a `wire` package?

geoknee (approved):

protocols/virtualdefund/virtualdefund.go
```diff
@@ -166,6 +169,14 @@ func (o *Objective) clone() Objective {
 	}
 	clone.MyRole = o.MyRole
 
+	// TODO: Properly clone the consensus channels
```
geoknee, line 172: I've started on #581. It might make sense to cherry pick [7b7724f](https://github.com/statechannels/go-nitro/pull/581/commits/7b7724f73ea1e49aec2ba977e8ec7b92ef8cde47) onto this branch, and then you can call the (unfinished at the moment) clone method here?

protocols/virtualdefund/virtualdefund.go
lalexgap: Removed in https://github.com/statechannels/go-nitro/pull/540/commits/1e7a20f0d30ba31c84a43da41854c91156d2e381

protocols/virtualdefund/virtualdefund.go
lalexgap: Removed in https://github.com/statechannels/go-nitro/pull/540/commits/1e7a20f0d30ba31c84a43da41854c91156d2e381

channel/consensus_channel/consensus_channel.go
lalexgap: Renamed in https://github.com/statechannels/go-nitro/pull/540/commits/53c0a077a286bc8dbf2655fde0d77a4f56a741c0

protocols/virtualdefund/virtualdefund.go
lalexgap: Nit: this sounds like imperative code again. Would it be better to name this (or at least describe this in the godoc) as "attempting to defund the ledger by generating side effects "?

This function modifies the `ledger` consensus channel by adding the proposal to the channel, so I think naming it `generateProposalToRemoveFromLedger` could be slightly misleading.

I've renamed this to `updateLedgerToRemoveGuarantee` which mirrors `updateLedgerWithGuarantee` in virtual funding.

geoknee: Right: I think what is throwing me is that it sounds like we are unilaterally updating the ledger. When all we are doing is updating our side and generating a message to send. But maybe it's too fussy to try and reflect that in the name! 

`updateLedgerToRemoveFromLedger` is definitely better than `defundLedger`.

protocols/virtualdefund/virtualdefund.go
geoknee: Update: #581 is pretty much finished and is just pending a review.

channel/consensus_channel/consensus_channel.go
lalexgap: `IncludesTarget` isn't used to determine if we sign a state so I don't think it needs to be concerned with safety. `IncludesTarget` is just used to check the if the state we've already signed includes a guarantee.

The proposal system in the consensus channel should enforce that we only sign a state with the specific outcome we expect.

protocols/virtualdefund/virtualdefund.go
lalexgap: If we [share code](https://github.com/statechannels/go-nitro/pull/540#discussion_r857812269) between `virtualfund` and `virtualdefund` we could probably include that there for now.

protocols/virtualdefund/virtualdefund.go
lalexgap: Currently this is just a private `constructor` used mostly for testing. I think we can kick this down the road and implement a proper constructor when we need to call it from the engine in https://github.com/statechannels/go-nitro/issues/564 ?

protocols/virtualdefund/helpers_test.go
lalexgap: I think we could refactor a bunch of these helpers to be shared between `virtualFund` and `virtualDefund`. I've also moved some of the assertion helpers to `internal` in https://github.com/statechannels/go-nitro/pull/540/commits/ee43f127973b396d63a270888056cde676a3e652 so they can be used by other tests.

protocols/virtualdefund/virtualdefund.go
lalexgap: Awesome! I've approved #581, merged it, and rebased this PR.

protocols/virtualdefund/virtualdefund.go
lalexgap: I'm going to defer this to a separate PR as `virtualdefund` has a couple value receivers that should be updated.

PR #541, opened 2022-04-14
Reinstate gas benchmarks

### LINKED ISSUE #518: Re-enable gas-benchmarks
We have most of the code in place, but the tests are not run in CI.

---

> * should a function for returning the last element of `VariablePart[]` be added? (e.g. for usage [here](https://github.com/statechannels/go-nitro/pull/508/files#diff-483aad04b1f90d19c78f0fd86f7ebb7869c6c194797c3d0774ce9e80dff5a2b8R91) and [here](https://github.com/statechannels/go-nitro/pull/508/files#diff-483aad04b1f90d19c78f0fd86f7ebb7869c6c194797c3d0774ce9e80dff5a2b8R191)) . The benefit from it is clearer code (repetitions are in a separate function), but the drawback is slightly increased gas usage (by 100 gas).

This could also be tackled in a follow-up PR. The links you gave don't seem to point to a particular line, so I'm not sure exactly where you propose to use such a function. In general I would say it is worth a little bit of gas to improve the quality of the code. 

This is reminding me that we should reintroduce some tracking of gas costs in CI, to make these decisions easier to make.

_Originally posted by @geoknee in https://github.com/statechannels/go-nitro/issues/508#issuecomment-1094779666_

geoknee:
Closes #518 

The "sad case virtual funded channel" case has been skipped, since we expect this to change considerably during upcoming contract changes.

### DIFF
--- a/nitro-protocol/gas-benchmarks/benchmark.test.ts
+++ b/nitro-protocol/gas-benchmarks/benchmark.test.ts
@@ -1,3 +1,4 @@
+import {encodeOutcome} from '../src';
 import {MAGIC_ADDRESS_INDICATING_ETH} from '../src/transactions';
 
 import {
@@ -34,7 +35,7 @@ async function addResidualTokenBalance(asset: string) {
 describe('Consumes the expected gas for deployments', () => {
   it(`when deploying the NitroAdjudicator`, async () => {
     await expect(await nitroAdjudicator.deployTransaction).toConsumeGas(
-      gasRequiredTo.deployInfrastructureContracts.vanillaNitro.NitroAdjudicator
+      gasRequiredTo.deployInfrastructureContracts.satp.NitroAdjudicator
     );
   });
 });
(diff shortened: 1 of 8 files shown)


nitro-protocol/gas-benchmarks/benchmark.test.ts
```diff
@@ -164,31 +164,32 @@ describe('Consumes the expected gas for sad-path exits', () => {
...
+      gasRequiredTo.ETHexitSadLedgerFunded.satp.challengeL +
+        gasRequiredTo.ETHexitSadLedgerFunded.satp.transferAllAssetsL +
+        gasRequiredTo.ETHexitSadLedgerFunded.satp.challengeX +
+        gasRequiredTo.ETHexitSadLedgerFunded.satp.transferAllAssetsX
+    ).toEqual(gasRequiredTo.ETHexitSadLedgerFunded.satp.total);
   });
 
-  it(`when exiting a virtual funded (with ETH) channel`, async () => {
+  // TODO unskip this test when the contracts are satp compatible
+  it.skip(`when exiting a virtual funded (with ETH) channel`, async () => {
```
geoknee, line 192: We should make an issue to unskip this.

nksazonov (approved):

PR #542, opened 2022-04-14
Update _hashState to be called with SingleAssetExit[] outcome

### LINKED ISSUE #532: Update `_hashState` to be called with `SingleAssetExit[] outcome`
> * update `_hashState` to be called with `SingleAssetExit[] outcome`

We should tackle this in a follow-up PR.

_Originally posted by @geoknee in https://github.com/statechannels/go-nitro/issues/508#issuecomment-1094779666_

nksazonov:
Closes #532.

### DIFF
--- a/nitro-protocol/contracts/ForceMove.sol
+++ b/nitro-protocol/contracts/ForceMove.sol
@@ -127,7 +127,7 @@ contract ForceMove is IForceMove, StatusManager {
         bytes32 challengeStateHash = _hashState(
             channelId,
             variablePartAB[0].appData,
-            Outcome.encodeExit(variablePartAB[0].outcome),
+            variablePartAB[0].outcome,
             turnNumRecord,
             variablePartAB[0].isFinal
         );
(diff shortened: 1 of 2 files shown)


geoknee (approved):
Great. I'd be keen to adopt the following merge strategy:

1. Wait for #541 to be merged.
2. Rebase this on `main`
3. See gas benchmarks fail
4. Update `gas.ts` so benchmarks pass
5. Understand the gas savings enabled by this change.

It's up to you though @nksazonov , as none of this ⬆️ is really necessary and will probably take about half an hour.

nksazonov:
> It's up to you though @nksazonov , as none of this arrow_up is really necessary and will probably take about half an hour.

This merge strategy seems logical though. We have a few similar issues, which should optimize gas usage, therefore such workflow will be consistent with them and enable to see gas improvements at ease.

geoknee:
Looks like a ~5% saving on most methods in this case 🎉 

Meta-comment: the DevEx for fixing up `gas.ts` could be nicer. We could automate updating the file with the "real numbers" (right now it is done manually).

nksazonov:
> Meta-comment: the DevEx for fixing up `gas.ts` could be nicer. We could automate updating the file with the "real numbers" (right now it is done manually).

That is a great idea indeed!

PR #545, opened 2022-04-14
Introduce workflow for Go <> Solidity tests

### LINKED ISSUE #486: Test `go-nitro`'s `state.Sign` against nitro-protocol contracts
`go-nitro` currently encodes states differently from `nitro-protocol`. This difference is addressed in https://github.com/statechannels/go-nitro/pull/427.

We should test the `go-nitro` encoding/decoding against the encoding/decoding performed by the `nitro-protocol` smart contracts.

geoknee:
Closes #486 .

This PR Introduces:
* a new package (sub package of the `chainservice`) named `NitroAdjudicator`
* shell script to generate Go bindings to the `NitroAdjudicator`(solidity) contract.
* The **output** of that shell script, which is an autogenerated file inserted into the `NitroAdjudicator` package.
* A new test `challenge_test.go` which uses said bindings, along With a [geth `SimulatedBackend`](https://goethereumbook.org/en/client-simulated/) to test the `challenge` method on the (new) Adjudicator contract. The test is mostly ported from `nitro-protocol/test/contracts/ForceMove/challenge.test.ts`

Plus:
* exporting / adding a full utilities to get this test to 
  * run (it is very fast), and to 
  * pass 🎉 

The upshots are:
* Our Go and solidity files are now coupled together. The Go code is now partially derived from the solidity code.
* The test integrates enough code to give us confidence that we are generating and recovering state signatures consistently in our client / on-chain code. 
* We can write more tests like this, to test other methods from a Go environment. 
* We can use the `NitroAdjudicator.go` in our production code when we write a proper chain service. I haven't dug into it much, but there looks to be some nice event / subscription entry points in there (as well as the deployment and transaction entry points I have used on this PR). 

TODO:
- [ ] get this to work in CI (requires installing solc and abigen so that we can regenerate the bindings and check they are the same as those checked in). I will tackle this in a separate PR #548 
- [ ] separate from other tests? I don't think this is necessary just yet.


readme.md
```diff
@@ -59,6 +59,24 @@ and executes logic from the `protocols` package. Data required for the secure cr
...
+We now use `solc` and `abigen` to generate Go bindings from our `*.sol` source files. This is achieved by running the `generate_adjudicator_bindings.sh` script at the top level of the repo. Because our `*.sol` files depend on external projects via `node_modules`, to run this script successfully you must:
+
+- have successfully run `npm install` in the `nitro-protocol` directory.
+- have [solc](https://docs.soliditylang.org/en/v0.8.13/installing-solidity.html) installed at the correct version (currently 0.7.4, see the CI or linting config for a hint if you think it may have changed)
+- have [abigen](https://geth.ethereum.org/docs/install-and-build/installing-geth) (a tool shipped with go-ethereum) installed.
+
+The resulting Go bindings file is _checked-in_ to the repository. Although it is auto-generated from on-chain source code, it effectively forms part of the off-chain source code.
+
+If you alter the contracts, you should regenerate the bindings file at check it in. A github action will run which will check that you have done this correctly.
+
```
geoknee, line 79: This addition to the readme explains what I am trying to do, here.

readme.md
```diff
@@ -59,6 +59,24 @@ and executes logic from the `protocols` package. Data required for the secure cr
...
+`go-nitro` was originally being developed "off-chain first". Alongside that development, we are now building out the on-chain component (solidity contracts): these are housed in the `nitro-protocol` directory. This directory contains an npm package with a hardhat / typechain / jest toolchain. This code originates from the nitro-protocol [npm package](https://www.npmjs.com/package/@statechannels/nitro-protocol), but is being modified / upgraded. The new version of the code is sometimes called nitro v2 or "satp".
+
+Initially, these two development streams will remain independent, with each having its own continuous integration checks.
+
+Over time, the development streams will become more coupled.
+
+We now use `solc` and `abigen` to generate Go bindings from our `*.sol` source files. This is achieved by running the `generate_adjudicator_bindings.sh` script at the top level of the repo. Because our `*.sol` files depend on external projects via `node_modules`, to run this script successfully you must:
+
+- have successfully run `npm install` in the `nitro-protocol` directory.
```
kerzhner, line 72: Any particular reason we are using `npm` instead of `yarn`?

generate-adjudicator-bindings.sh
```diff
@@ -0,0 +1,6 @@
+set -e
```
kerzhner, line 1: When I pulled down this feature branch, `generate-adjudicator-bindings.sh` is not executable by default

readme.md
```diff
@@ -59,6 +59,24 @@ and executes logic from the `protocols` package. Data required for the secure cr
...
+- have [solc](https://docs.soliditylang.org/en/v0.8.13/installing-solidity.html) installed at the correct version (currently 0.7.4, see the CI or linting config for a hint if you think it may have changed)
```
kerzhner, line 73: On my machine, `solc` is installed via homebrew. I am only able to install solidity 0.7.6 with homebrew as that is the latest 0.7 version.  Is it worth it to bump the contracts to the latest 0.7 version?

readme.md
kerzhner: When I try to run `npm install` in nitro-protocol directory I see:
```
➜  nitro-protocol git:(go-solidity-signature-test-rewrite) ✗ npm install
npm ERR! code ENOLOCAL
npm ERR! Could not install from "node_modules/@statechannels/devtools/node_modules/ganache-core/node_modules/web3-provider-engine/node_modules/eth-sig-util/ethereumjs-abi@git+https:/github.com/ethereumjs/ethereumjs-abi.git" as it does not contain a package.json file.
```

Any ideas on how to resolve that?

client/engine/chainservice/adjudicator/challenge_test.go
```diff
@@ -0,0 +1,146 @@
...
+	balance.SetString("10000000000000000000", 10) // 10 eth in wei
+
+	// Setup "blockchain"
+	gAlloc := map[common.Address]core.GenesisAccount{
+		address: {Balance: balance},
+	}
+	blockGasLimit := uint64(4712388)
+	sim := backends.NewSimulatedBackend(gAlloc, blockGasLimit)
+
+	// Deply Adjudicator
```
kerzhner, line 90:
```suggestion
	// Deploy Adjudicator
```

client/engine/chainservice/adjudicator/challenge_test.go
```diff
@@ -0,0 +1,146 @@
...
+	_, _, na, err := DeployNitroAdjudicator(auth, sim)
+
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Mine a block
+	sim.Commit()
```
kerzhner, line 98: Do you know why this is needed?

client/engine/chainservice/adjudicator/challenge_test.go
```diff
@@ -0,0 +1,146 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	header, err := sim.HeaderByNumber(context.Background(), receipt.BlockNumber)
+	if err != nil {
+		t.Fatal(err)
+	}
+	challengeTime := big.NewInt(int64(header.Time))
+
+	// Generate expectaiont
```
kerzhner, line 127:
```suggestion
	// Generate expectation
```

client/engine/chainservice/adjudicator/utils_test.go
```diff
@@ -0,0 +1,35 @@
...
+	"math/big"
+
+	ethAbi "github.com/ethereum/go-ethereum/accounts/abi"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/statechannels/go-nitro/abi"
+	"github.com/statechannels/go-nitro/channel/state"
+)
+
+func generateStatus(state state.State, finalizesAt *big.Int) ([]byte, error) {
+
```
kerzhner, line 13:
```suggestion
```

kerzhner (approved):
Good work, this PR looks reasonable to me! I am not able to try out `generate-adjudicator-bindings.sh` due to the npm issue mentioned.

readme.md
nksazonov: @kerzhner , I had the similar issue, but after removing `node_modules` folder and running `npm install` again it was resolved.

readme.md
nksazonov: You can use [`solc-select`](https://github.com/crytic/solc-select) to install multiple specific versions of `solc`.

generate-adjudicator-bindings.sh
nksazonov: Works fine for me though.

nksazonov (approved):

client/engine/chainservice/adjudicator/challenge.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+func hashChallengeMessage(s state.State) (types.Bytes32, error) {
+
+	digest, err := s.Hash()
+	if err != nil {
+		return types.Bytes32{}, err
+	}
+
+	encoded, err := ethAbi.Arguments{{Type: abi.Destination}, {Type: abi.String}}.Pack(digest, "forceMove")
```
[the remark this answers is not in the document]
geoknee: Its really a `Bytes32`, updated in [28cd990](https://github.com/statechannels/go-nitro/pull/545/commits/28cd990452eb6434be614d89d7311562ca745239)

generate-adjudicator-bindings.sh
geoknee: I think this means you need to explicitly run `sh generate-adjudicator-bindings.sh`? Happy to make it executable if you think that makes sense @kerzhner ?

readme.md
geoknee: One of the recommendations from our audit was to pin the solidity compiler version (so that we get a deterministic deployment from the audited source code). So I think it would be better for folks to be able to install a specific version of `solc`. There are instructions [here](https://docs.soliditylang.org/en/v0.8.13/installing-solidity.html#macos-package) for doing that with homebrew.

client/engine/chainservice/adjudicator/challenge_test.go
geoknee: This causes a new block to be mined https://goethereumbook.org/en/client-simulated/ (it doesn't happen automatically).

readme.md
kerzhner: Got it! On my machine, there is no `node_modules` folder to remove.

readme.md
kerzhner: Unfortunately, I haven't succeeded in installing Solidity version 0.7.4 from Homebrew. I am seeing an install error.

In general, I agree with pinning to a specific solidity version. For this specific case, I would be in favor of pinning at solidity 0.7.6 as the minor 0.7 version seems to have settled on patch version 0.7.6 a year and a half ago.

geoknee: We are planning on upgrading to `0.8` soon #556 .

PR #550, opened 2022-04-18
Implement `_lastVariablePart` helper function

### LINKED ISSUE #535: Introduce a `lastVariablePart` helper in `ForceMove`
Have a few doubts about the following:
- should a function for returning the last element of `VariablePart[]` be added? (e.g. for usage [here](https://github.com/statechannels/go-nitro/blob/main/nitro-protocol/contracts/ForceMove.sol#L64) and [here](https://github.com/statechannels/go-nitro/blob/main/nitro-protocol/contracts/ForceMove.sol#L91)) . The benefit from it is clearer code (repetitions are in a separate function), but the drawback is slightly increased gas usage (by 100 gas).

_Originally posted by @nksazonov in https://github.com/statechannels/go-nitro/issues/508#issuecomment-1094028710_

nksazonov:
Closes #535.

#### Gas cost changes:
Adds approximately 1% of gas cost to `exit sad` and `exit sad ledger funded` cases.
Slightly reduces deployment cost.

### DIFF
--- a/nitro-protocol/contracts/ForceMove.sol
+++ b/nitro-protocol/contracts/ForceMove.sol
@@ -61,7 +61,7 @@ contract ForceMove is IForceMove, StatusManager {
         );
 
         bytes32 channelId = _getChannelId(fixedPart);
-        uint48 largestTurnNum = variableParts[variableParts.length - 1].turnNum;
+        uint48 largestTurnNum = _lastVariablePart(variableParts).turnNum;
 
         if (_mode(channelId) == ChannelMode.Open) {
             _requireNonDecreasedTurnNumber(channelId, largestTurnNum);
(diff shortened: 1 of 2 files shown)


geoknee (approved):
This code will probably change when we tackle #496 -- because we are going to break the implicit equivalence between "last" (end of the array) and "latest" (greatest turn num). I think this change will probably make the changes easier to implement, though, so I'm in favour of merging this in. 

Side note: is `greatestSupportedState` better terminology than `latestSupportedState`?

nksazonov:
> This code will probably change when we tackle #496 -- because we are going to break the implicit equivalence between "last" (end of the array) and "latest" (greatest turn num).

Agree.

> Side note: is greatestSupportedState better terminology than latestSupportedState?

I don't think so. 'greatest' implies the biggest number in value, but also can be interpreted as biggest in size, thus has an ambiguity, which is not good for quick comprehension. 'latest' implies the last by time, which corresponds to satp execution flow.

PR #551, opened 2022-04-18
ref: remove on-chain signature fix

### LINKED ISSUE #497: Revert on-chain signature recovery to plain ecrecover
I agree. We will try to fix it from `go-nitro` client code ASAP.

_Originally posted by @nksazonov in https://github.com/statechannels/go-nitro/pull/427#discussion_r842575309_

nksazonov:
Closes #497.


nksazonov:
Waiting for #550 to be merged to update gas benchmark results.

geoknee:
Now that #545 is merged, do we expect any errors running the existing client code against the updated contracts on this branch? (I think we are expecting to need to update the client code?)

You'll want to rebase this branch anyway (for the gas benchmarks), so I guess we will find out!

geoknee (approved):

geoknee:
> Now that #545 is merged, do we expect any errors running the existing client code against the updated contracts on this branch? (I think we are expecting to need to update the client code?)
> 
> You'll want to rebase this branch anyway (for the gas benchmarks), so I guess we will find out!

@nksazonov I think I'm starting to see this cause an error https://github.com/statechannels/go-nitro/runs/6091975035?check_suite_focus=true#step:6:429 

We generate a random key in this test, which would explain why we only sometimes see an error.

nksazonov:
> We generate a random key in this test, which would explain why we only sometimes see an error.

So I assume that the correctness of the signature returned depends on the key itself.

PR #552, opened 2022-04-18
Bump contracts to solidity@0.7.6. Undo replacing `memory` with `calldata`

### LINKED ISSUE #527: Investigate keeping data in `calldata` if possible.
It might save some gas. 

---

I did a bit of digging: https://github.com/ethereum/solidity/issues/9160

At solidity 0.7.4, the functionality to copy data from `calldata` to `memory` for a struct (which we would need to do because e.g. `requireStateSupportedBy` takes a `FixedPart memory fixedPart`) was unimplemented. 

I believe it was implemented from 0.7.6.

_Originally posted by @geoknee in https://github.com/statechannels/go-nitro/pull/508#discussion_r848550512_

nksazonov:
Closes #527.


nksazonov:
Changing solidity version to 0.7.6 indeed solves `UnimplementedFeatureError`, but introduces a new one - `Stack too deep` in `challenge` function. Thus, I have changed `memory` to `calldata` everywhere except in `challenge`.

@geoknee , do you know a quick workaround to this problem?

geoknee:
I think you need to update this line to silence the build errors:

https://github.com/statechannels/go-nitro/blob/0d9236ab0cbe637d9ebed3b5c3eeff0172266eb4/nitro-protocol/.solhint.json#L4

geoknee:
> Changing solidity version to 0.7.6 indeed solves `UnimplementedFeatureError`, but introduces a new one - `Stack too deep` in `challenge` function

Yeah we are very familiar with this one 😞 . There are a few solutions, such as combining some of the function parameters into a struct, or introducing stricter scoping for local variables. https://soliditydeveloper.com/stacktoodeep . 

I think for now the gas savings seem tiny. So we should just leave `challenge` as is , for now. We can revisit the issue when #496 is done.

geoknee:
> Changing solidity version to 0.7.6 indeed solves `UnimplementedFeatureError`, but introduces a new one - `Stack too deep` in `challenge` function

Yeah we are very familiar with this one 😞 . There are a few solutions, such as combining some of the function parameters into a struct, or introducing stricter scoping for local variables. https://soliditydeveloper.com/stacktoodeep . 

I think for now the gas savings seem tiny. So we should just leave `challenge` as is , for now. We can revisit the issue when #496 is done.

nksazonov:
> I think you need to update this line to silence the build errors:
> 
> https://github.com/statechannels/go-nitro/blob/0d9236ab0cbe637d9ebed3b5c3eeff0172266eb4/nitro-protocol/.solhint.json#L4

Sure, thanks.

nksazonov:
@geoknee , what do you think of bumping contracts to the latest solidity version? It seems to have a lot of fixes and gas optimizations. Moreover, staying at old solidity version blocks us from using newer versions of libraries (e.g. openzeppellin), in which optimizations or even fixes are introduced.

geoknee:
> @geoknee , what do you think of bumping contracts to the latest solidity version? It seems to have a lot of fixes and gas optimizations. Moreover, staying at old solidity version blocks us from using newer versions of libraries (e.g. openzeppellin), in which optimizations or even fixes are introduced.

Yes it is probably a good idea to do that. The existing contracts have been audited, which I think generally disincentivizes changing the compiler version; but given the reasonably large changes upcoming in #496 and the other advantages you mention, I think I'm in favour of upgrading.

One nice thing will be that we will no longer need to import `SafeMath`, since that was absorbed into the language since `0.8.0`.

nksazonov:
> I think I'm in favour of upgrading.

As changing from 0.7.6 to 0.8.13 introduces a lot of breaking changes, I will do that in a separate pr.

geoknee:
@nksazonov looks like gas costs mostly increased under these changes? I think the original motivation for using `calldata` was to save gas. Do you have an idea about what is causing the increase? Is it the solidity version bump? Could we do an experiment to find out?

nksazonov:
> @nksazonov looks like gas costs mostly increased under these changes? I think the original motivation for using `calldata` was to save gas. Do you have an idea about what is causing the increase? Is it the solidity version bump? Could we do an experiment to find out?

Let me do two experiments: decrease version back to 0.7.4 and replace `calldata` with `memory` back.

nksazonov:
Results are confusing:

Replacing `calldata` to `memory`:
- Deploying NitroAdjudicator: -191_170 (-4.4%)
- ERC20/ETH happy exit: -72 gas (- 0.015%)
- ERC20 sad exit: -300 gas (- 0.15%)
- ETH sad exit: -602 gas (- 0.17%)

Downgrading to solidity 0.7.4:
Simply reverts all gas consumption to pre-this pr.

nksazonov:
The reason `calldata` is more expensive can lay in amount of its copying to `memory` inside external functions when they invoke internal ones.

geoknee:
In that case, perhaps best to close this? It has been a useful exercise, though!

nksazonov:
We undo changing `memory` to calldata made in  #508 . Except that only version bump.

geoknee (approved):

PR #555, opened 2022-04-19
Update readme

geoknee:


internal/testhelpers/testhelpers.go
```diff
@@ -1,3 +1,4 @@
+// Package testhelpers contains functions that pretty-print test failures.
```
geoknee, line 1:
```suggestion
// Package testhelpers contains functions which pretty-print test failures.
```

NiloCK: `that` vs `which` rundowns were not on my expected path for today, but I'll take it.

NiloCK (commented):
:+1:

PR #560, opened 2022-04-19
Fix action

geoknee:
this currently errors out.

TODO
- [x] get action running
- [ ] get action passing (there seems to be a mismatch between what I generate on my machine and what the action generates 😕


geoknee:
The reason the action is not passing is because the generated binding contains metadata derived from the compiler version used. So even though the compiler may be built from the same source code, it is a different binary on mac osx versus linux. 

https://docs.soliditylang.org/en/v0.7.4/metadata.html#encoding-of-the-metadata-hash-in-the-bytecode

geoknee:
https://github.com/ethereum/solidity/issues/1571

nksazonov (approved):

PR #566, opened 2022-04-20
Re-enable virtual funding integration test

lalexgap:
This PR re-enables the virtual funding integration test. It also makes a collection of small fixes to get the virtual funding integration test passing.

## Changes
- https://github.com/statechannels/go-nitro/pull/566/commits/ea2c1a769279d97dce9a4516205997459b0aff0d Ignore chain events if we don't have the channel. Right now the chain service returns chain events to every client, even if they're not interested in that channel. 
- https://github.com/statechannels/go-nitro/pull/566/commits/0e76ea9e1a48d13a5a8d6637a0ca5748cbecae7b Properly store the consensus channel when storing an objective.
- [619cc5b](https://github.com/statechannels/go-nitro/pull/566/commits/619cc5bc0df8d58a982f6de70bd735aa75d8c5c2) Previously we just declared an `ok` var and never set it. This updates `ok` based on the result of `getTwoPartyConsensusLedger`
- [0c27489](https://github.com/statechannels/go-nitro/pull/566/commits/0c2748904a73f7dad8064f9dd9deb34c40f892b2) Initialize a consensus channel before calling `UnmarshalJSON`. Previously we were just passing a nil pointer receiver.
- [f9e1057](https://github.com/statechannels/go-nitro/pull/566/commits/f9e10579507cf423153259019adaa4db46494966) get the turn number for the expected proposal based on the current consensus turn number. Previously we were just using 0.

---

#### Code quality

#### Project management

### DIFF
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -165,7 +165,10 @@ func (e *Engine) handleChainEvent(chainEvent chainservice.Event) (ObjectiveChang
 	e.logger.Printf("handling chain event %v", chainEvent)
 	objective, ok := e.store.GetObjectiveByChannelId(chainEvent.ChannelID())
 	if !ok {
-		return ObjectiveChangeEvent{}, &ErrUnhandledChainEvent{event: chainEvent, objective: objective, reason: "no objective for channel"}
+		// TODO: Right now the chain service returns chain events for ALL channels even those we aren't involved in
+		// for now we can ignore channels we aren't involved in
+		// in the future the chain service should allow us to register for specific channels
+		return ObjectiveChangeEvent{}, nil
 	}
 
 	eventHandler, ok := objective.(chainservice.ChainEventHandler)
--- a/client/engine/store/mockstore.go
+++ b/client/engine/store/mockstore.go
@@ -129,7 +129,12 @@ func (ms *MockStore) SetObjective(obj protocols.Objective) error {
 			if err != nil {
 				return fmt.Errorf("error setting channel %s from objective %s: %w", ch.Id, obj.Id(), err)
 			}
-
+		case *consensus_channel.ConsensusChannel:
+			ch := rel.(*consensus_channel.ConsensusChannel)
+			err := ms.SetConsensusChannel(ch)
+			if err != nil {
+				return fmt.Errorf("error setting consensus channel %s from objective %s: %w", ch.Id, obj.Id(), err)
+			}
 		default:
 			return fmt.Errorf("unexpected type: %T", rel)
 		}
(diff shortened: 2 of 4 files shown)


client/engine/engine.go
```diff
@@ -165,7 +165,10 @@ func (e *Engine) handleChainEvent(chainEvent chainservice.Event) (ObjectiveChang
 	e.logger.Printf("handling chain event %v", chainEvent)
 	objective, ok := e.store.GetObjectiveByChannelId(chainEvent.ChannelID())
 	if !ok {
-		return ObjectiveChangeEvent{}, &ErrUnhandledChainEvent{event: chainEvent, objective: objective, reason: "no objective for channel"}
+		// TODO: Right now the chain service returns chain events for ALL channels even those we aren't involved in
```
lalexgap, line 168: Create an issue for this.

client/engine/store/mockstore.go
```diff
@@ -129,7 +129,12 @@ func (ms *MockStore) SetObjective(obj protocols.Objective) error {
 			if err != nil {
 				return fmt.Errorf("error setting channel %s from objective %s: %w", ch.Id, obj.Id(), err)
 			}
-
+		case *consensus_channel.ConsensusChannel:
```
lalexgap, line 132: I didn't add a test for this as it will be tested once this [TODO](https://github.com/statechannels/go-nitro/blob/ag/vf-integration/client/engine/store/mockstore_test.go#L46:L46) is resolved

client/engine/store/mockstore.go
```diff
@@ -129,7 +129,12 @@ func (ms *MockStore) SetObjective(obj protocols.Objective) error {
...
+			ch := rel.(*consensus_channel.ConsensusChannel)
```
NiloCK, line 133: I'm not able to _suggest_ it here because of the specific change set, but I came across this the other day:

```
		switch ch := rel.(type) {
		case *channel.Channel:
			err := ms.SetChannel(ch)
			if err != nil {
				return fmt.Errorf("error setting channel %s from objective %s: %w", ch.Id, obj.Id(), err)
			}
		case *consensus_channel.ConsensusChannel:
			err := ms.SetConsensusChannel(ch)
			if err != nil {
				return fmt.Errorf("error setting consensus channel %s from objective %s: %w", ch.Id, obj.Id(), err)
			}
		default:
			return fmt.Errorf("unexpected type: %T", rel)
		}
```

this assigns the type-asserted `ch` (it has the correct type in each block) just once, and deduplicates some of the assertions.

NiloCK (approved):
LGTM. Pleasantly surprised at how smooth this was.

client/engine/store/mockstore.go
lalexgap: Updated in [81d0a07](https://github.com/statechannels/go-nitro/pull/566/commits/81d0a07ea4beb7e61e733cad48a2975c8f65b6d3)

PR #572, opened 2022-04-21
Add testLedgers helper functions

NiloCK:
This PR adds ledger-network generation functions which connect a list of `testactors.Actor`s via `consensus_channel`s.

In the immediate, it is a replacement part for now-broken generation of test VFOs (as consumed by the mockstore test). (Incremental toward #420 - reinstating mockstore test of setting/retrieving VFOs)

Going forward, these are potentially useful for stress-testing networks. IE, generating a large number of actors and connecting them in a synthetic network.

---

#### Code quality

#### Project management

### DIFF
--- a/internal/testdata/channels.go
+++ b/internal/testdata/channels.go
@@ -182,7 +182,7 @@ func createTestLedger(leader, follower testactors.Actor) TestLedger {
 		sigs,
 	)
 	if err != nil {
-		panic(fmt.Sprintf("error creating leader channel in testLedger: %v", err)
+		panic(fmt.Sprintf("error creating leader channel in testLedger: %v", err))
 	}
 	followCh, err := consensus_channel.NewFollowerChannel(
 		fp,
@@ -191,7 +191,7 @@ func createTestLedger(leader, follower testactors.Actor) TestLedger {
 		sigs,
 	)
 	if err != nil {
-		panic(fmt.Sprintf("error creating follwer channel in testLedger: %v", err)
+		panic(fmt.Sprintf("error creating follwer channel in testLedger: %v", err))
 	}
 
 	return TestLedger{leaderCh, followCh}
--- a/internal/testdata/channels.go
+++ b/internal/testdata/channels.go
@@ -99,7 +99,7 @@ func (l LedgerNetwork) GetLedgerLookup(seeker types.Address) virtualfund.GetTwoP
 // createLedgerNetwork returns active, funded consensus_channels connecting the supplied
 // actors according the the supplied edge list.
 //
-// Edges specify actors via their indeces in the actors slice,
+// Edges specify actors via their indices in the actors slice,
 // and each edge is ordered [leader, follower].
 func createLedgerNetwork(actors []testactors.Actor, edges [][2]int) LedgerNetwork {
 	if len(edges) == 0 {
--- a/internal/testdata/channels.go
+++ b/internal/testdata/channels.go
@@ -102,10 +102,6 @@ func (l LedgerNetwork) GetLedgerLookup(seeker types.Address) virtualfund.GetTwoP
 // Edges specify actors via their indices in the actors slice,
 // and each edge is ordered [leader, follower].
 func createLedgerNetwork(actors []testactors.Actor, edges [][2]int) LedgerNetwork {
-	if len(edges) == 0 {
-		return createLedgerPath(actors)
-	}
-
 	// naive connectedness check: does not detect, eg
 	// a--b  c--d (no path from a to c or d, etc)
 	for i, a := range actors {
--- a/internal/testdata/channels.go
+++ b/internal/testdata/channels.go
@@ -49,3 +49,150 @@ func mockConsensusChannel(counterparty types.Address) (ledger *consensus_channel
 	cc, _ := testObj.CreateConsensusChannel()
 	return cc, true
 }
+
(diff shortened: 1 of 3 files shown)


internal/testdata/channels.go
```diff
@@ -49,3 +49,150 @@ func mockConsensusChannel(counterparty types.Address) (ledger *consensus_channel
...
+			if ledger.LeaderView.Leader() == seeker &&
+				ledger.LeaderView.Follower() == counterparty {
+				return &ledger.LeaderView, true
+			}
+		}
+		return nil, false
+	}
+}
+
+// createLedgerNetwork returns active, funded consensus_channels connecting the supplied
```
geoknee, line 99: Super nit: what does "active" mean?

internal/testdata/channels.go
```diff
@@ -49,3 +49,150 @@ func mockConsensusChannel(counterparty types.Address) (ledger *consensus_channel
...
+// actors according the the supplied edge list.
+//
+// Edges specify actors via their indeces in the actors slice,
+// and each edge is ordered [leader, follower].
+func createLedgerNetwork(actors []testactors.Actor, edges [][2]int) LedgerNetwork {
+	if len(edges) == 0 {
+		return createLedgerPath(actors)
```
geoknee, line 106: I think it would be responsible to include this special case in the godoc? I.e. if I fail to specify any edges, I will in fact end up with a nontrivially connected network (with edges) -- which is somewhat unexpected?

internal/testdata/channels.go
```diff
@@ -49,3 +52,150 @@ func mockConsensusChannel(counterparty types.Address) (ledger *consensus_channel
...
+			}
+		}
+		return nil, false
+	}
+}
+
+// createLedgerNetwork returns active, funded consensus_channels connecting the supplied
+// actors according the the supplied edge list.
+//
+// Edges specify actors via their indeces in the actors slice,
```
geoknee, line 102:
```suggestion
// Edges specify actors via their indices in the actors slice,
```

internal/testdata/objectives.go
```diff
@@ -60,6 +61,10 @@ func genericDFO() directfund.Objective {
 
 func genericVFO() virtualfund.Objective {
```
geoknee, line 62: Some godoc here would be nice, since this constitutes quite a significant change to the side effects of this function. Something mentioning that a "linear" ledger network will be created?

internal/testdata/objectives.go
```diff
@@ -70,7 +75,15 @@ func genericVFO() virtualfund.Objective {
 		ts.Outcome,
 		ts.ChannelNonce.Int64(),
 	}
-	testVFO, err := virtualfund.NewObjective(request, Channels.MockTwoPartyLedger, Channels.MockConsensusChannel)
+
+	ledgerPath := createLedgerNetwork([]testactors.Actor{
+		testactors.Alice,
+		testactors.Irene,
+		testactors.Bob,
+	}, nil)
```
geoknee, line 83: Is it useful to construct this entire network but then only return one participant's "view" on it? I would have thought the real utility of your helpers was to bootstrap the ledger network plus an objective for each participant? Otherwise, if I want Bob's objective then I will create the ledger network all over again?

protocols/virtualfund/virtualfund.go
```diff
@@ -246,7 +245,7 @@ func constructFromState(
 		init.ToMyRight = &Connection{}
 
 		if consensusChannelToMyRight == nil {
-			return Objective{}, fmt.Errorf("non-bob virtualfund objective requires non-nil ledger channel")
```
geoknee, line 249: Seems like out of scope, but I will let you off :-P

geoknee (approved):
This is some powerful code. Ideally it would have its own unit tests, but I think we can punt those for now if we would like to unblock other work. Might be worth a new issue, though.

internal/testdata/channels.go
NiloCK: approximately:
- has a signed post-fund state
- is ready to `Propose`, `SignNextProposal`, etc

internal/testdata/objectives.go
NiloCK: I think this code could be reused / supplemented for that purpose, yes. Maybe via something like:

```
func (a Actor) vfo(counterparty Actor, network LedgerNetwor) virtualfund.Objective { ... }
```

The goal of the moment is just to accurately construct a single vfo which can be written and retrieved from the store. I found it easier to think about instantiating ledger channels in context of other ledger channels rather than as a one-off, and thought it'd be a good time to write this code.

protocols/virtualfund/virtualfund.go
NiloCK: Indeed. One of those commits that one forgets about...

internal/testdata/channels.go
NiloCK: I think better still is to remove the special case. see 9206839

internal/testdata/objectives.go
NiloCK: This is a little tricky because:

- (unless I'm mistaken) the changing side-effects aren't visible to the the caller
- this function is internal, and is exported via the struct `vfoCollection`, which does apply godoc to its `GenericVFO` member

Electing to leave as-is for now, but open to further conversation.

PR #576, opened 2022-04-22
Use objective Update in virtualfund_single_hop_test.go

kerzhner:
Prior to this PR, `virtualfund_single_hop_test` would manipulate internal objective data structures to progress the objective. This PR removes the manipulations and instead uses the public `Update` api.

### DIFF
--- a/protocols/virtualfund/helpers_test.go
+++ b/protocols/virtualfund/helpers_test.go
@@ -9,29 +9,33 @@ import (
 	"github.com/statechannels/go-nitro/types"
 )
 
-func fixedPart(left, right testactors.Actor) state.FixedPart {
-	return state.FixedPart{
+// prepareConsensusChannel prepares a consensus channel with a consensus outcome
+//  - allocating 6 to left
+//  - allocating 4 to right
+//  - including the given guarantees
+func prepareConsensusChannel(role uint, left, right testactors.Actor, guarantees ...consensus_channel.Guarantee) *consensus_channel.ConsensusChannel {
+	return prepareConsensusChannelHelper(role, left, right, 6, 4, 1, guarantees...)
+}
+
+func consensusStateSignatures(left, right testactors.Actor, guarantees ...consensus_channel.Guarantee) [2]state.Signature {
+	return prepareConsensusChannelHelper(0, left, right, 0, 0, 2, guarantees...).Signatures()
+}
+
+func prepareConsensusChannelHelper(role uint, left, right testactors.Actor, leftBalance, rightBalance, turnNum int, guarantees ...consensus_channel.Guarantee) *consensus_channel.ConsensusChannel {
+	fp := state.FixedPart{
 		ChainId:           big.NewInt(9001),
 		Participants:      []types.Address{left.Address, right.Address},
 		ChannelNonce:      big.NewInt(0),
 		AppDefinition:     types.Address{},
 		ChallengeDuration: big.NewInt(45),
 	}
-}
 
-// prepareConsensusChannel prepares a consensus channel with a consensus outcome
-//  - allocating 6 to left
-//  - allocating 4 to right
-//  - including the given guarantees
-func prepareConsensusChannel(role uint, left, right testactors.Actor, guarantees ...consensus_channel.Guarantee) *consensus_channel.ConsensusChannel {
-	fp := fixedPart(left, right)
-
-	leftBal := consensus_channel.NewBalance(left.Destination(), big.NewInt(6))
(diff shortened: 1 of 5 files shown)


protocols/virtualfund/virtualfund_single_hop_test.go
```diff
@@ -172,26 +172,29 @@ func TestClone(t *testing.T) {
 	}
 }
 
-func collectPeerSignaturesOnSetupState(V *channel.SingleHopVirtualChannel, myRole uint, prefund bool) {
+func collectPeerSignaturesOnSetupState(v *channel.SingleHopVirtualChannel, myRole uint, prefund bool) *channel.SingleHopVirtualChannel {
```
NiloCK, line 175: This is a nice workflow. As I understand it:
- cloning the test objective's channel
- manually progressing the cloned channel
- generating update events with the progressed, cloned channel
- applying the test objective's `Update` to the generated event

Probably this function wants a new name though - it no longer collects the state signatures onto the passed-in channel. Maybe something like `peerSignedClone`.

NiloCK, line 175: Makes the non-modifying intent explicit.

```suggestion
func collectPeerSignaturesOnSetupState(v channel.SingleHopVirtualChannel, myRole uint, prefund bool) *channel.SingleHopVirtualChannel {
```

protocols/virtualfund/virtualfund_single_hop_test.go
```diff
@@ -172,26 +172,29 @@ func TestClone(t *testing.T) {
...
 	var state state.State
 	if prefund {
-		state = V.PreFundState()
+		state = v.PreFundState()
 	} else {
-		state = V.PostFundState()
+		state = v.PostFundState()
 	}
```
NiloCK, line 181: This code is correct, but it made me think / investigate a bit to convince myself that no side effects leaked onto `v`.

I'm not allowed to make this suggestion, but something like 

```
func collectPeerSignaturesOnSetupState(v channel.SingleHopVirtualChannel, myRole uint, prefund bool) *channel.SingleHopVirtualChannel {
	withSigs := v.Clone()

	var state state.State
	if prefund {
		state = withSigs.PreFundState()
	} else {
		state = withSigs.PostFundState()
	}

	if myRole != alice.Role {
		aliceSig, _ := state.Sign(alice.PrivateKey)
		withSigs.AddStateWithSignature(state, aliceSig)
	}
	if myRole != p1.Role {
		p1Sig, _ := state.Sign(p1.PrivateKey)
		withSigs.AddStateWithSignature(state, p1Sig)
	}
	if myRole != bob.Role {
		bobSig, _ := state.Sign(bob.PrivateKey)
		withSigs.AddStateWithSignature(state, bobSig)
	}
	return withSigs
}
```

is slightly more explicit.

NiloCK (approved):
This is a nice `Update` to the test. It moves these functions a little in the direction of integration tests (Update and Crank are ping-ponging), which I think is fine.

kerzhner:
> This is a nice `Update` to the test. It moves these functions a little in the direction of integration tests (Update and Crank are ping-ponging), which I think is fine.

I considered separating `Update` and `Crank` into their own functions. While that approach does isolate the tested behavior, the setup overhead seemed not worth it.

protocols/virtualfund/helpers_test.go
```diff
@@ -14,6 +14,15 @@ import (
...
 func prepareConsensusChannel(role uint, left, right testactors.Actor, guarantees ...consensus_channel.Guarantee) *consensus_channel.ConsensusChannel {
+	return prepareConsensusChannelHelper(role, left, right, 6, 4, 1, guarantees...)
+}
+
+// consensusStateSignatures prepares a consensus channel with a consensus outcome and returns the signatures on the consensus state
+func consensusStateSignatures(left, right testactors.Actor, guarantees ...consensus_channel.Guarantee) [2]state.Signature {
+	return prepareConsensusChannelHelper(0, left, right, 0, 0, 2, guarantees...).Signatures()
+}
+
+func prepareConsensusChannelHelper(role uint, left, right testactors.Actor, leftBalance, rightBalance, turnNum int, guarantees ...consensus_channel.Guarantee) *consensus_channel.ConsensusChannel {
```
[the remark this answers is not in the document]
kerzhner: Good point. Filed https://github.com/statechannels/go-nitro/issues/596.

protocols/virtualfund/virtualfund_single_hop_test.go
kerzhner: Done in a9a0dde.

PR #577, opened 2022-04-22
Fix eslint errors, run eslint for all fiiles

### LINKED ISSUE #554: Fix eslint errors, add .js files to eslint
After adding eslint workflow, not all files have been fixed (e.g. `hardhat.config.ts`). Moreover, several files (configs mostly) have `.js` extension, which means they are not checked by eslint.

nksazonov:
Closes #554.


geoknee:
Can we update these lines so that the we enforce linting on all files on every PR?

https://github.com/statechannels/go-nitro/blob/main/nitro-protocol/package.json#L12-L13

geoknee:
Also, I would like to understand whether this needs to be changed:

https://github.com/statechannels/go-nitro/blob/c7447471d39274d4e226b61d6fe73931510eb844/nitro-protocol/package.json#L15

1. Does `npx solhint --fix './contracts/**/*.sol' work` ?
2. I tried running it on commit 1389b3b but nothing changed. Also the `solhint:check` seemed to pass?

DOes

nksazonov:
> Can we update these lines so that the we enforce linting on all files on every PR?
> 
> https://github.com/statechannels/go-nitro/blob/c7447471d39274d4e226b61d6fe73931510eb844/nitro-protocol/package.json#L12-L13

It requires either a lot of manual fixes or ignoring failing rules.
The most common failing rules are:
```sh
warning  Unexpected any. Specify a different type    @typescript-eslint/no-explicit-any
warning  Forbidden non-null assertion                @typescript-eslint/no-non-null-assertion
error  'require' is not defined                      no-undef
error    Don't use `Object` as a type. The `Object` type actually means "any non-nullish value", so it is marginally better than `unknown`.
- If you want a type meaning "any object", you probably want `Record<string, unknown>` instead.
- If you want a type meaning "any value", you probably want `unknown` instead  @typescript-eslint/ban-type
```

nksazonov:
> Does npx solhint --fix './contracts/**/*.sol' work ?

Nope. It just silently does nothing.

> I tried running it on commit https://github.com/statechannels/go-nitro/commit/1389b3bcf5cc49c3b86e4f5eeaead4c99d10471b but nothing changed. Also the solhint:check seemed to pass?

It seems to does not work at all.
1389b3b has wrong name - I run `prettier:write`

geoknee:
> It requires either a lot of manual fixes or ignoring failing rules.

Ok -- so does [1389b3b](https://github.com/statechannels/go-nitro/pull/577/commits/1389b3bcf5cc49c3b86e4f5eeaead4c99d10471b) just fix _some_ of the problems, then? Or does it just fix _some_ of the files?

I think also #577 was calling for `.js` files to be added to the linting check, but I don't see any change to the linting configuration. I'm just trying to get an idea of the scope of the changes on this PR.

nksazonov:
> Ok -- so does [1389b3b](https://github.com/statechannels/go-nitro/pull/577/commits/1389b3bcf5cc49c3b86e4f5eeaead4c99d10471b) just fix some of the problems, then? Or does it just fix some of the files?

It fixes what can be fixed automatically. 

> I think also https://github.com/statechannels/go-nitro/issues/554 was calling for .js files to be added to the linting check, but I don't see any change to the linting configuration. I'm just trying to get an idea of the scope of the changes on this PR.

Does any changes have to be made to `.eslintrc` for eslint to work with .js files?

nksazonov:
My first intention was to run eslint for all files, while now I am trying to select a strategy for fixing eslint errors.
I believe, we can ignore `@typescript-eslint/no-explicit-any` and fix other errors manually. What do you think @geoknee ?

geoknee:
> My first intention was to run eslint for all files, while now I am trying to select a strategy for fixing eslint errors. I believe, we can ignore `@typescript-eslint/no-explicit-any` and fix other errors manually. What do you think @geoknee ?

Yes this sounds good to me. Is it possible to ignore that rule only for non-src files? In other words, it would be nice to not slacken the checks for files which currently pass them.

geoknee:
Relevant https://github.com/statechannels/go-nitro/issues/578

nksazonov:
> Is it possible to ignore that rule only for non-src files? In other words, it would be nice to not slacken the checks for files which currently pass them.

Fully agree with you. Will try to find out how to implement it.

geoknee (approved):
👍

Issue #574, opened 2022-04-21
Break down the `store` interface


geoknee:
Into e.g.

```go
type ChannelGetter interface {
GetChannelById()
}
```

This interface can then be imported by individual protocols, instead of duplicating the function signatures like this: 

https://github.com/statechannels/go-nitro/blob/1f3cbe4d790b694346328830dc86459447ad87d2/protocols/virtualfund/virtualfund.go#L494-L498

In production, we can pass the entire `store` (which embeds a `ChannelGetter`). In tests we only have to mock the getter itself. 

Hopefully we avoid cyclic dependencies.

geoknee:
Without some refactoring, we will hit a cyclic dependency since the mockstore imports from individual protocol implementations.

PR #579, opened 2022-04-22
Experiment with breaking up the store

geoknee:
This is quite far from being mergable, but I wanted to share some experimentation I have done towards #574.

I had to put the store interface in its own package to avoid import cycles. 

Quite a lot of our (test in particular) code is tightly coupled to the use of **function types** rather than **interfaces**. This has made the work a bit painful and it no longer seems worth the effort to me.

### DIFF
--- a/protocols/virtualfund/virtualfund.go
+++ b/protocols/virtualfund/virtualfund.go
@@ -12,6 +12,7 @@ import (
 	"github.com/statechannels/go-nitro/channel/consensus_channel"
 	"github.com/statechannels/go-nitro/channel/state"
 	"github.com/statechannels/go-nitro/channel/state/outcome"
+	"github.com/statechannels/go-nitro/client/engine/store/store_interface"
 
 	"github.com/statechannels/go-nitro/protocols"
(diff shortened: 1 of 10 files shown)


protocols/virtualfund/virtualfund.go
```diff
@@ -490,19 +491,12 @@ func (o *Objective) isBob() bool {
...
 // It accepts the message, myAddress, and a function to to retrieve ledgers from a store.
 func ConstructObjectiveFromMessage(
 	m protocols.Message,
 	myAddress types.Address,
-	getTwoPartyLedger GetTwoPartyLedgerFunction,
-	getTwoPartyConsensusLedger GetTwoPartyConsensusLedgerFunction,
+	getter store_interface.ConsensusChannelGetter,
 ) (Objective, error) {
 	if len(m.SignedStates) != 1 {
 		return Objective{}, errors.New("expected exactly one signed state in the message")
```
geoknee, line 502: This is the core change that drives all of the others!

geoknee:
We might want to return to this.... some day.

Issue #558, opened 2022-04-19
Rewrite our example applications to implement the new `ForceMoveApp` interface & update unit tests


geoknee:

nksazonov:
Solved with #676.

Issue #546, opened 2022-04-14
Automate updating gas benchmark results


nksazonov:
Meta-comment: the DevEx for fixing up `gas.ts` could be nicer. We could automate updating the file with the "real numbers" (right now it is done manually).

_Originally posted by @geoknee in https://github.com/statechannels/go-nitro/issues/542#issuecomment-1098926419_

Issue #543, opened 2022-04-14
Unskip "exiting virtual funded channel" gas benchmark test 


nksazonov:
We should make an issue to unskip this.

_Originally posted by @geoknee in https://github.com/statechannels/go-nitro/pull/541#discussion_r850226828_

geoknee:
Via #831

Issue #498, opened 2022-04-07
Add `tsc` to continuous integration checks (and fix all errors)


geoknee:
Maybe outside the scope of this work, but it might be nice to get to zero errors. Also, we want to gitignore the resulting `dist` folder.

_Originally posted by @geoknee in https://github.com/statechannels/go-nitro/pull/427#discussion_r842763887_

PR #585, opened 2022-04-25
Integrate `latestSupportedState` into `ForceMove` and DApps

### LINKED ISSUE #557: Modify the `ForceMoveApp` interface to support `latestSupportedState` and integrate it into `ForceMove.sol`

nksazonov:
Closes #557 and #558. 

This PR is the first half of migration to nitro-v2 (aka satp) and consists of #786, #754, #748, #679,  #676 and #557.
The second half will include #785 , #771 , #768 , #760 , #764 , #546 , #543 and #498.

It was decided to separate migration to nitro-v2 into 2 PRs to enable nitro protocol golang code parallel development.


nitro-protocol/contracts/ForceMove.sol
```diff
@@ -457,126 +473,79 @@ contract ForceMove is IForceMove, StatusManager {
...
+     * @dev Construct SignedVariablePart array from supplied variable parts and signatures.
+     * @param variableParts Variable parts to construct signed ones from.
+     * @param sigs Signatures to combine with variable parts.
+     * @param whoSignedWhat participant[i] signed stateHashes[whoSignedWhat[i]]
+     * @return SignedVariablePart array.
      */
-    function _requireValidTransitionChain(
-        // returns stateHashes array if valid
-        // else, reverts
+    function _bindSigsToVariableParts(
```
geoknee, line 499: I think we can avoid the need for this: the actual contract API should accept signed variable parts, and these can be constructed off chain (in the client code).

nitro-protocol/contracts/ForceMove.sol
```diff
@@ -157,10 +157,26 @@ contract ForceMove is IForceMove, StatusManager {
             'Signer not authorized mover'
         );
 
-        _requireValidTransition(
-            fixedPart.participants.length,
-            variablePartAB,
-            fixedPart.appDefinition
+        VariablePart[] memory variableParts = new VariablePart[](2);
+        variableParts[0] = variablePartAB[0];
+        variableParts[1] = variablePartAB[1];
```
geoknee, line 162: I think we may be able to remove `respond` from the on chain API.  It seems closely coupled to the idea of a valid transition, which we are deprecating.

geoknee (commented):
This is looking good so far!

nitro-protocol/contracts/ForceMove.sol
nksazonov: Agree on this. Moreover, the actual contract API accepting signed variable parts would make it even more intuitive and convenient.

nitro-protocol/contracts/ForceMove.sol
nksazonov: Do you suggest `checkpoint` being used instead of `respond`?

geoknee: Exactly!

nitro-protocol/contracts/ForceMove.sol
```diff
@@ -451,132 +467,85 @@ contract ForceMove is IForceMove, StatusManager {
...
-                _lastVariablePart(variableParts).turnNum,
-                fixedPart.participants,
-                stateHashes,
-                sigs,
-                whoSignedWhat
-            ),
-            'Invalid signatures'
-        );
+        // enforcing the latest supported state being in the last slot of the array
+        _requireVariablePartIsLast(latestVariablePart, variableParts);
```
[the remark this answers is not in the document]
nksazonov: This needs to be enforced explicitly, otherwise undesired behavior could appear.

Let's take [`challenge` method](https://github.com/statechannels/go-nitro/blob/integrate-latest-supported-state/nitro-protocol/contracts/ForceMove.sol#L45) for example:

Firstly we look up `largestTurnNum` **of the last supplied variable part**
``` solidity
uint48 largestTurnNum = _lastVariablePart(signedVariableParts).turnNum;
```
for future checks
``` solidity
if (_mode(channelId) == ChannelMode.Open) {
    // using largestTurnNum of the last VariablePart
    _requireNonDecreasedTurnNumber(channelId, largestTurnNum);
} else if (_mode(channelId) == ChannelMode.Challenge) {
    // using largestTurnNum of the last VariablePart
    _requireIncreasedTurnNumber(channelId, largestTurnNum);
} else {
    _requireChannelNotFinalized(channelId);
}
```
Thus, we _assume_ that `_requireStateSupportedBy` (and therefore `latestSupportedState`) perceive last variable part as latest supported one.

Suppose developers described such logic, that _`requireStateSupportedBy` returned the hash **of not the last state**. Then, by the means of dev's DApp, not the last state should be marked as challenging, but that is not what happens:
``` solidity
statusOf[channelId] = _generateStatus(
        //  \/ data from the last VariablePart \/
        largestTurnNum,
        uint48(block.timestamp) + fixedPart.challengeDuration,
        //  \/ data from NOT the last VariablePart  \/
        supportedStateHash,
        //  \/ data from the last VariablePart \/
        _hashOutcome(_lastVariablePart(signedVariableParts).outcome)
    )
);
```
Parts of the last VariablePart alongside with hash of NOT the last state are saved, which will lead to behavior errors.

Described undesired logic can be easily fixed by only using the data gained from `_requireStateSupportedBy`, however that means we are allowing dev's DApp to return not only the last state, which is the opposite of what we are trying to enforce.

Having said that, the only way I see we enforce DApp returning only the last state is by explicitly requiring so.

nitro-protocol/contracts/CountingApp.sol
```diff
@@ -25,23 +25,17 @@ contract CountingApp is IForceMoveApp {
...
+     * @dev Encodes application-specific rules for a particular ForceMove-compliant state channel.
+     * @param fixedPart Fixed part of the state channel.
+     * @param signedVariableParts Array of variable parts to find the latest of.
+     * @return VariablePart Latest supported by application variable part from supplied array.
+     */    
+    function latestSupportedState(
+        FixedPart calldata fixedPart,
+        SignedVariablePart[] calldata signedVariableParts
+    ) external pure override returns (VariablePart memory) {
+        // TODO
```
geoknee, line 38: You could reference issue #558 here.

nitro-protocol/contracts/NitroAdjudicator.sol
```diff
@@ -101,20 +92,19 @@ contract NitroAdjudicator is ForceMove, MultiAssetHolder {
...
+     * @dev Encodes application-specific rules for a particular ForceMove-compliant state channel.
+     * @param fixedPart Fixed part of the state channel.
+     * @param signedVariableParts Array of variable parts to find the latest of.
+     * @return VariablePart Latest supported by application variable part from supplied array.
+     */    
+    function latestSupportedState(
+        FixedPart calldata fixedPart,
+        SignedVariablePart[] calldata signedVariableParts
+    ) external pure returns (VariablePart memory) {
+        // copy to memory explicitly to avoid `Stack to deep` error
```
geoknee, line 105: This could also have an issue to track it? I guess it is a TODO to return to this at some point?

nitro-protocol/contracts/interfaces/IForceMove.sol
```diff
@@ -2,97 +2,45 @@
...
-    struct State {
-        // participants sign the hash of this
-        bytes32 channelId; // keccack(chainId,participants,channelNonce,appDefinition,challengeDuration)
-        bytes appData;
-        bytes outcome;
-        uint48 turnNum;
-        bool isFinal;
-    }
-
+interface IForceMove is INitroTypes {
```
geoknee, line 10: The natspec above needs modifying.

nitro-protocol/contracts/interfaces/IForceMove.sol
```diff
@@ -2,97 +2,45 @@
...
      * @dev Registers a challenge against a state channel. A challenge will either prompt another participant into clearing the challenge (via one of the other methods), or cause the channel to finalize at a specific time.
      * @param fixedPart Data describing properties of the state channel that do not change with state updates.
-     * @param variableParts An ordered array of structs, each decribing the properties of the state channel that may change with each state update. Length is from 1 to the number of participants (inclusive).
-     * @param sigs An array of signatures that support the state with the `largestTurnNum`. There must be one for each participant, e.g.: [sig-from-p0, sig-from-p1, ...]
-     * @param whoSignedWhat An array denoting which participant has signed which state: `participant[i]` signed the state with index `whoSignedWhat[i]`.
+     * @param signedVariableParts An ordered array of structs, that can be signed by any number of participants, each struct decribing the properties of the state channel that may change with each state update. Length is from 1 to the number of participants (inclusive).
```
geoknee, line 15:
```suggestion
     * @param signedVariableParts An ordered array of structs, that can be signed by any number of participants, each struct describing the properties of the state channel that may change with each state update.
```
I don't think we need to say this (and I'm not 💯 % sure it is true?)

nitro-protocol/contracts/interfaces/IForceMove.sol
```diff
@@ -2,97 +2,45 @@
...
      * @dev Overwrites the `turnNumRecord` stored against a channel by providing a state with higher turn number, supported by a signature from each participant.
      * @param fixedPart Data describing properties of the state channel that do not change with state updates.
-     * @param variableParts An ordered array of structs, each decribing the properties of the state channel that may change with each state update.
-     * @param sigs An array of signatures that support the state with the `largestTurnNum`: one for each participant, in participant order (e.g. [sig of participant[0], sig of participant[1], ...]).
-     * @param whoSignedWhat An array denoting which participant has signed which state: `participant[i]` signed the state with index `whoSignedWhat[i]`.
+     * @param signedVariableParts An ordered array of structs, that can be signed by any number of participants, each struct decribing the properties of the state channel that may change with each state update. Length is from 1 to the number of participants (inclusive).
```
geoknee, line 28:
```suggestion
     * @param signedVariableParts An ordered array of structs, that can be signed by any number of participants, each struct describing the properties of the state channel that may change with each state update.
```

nitro-protocol/contracts/interfaces/IForceMove.sol
```diff
@@ -104,19 +52,15 @@ interface IForceMove {
      * @param finalizesAt The unix timestamp when `channelId` will finalize.
      * @param isFinal Boolean denoting whether the challenge state is final.
      * @param fixedPart Data describing properties of the state channel that do not change with state updates.
-     * @param variableParts An ordered array of structs, each decribing the properties of the state channel that may change with each state update.
-     * @param sigs A list of Signatures that supported the challenge: one for each participant, in participant order (e.g. [sig of participant[0], sig of participant[1], ...]).
-     * @param whoSignedWhat Indexing information to identify which signature was by which participant
+     * @param signedVariableParts An ordered array of structs, that can be signed by any number of participants, each struct decribing the properties of the state channel that may change with each state update. Length is from 1 to the number of participants (inclusive).
```
geoknee, line 55:
```suggestion
     * @param signedVariableParts An ordered array of structs, that can be signed by any number of participants, each struct describing the properties of the state channel that may change with each state update.
```

nitro-protocol/contracts/interfaces/INitroTypes.sol
```diff
@@ -0,0 +1,42 @@
...
+        uint256 chainId;
+        address[] participants;
+        uint48 channelNonce;
+        address appDefinition;
+        uint48 challengeDuration;
+    }
+
+    struct State {
+        // participants sign the hash of this
+        bytes32 channelId; // keccack(chainId,participants,channelNonce,appDefinition,challengeDuration)
```
geoknee, line 36:
```suggestion
        bytes32 channelId; // keccack(FixedPart)
```
?

nitro-protocol/contracts/ForceMove.sol
```diff
@@ -39,28 +39,22 @@ contract ForceMove is IForceMove, StatusManager {
...
      * @dev Registers a challenge against a state channel. A challenge will either prompt another participant into clearing the challenge (via one of the other methods), or cause the channel to finalize at a specific time.
      * @param fixedPart Data describing properties of the state channel that do not change with state updates.
-     * @param variableParts An ordered array of structs, each decribing the properties of the state channel that may change with each state update. Length is from 1 to the number of participants (inclusive).
-     * @param sigs An array of signatures that support the state with the `largestTurnNum`. There must be one for each participant, e.g.: [sig-from-p0, sig-from-p1, ...]
-     * @param whoSignedWhat An array denoting which participant has signed which state: `participant[i]` signed the state with index `whoSignedWhat[i]`.
+     * @param signedVariableParts An ordered array of structs, that can be signed by any number of participants, each struct decribing the properties of the state channel that may change with each state update. Length is from 1 to the number of participants (inclusive).
```
geoknee, line 42:
```suggestion
     * @param signedVariableParts An ordered array of structs, that can be signed by any number of participants, each struct describing the properties of the state channel that may change with each state update.
```

nitro-protocol/contracts/ForceMove.sol
```diff
@@ -299,24 +180,16 @@ contract ForceMove is IForceMove, StatusManager {
...
+     * @dev Validates related to participants and states input for several external methods.
+     * @param numParticipants Length of the participants array.
+     * @param numStates Number of states submitted.
      */
     function requireValidInput(
         uint256 numParticipants,
-        uint256 numStates,
-        uint256 numSigs,
-        uint256 numWhoSignedWhats
+        uint256 numStates
```
geoknee, line 190: We should think about whether this check is necessary or not. 

The old check was important, but I think we don't need this check anymore since we are making the protocol much more open with respect to what can constitute a support proof. In particular, we defer to the application rules to decide. It might be that most (or even all) applications would have the same check inside. But it probably doesn't belong here. 

There might also be an interesting application that has more states in the support proof than there are participants in the channel. I can't think of an example. But I also can't think why this could be ruled out -- can you @nksazonov or @andrewgordstewart ?

geoknee (commented):
👍  This is looking great!

I think we can move on to #558 now. Let's keep this PR open until we can get all the applications reinstated and unit tests passing.

nitro-protocol/contracts/ForceMove.sol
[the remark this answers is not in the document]
geoknee: You're right about this. However, I am curious what you think about this approach #666. It may go against what I said before 🤦 but I think it might help reduce confusion around "last" (`arr.length-1`) and "latest" (greatest turn number).

nitro-protocol/contracts/interfaces/IForceMove.sol
nksazonov: Yeah, I was not sure either. I think, it is better to specify all parameter restrictions in one place in docs, but not in every function.

nitro-protocol/contracts/ForceMove.sol
nksazonov: Agree with @andrewgordstewart . This check belongs to deciding what constitutes a support proof, which we give away to DApp.

nitro-protocol/contracts/ForceMove.sol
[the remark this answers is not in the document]
nksazonov: #666 suggests `latestSupportedState` can return `VariablePart` with not the greatest turn number amongst what was passed to this method. This can happen in a situation, where first N `VariableParts` constitutes a support proof, but the last M do not fall under the rules of DApp, thus are not in support proof. Let's call this situation "sub-support proof". So latest `VariablePart` in provided array (expected support proof) will not be equal to latest one in sub-support proof.

Therefore, confusions can emerge. Imagine participants want to conclude a channel and pass a support proof to `latestSupportedState`, expecting the channel to conclude with outcome specified in the latest `VariablePart`. However, by the logic specified in DApp, latest `VariablePart` appears not to be in support proof (either because participants made a mistake and DApp returned sub-support proof or DApp is malicious), consequently, not the latest `VariablePart` is returned and channel concludes with outdated outcome.

This means, that by not explicitly requiring `latestSupportedState` to return the latest `VariablePart`, we shift responsibility of ensuring channel will conclude / make a checkpoint / challenge with a desired (latest provided) `VariablePart` from `ForceMove` to the participants, who are prone to errors. And consequently, another possible _vector of attack / ability to make a mistake_ appears.

Or we can lay this responsibility on `ForceMove` by requiring `signedVariableParts` array to be increasingly sorted by turn number and `latestSupportedState` to return the last (thus, expected latest) `VariablePart`.

nitro-protocol/contracts/CountingApp.sol
nksazonov: Do you mean referencing in comments?

nitro-protocol/contracts/NitroAdjudicator.sol
nksazonov: Using `memory` instead of `calldata` adds up to 100 gas to `conclude`, but `NitroAdjudicator` deployment consumed 100k less gas.

This made me think about compiler optimization runs and the question arose: How was number "200" for optimization runs chosen?
I have reverted to using `calldata`, changed runs to "1000" and gas savings on functions are 500-800, but deployment consumes 50k more.

I believe, gas costs on functions are more important than on deployment, thus it would be logical to increase optimization runs and use `calldata` with internal copying to `memory` (as presented in commented piece of code).

What do you think, @geoknee ?

geoknee: I think 200 is just the default value, it wasn't chosen by us for any particular reason. 

Yes I agree that deployment costs are less important than function calls. So your proposed approach makes sense.

nksazonov:
While upgrading `EmbeddedApplication` to comply with `latestSupportedState` I realised it would be wise to include a bit mask field `signedBy` in `SignedVariablePart` for developers to be more convenient when implementing signature check logic. You have already implemented an example of it in `EmbeddedApplication`.

But this introduces a limitation: the maximum amount of bytes available in Integer type is 256, thus allowing only 256 participants in a channel. There was a similar limitation in `requireValidInput`, but we had agreed to remove this function.

@geoknee , why was this limitation present and what do you think of bit mask field `signedBy`?

nitro-protocol/contracts/NitroAdjudicator.sol
nksazonov: So then I will change the comment to something more precise and not related to TODO.
Re optimization runs: I believe it is not related to this PR at all, thus will be put in a separate issue and tackled once this is merged.

nitro-protocol/contracts/ForceMove.sol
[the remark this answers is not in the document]
nksazonov: Logical continuation of this thread is [#764](https://github.com/statechannels/go-nitro/discussions/764).

PR #597, opened 2022-04-26
Refactor Update on Objective interface

lalexgap:
Now that the engine is responsible for making sure proposals and states are handled in a specific order, it makes sense to simplify `Update` to only deal with one value at a time.

To accomplish this I've split `Update` into `UpdateWithState` and `UpdateWithProposal`.  We no longer deal with an `ObjectiveUpdateEvent`. This simplifies the code slightly.

### DIFF
--- a/protocols/directdefund/directdefund.go
+++ b/protocols/directdefund/directdefund.go
@@ -171,28 +171,31 @@ func (o Objective) Related() []protocols.Storable {
 	return []protocols.Storable{o.C}
 }
 
-// Update receives an ObjectiveEvent, applies all applicable event data to the DirectDefundingObjective,
-// and returns the updated objective
-func (o Objective) Update(event protocols.ObjectiveEvent) (protocols.Objective, error) {
-	if o.Id() != event.ObjectiveId {
-		return &o, fmt.Errorf("event and objective Ids do not match: %s and %s respectively", string(event.ObjectiveId), string(o.Id()))
-	}
+func (o Objective) UpdateWithProposal(sp consensus_channel.SignedProposal) (protocols.Objective, error) {
+	return &o, fmt.Errorf("direct funding does not support proposals")
+}
 
-	if len(event.SignedStates) > 0 {
-		for _, ss := range event.SignedStates {
-			if !ss.State().IsFinal {
-				return &o, errors.New("direct defund objective can only be updated with final states")
-			}
-			if o.finalTurnNum != ss.State().TurnNum {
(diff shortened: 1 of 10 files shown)


protocols/directdefund/directdefund.go
```diff
@@ -163,28 +164,31 @@ func (o Objective) Related() []protocols.Storable {
...
-				return &o, errors.New("direct defund objective can only be updated with final states")
-			}
-			if o.finalTurnNum != ss.State().TurnNum {
-				return &o, fmt.Errorf("expected state with turn number %d, received turn number %d", o.finalTurnNum, ss.State().TurnNum)
-			}
-		}
-	}
+func (o Objective) UpdateWithProposal(sp consensus_channel.SignedProposal) (protocols.Objective, error) {
+	return &o, fmt.Errorf("direct funding does not support proposals")
+}
```
geoknee, line 169: This is unfortunate... but the alternatives I can think of don't seem to work. I thought about trying to put a generic method into the `Objective` interface, but that is not allowed. 

We could have `Update(interface{})` and put a type switch inside it. But that also seems bad.

As long as we don't anticipate too many different `UpdateWith*` methods on the `Objective` interface, what you have here is probably tolerable.

protocols/virtualdefund/virtualdefund.go
```diff
@@ -360,50 +360,48 @@ func (o Objective) validateSignature(sig state.Signature, participantIndex uint)
...
 }
 
-// Update receives an protocols.ObjectiveEvent, applies all applicable event data to the VirtualDefundObjective,
-// and returns the updated state.
-func (o Objective) Update(event protocols.ObjectiveEvent) (protocols.Objective, error) {
-	if o.Id() != event.ObjectiveId {
-		return &o, fmt.Errorf("event and objective Ids do not match: %s and %s respectively", string(event.ObjectiveId), string(o.Id()))
-	}
-
+func (o Objective) UpdateWithState(ss state.SignedState) (protocols.Objective, error) {
```
geoknee, line 363: Somewhat outside of scope:

I think we could make all objective methods have a pointer receiver.

protocols/directdefund/directdefund.go
lalexgap: > I thought about trying to put a generic method into the Objective interface, but that is not allowed.
I also tried that and was similarly disappointed :( 

This seems the least worst option I could come up with.

lalexgap:
Closing in favour #626

PR #583, opened 2022-04-25
Cmk.420 remove tpl getter

NiloCK:
closes #573, toward #420, #429

PR: 
- removes the unused `GetTwoPartyConsensusLedger` from `virtualfund.ConstructObjectiveFromMessage`
- removes the (now) unused function type `GetTwoPartyLedger`
- shortens the name `GetTwoPartyConsensusLedgerFunction` to `GetLedgerFunction`
- refactors a number of overly specified references to ConsensusChannels to read as references to LedgerChannels

---

#### Code quality

- [x] I have written clear commit messages
- [ ] I have performed a self-review of my own code
- [ ] This change does not have an unduly wide scope
  - bled into a couple of separate issues accidentally
  - still a straightforward refactor PR: zero functional changes
- [x] I have separated logic changes from refactor changes (formatting, renames, etc.)
- [x] I have commented my code wherever necessary (can be 0)
- [ ] ~I have added tests that prove my fix is effective or that my feature works, if necessary~
- [x] I have added comprehensive [godoc](https://go.dev/blog/godoc) comments 

#### Project management

- [x] I have assigned myself to this PR
- [x] I have linked the appropriate github issue
- [x] I have assigned this PR to the appropriate GitHub project
- [x] I have assigned this PR to the appropriate GitHub Milestone

PR #610, opened 2022-04-27
rename consensus channel to ledger channel

lalexgap:
Originally I think we chose the name of `consensus_channel` because we already had a `TwoPartyLedgerChannel`. Now that `TwoPartyLedgerChannel`s are removed we can rename `consensus_channel` to `ledger`.

Ledger is the term we use in the paper and in our discussions so I think it makes sense for the code to reflect that.

### DIFF
--- a/channel/ledger/clone_test.go
+++ b/channel/ledger/clone_test.go
@@ -1,4 +1,4 @@
-package consensus_channel
+package ledger
 
 import (
 	"math/big"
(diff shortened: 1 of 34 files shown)


geoknee:
I think we want to do this at some point. I'm not sure exactly when the right moment is, though... 

https://github.com/statechannels/go-nitro/pull/583#issuecomment-1109867867

* this could cause quite a lot of merge conflicts if there's a lot of development around consensus channels ongoing
* I think we have quite a lot of comments in the code base that reference "consensus channels" so we would want to catch those, as well.

PR #611, opened 2022-04-28
Destroy Channel / ConsensusChannel when governance is handed over

### LINKED ISSUE #598: Ensure data is neither lost nor replicated during shared governance by `Channel` and `ConsensusChannel`
https://www.notion.so/statechannels/Channels-ConsensusChannels-Objectives-9701e24e2bc1491a83fae375e4e4c64a#9c41424d07694150bdc6cf373309757a

For now we will avoid replication by having "butt joints" (that is, no overlap) between `Channel` and `ConsensusChannel`.

We will avoid data loss by copying data from the one type to the other when they hand off governance of a channel.

### LINKED ISSUE #607: Destroy Consensus Channel when direct defunding
This should become a new issue or be tracked in #598

_Originally posted by @geoknee in https://github.com/statechannels/go-nitro/pull/602#discussion_r860055312_

geoknee:
Closes #607 
Closes #598

### DIFF
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -253,6 +253,8 @@ func (e *Engine) handleAPIEvent(apiEvent APIEvent) (ObjectiveChangeEvent, error)
 			if err != nil {
 				return ObjectiveChangeEvent{}, fmt.Errorf("handleAPIEvent: Could not create objective for %+v: %w", request, err)
 			}
+			// If ddfo creation was successful, destroy the consensus channel to prevent it being used (a Channel will now take over governance)
+			e.store.DestroyConsensusChannel(request.ChannelId)
 			return e.attemptProgress(&ddfo)
 
 		default:
@@ -347,6 +349,8 @@ func (e Engine) SpawnConsensusChannelIfDirectFundObjective(crankedObjective prot
 		if err != nil {
 			return fmt.Errorf("could not store consensus channel for objective %s: %w", crankedObjective.Id(), err)
 		}
+		// Destroy the channel since the consensus channel takes over governance:
+		e.store.DestroyChannel(c.Id)
 	}
 	return nil
 }
@@ -395,9 +399,8 @@ func (e *Engine) constructObjectiveFromMessage(id protocols.ObjectiveId, ss stat
 		if err != nil {
 			return &directdefund.Objective{}, fmt.Errorf("could not create direct defund objective from message: %w", err)
 		}
-
-		// TODO Destroy / Disable / Disown the associated ConsensusChannel
-
+		// If ddfo creation was successful, destroy the consensus channel to prevent it being used (a Channel will now take over governance)
+		e.store.DestroyConsensusChannel(ddfo.C.Id)
 		return &ddfo, nil
 
 	default:
--- a/client/engine/store/memstore.go
+++ b/client/engine/store/memstore.go
@@ -167,7 +167,7 @@ func (ms *MemStore) SetChannel(ch *channel.Channel) error {
 
 // DestroyChannel deletes the channel with id id.
 func (ms *MemStore) DestroyChannel(id types.Destination) {
-	ms.channels.m.Delete(id)
+	ms.channels.m.Delete(id.String())
 }
 
 // SetConsensusChannel sets the channel in the store.
@@ -184,7 +184,7 @@ func (ms *MemStore) SetConsensusChannel(ch *consensus_channel.ConsensusChannel) 
 
 // DestroyChannel deletes the channel with id id.
 func (ms *MemStore) DestroyConsensusChannel(id types.Destination) {
-	ms.consensusChannels.m.Delete(id)
+	ms.consensusChannels.m.Delete(id.String())
 }
 
 // GetChannelById retrieves the channel with the supplied id, if it exists.
--- a/client/engine/store/memstore.go
+++ b/client/engine/store/memstore.go
@@ -51,7 +51,7 @@ func (o *syncMap[T]) Store(key string, data T) {
 	o.m.Store(key, data)
 }
 
-// Delete cleats the value for a key.
+// Delete deletes the value for a key.
 func (o *syncMap[T]) Delete(key string) {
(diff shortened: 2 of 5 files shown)


client/engine/engine.go
```diff
@@ -253,6 +253,8 @@ func (e *Engine) handleAPIEvent(apiEvent APIEvent) (ObjectiveChangeEvent, error)
 			if err != nil {
 				return ObjectiveChangeEvent{}, fmt.Errorf("handleAPIEvent: Could not create objective for %+v: %w", request, err)
 			}
+			// If ddfo creation was successful, destroy the consensus channel to prevent it being used (a Channel will now take over governance)
+			e.store.DestroyConsensusChannel(request.ChannelId)
```
geoknee, line 257: Unfortunately, there will be a nonzero amount of time between destroying the `ConsensusChannel`, and writing the `Channel` to the store. We can improve that here by using a `defer` statement...

client/engine/engine.go
```diff
@@ -395,9 +399,8 @@ func (e *Engine) constructObjectiveFromMessage(id protocols.ObjectiveId, ss stat
 		if err != nil {
 			return &directdefund.Objective{}, fmt.Errorf("could not create direct defund objective from message: %w", err)
 		}
-
-		// TODO Destroy / Disable / Disown the associated ConsensusChannel
-
+		// If ddfo creation was successful, destroy the consensus channel to prevent it being used (a Channel will now take over governance)
+		e.store.DestroyConsensusChannel(ddfo.C.Id)
```
geoknee, line 403: ... but we can't pull that trick here.

geoknee: We could make an issue to try and make the composite operation atomic in the store -- i.e. spawn one type and delete the other at the same time.

client/engine/store/memstore.go
```diff
@@ -51,6 +51,11 @@ func (o *syncMap[T]) Store(key string, data T) {
 	o.m.Store(key, data)
 }
 
+// Delete cleats the value for a key.
```
NiloCK, line 54:
```suggestion
// Delete deletes the value for a key.
```

client/engine/store/memstore.go
```diff
@@ -51,6 +51,11 @@ func (o *syncMap[T]) Store(key string, data T) {
...
+func (o *syncMap[T]) Delete(key string) {
```
NiloCK, line 55: Is there a case to be made for `Archive` rather than `Delete`?

client_test/directfund_integration_test.go
```diff
@@ -84,6 +84,10 @@ func TestDirectFundIntegration(t *testing.T) {
 			t.Fatal("Expected nonzero on chain funding, but got zero")
 		}
 
+		if _, channelStillInStore := store.GetChannelById(con.Id); channelStillInStore {
+			t.Fatalf("Expected channel to have been destroyed in %v's store, but it was not", store.GetAddress())
```
NiloCK, line 88:
```suggestion
			t.Fatalf("Expected the channel.Channel data to have been destroyed in %v's store, but it was not", store.GetAddress())
```

This suggestion tightly couples the error message to current package / struct names, but might be worth it to provide context.

The check could also be moved closer to the check that the ConsensusChannel (ledger channel) **does** exist, with a comment similar to the one introduced in this PR in `directdefund_integration_test.go`. Something like:

```
// Ensure that we have a consensus channel in the store
// And that we no longer have a regular channel
```

client_test/directdefund_integration_test.go
```diff
@@ -27,10 +28,31 @@ func TestDirectDefundIntegration(t *testing.T) {
...
+		if c.OnChainFunding.IsNonZero() {
+			t.Fatal("Expected zero on chain funding, but got nonzero")
+		}
+
+		_, err := store.GetConsensusChannelById(channelId)
+		if consensusChannelStillInStore := (err == nil); consensusChannelStillInStore {
+			t.Fatalf("Expected ConsensusChannel to have been destroyed in %v's store, but it was not", store.GetAddress())
+		}
+
+	}
```
NiloCK, line 56: large diff here because the loop identifier `store` blocked access to `store.ErrNoSuchChannel`.

```suggestion
	for _, clientStore := range []store.Store{storeA, storeB} {

		c, channelInStore := clientStore.GetChannelById(channelId)

		if !channelInStore {
			t.Fatalf("expected a Channel to have been created")
		}

		if c.OnChainFunding.IsNonZero() {
			t.Fatal("Expected zero on chain funding, but got nonzero")
		}

		_, err := clientStore.GetConsensusChannelById(channelId)
		if err != store.ErrNoSuchChannel {
			t.Fatalf("Expected ConsensusChannel to have been destroyed in %v's store, but it was not", clientStore.GetAddress())
		}

	}
```

client/engine/engine.go
NiloCK: I am in favour. If it is incorrect for a `channel.Channel` and a `consensus_channel.ConsensusChannel` with the same IDs to both exist, then SetConsensusChannel(c) itself should check / destroy a conflicting `channel.Channel`.

Lower-level, they could share the same namespace for keys, so that it's entirely impossible for both to exist.

NiloCK (approved):
:+1:

client/engine/store/memstore.go
geoknee: Is `Archive` an actual, or imagined function? And are you suggesting a different behaviour of just a different name?

Of course some kind of "soft delete' would seem safest... but I think unnecessary because replacement by the other type should preserve all of the important information.

client/engine/engine.go
geoknee: I'll merge this for now, since it gets the test assertions in place and an "ok" implementation. Filed #613

client/engine/store/memstore.go
NiloCK: Yes, I'm imagining `Archive` as a soft-delete. More like adding a `deleted: true` flag.

Issue #213, opened 2022-02-03
Implement a permissions system so that incoming (proposed) objectives are not blindly approved


geoknee:
We might want to create an issue to track this

_Originally posted by @lalexgap in https://github.com/statechannels/go-nitro/pull/202#discussion_r797971763_

geoknee:
Via #739

PR #612, opened 2022-04-28
WIP virtual defunding integration

geoknee:


protocols/virtualdefund/virtualdefund_test.go
```diff
@@ -32,9 +32,15 @@ func TestCrank(t *testing.T) {
 }
 
 func TestInvalidUpdate(t *testing.T) {
+	t.Skip() // TODO
```
geoknee, line 35: @NiloCK would be great to split off the work to unskip these tests.

protocols/virtualdefund/virtualdefund.go
```diff
@@ -106,6 +107,29 @@ func NewObjective(preApprove bool,
...
+func ConstructObjectiveFromState(
+	initialState state.State,
+	getChannel GetChannelByIdFunction,
+	getTwoPartyConsensusLedger GetTwoPartyConsensusLedgerFunction,
+) (Objective, error) {
+	channelId, err := initialState.ChannelId()
+	if err != nil {
+		return Objective{}, err
+	}
+	paidToBob := big.NewInt(0) // TODO how to set this properly?
```
geoknee, line 121: In standup today we agreed that this issue should be punted pending discussion #617 .

If it is the case that we need to "'query the consuming application to get/verify `paidToBob`" (rather than querying `V := getChannel(id)` then we can put a do that via the planned "objective approval system" #213 . We could put a TODO in to show that.

lalexgap:
Closing this in favour of my branch #618

PR #619, opened 2022-04-28
Fix Remove Proposal Serde

lalexgap:
Since `turnNum` is private on the `Remove` proposal , `Remove` proposals were not being properly serialized/deserialized. This resulted in the turn num always getting set to 0 on a remove proposal when it was sent across the wire.

This PR does two things:
- Enhances or serde tests so we are now testing the remove proposal serde
- Add serde logic for `Remove` so it uses a `jsonRemove` to store private fields.


channel/consensus_channel/serde.go
```diff
@@ -42,6 +42,41 @@ func (a *Add) UnmarshalJSON(data []byte) error {
...
+// making it suitable for serialization
+// embedded structs are moved to name fields for easier serialization
+type jsonRemove struct {
+	TurnNum     uint64
+	Target      types.Destination
+	LeftAmount  *big.Int
+	RightAmount *big.Int
+}
+
+// MarshalJSON returns a JSON representation of the Add
```
kerzhner, line 55:
```suggestion
// MarshalJSON returns a JSON representation of the Remove
```

kerzhner (approved):

PR #618, opened 2022-04-28
Virtual defunding integration

lalexgap:
This PR introduces a virtual defunding integration test as well as various changes to get the test up and running.

Fixes #651
Fixes #642
Fixes #637

TODO:
- [x] ~Investigate flickering testhttps://github.com/statechannels/go-nitro/runs/6277375843?check_suite_focus=true~ Created https://github.com/statechannels/go-nitro/issues/637
- [x] Resolve #637 and update the test to use multiple virtual channels https://github.com/statechannels/go-nitro/pull/644
- [x] Use non-zero paidToBob
- [x] Check that ledger channel has expected outcome[](https://github.com/statechannels/go-nitro/issues/new)(https://github.com/statechannels/go-nitro/pull/618/commits/2a0d37791e5ba20493c3c776881b514ebc33cff3)
- [x] [Unskip tests](https://github.com/statechannels/go-nitro/issues/643)
- [x] Add message delay test https://github.com/statechannels/go-nitro/pull/618/commits/421c33ed8edea2f099ad51b90f03f39a03fc2b6a

```
<snip>
0xBBB676: 19:01:09.276606 engine.go:127: Objective VirtualDefund-0xa2a3a731b9d2650365e5071237e3ea6569a5ae11584b19af8047f2f8645670e2 is complete & returned to API
0xAAA662: 19:01:09.276964 engine.go:143: Handling inbound message {To:0xAAA6628Ec44A8a742987EF3A114dDFE2D4F7aDCE Proposals:[] States:[{ObjectiveId:VirtualDefund-0xa2a3a731b9d2650365e5071237e3ea6569a5ae11584b19af8047f2f8645670e2 ChannelId:0xa2a3a731b9d2650365e5071237e3ea6569a5ae11584b19af8047f2f8645670e2 TurnNum:3}]}
0xAAA662: 19:01:09.277516 engine.go:330: Objective VirtualDefund-0xa2a3a731b9d2650365e5071237e3ea6569a5ae11584b19af8047f2f8645670e2 is WaitingForCompleteLedgerDefunding
0x111A00: 19:01:09.285471 engine.go:330: Objective VirtualDefund-0xa2a3a731b9d2650365e5071237e3ea6569a5ae11584b19af8047f2f8645670e2 is WaitingForCompleteLedgerDefunding
[...]

PR #620, opened 2022-04-29
refactor: remove turn num from add/remove

### LINKED ISSUE #515: Propose API is confusing
We could create an issue for that!

_Originally posted by @lalexgap in https://github.com/statechannels/go-nitro/pull/513#discussion_r846463348_

### LINKED ISSUE #623: Do not use hardcoded turn number for remove proposal
Our consensus_channel [logic](https://github.com/statechannels/go-nitro/blob/1b4e0577a07b8d3944f75a4469de6a50621d4b59/channel/consensus_channel/consensus_channel.go#L751)  assumes that a proposal will have a turn number **after** the current consensus turn.

By using a hardcoded value of 3 this works for the simple case of a ledger channel funding one virtual channel, but this will break down with more payments. We should not be using a specific turn number when we're generating/ accepting proposals

_Originally posted by @lalexgap in https://github.com/statechannels/go-nitro/pull/618#discussion_r861243483_

lalexgap:
Fixes #623 
Fixes #515

This PR refactors proposals by moving `TurnNum` from the `Add`/`Remove` to the `SignedProposal`.

This helps simplify the logic around proposals. In cases like `SignNextProposal` and `Propose` we were ignoring or overwriting the turn number on the proposal. Now, the consensus channel just sets the turn number on the `SignedProposal` when it signs the proposal. The turn number is validated by the follower when they receive the signed proposal. We still use the `TurnNum` to sort the `SignedProposals` so they are processed correctly.

This also resolves an issue where our `virtualdefund` protocol was using a fixed turn number (`FinalTurnNum=2`) when creating the `Remove` proposals. This resulted in virtual defunding failing if the ledger channel funded multiple virtual channels.  This will be tested in #618.


channel/consensus_channel/consensus_channel.go
```diff
@@ -526,43 +526,21 @@ func (p *Proposal) Type() ProposalType {
...
 // ChannelId returns the id of the ConsensusChannel which receive the proposal.
 func (p SignedProposal) ChannelId() types.Destination {
 	return p.Proposal.ChannelID
 }
 
-// TurnNum returns the turn number of the proposal.
-func (p SignedProposal) TurnNum() uint64 {
-	return p.Proposal.TurnNum()
+// SortInfo returns the channelId and turn number so the proposal can be easily sorted.
+func (p SignedProposal) SortInfo() (types.Destination, uint64) {
```
lalexgap, line 540: Using `SortInfo` instead of `TurnNum()` and `ChannelId()` makes it easier to avoid naming conflicts with public struct fields.

channel/consensus_channel/consensus_channel.go
```diff
@@ -587,19 +565,19 @@ func (p *Proposal) Target() types.Destination {
 type SignedProposal struct {
 	state.Signature
 	Proposal Proposal
+	TurnNum  uint64
```
lalexgap, line 568: This is the key change, by moving the `TurnNum` to the `SignedProposal` means we only have to deal with a turn number when we are signing/validating a signed proposal.

This lets the follower easily check if a proposal matches what they're expecting without specifying a turn number. This means the leader can solely be responsible for choosing the turn order for proposals.

We can still use the `TurnNum` to sort the proposals so they are returned in order.

channel/consensus_channel/leader_channel.go
```diff
@@ -51,25 +55,19 @@ func (c *ConsensusChannel) IsProposedNext(g Guarantee) (bool, error) {
...
 	if c.MyIndex != Leader {
 		return SignedProposal{}, ErrNotLeader
 	}
 
-	// TODO: the Propose API should be less confusing!
-	// Currently, the TurnNum is ignored, and Propose could easily
-	// return the same ChannelId that it's been passed
-	proposal.ChannelID = c.Id
-
+	if proposal.ChannelID != c.Id {
```
lalexgap, line 63: I think it makes more sense to enforce the channelId is correct, instead of overwriting it.

channel/consensus_channel/consensus_channel.go
```diff
@@ -526,43 +526,21 @@ func (p *Proposal) Type() ProposalType {
...
-		return p.ToAdd.turnNum
-	} else {
-		return p.ToRemove.turnNum
-	}
-}
-
 // Equal returns true if the supplied Proposal is deeply equal to the receiver, false otherwise.
 func (p *Proposal) Equal(q *Proposal) bool {
-	return p.ToAdd.equal(q.ToAdd) && p.ToRemove.equal(q.ToRemove)
+	return p.ChannelID == q.ChannelID && p.ToAdd.equal(q.ToAdd) && p.ToRemove.equal(q.ToRemove)
```
lalexgap, line 531: I think we forgot to add a check for `ChannelID` here.

channel/consensus_channel/leader_channel.go
NiloCK: Agreed. Silently overwriting the input is not expected. A different API surface which didn't allow for a passed-in channelID would take responsibility for setting that ID, but that's not what this method is.

channel/consensus_channel/consensus_channel.go
NiloCK: Main concern: is the turnNum still implicitly part of the data endorsed by the `state.Signature` here, or has this change introduced opportunity for persons to "spoof" turn numbers in some way.

My observation: looks fine / safe.

The signature generated in `Propose()` is on the the updated `vars` object, which has handled the proposal and bumped its turnNumber.

channel/consensus_channel/leader_channel_test.go
```diff
@@ -207,10 +206,9 @@ func TestLeaderChannel(t *testing.T) {
 			guarantee(vAmount, channel1Id, alice, bob),
 		)
 		c := testChannel(startingOutcome, emptyQueue())
-		proposalMade := createAdd(cId, 1, targetChannel)
-		expectedSp := aliceSignedProposal(c.current.Vars, proposalMade).SignedProposal
+		proposalMade := createAdd(cId, targetChannel)
+		expectedSp := aliceSignedProposal(c.current.Vars, proposalMade, 1).SignedProposal
 
-		proposalMade.SetTurnNum(9001)
```
NiloCK, line 213: This test case can be removed now? the block

```
{
		msg := "ok:provided turn number is ignored"
		startingOutcome := makeOutcome(
			allocation(alice, aBal-amountAdded),
			allocation(bob, bBal),
			guarantee(vAmount, channel1Id, alice, bob),
		)
		c := testChannel(startingOutcome, emptyQueue())
		proposalMade := createAdd(cId, targetChannel)
		expectedSp := aliceSignedProposal(c.current.Vars, proposalMade, 1).SignedProposal

		t.Run(msg, testPropose(c, proposalMade, expectedSp, nil))
	}
```

NiloCK (approved):
LGTM!

channel/state/signedstate.go
```diff
@@ -175,3 +175,10 @@ func (ss SignedState) ChannelId() types.Destination {
 func (ss SignedState) TurnNum() uint64 {
 	return ss.state.TurnNum
 }
+
+// SortInfo returns the channel id and turn number of the state, so the state can be easily sorted.
+func (ss SignedState) SortInfo() (types.Destination, uint64) {
+	cId, _ := ss.State().ChannelId()
```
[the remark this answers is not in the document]
lalexgap: I agree! Perhaps `ChannelID` should just `panic` and the channel constructor could be responsible for returning an error if the fixed part is invalid?

channel/consensus_channel/leader_channel_test.go
lalexgap: Good catch! Removed in https://github.com/statechannels/go-nitro/pull/620/commits/cf1888ae7f85e54734118c6fcbe0edccd423ebac

channel/consensus_channel/consensus_channel.go
lalexgap: > Main concern: is the turnNum still implicitly part of the data endorsed by the `state.Signature` here, or has this change introduced opportunity for persons to "spoof" turn numbers in some way.

The turn number is still part of the `state.Signature` so we're still validating the turn number and preventing spoofing.

PR #624, opened 2022-04-29
Switch Actor Address from field to function

kerzhner:
An actor's address is determined by the private key. Switching to a function has the following benefits:
- Since the address is calculated, we avoid the possibility of an actor where the private key and the address do not match.
- Simplifies adding a new actor.


NiloCK (approved):
Nice idea & implementation. Glad to see the vanity addresses added as visible godoc.

PR #626, opened 2022-05-02
restrict ObjectiveEvent to single proposal/state

NiloCK:
This is an alternative to #597.

---

#### Code quality

#### Project management

### DIFF
--- a/channel/channel.go
+++ b/channel/channel.go
@@ -266,19 +266,6 @@ func (c *Channel) AddSignedState(ss state.SignedState) bool {
 	return true
 }
 
-// AddSignedStates adds each signed state in the passed slice. It returns true if all signed states were added successfully, false otherwise.
-// If one or more signed states fails to be added, this does not prevent other signed states from being added.
-func (c *Channel) AddSignedStates(sss []state.SignedState) bool {
-	allOk := true
-	for _, ss := range sss {
-		ok := c.AddSignedState(ss)
-		if !ok {
-			allOk = false
-		}
-	}
-	return allOk
-}
-
 // SignAndAddPrefund signs and adds the prefund state for the channel, returning a state.SignedState suitable for sending to peers.
 func (c *Channel) SignAndAddPrefund(sk *[]byte) (state.SignedState, error) {
 	return c.SignAndAddState(c.PreFundState(), sk)
--- a/protocols/interfaces.go
+++ b/protocols/interfaces.go
@@ -48,9 +48,9 @@ type AdjudicationStatus struct {
 
 // ObjectiveEvent holds information used to update an Objective. Some fields may be nil.
 type ObjectiveEvent struct {
-	ObjectiveId     ObjectiveId
-	SignedStates    []state.SignedState
-	SignedProposals []consensus_channel.SignedProposal
+	ObjectiveId    ObjectiveId
+	SignedState    state.SignedState
+	SignedProposal consensus_channel.SignedProposal
 }
 
 // Storable is an object that can be stored by the store.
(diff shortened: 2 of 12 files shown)


channel/channel.go
```diff
@@ -266,19 +266,6 @@ func (c *Channel) AddSignedState(ss state.SignedState) bool {
 	return true
 }
 
-// AddSignedStates adds each signed state in the passed slice. It returns true if all signed states were added successfully, false otherwise.
-// If one or more signed states fails to be added, this does not prevent other signed states from being added.
-func (c *Channel) AddSignedStates(sss []state.SignedState) bool {
```
NiloCK, line 271: Interesting side effect of this change is that this method was no longer used.

protocols/virtualdefund/virtualdefund.go
```diff
@@ -366,7 +366,7 @@ func (o Objective) Update(event protocols.ObjectiveEvent) (protocols.Objective,
 
 	updated := o.clone()
 
-	for _, ss := range event.SignedStates {
+	if ss := event.SignedState; len(ss.Signatures()) != 0 {
```
NiloCK, line 369: these checks guard against attempts at processing zero-valued SignStates

lalexgap: This is outside the scope of this PR but should we have a canonical way of checking if a `SignedState` is a zero value? Here's another place we perform a similar [check](https://github.com/statechannels/go-nitro/blob/ag/fix-wait-for-completed-objective-ids/protocols/messages.go#L35:L35)

protocols/interfaces.go
```diff
@@ -48,9 +48,9 @@ type AdjudicationStatus struct {
 
 // ObjectiveEvent holds information used to update an Objective. Some fields may be nil.
 type ObjectiveEvent struct {
-	ObjectiveId     ObjectiveId
-	SignedStates    []state.SignedState
-	SignedProposals []consensus_channel.SignedProposal
+	ObjectiveId    ObjectiveId
+	SignedState    state.SignedState
```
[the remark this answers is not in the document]
lalexgap: Do we have any constraints on whether an `ObjectiveEvent` can contain both a `SignedProposal` and a `SignedState`?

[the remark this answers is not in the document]
NiloCK: This is checked by the engine when processing messages. msg.SignedStates and msg.Proposals run though [`msg.Type()`](https://github.com/statechannels/go-nitro/blob/9b3db3d9e199f7a8d73c777713ea5a74cb79bb7a/protocols/messages.go#L44) to create the events from the message payloads.

protocols/virtualdefund/virtualdefund.go
NiloCK: I don't know if it's better to have an `IsNonZero` method or to have pointer structs that enable nil checks.

protocols/interfaces.go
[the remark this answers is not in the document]
lalexgap: Yup, due to our messaging format we expect `ObjectiveEvent` to contain either a `SignedState` or `SignedProposal`.

However it'd be nice if `Update` wasn't implicitly coupled to the message format. If we rely on the expectation that an `ObjectiveEvent` is either a `SignedProposal` or a `SignedState` I think we should make that expectation explicit for the `Update` API. We could this by introducing a `Type` function similar to [messages](https://github.com/statechannels/go-nitro/blob/9b3db3d9e199f7a8d73c777713ea5a74cb79bb7a/protocols/messages.go#L44) but this would add more boilerplate.

lalexgap (approved):

PR #628, opened 2022-05-02
waitTimeForCompletedObjectiveIds only waited for one objective id

lalexgap:
`waitTimeForCompletedObjectiveIds` would iterate through the `completed` map to determine if it was all done or not. The problem is that the `completed` map only gets populated when an objective is complete. This means that `waitTimeForCompletedObjectiveIds` will return after one objective is completed, since the `isDone = isDone && completed[id]` will always return true.

This PR solves this by iterating through the `ids` passed into `waitTimeForCompletedObjectiveIds`. That way we're checking that all objectives are complete.


NiloCK (approved):
Good catch. Fix looks good to me.

Issue #568, opened 2022-04-20
If the engine panics no logs are written out


lalexgap:
Currently when the engine panics we don't write the file to the log destination. Since the `engine` panics on any [error](https://github.com/statechannels/go-nitro/blob/5e4b8b105e193b92fc81818a8dadfe7cce9bd3d8/client/engine/engine.go#L115), it means we won't have logs available when we're debugging the error.

geoknee:
https://github.com/golang/go/issues/41355

PR #404, opened 2022-03-25
Write logs to a `bytes.Buffer` during tests, and flush to a file when cleaning up the test

geoknee:
Closes #375, but reveals a different problem: 

```
> go test ./client_test -count=1000
--- FAIL: TestMultiPartyVirtualFundIntegration (1.02s)
    helpers_test.go:61: Objective ids [VirtualFund-0xfc5ee41721a620147acd2814517c0c06618add28bd0b87d0b5a9377e21247b42] failed to complete on client 0xBBB676f9cFF8D242e9eaC39D063848807d3D1D94 within 1s
--- FAIL: TestMultiPartyVirtualFundIntegration (1.02s)
    helpers_test.go:61: Objective ids [VirtualFund-0x56e484152ee7b502e7f54f34c2893daaaed11f6294e64bb73beffbaec60d3b37] failed to complete on client 0xBBB676f9cFF8D242e9eaC39D063848807d3D1D94 within 1s
--- FAIL: TestMultiPartyVirtualFundIntegration (1.03s)
    helpers_test.go:61: Objective ids [VirtualFund-0xd47397b0e9acb2e5223437b01193da1e8f93fe469c59fd1c67b1e26ba6bf019c] failed to complete on client 0xBBB676f9cFF8D242e9eaC39D063848807d3D1D94 within 1s
FAIL
FAIL	github.com/statechannels/go-nitro/client_test	60.043s
FAIL

```

I think it is still worth merging because it shortens the time a log file is kept open for, and keeps IO out of the main run time of the test. 

---

#### Code quality

- [x] I have written clear commit messages
- [x] I have performed a self-review of my own code
- [x] This change does not have an unduly wide scope
- [x] I have separated logic changes from refactor changes (formatting, renames, etc.)
- [x] I have commented my code wherever necessary (can be 0)
- [x] I have added tests that prove my fix is effective or that my feature works, if necessary
- [x] I have added comprehensive [godoc](https://go.dev/blog/godoc) comments 

#### Project management

- [x] I have assigned myself to this PR
- [x] I have linked the appropriate github issue
- [x] I have assigned this PR to the appropriate GitHub project
- [x] I have assigned this PR to the appropriate GitHub Milestone

Issue #375, opened 2022-03-22
Running Large amount of tests fails due to open log file handles


lalexgap:
```
❯ go test ./client_test -count=1000
2022/03/21 18:33:34 open ../artifacts/virtualfund_client_test.log: too many open files
```

andrewgordstewart:
I believe the current integration test creates one open file `f` per client, but each `f` is opening the same filename.

geoknee:
@lalexgap can you give a bit more context on why you might want to run so many tests concurrently?

geoknee:
Discussed in standup: the `-count=N` flag seems pretty useful for unit tests, but maybe less so for integration tests which may be more resource hungry (think log files, database connections, etc). So one solution may be to silo integration tests away from unit tests, so they can be run with different test parameters.

PR #629, opened 2022-05-02
Ensure logs get written out on failure

lalexgap:
Fixes https://github.com/statechannels/go-nitro/issues/568

We made some changes in https://github.com/statechannels/go-nitro/pull/404 to allow for more concurrent tests to run and log out to the log files.  We a keep a buffer in memory that we write to and we flush that out in `t.Cleanup`

However if a `panic` happens (like for the current `engine` error handling) the clean up function does not get triggered and the log doesn't get written out. This means that when an error occurs we won't get the log files to aid in debugging 😢 
(Here's an [example](https://github.com/statechannels/go-nitro/actions/runs/2259743068))

To resolve this I've just reverted the client tests to write directly to the file instead of writing to a buffer. This means that less client tests can run concurrently (using the `go test -count` flag) but it does mean that we will always generate log files.

### DIFF
--- a/client_test/directdefund_integration_test.go
+++ b/client_test/directdefund_integration_test.go
@@ -2,7 +2,6 @@
 package client_test // import "github.com/statechannels/go-nitro/client_test"
 
 import (
-	"bytes"
 	"testing"
 
 	"github.com/statechannels/go-nitro/client"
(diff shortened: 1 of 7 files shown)


NiloCK (approved):
Was it only on the `--count` flag that we were experiencing the open file congestion issue?

lalexgap:
> Was it only on the `--count` flag that we were experiencing the open file congestion issue?

I believe so. I opened the original [issue](https://github.com/statechannels/go-nitro/issues/375) when using quite large values with the count flag.

PR #630, opened 2022-05-03
testmessageservice: serialize outgoing messages, deserialize incoming ones

geoknee:
This :
* will make it possible to wrap the messages more easily (for debugging / instrumentation purposes)
* makes it more realistic

I've also done some refactoring which I think makes things easier to read and understand.

### DIFF
--- a/client/engine/messageservice/test-messageservice.go
+++ b/client/engine/messageservice/test-messageservice.go
@@ -58,8 +58,8 @@ func NewTestMessageService(address types.Address, broker Broker, maxDelay time.D
 	}
 
 	tms.connect(broker)
-	tms.routeInbound()
-	tms.routeOutbound(broker)
+	go tms.routeInbound()
+	go tms.routeOutbound(broker)
 
 	return tms
 }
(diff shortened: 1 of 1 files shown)


client/engine/messageservice/test-messageservice.go
```diff
@@ -83,46 +89,51 @@ func (t TestMessageService) dispatchMessage(message protocols.Message, b Broker)
...
+	}
+}
 
-	b.services[t.address] = t
+// routeFromPeers listens for messages from peers, deserializes them and feeds them to the engine
+func (tms TestMessageService) routeFromPeers() {
+	for message := range tms.fromPeers {
+		msg, err := protocols.DeserializeMessage(string(message))
+		if err != nil {
+			panic(err)
```
NiloCK, line 116:
```suggestion
			panic(fmt.Errorf("could not deserialize message :%w", err))
```

client/engine/messageservice/test-messageservice.go
```diff
@@ -25,6 +25,9 @@ type TestMessageService struct {
 	in       chan protocols.Message // for recieving messages from engine
 	out      chan protocols.Message // for sending message to engine
 	maxDelay time.Duration          // the max delay for messages
+
+	// connection with Peers:
+	fromPeers chan []byte // for receiving serialized messages from peers
```
NiloCK, line 30: a little nit for your information. In this struct:
- the comment `// connection to Engine:` is interpreted as documentation for `in`
- `// connection with Peers:` as documentation for `fromPeers`
- The actual docstrings intended for each field (eg, `// for receiving serialized messages from peers`) are omitted.

Codebase is full of these, as they are a natural way to document things in js/ts, but we should attempt to habituate ourselves to the godoc style

PR #631, opened 2022-05-03
Git ignore artifacts directory

kerzhner:
Do not ignore `.log` files. This helps developers notice if tests are sprinkling log files outside of the artifacts directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,5 +14,8 @@
 # Dependency directories (remove the comment below to include it)
 # vendor/
 
-# Logs
-*.log
+# The artifacts contains only logs, for now
+artifacts
+
+# OSX specific files
+.DS_Store


.gitignore
```diff
@@ -14,5 +14,8 @@
...
 # vendor/
 
-# Logs
-*.log
\ No newline at end of file
+# The artifacts contains only logs, for now
+artifacts
+
+# OSX specific files
+.DS_Store
```
NiloCK, line 21: As the person who doesn't know: What's that?

NiloCK (approved):

.gitignore
kerzhner: At the risk of coming off brisk: I would recommend googling this :). These files pop up on my machine after certain command line operations.

PR #634, opened 2022-05-03
Consistent client test names and log files

kerzhner:
- Integration removed from names since all client tests are integration tests.
- Log file names are snake case test-function names.


lalexgap (approved):
👍  Could we rename `virtual_fund_message_delay_test.go` to  `virtualfund_message_delay_test.go` to be consistent with the other file names?

kerzhner:
> 👍 Could we rename `virtual_fund_message_delay_test.go` to `virtualfund_message_delay_test.go` to be consistent with the other file names?

Addressed in https://github.com/statechannels/go-nitro/pull/636.

PR #636, opened 2022-05-03
Rename client_test files

kerzhner:
Each file name is the camel case of the main test function with 'test' shifted to the end.


geoknee (commented):
I'm onboard with removing `_integration`, but I would prefer to not introduce an extra `_`  e.g taking. `virtualfund`=>`virtual_fund`. See https://github.com/statechannels/go-nitro/pull/634#pullrequestreview-960840028

kerzhner:
> > should we switch our test case functions from `VirtualFund` to `Virtualfund` then?
> 
> Good point -- I think probably yes?

`Virtualfund` and `Directfund` are not as readable as `VirtualFund` and `DirectFund`, so I decided to not change casing on the test function names. Also, is seems like the rest of the codebase uses Pascal or Camel casing for these types of functions and variables.

Extra `_` has been removed from file names per @geoknee preference. I am ok with the inconsistency between test case names and test file names.

kerzhner:
> 👍 with a nit: [multiparty](https://www.dictionary.com/browse/multiparty) seems to be a word. Take or leave.

Thanks! My goal is to stay consistent in mapping test function names to file names.

PR #638, opened 2022-05-03
Stop follower from sending proposals signed by leader

### LINKED ISSUE #635: The follower can send signed proposals with the leader's signature
Right now the `ProposalQueue` for both the leader and follower will only ever contain proposals signed by the leader.

The follower only signs the proposal in `signNextProposal` and then sends that out in a message.

The problem is that the [CreateSignedProposalMessage](https://github.com/statechannels/go-nitro/blob/ag/virtual-defunding-integration/protocols/messages.go#L220:L220) always tries to send whatever is in the proposal queue.

This results in the follower sending proposals with the leader's signatures!  This causes the leader's receive function to fail with a `WrongSigner` error

lalexgap:
Towards #618.

Fixes #635.

Previously `CreateSignedProposalMessage` would include the `ProposalQueue` when sending out signed proposals. However since the `ProposalQueue` only contains the leader signatures this means that the follower could send out signed proposals with the leader's signatures.

This PR fixes this by only including the proposal queue when the leader is sending messages. The follower just sends individual signed proposals.

### DIFF
--- a/protocols/messages.go
+++ b/protocols/messages.go
@@ -211,13 +211,8 @@ func SummarizeMessage(m Message) MessageSummary {
 
 // CreateSignedProposalMessage returns a signed proposal message addressed to the counterparty in the given ledger
 // It contains the provided signed proposals and any proposals in the proposal queue.
-func CreateSignedProposalMessage(ledger *consensus_channel.ConsensusChannel, sp ...consensus_channel.SignedProposal) Message {
-	recipient := ledger.Leader()
-	if ledger.IsLeader() {
-		recipient = ledger.Follower()
-	}
-	// Append the proposals to any existing proposals in the queue
-	proposals := append(ledger.ProposalQueue(), sp...)
+func CreateSignedProposalMessage(recipient types.Address, proposals ...consensus_channel.SignedProposal) Message {
+
 	payloads := make([]MessagePayload, len(proposals))
 	for i, sp := range proposals {
 		id := getProposalObjectiveId(sp.Proposal)
--- a/protocols/virtualfund/virtualfund.go
+++ b/protocols/virtualfund/virtualfund.go
@@ -580,7 +580,8 @@ func (o *Objective) proposeLedgerUpdate(connection Connection, sk *[]byte) (prot
 		return protocols.SideEffects{}, err
 	}
 
-	message := protocols.CreateSignedProposalMessage(connection.Channel)
+	recipient := ledger.Follower()
+	message := protocols.CreateSignedProposalMessage(recipient, connection.Channel.ProposalQueue()...)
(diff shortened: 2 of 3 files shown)


protocols/virtualfund/virtualfund.go
```diff
@@ -580,7 +580,8 @@ func (o *Objective) proposeLedgerUpdate(connection Connection, sk *[]byte) (prot
 		return protocols.SideEffects{}, err
 	}
 
-	message := protocols.CreateSignedProposalMessage(connection.Channel)
+	recipient := ledger.Follower()
```
geoknee, line 583: On L573 we refer to "proposer" but I think it should be "leader".

protocols/messages.go
```diff
@@ -211,13 +211,8 @@ func SummarizeMessage(m Message) MessageSummary {
...
 // CreateSignedProposalMessage returns a signed proposal message addressed to the counterparty in the given ledger
 // It contains the provided signed proposals and any proposals in the proposal queue.
-func CreateSignedProposalMessage(ledger *consensus_channel.ConsensusChannel, sp ...consensus_channel.SignedProposal) Message {
-	recipient := ledger.Leader()
-	if ledger.IsLeader() {
-		recipient = ledger.Follower()
-	}
-	// Append the proposals to any existing proposals in the queue
-	proposals := append(ledger.ProposalQueue(), sp...)
+func CreateSignedProposalMessage(recipient types.Address, proposals ...consensus_channel.SignedProposal) Message {
```
geoknee, line 214: I wonder if we could improve some of the function names. I don't have a great suggestion, but itt seems awkward that the follower creates a signed "proposal" message. It sounds like they are making a proposal, which they aren't.

Is the follower sending a "countersigned proposal"? 

Are both roles merely sending a LedgerProtocolMessage?

geoknee (approved):

protocols/virtualfund/virtualfund.go
lalexgap: Resolved in https://github.com/statechannels/go-nitro/pull/638/commits/76932e82ce4448c025e2b45fa04b543497dc4aa5

protocols/messages.go
lalexgap: >I wonder if we could improve some of the function names. I don't have a great suggestion, but it seems awkward that the follower creates a signed "proposal" message. It sounds like they are making a proposal, which they aren't.

I think the problem is that a `SignedProposal` is overloaded. We used a `SignedProposal` in a message to either:
- send a new proposal from the leader to the follower
- accept a proposal as the follower by sending out a `SignedProposal` with the follower's signature.

Maybe we should look at renaming `SignedProposal` in the message so it's less confusing?

PR #640, opened 2022-05-03
Only sign proposal if it is next in the queue in virtual defunding

### LINKED ISSUE #639: Virtual defund will call signNextProposal even if our proposal isn't next
The follower will call signNextProposal if the [proposal queue contains the proposal we want](https://github.com/statechannels/go-nitro/blob/b426f009f463c5e62333a9b5772e3ab53248d340/protocols/virtualdefund/virtualdefund.go#L287). However the **next** proposal might not be one we're interested in.

Like virtual funding we should only sign the next proposal if the next proposal is for our objective.

lalexgap:
Towards #618.

Fixes #639.

Updated virtual defunding to follower a similar pattern as virtual funding and only call `signNextProposal` if the proposal next in the queue is what we want.

### DIFF
--- a/channel/consensus_channel/consensus_channel.go
+++ b/channel/consensus_channel/consensus_channel.go
@@ -133,8 +133,8 @@ func (c *ConsensusChannel) IncludesTarget(target types.Destination) bool {
 	return c.current.Outcome.includesTarget(target)
 }
 
-// HasRemovalBeenProposedFor returns whether or not a proposal exists to remove the guaranatee for the target.
-func (c *ConsensusChannel) HasRemovalBeenProposedFor(target types.Destination) bool {
+// HasRemovalBeenProposed returns whether or not a proposal exists to remove the guaranatee for the target.
+func (c *ConsensusChannel) HasRemovalBeenProposed(target types.Destination) bool {
 	for _, p := range c.proposalQueue {
 		if p.Proposal.Type() == RemoveProposal {
 			remove := p.Proposal.ToRemove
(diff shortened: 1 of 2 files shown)


geoknee (approved):

channel/consensus_channel/consensus_channel.go
```diff
@@ -146,6 +146,18 @@ func (c *ConsensusChannel) HasRemovalBeenProposedFor(target types.Destination) b
 	return false
 }
 
+// HasRemovalBeenProposedNext returns whether or not the next proposal in the queue is a remove proposal for the given target
+func (c *ConsensusChannel) HasRemovalBeenProposedNext(target types.Destination) bool {
+
+	if len(c.proposalQueue) == 0 {
```
geoknee, line 152: I'm looking at this line https://github.com/statechannels/go-nitro/blob/b426f009f463c5e62333a9b5772e3ab53248d340/channel/consensus_channel/leader_channel.go#L24 and wondering if there is an opportunity to increase consistency by reusing `latestProposedVars()`?

geoknee (approved):

channel/consensus_channel/consensus_channel.go
geoknee: Although `latestProposedVars()` is a little difficult to follow 😟

lalexgap: `latestProposedVars` is a bit more difficult for removals since we're looking for an absence of a guarantee. I think that would still work but it seems easier to just check the proposal queue.

PR #641, opened 2022-05-03
Add Completed status for Objective and ignore completed objectives

lalexgap:
Towards #618 

Adds a completed status for objectives so we can ignore completed objectives instead of trying to process any messages for them.

`Crank` now updates the status to `Completed` once an objective is done.

### DIFF
--- a/client/engine/store/memstore.go
+++ b/client/engine/store/memstore.go
@@ -98,7 +98,7 @@ func (ms *MemStore) SetObjective(obj protocols.Objective) error {
 
 	// Objective ownership can only be transferred if the channel is not owned by another objective
 	prevOwner, isOwned := ms.channelToObjective.Load(obj.OwnsChannel().String())
-	if status := obj.GetStatus(); status == protocols.Approved || status == protocols.Completed {
+	if status := obj.GetStatus(); status == protocols.Approved {
 		if !isOwned {
 			ms.channelToObjective.Store(obj.OwnsChannel().String(), obj.Id())
 		}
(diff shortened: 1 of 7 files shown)


client/engine/store/memstore.go
```diff
@@ -98,7 +98,7 @@ func (ms *MemStore) SetObjective(obj protocols.Objective) error {
 
 	// Objective ownership can only be transferred if the channel is not owned by another objective
 	prevOwner, isOwned := ms.channelToObjective.Load(obj.OwnsChannel().String())
-	if obj.GetStatus() == protocols.Approved {
+	if status := obj.GetStatus(); status == protocols.Approved || status == protocols.Completed {
```
geoknee, line 101: The comment on L99 could be improved to reflect what is going on. 

In order to have the objective being stored successfully take ownership of `.OwnsChannel`* , the objective must either be approved or completed and the channel must not be owned by any other objective.

I'm a bit curious about why we need to include the "completed" case. Won't the objective want to relinquish ownership of the channel if it is completed.

* Isn't `.OwnsChannel` poorly named?

https://github.com/statechannels/go-nitro/blob/b426f009f463c5e62333a9b5772e3ab53248d340/protocols/interfaces.go#L81-L82

 The source of truth about whether the objective actually owns the channel is the store itself. Should we rename it to `PrincipalChannel` or something?

geoknee (approved):

client/engine/store/memstore.go
lalexgap: > I'm a bit curious about why we need to include the "completed" case. Won't the objective want to relinquish ownership of the channel if it is completed?

I think you're right. I made this change without too much though. I've updated that in a4a9d84.

Issue #609, opened 2022-04-27
Publish v0.0.1


NiloCK:
At some appropriate time we should pin an initial version for go-nitro.

- semver applies per its regular usage
- versioning is managed by git tags:
    - on some acceptable commit of main, run `git tag v0.0.1`, `git push origin v0.0.1` for our init version
    - visit [pkg.go.dev/statechannels/go-nitro](http://pkg.go.dev/statechannels/go-nitro) to ensure that it has picked up the tagged version
- the **special rules** I had remembered concerned major version bumps from `n>1` → `n+1`. (IE, this rule does not apply to the `0.x.y` → `1.0.0` major bump):
    - for these later version bumps, the declared module name in `go.mod` must match the major version number in a specific fashion: `[module github.com/statechannels/go-nitro/v2](http://github.com/statechannels/go-nitro/v2)` for versions v2.x.y, `.../v3` for versions v3.x.y, etc.
    - this rule exists more or less to add friction to the upgrade path across major versions bumps, so that module consumers do not accidentally / automatically pull down breaking changes.

geoknee:
Similar to #390.

PR #649, opened 2022-05-05
Consider: encapsulate protocol internals

NiloCK:
As a consideration ahead of #609 (publish v0.0.1):

**Motivating Question:**

What is it that we intend to expose for consuming applications? What do consuming applications "need to know"?

**Changes**

PR moves all code other than `client` and its descendants into the `internal` folder, hiding them from consuming applications & ~excluding them from the generated documentation for the package~ removing them from top-level of the generated documentation.

PR splits the work into two commits to preserve git history / line blame:
- git mv ./x ./internal/x
- global find/replace `go-nitro/x` -> `go-nitro/internal/x`

existing test suite / linting runs & passes.

- [x] Is it a significant change to architecture or design?

Very significant change to the exported package!

---

#### Code quality

- [ ] ~I have commented my code wherever necessary (can be 0)~ no new code
- [ ] ~I have added comprehensive [godoc](https://go.dev/blog/godoc) comments~

#### Project management


geoknee:
We've found it useful in the past to expose things like solidity source code and client utilities (other than the full client itself) -- for example the `@statechannels/nitro-protocol` npm package. Interested parties could tinker with those components (extend them, learn from them, build their own applications / clients). This was especially useful during development of the full client, i.e. before it was ready to be used. 

As another example, it is useful in our project to be able to import things like abi encoding utilities from `go-ethereum` -- so I think we benefit from that project's choice _not_ to house those utilities in an `internal` folder. 

That said, I think as the `go-nitro` client API develops we might be able to hide say the `protocols` folder. That seems tightly coupled to the client / engine anyway. But even then I'm not totally sure we would to prevent folks directly importing it.

It's an interesting suggestion, but to me it feels like we would stand to lose slightly more than we would gain from implementing it.

NiloCK:
I think I mostly agree, but some devil's advocacy:
- there's a difference in _kind_ between geth and go-nitro. geth is itself an application (in particular it's the base application for a large ecosystem), but go-nitro is intended as a library
- taking the sales perspective: is it easier for people to adopt a large bundle of complicated protocols or a little widget that pays for things quickly?

A different option: slicing `client` and its children into their own package. This
- leaves all of the protocol & utility code accessible
- has even stronger encapsulation properties for applications importing `nitro-client`

NiloCK:
> I'm not sure I follow. Do you mean to put this code in a separate _module_?

Whoops. yes, that's what I meant.

PR #657, opened 2022-05-06
Minor fixes / refactors

NiloCK:
This PR is the result of my new local backlogging system for issues too small to be logged as issues. It is scattershot, but each commit is
- self-contained
- small
- (I think) uncontroversial

Read per-commit.

---

#### Code quality

  - See above
- [ ] ~I have added tests that prove my fix is effective or that my feature works, if necessary~
- [ ] ~I have added comprehensive [godoc](https://go.dev/blog/godoc) comments~

#### Project management

- [ ] ~I have linked the appropriate github issue~

### DIFF
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -343,7 +343,7 @@ func (e *Engine) attemptProgress(objective protocols.Objective) (outgoing Object
 	return
 }
 
-// SpawnConsensusChannelIfDirectFundObjective will attempt to create and store a ConsensusChannel derived from the supplied Objective iff it is a directfund.Objective.
+// SpawnConsensusChannelIfDirectFundObjective will attempt to create and store a ConsensusChannel derived from the supplied Objective if it is a directfund.Objective.
 //
 // The associated Channel will remain in the store.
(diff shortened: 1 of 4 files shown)


client/engine/engine.go
```diff
@@ -343,7 +343,7 @@ func (e *Engine) attemptProgress(objective protocols.Objective) (outgoing Object
 	return
 }
 
-// SpawnConsensusChannelIfDirectFundObjective will attempt to create and store a ConsensusChannel derived from the supplied Objective iff it is a directfund.Objective.
```
geoknee, line 346: I think I wrote this, and I think maybe it wasn't a typo -- I may have meant "if and only if". I think your change here is good, regardless.

geoknee (approved):

client/engine/engine.go
NiloCK: I checked the authorship. If it had been Andrew I would have been **sure** that it wasn't a typo. Being you I settled on 50/50.

But in any case I think the "plain Enghlish" is better in places like this.

PR #658, opened 2022-05-09
Do not return error when computing ChannelId

### LINKED ISSUE #625: Should ChannelID() return errors?
Swallowing these errors doesn't seem great, but also the error returns on fp.ChannelID() are more or less impossible except on init for a channel. I don't have any suggestion :/

I think this annoyance has been brought up before.

_Originally posted by @NiloCK in https://github.com/statechannels/go-nitro/pull/620#discussion_r862902239_

geoknee:
Closes #625 

The error is currently being ignored in the majority of cases, anyway. 

By introducing a `FixedPart.Validate()` method, we can continue to catch malformed fixed parts when it matters.

### DIFF
--- a/channel/channel.go
+++ b/channel/channel.go
@@ -242,7 +242,7 @@ func (c *Channel) AddSignedState(ss state.SignedState) bool {
 
 	s := ss.State()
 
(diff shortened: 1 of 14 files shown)


lalexgap (approved):
Yay!!

PR #659, opened 2022-05-09
fix version of golangci-lint

geoknee:
This prevents unannounced / unexpected linting errors from appearing due to newer versions of the linting tool being published. 

Context: I tried upgrading my local linter, but seemed to find quite a lot of false positives. So I think the cutting edge version is not so great. Hence the decision to fix the version for now. 

More context: There are several linters which are "disabled because of go1.18". Therefore we expect quite a lot of significant changes to linting in the short term.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -17,7 +17,7 @@ jobs:
         uses: golangci/golangci-lint-action@v3.1.0
         with:
           # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
-          version: v1.29
+          version: v1.45.2
 
       - name: make bin folder
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/go.yml
```diff
@@ -15,6 +15,9 @@ jobs:
           go-version: "^1.18.0"
       - name: Run golangci-lint
         uses: golangci/golangci-lint-action@v3.1.0
+        with:
+          # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
+          version: v1.29
```
geoknee, line 20: This is wrong. It is both a very old version number, and not related to the version of golangci-lint :-(

NiloCK (approved):
:+1: 

(github.com is behaving badly for me today!)

PR #664, opened 2022-05-09
Increase test coverage

lalexgap:
Increase test coverage by adding and modifying some tests.


channel/channel_test.go
```diff
@@ -382,3 +378,16 @@ func TestSerde(t *testing.T) {
 		t.Fatalf("incorrect json unmarshaling, expected \n%+v got \n%+v", someChannel, got)
 	}
 }
+
+// signState generates a signature on the signed state with the supplied key, and adds that signature.
+// If an error occurs the function panics
+func signState(ss *state.SignedState, secretKey *[]byte) {
```
geoknee, line 384: Can you reuse `testhelpers.SignState()`?

geoknee (approved):

channel/channel_test.go
geoknee: [fe5f50b](https://github.com/statechannels/go-nitro/pull/664/commits/fe5f50b0a0ff3b4c51ecc07554be2803ec976657)

PR #665, opened 2022-05-09
rename OwnsChannel to PrincipalChannel

### LINKED ISSUE #648: OwnsChannel is poorly named
> *Isn't `.OwnsChannel` poorly named?
> 
> https://github.com/statechannels/go-nitro/blob/b426f009f463c5e62333a9b5772e3ab53248d340/protocols/interfaces.go#L81-L82
> 
> The source of truth about whether the objective actually owns the channel is the store itself. Should we rename it to `PrincipalChannel` or something?

lalexgap:
Fixes #648 

Renames `OwnsChannel` to` PrincipalChannel`. If anyone can think of a better name than `PrincipalChannel` let me know!

### DIFF
--- a/protocols/directdefund/directdefund.go
+++ b/protocols/directdefund/directdefund.go
@@ -159,8 +159,8 @@ func (o Objective) Reject() protocols.Objective {
 	return &updated
 }
 
-// OwnsChannel returns the channel that the objective is funding.
-func (ddo Objective) OwnsChannel() types.Destination {
+// PrincipalChannel returns the channel that the objective is funding.
+func (ddo Objective) PrincipalChannel() types.Destination {
 	return ddo.C.Id
 }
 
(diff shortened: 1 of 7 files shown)


protocols/directdefund/directdefund.go
```diff
@@ -159,8 +159,8 @@ func (o Objective) Reject() protocols.Objective {
 	return &updated
 }
 
-// OwnsChannel returns the channel that the objective is funding.
-func (ddo Objective) OwnsChannel() types.Destination {
+// PrincipalChannel returns the channel that the objective is funding.
+func (ddo Objective) PrincipalChannel() types.Destination {
```
geoknee, line 163: We probably want all of these godoc lines to match, and to do a better job at explaining what this is. I think we probably want an ADR, too.

lalexgap:
Closing this for now as we think further on channel ownership

PR #666, opened 2022-05-10
remove _requireVariablePartIsLast

geoknee:
This PR also steers away from the confusing terminology of "latest" (meaning greatest turn number) and "last" (meaning greatest index in the call data array).


geoknee:
See #764

geoknee:
Closing in favour of #764

PR #667, opened 2022-05-10
application returns index, not state

geoknee:
This is a gas optimisation only -- not sure how much it might save, though.


geoknee:
See #764

PR #668, opened 2022-05-10
Check for existing ledger channels when constructing direct fund objective

### LINKED ISSUE #627: Virtual funding fails if there are multiple ledger channels between participants
Currently the store just returns the [first ledger channel ](https://github.com/statechannels/go-nitro/blob/9b3db3d9e199f7a8d73c777713ea5a74cb79bb7a/client/engine/store/memstore.go#L195) with a counter party it finds.

This means if there are multiple ledger channels between participants each participant could end up choosing different ledger channels for an objective.

Perhaps for now we should just limit the amount of ledger channels with a counterparty to 1?

lalexgap:
Fixes #627

Adds a check when creating a direct funding objective that no other ledger channels exist with the counterparty.

We check both for `channel.Channel` for any "in-flight" direct funding objectives as well as `consenus_channel.ConsensusChannel` for any completed direct funding objectives.

This prevents us from creating multiple ledger channels with a counterparty and not knowing which channel to choose when virtual funding.

### DIFF
--- a/internal/testdata/channels.go
+++ b/internal/testdata/channels.go
@@ -24,16 +24,8 @@ var Channels channelCollection = channelCollection{
 
 func mockConsensusChannel(counterparty types.Address) (ledger *consensus_channel.ConsensusChannel, ok bool) {
 	ts := testState.Clone()
-	request := directfund.ObjectiveRequest{
-		MyAddress:         ts.Participants[0],
-		CounterParty:      ts.Participants[1],
-		AppData:           ts.AppData,
-		AppDefinition:     ts.AppDefinition,
-		ChallengeDuration: ts.ChallengeDuration,
-		Nonce:             ts.ChannelNonce.Int64(),
-		Outcome:           ts.Outcome,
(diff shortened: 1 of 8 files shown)


internal/testdata/channels.go
```diff
@@ -100,45 +92,6 @@ func (l LedgerNetwork) GetLedgerLookup(seeker types.Address) virtualfund.GetTwoP
 	}
 }
 
-// createLedgerNetwork returns active, funded consensus_channels connecting the supplied
-// actors according the the supplied edge list.
-//
-// Edges specify actors via their indices in the actors slice,
-// and each edge is ordered [leader, follower].
-func createLedgerNetwork(actors []testactors.Actor, edges [][2]int) LedgerNetwork {
```
geoknee, line 108: This change seems unrelated: are you removing because it is currently unused?

If so, just looping in @NiloCK who I think put some effort into this.

geoknee (approved):

PR #673, opened 2022-05-11
update readme with completion status

lalexgap:


readme.md
```diff
@@ -17,15 +17,15 @@ The following roadmap gives an idea of the various packages that compose the `go
...
 │       ├── outcome ✅         # define how funds are dispersed when a channel closes
-├── client 🚧                  # exposes an API to the consuming application
-│   └── engine 🚧              # coordinate the client components, runs the protocols
-│       ├── chainservice 🚧    # watch the chain and submit transactions
-│       ├── messageservice 🚧  # send and recieves messages from peers
-│       └── store 🚧           # store keys, state updates and other critical data
-├── client_test 🚧             # integration tests involving multiple clients
+├── client ✅                  # exposes an API to the consuming application
+│   └── engine ✅              # coordinate the client components, runs the protocols
+│       ├── chainservice ✅   # watch the chain and submit transactions
```
lalexgap, line 25: I wasn't sure if these should be marked as complete or not since we just have a mock-chain and a memory store.

readme.md
```diff
@@ -45,7 +45,8 @@ Milestones that we hope to hit in the coming weeks:
 | --------------------------------------|--- | --------|
 | Directly Fund an (Alice, Bob) Channel | mock | ✅ |
 | Virtually Fund an (Alice, Bob) Channel through a single Intermediary | mock | ✅ |
-| Virtually Fund an (Alice, Bob) and an (Alice, Brian) Channel through a single Intermediary | mock | 🟡|
+| Virtually Fund an (Alice, Bob) and an (Alice, Brian) Channel through a single Intermediary | mock | ✅ |
+| Virtually Defund an (Alice, Bob) Channel through a single Intermediary | mock | ✅ |
 | Stream micropayments fron Alice to Bob | mock | |
```
lalexgap, line 50: Should we drop this from the readme for now?

readme.md
geoknee: I think we should leave them as "in progress" until we have the "proper" services built.

readme.md
geoknee: I guess we aren't 100% sure if we `go-nitro` clients will actually be responsible for payments or not. So we can drop this until we decide on that.

PR #675, opened 2022-05-12
Some improvements to "large scale instrumented test"

geoknee:
* Uses a global "pretty print dictionary", managed by the test script, to prettify the output logs (which are reverted to contain un-clipped addresses and destinations). 
* Removes the need to run GoVector CLI as an external dependency (it was a very lightweight utility which I have rewritten)
* Enables us greater control of the order of actors in the resulting shiviz visualisation
* Focuses vector clock logging to _after_ the ledger channels have been created.
* unskips the test so everyone will now run it

### DIFF
--- a/client/engine/messageservice/vectorclock_testmessageservice.go
+++ b/client/engine/messageservice/vectorclock_testmessageservice.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/DistributedClocks/GoVector/govec"
+	"github.com/statechannels/go-nitro/channel/consensus_channel"
 	"github.com/statechannels/go-nitro/protocols"
 	"github.com/statechannels/go-nitro/types"
 )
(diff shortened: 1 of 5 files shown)


client/engine/messageservice/vectorclock_testmessageservice.go
```diff
@@ -76,23 +75,19 @@ func (t VectorClockTestMessageService) dispatchMessage(message protocols.Message
...
 	for _, entry := range msg.SignedStates() {
 		summary += `send `
-		if len(entry.Payload.State().Participants) == 3 {
-			summary += `V`
-		} else {
-			summary += `L`
-		}
 		_, turnNum := entry.Payload.SortInfo()
-		summary += fmt.Sprint(entry.Payload.ChannelId())[1:8]
-		summary += fmt.Sprint(turnNum)
```
geoknee, line 95: This line is unintended.

client/engine/messageservice/vectorclock_testmessageservice.go
```diff
@@ -76,23 +75,19 @@ func (t VectorClockTestMessageService) dispatchMessage(message protocols.Message
 // summarizeMessageSend returns a string which tersely summarizes the supplied message.
 // It may be used to make logs more readable.
 func summarizeMessageSend(msg protocols.Message) string {
-	summary := ""
+	str, _ := msg.Serialize()
+	size := len([]byte(str))
+	summary := fmt.Sprint(size) + "B:"
 	for _, entry := range msg.SignedProposals() {
```
lalexgap, line 81: Should we update this to handle `Removal`s as well?

geoknee: Good call [25ae9b7](https://github.com/statechannels/go-nitro/pull/675/commits/25ae9b79cebf753ad8aa916aaa62352b43db6662)

PR #678, opened 2022-05-12
Remove MyAddress from ObjectiveRequest

### LINKED ISSUE #677: Remove MyAddress from API (if possible)
This just seems a bit nuts:

```go
	// Construct Nitro Client
	retrieverNitroClient := client.New(messageservice, chainservice, store, logDestination)

	// Prepare a request to connect with (e.g.) a Retrieval Provider
	request := directfund.ObjectiveRequest{
		MyAddress:         *retrieverNitroClient.Address,
		CounterParty:      common.HexToAddress("0xBBB676f9cFF8D242e9eaC39D063848807d3D1D94"),
		Outcome:           outcome.Exit{},  // Soon: Me: 100 FIL, Them: 0
		AppDefinition:     types.Address{}, // Soon: On-chain actor / smart contract for virtual channels
		AppData:           types.Bytes{},
		ChallengeDuration: big.NewInt(100),
		Nonce:             rand.Int63(),
	}

	// Make the appropriate API call
	response := retrieverNitroClient.CreateDirectChannel(request)
```

I need to ask the nitro client my address, so that I can simply pass it back all over again?

lalexgap:
Fixes #677 

Removes `MyAddress` from the `ObjectiveRequest` so the client API doesn't require it to be passed in.

I've also done a slight refactor of the `NewObjective` constructors so they all take in arguments in the same order.

### DIFF
--- a/protocols/virtualfund/virtualfund.go
+++ b/protocols/virtualfund/virtualfund.go
@@ -129,7 +129,7 @@ type Objective struct {
 }
 
 // NewObjective creates a new virtual funding objective from a given request.
-func NewObjective(myAddress types.Address, request ObjectiveRequest, getTwoPartyConsensusLedger GetTwoPartyConsensusLedgerFunction) (Objective, error) {
+func NewObjective(request ObjectiveRequest, preApprove bool, myAddress types.Address, getTwoPartyConsensusLedger GetTwoPartyConsensusLedgerFunction) (Objective, error) {
 	rightCC, ok := getTwoPartyConsensusLedger(request.Intermediary)
 
 	if !ok {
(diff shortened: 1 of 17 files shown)


protocols/virtualfund/virtualfund.go
```diff
@@ -678,9 +677,9 @@ type ObjectiveResponse struct {
 }
 
 // Response computes and returns the appropriate response from the request.
-func (r ObjectiveRequest) Response() ObjectiveResponse {
+func (r ObjectiveRequest) Response(myAddress types.Address) ObjectiveResponse {
```
geoknee, line 680: It seems a little awkward to have to pass this in, now -- I suppose that is a consequence of wanting to compute the channelId synchronously (i.e. as soon as the request is submitted, and before any store access or other async things happen)? Still, seems less bad than having to pass it across the API.

geoknee (approved):

PR #681, opened 2022-05-13
Create basic TCP message service

lalexgap:
This PR implements a basic message service that works over TCP. The main motivation is allowing us to set up test scenarios that can have client's running in different processes/machines that can communicate with each-other. 

Currently this message service assumes that the address of all other messages services will be known when creating the service.

### DIFF
--- a/client/engine/messageservice/simple-tcp/simpletcp-messageservice.go
+++ b/client/engine/messageservice/simple-tcp/simpletcp-messageservice.go
@@ -1,0 +1,157 @@
+package simpletcp
+
+import (
+	"bufio"
+	"fmt"
+	"net"
+
+	"github.com/statechannels/go-nitro/protocols"
+	"github.com/statechannels/go-nitro/types"
+)
+
+const (
+	CONN_TYPE = "tcp"
+	DELIMETER = '\n'
+)
+
+// SimpleTCPMessageService is a rudimentary message service that uses TCP to send and receive messages
+type SimpleTCPMessageService struct {
+	in  chan protocols.Message // for receiving messages from engine
+	out chan protocols.Message // for sending message to engine
+
+	peers map[types.Address]string
+
+	listener net.Listener // The listener for incoming connections on our port
+
+	quit chan struct{} // quit is used to signal the goroutine to stop
+
+}
+
+// NewTestMessageService returns a running SimpleTcpMessageService listening on the given url
+func NewSimpleTCPMessageService(myUrl string, peers map[types.Address]string) *SimpleTCPMessageService {
+
+	l, err := net.Listen(CONN_TYPE, myUrl)
+	if err != nil {
+		panic(err)
+	}
+	h := &SimpleTCPMessageService{
+		in:    make(chan protocols.Message, 5),
+		out:   make(chan protocols.Message, 5),
+		peers: peers,
+
+		listener: l,
+		quit:     make(chan struct{}),
+	}
+
+	go h.listenForIncoming()
+	go h.listenForOutgoing()
+
+	return h
+
+}
+
+// listenForOutgoing listens for any messages that should be sent out to peers and sends them
+func (s *SimpleTCPMessageService) listenForOutgoing() {
+	for {
+		select {
+		case <-s.quit:
+			return
+
+		case m := <-s.in:
+			{
+				peer, ok := s.peers[m.To]
+
+				if !ok {
+					panic(fmt.Errorf("no peer port for %s", m.To))
+				}
+
+				raw, err := m.Serialize()
+
+				// Append the delimiter to the message to indicate the end of the message
+				raw = fmt.Sprintf("%s%c", raw, DELIMETER)
+
+				if err != nil {
+					panic(err)
+				}
+
+				conn, err := net.Dial(CONN_TYPE, peer)
+				if err != nil {
+					select {
+					case <-s.quit: // If we are quitting we can ignore the error
+						return
+					default:
+						panic(err)
+					}
+				}
+				_, err = conn.Write([]byte(raw))
+				if err != nil {
+					select {
+					case <-s.quit: // If we are quitting we can ignore the error
+						return
+					default:
+						panic(err)
+					}
+				}
+				conn.Close()
+
+			}
+
+		}
+	}
+
+}
+
+// listenForIncoming listens for any incoming messages from other peers
+func (s *SimpleTCPMessageService) listenForIncoming() {
+	for {
+		conn, err := s.listener.Accept()
+
+		if err != nil {
+			select {
+			case <-s.quit:
+				return
+			default:
+				panic(err)
+			}
+		}
+
+		raw, err := bufio.NewReader(conn).ReadString(DELIMETER)
+
+		if err != nil {
+			select {
+			case <-s.quit: // If we are quitting we can ignore the error
+				return
+			default:
+				panic(err)
+			}
+
+		}
+		m, err := protocols.DeserializeMessage(raw)
+		if err != nil {
+			panic(err)
+		}
+		s.out <- m
+
+		conn.Close()
+
+	}
+
+}
+
+func (s *SimpleTCPMessageService) Out() <-chan protocols.Message {
+	return s.out
+}
+
+func (s *SimpleTCPMessageService) In() chan<- protocols.Message {
+	return s.in
+}
+
+// Close closes the SimpleTCPMessageService
+func (s *SimpleTCPMessageService) Close() {
+	close(s.quit)
+	s.listener.Close()
+
+	close(s.in)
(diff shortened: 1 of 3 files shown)


client/engine/messageservice/simple-tcp/simpletcp-messageservice.go
```diff
@@ -0,0 +1,90 @@
...
+		peer, ok := s.peers[m.To]
+
+		if !ok {
+			panic(fmt.Errorf("no peer port for %s", m.To))
+		}
+		raw, err := m.Serialize()
+		if err != nil {
+			panic(err)
+		}
+		conn, err := net.Dial(connectionType, peer)
```
geoknee, line 52: Here you've got a new connection per outgoing message. 

It is interesting to think about having a persistent TCP connection between peers, that could be reused to send messages. (I'm not suggesting we do this now).

In that model, I believe messages would be delivered to peers in the same order they were sent. This is relevant to this RFC https://www.notion.so/statechannels/RFC-14-Lean-on-message-delivery-guarantees-to-simplify-engine-protocols-81e11b814cfa46babae957dcdfdceae5.

lalexgap: Interesting! That would guarantee that messages from a specific peer arrive in order. We would still have to deal with messages from different peers arriving in different orders (IE: We could get 3 messages from Bob before we get the first from Irene) but that might be fine.

It would introduce a little book keeping as a message service would have to negotiate a port for each peer instead of using a fixed port, but that probably shouldn't be too hard either.

lalexgap:
Moving this back to draft since I noticed that we're constantly reading empty messages

lalexgap:
> Moving this back to draft since I noticed that we're constantly reading empty messages

This seemed to be caused by closing the in/out channels of the message service. This is resolved in https://github.com/statechannels/go-nitro/pull/681/commits/2f0d9fd38d80a11a371d58017d67c9ab0a5f8f8b

client_test/directfund_simple_tcp_test.go
```diff
@@ -0,0 +1,46 @@
...
+	simpletcp "github.com/statechannels/go-nitro/client/engine/messageservice/simple-tcp"
+	"github.com/statechannels/go-nitro/client/engine/store"
+	"github.com/statechannels/go-nitro/crypto"
+	"github.com/statechannels/go-nitro/types"
+)
+
+// setupClientWithSimpleTCP is a helper function that contructs a client and returns the new client and its store.
+func setupClientWithSimpleTCP(pk []byte, chain chainservice.MockChain, peers map[types.Address]string, logDestination io.Writer, meanMessageDelay time.Duration) (client.Client, *simpletcp.SimpleTCPMessageService) {
+	myAddress := crypto.GetAddressFromSecretKeyBytes(pk)
+	chain.Subscribe(myAddress)
```
geoknee, line 20: I think you don't need these calls since #687

client_test/directfund_simple_tcp_test.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/statechannels/go-nitro/client"
+	"github.com/statechannels/go-nitro/client/engine/chainservice"
+	simpletcp "github.com/statechannels/go-nitro/client/engine/messageservice/simple-tcp"
+	"github.com/statechannels/go-nitro/client/engine/store"
+	"github.com/statechannels/go-nitro/crypto"
+	"github.com/statechannels/go-nitro/types"
+)
+
+// setupClientWithSimpleTCP is a helper function that contructs a client and returns the new client and its store.
+func setupClientWithSimpleTCP(pk []byte, chain chainservice.MockChain, peers map[types.Address]string, logDestination io.Writer, meanMessageDelay time.Duration) (client.Client, *simpletcp.SimpleTCPMessageService) {
```
geoknee, line 18: How can we organise our integration tests, such that we aren't repeating ourselves? We probably don't want to write a new test for every (protocol, message service, store, chain service) combination. At least not manually.

Perhaps if we were careful enough with the test setup /teardown interfaces, we could actually programmatically try out all of the combinations? (I realise the setup and teardown are currently a little different between the two message services).

Another option would be just to drop this TCP service in over the existing `chan` based service in all of our integration tests. A did some quick experiments and it seems to be just as fast. I wonder if there is much value keeping the `chan` based service around, since the TCP one is probably more realistic. What are your thoughts?

client/engine/messageservice/simple-tcp/simpletcp-messageservice.go
```diff
@@ -0,0 +1,154 @@
...
+				raw = fmt.Sprintf("%s%c", raw, DELIMETER)
+
+				if err != nil {
+					panic(err)
+				}
+
+				conn, err := net.Dial(CONN_TYPE, peer)
+				if err != nil {
+					select {
+					case <-s.quit: // If we are quitting we can ignore the error
```
geoknee, line 80: This pattern (checking first whether a quit signal has been sent, and if it hasn't, checking _again_ if we hit an error condition) is repeated a lot -- is it unavoidable, do you think?

geoknee (approved):
🌋

client/engine/messageservice/simple-tcp/simpletcp-messageservice.go
```diff
@@ -0,0 +1,154 @@
...
+
+const (
+	CONN_TYPE = "tcp"
+	DELIMETER = '\n'
+)
+
+// SimpleTCPMessageService is a rudimentary message service that uses TCP to send and receive messages
+type SimpleTCPMessageService struct {
+	in  chan protocols.Message // for receiving messages from engine
+	out chan protocols.Message // for sending message to engine
```
NiloCK, line 20:
```suggestion
	// for receiving messages from engine
	in  chan protocols.Message
	// for sending message to engine
	out chan protocols.Message 
```

client/engine/messageservice/simple-tcp/simpletcp-messageservice.go
```diff
@@ -0,0 +1,154 @@
...
+
+			}
+
+		}
+	}
+
+}
+
+// listenForIncoming listens for any incoming messages from other peers
+func (s *SimpleTCPMessageService) listenForIncoming() {
```
NiloCK, line 105: These guys could be cleaned up a bit with something like:

```suggestion
				if err != nil {
					s.panic(err)
				}
				conn.Close()

			}

		}
	}

}

// panic swallows an error if quitting, otherwise panics
func (s *SimpleTCPMessageService) panic(err error) {
	select {
	case <-s.quit: // If we are quitting we can ignore the error
		return
	default:
		panic(err)
	}
}

// listenForIncoming listens for any incoming messages from other peers
func (s *SimpleTCPMessageService) listenForIncoming() {
```

NiloCK (approved):
:1st_place_medal:

client/engine/messageservice/simple-tcp/simpletcp-messageservice.go
lalexgap: Wouldn't the `return` statement in `s.panic` just return from the `s.panic` function and not the parent function calling `s.panic`?


I guess we could do something like 
```
if err != nil {
	s.panic(err)
         return
	}
```

in the parent function.

client_test/directfund_simple_tcp_test.go
lalexgap: I think switching to the TCP message service would make sense. I'd be in favour of just deprecating the `chan` based service so we only have to maintain the one message service if we could.  We'd need to add message delay logic for the message delay test and probably want to create a vectorclock version as well?

I think by default tests in a package are [not run in parallel](https://stackoverflow.com/questions/24375966/does-go-test-run-unit-tests-concurrently) so I think we'd avoid messages services fighting over a TCP port, but it is something to keep in mind.

In general I think we could also refactor some of our integration tests into one file. Currently the `virtualfund_test`, `virtualfund_with_message_delay` and `virtualfund_multiparty_test` all have their own test files but they're essentially similar tests with some different parameters passed in.

client_test/directfund_simple_tcp_test.go
lalexgap: https://github.com/statechannels/go-nitro/pull/681/commits/d03a68834a5f5ab76f6c05e52bd3a3834d86cc7a

client/engine/messageservice/simple-tcp/simpletcp-messageservice.go
lalexgap: I've tried to clean this up in https://github.com/statechannels/go-nitro/pull/681/commits/2c6f02021a4d7073037036fbc77f49bebd22cc6f. I think anytime we're dealing with the `listener` we have to deal with errors that might arise from calling `s.listener.Close()`

PR #686, opened 2022-05-16
return a single objective per ledger channel...

### LINKED ISSUE #684: Return a single `Related` objective per ledger channel
Currently,
- `attemptProgressForRelatedObjectives` receives objectiveIDs for all proposals in a given ledger channel's queue.
- it then calls `attemptProgress` on each of them

This is wasteful, (and may invite syncing / channel lock bottlenecks) since only the first item in the proposal queue is ready to be incorporated into the ledger channel's `current` state.

Instead, the `relatedObjectives` function should return only the "leading" objectives whose proposals are ready to be the included in their given ledger channels.

### LINKED ISSUE #685: Rename `related` in context of `attemptProgressForRelatedObjectives`
Currently, objectives have a `Related()` function which returns `Storable` objects, which all happen to be `channel`s of one variety or another.

`engine.findRelatedObjectives(someObjective)` returns other `Objectives`.

This is a pretty confusing overload, and in both cases `Related` is probably not as descriptive as we might like.

NiloCK:
when searching for unblocked objectives.

Closes #684, closes #685. Change-set is small. See issues for motivation.

---

#### Code quality

- [ ] ~I have added tests that prove my fix is effective or that my feature works, if necessary~ Existing tests run

#### Project management

### DIFF
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -169,7 +169,7 @@ func (e *Engine) handleMessage(message protocols.Message) (ObjectiveChangeEvent,
 		}
 		allCompleted.CompletedObjectives = append(allCompleted.CompletedObjectives, progressEvent.CompletedObjectives...)
 
-		relatedObjectiveCompletions, err := e.attemptProgressForRelatedObjectives(&updatedObjective)
+		relatedObjectiveCompletions, err := e.attemptProgressForPendingObjectives(&updatedObjective)
 		if err != nil {
 			return ObjectiveChangeEvent{}, err
 		}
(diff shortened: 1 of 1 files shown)


lalexgap:
> Second thoughts:
> 
> In this PR, a ledger channel with more than one queued proposal might not ever "surface" the 2nd, 3rd, ... Nth queued proposal.

I think we might want to watch out for this.  Let's say we have three virtual channels: V1, V2, V3 all funded by some ledger L. Let's us assume we're Bob:
- We send out our prefund for V1,V2,V3
- We receive prefund for V2 and V3
- We receive proposals for L for V1,V2, and V3
- We receive prefund for for V1

I think in that scenario Bob would never call `attemptProgress` for V3 and would get stuck.

> Previously, all queued proposals were surfaced, and `attemptProgress` called on each in sequential order. It's not obvious to me though that the later objectives (& their ledger channels) will be in the correct state to progress when they are cranked.

 I think we can expect the objectives/ledger channels to **possibly** be in the correct state to crank. If they're not then there's a message we haven't yet received that will trigger us to crank it.

We should be storing and refetching the ledger between each crank, so I think it's quite possible that later objectives will be ready to be cranked.

PR #689, opened 2022-05-17
Return NextProposal as side-effect of ledger funding `Crank`s

### LINKED ISSUE #684: Return a single `Related` objective per ledger channel

### LINKED ISSUE #685: Rename `related` in context of `attemptProgressForRelatedObjectives`

NiloCK:
This is another option re [messages arriving out of order can cause virtual protocols to stall](https://www.notion.so/statechannels/Messages-arriving-out-of-order-can-cause-virtual-protocols-to-stall-fc5fe7a1121f4b289a6f6384c1ed4429). 

Closes #684 - this workflow achieves the goal of returning / processing a single objective at a time
Closes #685, by way of refactor

**Changes**

PR introduces an internal `fromLedger` channel for the engine to execute "flywheel" cranks - cranking virtual funding objective `B` that has become freed up during the execution of another virtual protocol `A` after `A` alters the state of one or more ledger channels.

Specifically, `vfo.Crank` and `vdfo.Crank` now each return a `ProposalToProcess` when they make adjustments to any ledger channel's proposalQueue. Execution of this sideEffect is just passing the proposal into the `fromLedger` channel, where it is handled in the the engine's main `Run` loop in the same manner as messages from the `FromAPI`, `fromChain`, and `fromMsg` channels.

This:
- removes the potentially confusing/surprising chain of objective cranks that can be started by `handleMessage`
- eliminates the [problem](https://github.com/statechannels/go-nitro/pull/686#issuecomment-1127889458) of discovering sequential queued proposals (this one recursese - the `handelProposal` crank will return more Proposals if they are ready)
- makes the `Run` loop the starting point for each `Crank`

**Incidental Discovery**
Implementing this made it clear that only ledger-channel followers need ever re-crank objectives after the ledger channel's proposalQueue is trimmed. This wasn't obvious to me.

**How Has This Been Tested?**

This is a refactor. Existing tests pass.

- [ ] Is it a significant change to architecture or design?
  - borderline 

---

#### Code quality

#### Project management

### DIFF
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -457,50 +457,6 @@ func (e *Engine) constructObjectiveFromMessage(id protocols.ObjectiveId, ss stat
 
 }
 
-// TODO: These were implemented to quickly resolve https://github.com/statechannels/go-nitro/issues/637
-// This may not be the long term approach we use.
-// See https://www.notion.so/statechannels/Messages-arriving-out-of-order-can-cause-virtual-protocols-to-stall-fc5fe7a1121f4b289a6f6384c1ed4429
-
-// attemptProgressForRelatedObjectives attempts to progress any objectives that may be related to the objective that was just cranked
-// An objective is related when it shares a ledger channel with the provided objective.
-// This allows progress to made on other objectives that may be unblocked after processing the updatedObjective.
-func (e *Engine) attemptProgressForRelatedObjectives(updatedObjective *protocols.Objective) (ObjectiveChangeEvent, error) {
-	allCompleted := ObjectiveChangeEvent{}
-	relatedIds, err := e.findRelatedObjectives(*updatedObjective)
-	if err != nil {
-		return ObjectiveChangeEvent{}, err
-	}
-	for _, id := range relatedIds {
-		related, err := e.store.GetObjectiveById(id)
-		if err != nil {
-			return ObjectiveChangeEvent{}, err
-		}
-		progressEvent, err := e.attemptProgress(related)
-		if err != nil {
-			return ObjectiveChangeEvent{}, err
-		}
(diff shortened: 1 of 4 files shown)


protocols/interfaces.go
```diff
@@ -33,6 +33,7 @@ type ChainTransaction struct {
 type SideEffects struct {
 	MessagesToSend       []Message
 	TransactionsToSubmit []ChainTransaction
+	ProposalsToProcess   []consensus_channel.Proposal
```
lalexgap, line 36: Will this slice ever contain more than one element?

NiloCK: If an intermediary is follower in two ledger channels it could contain a proposal from each side.

lalexgap (approved):
This seems like a nice solution to me.

client/engine/engine.go
```diff
@@ -112,6 +113,9 @@ func (e *Engine) Run() {
 
 		case message := <-e.fromMsg:
 			res, err = e.handleMessage(message)
+
+		case id := <-e.fromLedger:
```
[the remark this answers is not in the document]
NiloCK: Good question! On a previous commit it had been a channel of objective IDs.

PR #687, opened 2022-05-17
subscribe to chain during chainservice construction

geoknee:
This is a small optimization to the overall `go-nitro` API. 

However, I am seeing an unexpected test failure locally: 

```
panic: client 0xAAA6628Ec44A8a742987EF3A114dDFE2D4F7aDCE has no connection to client 0xBBB676f9cFF8D242e9eaC39D063848807d3D1D94

goroutine 69 [running]:
github.com/statechannels/go-nitro/client/engine/messageservice.TestMessageService.dispatchMessage({{0xaa, 0xa6, 0x62, 0x8e, 0xc4, 0x4a, 0x8a, 0x74, 0x29, 0x87, ...}, ...}, ...)
	/Users/georgeknee/statechannels/go-nitro/client/engine/messageservice/test-messageservice.go:94 +0x279
created by github.com/statechannels/go-nitro/client/engine/messageservice.TestMessageService.routeToPeers
	/Users/georgeknee/statechannels/go-nitro/client/engine/messageservice/test-messageservice.go:107 +0x54
FAIL	github.com/statechannels/go-nitro/client_test	0.234s
```

It seems to pass on CI, though 🤔 .

PR #690, opened 2022-05-17
Hide `simpleChainService` struct

NiloCK:
Following #687, another minor api simplification.

`NewSimpleChainService` is the consumption point for this file and the intended method to create a simple chain service. Hiding the struct prevents callers from attempting struct literal instantiation.


geoknee (approved):
💪

PR #691, opened 2022-05-17
Expose Transactions a MockChain receives

lalexgap:
This PR allows a `transListener` to be passed in when construcing a `MockChain`. This listener chan can be used to listen for all transactions that a mock chain receives.

This is done to we can easily share transactions between MockChains running on different processes/machines via the [testground sync service](https://github.com/statechannels/go-nitro-test-plans/blob/main/simple-direct/main.go#L163:L163). We use the `transListener` to listen for transactions from one `MockChain` and send them to the `In` chan of another `MockChain` running on a different process/machine.

This is a short term approach to unblock work on a testground integration test. Eventually this will be removed when we're using a proper `geth/ganache` instance that we can share between clients.

### DIFF
--- a/client/engine/chainservice/mockchain.go
+++ b/client/engine/chainservice/mockchain.go
@@ -13,8 +13,9 @@ type MockChain struct {
 	out map[types.Address]chan Event    // out is a mapping with a chan for each connected ChainService, used to send Events to that service
 	in  chan protocols.ChainTransaction // in is the chan used to receive Transactions from multiple ChainServices
 
-	holdings map[types.Destination]types.Funds // holdings tracks funds for each channel
-	blockNum uint64
+	transListener chan protocols.ChainTransaction   // this is used to broadcast transactions that have been received
+	holdings      map[types.Destination]types.Funds // holdings tracks funds for each channel
+	blockNum      uint64
 }
 
 // Out returns the out chan for a particular ChainService, and narrows the type so that external consumers may only receive on it.
@@ -29,11 +30,18 @@ func (mc MockChain) In() chan<- protocols.ChainTransaction {
 
 // NewMockChain returns a new MockChain.
 func NewMockChain() MockChain {
+	return NewMockChainWithTransactionListener(nil)
+}
+
+// NewMockChainWithTransactionListener returns a new MockChain with the supplied transaction listener.
+// The transaction listener will receive all transactions that are sent to the MockChain.
+func NewMockChainWithTransactionListener(transactionListener chan protocols.ChainTransaction) MockChain {
 
 	mc := MockChain{}
 	mc.out = make(map[types.Address]chan Event)
 	mc.in = make(chan protocols.ChainTransaction)
 	mc.holdings = make(map[types.Destination]types.Funds)
+	mc.transListener = transactionListener
 	mc.blockNum = 1
 
 	go mc.Run()
@@ -49,8 +57,20 @@ func (mc *MockChain) Subscribe(a types.Address) {
 // Run starts a listener for transactions on the MockChain's in chan.
 func (mc MockChain) Run() {
 	for tx := range mc.in {
+		mc.sendToTransListener(tx)
 		mc.blockNum++
 		mc.handleTx(tx)
+	}
+}
+
+// sendToTransListener sends the transaction to the transListener if not nil and the chan is not full.
+func (mc *MockChain) sendToTransListener(tx protocols.ChainTransaction) {
+	if mc.transListener != nil {
+		// Send to transListener and ignore if the chan is full
+		select {
+		case mc.transListener <- tx:
+		default:
+		}
 	}
 }
 
--- a/client/engine/messageservice/simple-tcp/simpletcp-messageservice.go
+++ b/client/engine/messageservice/simple-tcp/simpletcp-messageservice.go
@@ -118,8 +118,6 @@ func (s *SimpleTCPMessageService) listenForIncoming() {
 		}
 		s.out <- m
 
-		conn.Close()
-
 	}
 
 }


lalexgap:
Moving this back to draft as we can block if no one is consuming from TransactionReceived

lalexgap:
> Moving this back to draft as we can block if no one is consuming from TransactionReceived

https://github.com/statechannels/go-nitro/pull/691/commits/f764ba7f5921f1197871a124d3b2f6b111da14b0 will just drop the tx if the chan is blocked.

geoknee:
1. I'm wondering whether we can have a single mock chain instance, rather than having one for each go-nitro Client and synchronizing them? Then, the geth/ganache solution (as well as a "real" blockchain) can just drop in for that.
2. Related: Would it make sense to have the mock chain service communicate with chainservices via TCP, rather than `chans`, similar to the new TCPMessagingService ?

lalexgap:
> 1. I'm wondering whether we can have a single mock chain instance, rather than having one for each go-nitro Client and synchronizing them? Then, the geth/ganache solution (as well as a "real" blockchain) can just drop in for that.

This change is a bit of a fudge mostly to unblock the testground. 😛 

I'm slightly hesitant to write a distributed `MockChain` as I think most of that logic will eventually be handled by `ganache` or `geth`. I also don't think it makes the test much on how "realistic" the test is, since the `MockChain` implementation is so simplistic and very different from blockchain behaviour already. (You couldalso  view a `MockChain` instance as a node in the blockchain 🤔 )
 
Having a distributed `MockChain` would make the testground integration test a bit cleaner since we wouldn't have to handle distributing transaction to other chains though.

> 2. Related: Would it make sense to have the mock chain service communicate with chainservices via TCP, rather than chans, similar to the new TCPMessagingService ?

I don't think there'd be a huge benefit to switching to TCP messaging, as eventually that logic would all be replaced by communication using JSON-RPC. By writing our own TCP messaging we'd be writing a chain-service tied to the implementation of our `MockChain`.

I'd personally be in favour of using this kludge for now and spend more time thinking about how we want to change the `MockChain` to be more realistic. Could we start using `geth` in some capacity? Or should we wrap our existing `MockChain` in a JSON-RPC interface?

geoknee:
> This change is a bit of a fudge mostly to unblock the testground. 😛
> 
> I'm slightly hesitant to write a distributed `MockChain` as I think most of that logic will eventually be handled by `ganache` or `geth`. I also don't think it makes the test much on how "realistic" the test is, since the `MockChain` implementation is so simplistic and very different from blockchain behaviour already. (You couldalso view a `MockChain` instance as a node in the blockchain 🤔 )

Fair enough!

> 
> Having a distributed `MockChain` would make the testground integration test a bit cleaner since we wouldn't have to handle distributing transaction to other chains though.
> 
> > 2. Related: Would it make sense to have the mock chain service communicate with chainservices via TCP, rather than chans, similar to the new TCPMessagingService ?
> 
> I don't think there'd be a huge benefit to switching to TCP messaging, as eventually that logic would all be replaced by communication using JSON-RPC. By writing our own TCP messaging we'd be writing a chain-service tied to the implementation of our `MockChain`.
> ...Or should we wrap our existing `MockChain` in a JSON-RPC interface?

I guess what I'm proposing that we have "out of process" messaging between the mock chain and the various chain services. It wouldn't need to be TCP, but it could be. I think we definitely could move towards the JSON-RPC interface, and this would make things more realistic. That is [going to require an underlying connection](https://pkg.go.dev/net/rpc/jsonrpc#NewClient) (which could be / might as well be a TCP connection). 




> 
> I'd personally be in favour of using this kludge for now and spend more time thinking about how we want to change the `MockChain` to be more realistic. Could we start using `geth` in some capacity? 

Yes I think we could start working towards this.

geoknee:
Edit: we may want to use WebSockets instead of TCP, depending on how we plan on having `go-nitro` clients get event information from the chain https://geth.ethereum.org/docs/rpc/pubsub .

lalexgap:
> > This change is a bit of a fudge mostly to unblock the testground. 😛
> > I'm slightly hesitant to write a distributed `MockChain` as I think most of that logic will eventually be handled by `ganache` or `geth`. I also don't think it makes the test much on how "realistic" the test is, since the `MockChain` implementation is so simplistic and very different from blockchain behaviour already. (You couldalso view a `MockChain` instance as a node in the blockchain 🤔 )
> 
> Fair enough!
> 
> > Having a distributed `MockChain` would make the testground integration test a bit cleaner since we wouldn't have to handle distributing transaction to other chains though.
> > > 2. Related: Would it make sense to have the mock chain service communicate with chainservices via TCP, rather than chans, similar to the new TCPMessagingService ?
> > 
> > 
> > I don't think there'd be a huge benefit to switching to TCP messaging, as eventually that logic would all be replaced by communication using JSON-RPC. By writing our own TCP messaging we'd be writing a chain-service tied to the implementation of our `MockChain`.
> > ...Or should we wrap our existing `MockChain` in a JSON-RPC interface?
> 
> I guess what I'm proposing that we have "out of process" messaging between the mock chain and the various chain services. It wouldn't need to be TCP, but it could be. I think we definitely could move towards the JSON-RPC interface, and this would make things more realistic. That is [going to require an underlying connection](https://pkg.go.dev/net/rpc/jsonrpc#NewClient) (which could be / might as well be a TCP connection).
> 
> > I'd personally be in favour of using this kludge for now and spend more time thinking about how we want to change the `MockChain` to be more realistic. Could we start using `geth` in some capacity?
> 
> Yes I think we could start working towards this. As you say, it might be just as well to shoot straight to this solution, rather than building a realistic "mock chain" ourselves.

Could we merge this in for now (I could flag it with a TODO?) just to support the current work in https://github.com/statechannels/go-nitro-test-plan ?

geoknee (approved):

PR #693, opened 2022-05-18
update architecture diagram

geoknee:
Following #689


NiloCK (approved):
Nice catch. This had occurred to me at a point but then escaped me.

PR #702, opened 2022-05-19
Handle TCP connections in a go-routine

lalexgap:
By handling a the TCP connection in a go routine we avoid blocking new connections until we're done with the current one.

Based on the integration test in testground this prevents the message services failing with a `connection refused` error.


lalexgap:
Closing this as it just introduces another problem.

Issue #700, opened 2022-05-19
Test flicker: TestVirtualDefundIntegrationWithMessageDelay


geoknee:
See here on a PR that only changed documentation: https://github.com/statechannels/go-nitro/runs/6496315804?check_suite_focus=true
```
-test.shuffle 1652904478928870717
--- FAIL: TestVirtualDefundIntegrationWithMessageDelay (3.04s)
    helpers_test.go:61: Objective ids [VirtualDefund-0x135621343d7646514e33cad0c62e7ac66e2719f527af9[18](https://github.com/statechannels/go-nitro/runs/6496315804?check_suite_focus=true#step:6:19)6ba0[19](https://github.com/statechannels/go-nitro/runs/6496315804?check_suite_focus=true#step:6:20)b5b43b9d967] failed to complete on client 0xBBB676f9cFF8D242e9eaC39D063848807d3D1D94 within 2s
--- FAIL: TestVirtualDefundIntegrationWithMessageDelay (2.94s)
    helpers_test.go:61: Objective ids [VirtualDefund-0x0f0a085e6c272c1b22dcaa244db7285d988e7862412f930b4444d4ae9416c43d] failed to complete on client 0xBBB676f9cFF8D242e9eaC39D063848807d3D1D94 within 2s
FAIL
FAIL	github.com/statechannels/go-nitro/client_test	10.[23](https://github.com/statechannels/go-nitro/runs/6496315804?check_suite_focus=true#step:6:24)4s
```

geoknee:
The timeouts in this test could do with an overhaul:

Firstly, we define a special "long" timeout for this test due to the introduction of delays:

https://github.com/statechannels/go-nitro/blob/26a4ae3f30b3d79135b0fcb98d1ca4c1efd4e017/client_test/virtualfund_with_message_delays_test.go#L23-L24

 However, this special timeout is actually shorter than the default (which we recently lengthened):

https://github.com/statechannels/go-nitro/blob/d883262259a0399bab2f87957adc0492902d6498/client_test/helpers_test.go#L19-L20

Furthermore, we don't use the "special" timeout throughout the test:

https://github.com/statechannels/go-nitro/blob/26a4ae3f30b3d79135b0fcb98d1ca4c1efd4e017/client_test/virtualfund_with_message_delays_test.go#L37-L38

(these calls end up using a default timeout). It seems to be these lines which are actually causing the problem (despite the longer timeout, which is only longer by accident at the moment).

[...]

PR #705, opened 2022-05-20
Increase the default timeout for objectives to complete

geoknee:
Attempt to address #700.

I think we can justify this "fix" by adopting a strategy where correctness and performance are tested separately. 

We have a "benchmark" test (although since #688 the output of the test is not save anywhere), we should file an issue for that) for testing performance. And it is something we plan to dig much deeper into soon. 

It seems clear that a few seconds is sometimes not enough to let a test run its course. It may well indicate a bottleneck, but it _probably_ doesn't indicate a bug.

### DIFF
--- a/client_test/helpers_test.go
+++ b/client_test/helpers_test.go
@@ -16,7 +16,7 @@ import (
 	"github.com/statechannels/go-nitro/protocols"
 )
 
-const defaultTimeout = time.Second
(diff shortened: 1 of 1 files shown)


lalexgap (approved):
We could also consider reducing the amount of channels the test are trying to create. That way the CI would still completely pretty quickly

PR #710, opened 2022-05-25
remove "pause" from eth_chainservice_test.go

geoknee:
This is achieved by introducing a method on the eth_chainservice which returns a chan where a done signal will be sent. We can block the test on pulling from this channel, instead of waiting for a fixed period of time.

### DIFF
--- a/client/engine/chainservice/eth_chainservice.go
+++ b/client/engine/chainservice/eth_chainservice.go
@@ -48,6 +48,7 @@ func NewEthChainService(na *NitroAdjudicator.NitroAdjudicator, naAddress common.
 	return ecs
 }
 
+// SendTransaction sends the transaction and blocks until it has been submitted.
 func (cc *EthChainService) SendTransaction(tx protocols.ChainTransaction) {
 	c := make(chan struct{})
 	cc.in <- TransactionWithStatusChan{ChainTransaction: tx, done: c}


client/engine/chainservice/eth_chainservice.go
```diff
@@ -21,16 +21,25 @@ type eventSource interface {
 	SubscribeFilterLogs(ctx context.Context, query ethereum.FilterQuery, ch chan<- ethTypes.Log) (ethereum.Subscription, error)
 }
 
+type TransactionWithStatusChan struct {
+	protocols.ChainTransaction
+	done chan struct{}
```
kerzhner, line 26: Should the `done` channel emit a boolean? Or is the thought that the `done` channel eventually emits details about the submitted transaction?

client/engine/chainservice/eth_chainservice.go
```diff
@@ -39,7 +48,13 @@ func NewEthChainService(na *NitroAdjudicator.NitroAdjudicator, naAddress common.
 	return ecs
 }
 
-func (cc EthChainService) In() chan<- protocols.ChainTransaction {
+func (cc *EthChainService) Input(tx protocols.ChainTransaction) chan struct{} {
```
kerzhner, line 51: Do you have a vision for the `In` vs `Input` APIs?

client/engine/chainservice/eth_chainservice.go
geoknee: I'm not sure. What do you think would make the most sense?

client/engine/chainservice/eth_chainservice.go
geoknee: Not really, very open to suggestions. The main thing was to not expose the input `chan` directly. By wrapping in a function call, that function can `make` the `done` chan and wrap it around the `tx` (saving the consumer from having to do that).

client/engine/chainservice/eth_chainservice.go
kerzhner: Maybe we should use a boolean for now as that is the simplest primitive for the current needs. We can change the interface later as required.

client/engine/chainservice/eth_chainservice.go
kerzhner: Thanks for the explanation! Do you have any background on the design decision for the chain service to accept inputs via a channel (as opposed to a function call)? Specifically, I am thinking of [this API](https://github.com/statechannels/go-nitro/blob/42e25a802693f8daea10f5b3b5e5638e6d721082/client/engine/chainservice/simplechainservice.go#L38).

client/engine/chainservice/eth_chainservice.go
geoknee: OK. I think the actual simplest thing is the empty struct that is there now ⬆️ . This way, the chain service is signalling that it tried to send the transaction, and nothing more (no success / failure information or transaction hash or anything).

kerzhner (approved):

client/engine/chainservice/eth_chainservice.go
geoknee: That's a good question. That decision was made very early on, and there may not be a strong justification for it. I think probably the original idea was loosely based on transaction submission being an async process.

But now we come to revisit the design, and it seems that we would _like_ the consumer to _block_ until the transaction is submitted (e.g. `NitroAdjudicator.Deposit()` is called). So this could be an argument for not using channels at all.

geoknee: Or at least we would like the consumer to be able to block if it wants to. The current system on `main` is really "fire and forget".

PR #716, opened 2022-05-25
libp2p message service

lalexgap:
This is a proof of concept of a mesage service using libp2p. I'm currently running into lots of errors with the [Simple TCP Message service](https://github.com/statechannels/go-nitro/issues/717) so I wanted to try a different protocol.


lalexgap:
Closing this for now. Our testground test has an[ equivalent version](https://github.com/statechannels/go-nitro-test-plan/blob/main/message-service.go) of this message service.

PR #722, opened 2022-05-27
remove `client_test/virtualfund_benchmark_test.go`

### LINKED ISSUE #708: Save output of `TestVirtualFundBenchmark` to CI
Or else delete the test if we plan on measuring performance elsewhere.

geoknee:
This test is designed to compute a benchmark and print it to the screen. However, unless tests are run on CI with the -verbose flag (they currently aren't), we won't see the benchmark value printed.

Given the choice of getting that value persisted on CI, versus deleting the test in favour of benchmarking via our testground repo, it is much easier to do the former!

Closes #708


NiloCK (approved):
:+1:

Issue #704, opened 2022-05-19
The message service should buffer more messages


lalexgap:
Right now a message service is using a relatively[small buffered channel](https://github.com/statechannels/go-nitro/blob/220ef0ef8418380c74aa9fc12530dc85f2332621/client/engine/messageservice/simple-tcp/simpletcp-messageservice.go#L38) for in and out. 

With such small buffers it's easy for them to get filled up which means the message service or engine can get blocked on writing to these channels. I think we can even run into a deadlock situation.

Using a larger buffered channel for `in` and `out` helps address this but it seems like a code smell to just increase it. Should the message service be doing it's own buffering somehow?

geoknee:
An alternative is to fix #697 and then this problem goes away.

lalexgap:
I think is resolved by #723

PR #723, opened 2022-05-27
Change engine -> {msg,chain} service comms to be synchronous 

### LINKED ISSUE #720: Audit use of channels in the engine
The chain service and the messaging service currently communicate _in both directions_ with `go chans`. 

It might make sense to use synchronous function calls instead when the engine needs to send information out to either service. 

We could keep channels for the other direction, which makes more sense.

geoknee:
Closes #720 . 

Possibly relevant to #704 , #700 .

Pros:
- easier to debug this code because fewer go routines
- easier to reason about control flow of the engine
- lower probability to have a race condition?

Cons
- _might_ have an adverse effect on performance (anecdotal evidence)
- may increase probability for a dead lock?

### DIFF
--- a/client/engine/chainservice/chainservice.go
+++ b/client/engine/chainservice/chainservice.go
@@ -46,6 +46,6 @@ type ChainEventHandler interface {
 type ChainService interface {
 	// Out returns a chan for receiving events from the chain service
 	Out() <-chan Event
-	// Send is for sending transactions with the message service
+	// Send is for sending transactions with the chain service
 	Send(protocols.ChainTransaction)
 }
--- a/client/engine/chainservice/chainservice.go
+++ b/client/engine/chainservice/chainservice.go
@@ -44,6 +44,8 @@ type ChainEventHandler interface {
 }
 
 type ChainService interface {
+	// Out returns a chan for receiving events from the chain service
 	Out() <-chan Event
-	In() chan<- protocols.ChainTransaction
+	// Send is for sending transactions with the message service
+	Send(protocols.ChainTransaction)
 }
--- a/client/engine/chainservice/simplechainservice.go
+++ b/client/engine/chainservice/simplechainservice.go
@@ -33,7 +33,7 @@ func (mcs simpleChainService) Out() <-chan Event {
 
 // Send pipes transactions to the MockChain
 func (mcs simpleChainService) Send(tx protocols.ChainTransaction) {
-	mcs.chain.In() <- tx // TODO block until this has been successful / convert chain.In to a sync call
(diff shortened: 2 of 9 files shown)


client/engine/engine.go
```diff
@@ -325,11 +324,11 @@ func (e *Engine) handleAPIEvent(apiEvent APIEvent) (ObjectiveChangeEvent, error)
 func (e *Engine) executeSideEffects(sideEffects protocols.SideEffects) {
 	for _, message := range sideEffects.MessagesToSend {
 		e.logger.Printf("Sending message %+v", protocols.SummarizeMessage(message))
-		e.toMsg <- message
+		e.msg.Send(message)
 	}
 	for _, tx := range sideEffects.TransactionsToSubmit {
 		e.logger.Printf("Sending chain transaction for channel %s", tx.ChannelId)
-		e.toChain <- tx
+		e.chain.Send(tx)
```
geoknee, line 331: This is the nub of the change.

client/engine/chainservice/chainservice.go
```diff
@@ -44,6 +44,8 @@ type ChainEventHandler interface {
 }
 
 type ChainService interface {
+	// Out returns a chan for receiving events from the chain service
 	Out() <-chan Event
-	In() chan<- protocols.ChainTransaction
+	// Send is for sending transactions with the message service
```
NiloCK, line 49:
```suggestion
	// Send is for sending transactions with the chain service
```

client/engine/chainservice/simplechainservice.go
```diff
@@ -34,16 +31,9 @@ func (mcs simpleChainService) Out() <-chan Event {
...
-}
-
-// forwardTransactions pipes transactions to the MockChain
-func (mcs simpleChainService) forwardTransactions() {
-	for tx := range mcs.in {
-		mcs.chain.In() <- tx
-	}
+// Send pipes transactions to the MockChain
+func (mcs simpleChainService) Send(tx protocols.ChainTransaction) {
+	mcs.chain.In() <- tx // TODO block until this has been successful / convert chain.In to a sync call
```
NiloCK, line 36: I believe this does block until successful. 

https://go.dev/tour/concurrency/2:
> By default, sends and receives block until the other side is ready.

ie, this code does not advance until `In()` is ready to consume `tx`.

```suggestion
	mcs.chain.In() <- tx // TODO convert chain.In to a sync call
```

NiloCK (approved):
This also LGTM.

I am seeing a promising reduction in test failures locally with this branch. :crossed_fingers: 

(It is disturbingly easy to autocomplete a middle finger when attempting those crossed fingers.)

PR #727, opened 2022-05-27
Fix DirectFund and DirectDefund objective serialization

kerzhner:


protocols/directdefund/directdefund.go
```diff
@@ -310,6 +310,7 @@ func (o *Objective) clone() Objective {
 	cClone := o.C.Clone()
 	clone.C = cClone
 	clone.finalTurnNum = o.finalTurnNum
+	clone.transactionSubmitted = o.transactionSubmitted
```
geoknee, line 313: It would be great to have some safety net to prevent this mistake in future. 

We should either put one in place, or record our best knowledge about why it is difficult / impossible.

geoknee (approved):

PR #731, opened 2022-05-27
Use a `sync.Map` in the `MockChain` and make all methods pass by reference

### LINKED ISSUE #730: `go test ./client_test -run TestDirectFund -race` FAILS

geoknee:
Closes #730

### DIFF
--- a/client/engine/chainservice/mockchain.go
+++ b/client/engine/chainservice/mockchain.go
@@ -1,6 +1,7 @@
 package chainservice
 
 import (
+	"github.com/statechannels/go-nitro/client/engine/store/safesync"
 	"github.com/statechannels/go-nitro/protocols"
 	"github.com/statechannels/go-nitro/types"
 )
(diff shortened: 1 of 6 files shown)


client/engine/chainservice/mockchain.go
```diff
@@ -102,10 +107,10 @@ func (mc MockChain) handleTx(tx protocols.ChainTransaction) {
 		panic("unexpected chain transaction")
 	}
 
-	for _, out := range mc.out {
-		attemptSend(out, event)
-	}
-
+	mc.out.Range(func(key string, value chan Event) bool {
```
NiloCK, line 110: nit: blank identifier makes the non-use explicit.

```suggestion
	mc.out.Range(func(_ string, value chan Event) bool {
```

NiloCK (approved):
LGTM.

Sadly, I still see the flicker locally with this branch, but these sorts of issues are definite suspects.

PR #732, opened 2022-05-27
Revamp ChainService interface and remove SimpleChainService

kerzhner:
This PR revamps the Chain Service interface. The most consequential change is that the Chain Service now accepts transactions via a `SendTransaction` method instead of an input channel.

While changing the interface, I decided to remove the `SimpleChainService`. `SimpleChainService` is a thin wrapper around `MockChain` that added glue code. In my opinion, using the `MockChain` directly made it easier to reason about how the test code uses the chain service. I am open to other views though.


client/engine/chainservice/mockchain.go
```diff
@@ -16,7 +16,7 @@ type MockChain struct {
 
 	transListener chan protocols.ChainTransaction   // this is used to broadcast transactions that have been received
 	holdings      map[types.Destination]types.Funds // holdings tracks funds for each channel
-	blockNum      *uint64
+	blockNum      *uint64                           // MockChain is often passed around by value. The pointer allows for shared state.
```
geoknee: I've looked into passing the mock chain by reference here #731 .

kerzhner: Gotcha. My goal here is to make it harder to create unexpected behavior if MockChain is passed by value. When MockChain is passed by value and the MockChain struct has primitive (as opposed to pointer) fields, it's possible to run into deadlock since MockChain values will not share the block number.

kerzhner:
Closing in favor of https://github.com/statechannels/go-nitro/pull/733.

PR #734, opened 2022-05-31
Keep track of latest block number in direct defunding

lalexgap:
This updates direct defunding to keep track of the latest block number and ignore events with a stale block number.

This follows the same pattern that we use in direct funding.


lalexgap:
This doesn't fully protect us as when we initialize the defund objective `latestBlockNum=0`. So that means if the first chain event we receive is a stale `Deposit` event we can still error.

I think we can either:
- Update direct defunding to ignore deposit events instead of panicking
- Accept a `latestBlockNum` when constructing the objective.

Issue #735, opened 2022-05-31
Direct Defunding can fail if it receives a stale deposit chain event


lalexgap:
This doesn't fully protect us as when we initialize the defund objective `latestBlockNum=0`. So that means if the first chain event we receive is a stale `Deposit` event we can still error.

I think we can either:
- Update direct defunding to ignore deposit events instead of panicking
- Accept a `latestBlockNum` when constructing the objective.

_Originally posted by @lalexgap in https://github.com/statechannels/go-nitro/issues/734#issuecomment-1142444486_

lalexgap:
Closing for now as I think the stale events were due to testground

PR #736, opened 2022-05-31
Ignore deposit events in direct defunding

lalexgap:
This is a quick workaround for #735. Instead of throwing an error when receiving a deposit event we just ignore it.

### DIFF
--- a/protocols/directdefund/directdefund.go
+++ b/protocols/directdefund/directdefund.go
@@ -214,6 +214,10 @@ func (o *Objective) UpdateWithChainEvent(event chainservice.Event) (protocols.Ob
 				updated.C.OnChainFunding = e.Holdings.Clone()
 			}
 		}
+	case chainservice.DepositedEvent:
+		// TODO: This is a workaround for https://github.com/statechannels/go-nitro/issues/735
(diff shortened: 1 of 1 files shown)


protocols/directdefund/directdefund.go
```diff
@@ -214,6 +214,10 @@ func (o *Objective) UpdateWithChainEvent(event chainservice.Event) (protocols.Ob
 				updated.C.OnChainFunding = e.Holdings.Clone()
 			}
 		}
+	case chainservice.DepositedEvent:
+		// TODO: This is a workaround for https://github.com/statechannels/go-nitro/issues/735
+		// Eventually we should handle this by checking block numbers
```
geoknee, line 219: Can you clarify how that solution will work? Is there anything "wrong" this approach / could we remove this todo?

geoknee (approved):

protocols/directdefund/directdefund.go
lalexgap: Good point, I think this is a probably a valid way to deal with it.

lalexgap:
Closing this for now as I think the stale events were due to a bug with the testground test. I'll reopen this if the issue reappears.

PR #739, opened 2022-06-02
Introduce a PolicyMaker

andrewgordstewart:
This PR introduces a PolicyMaker, which is an object that is used by the engine to decide whether or not to approve or reject an objective that's created via an incoming SignedState.

We also considered having the client emit events when an objective is created, and introducing ApproveObjective/RejectObjective. I preferred injecting a PolicyMaker, as it was an easier solution in the short-term.

**What's not finished**
If Alice starts an objective involving Bob, and Bob rejects it, then Alice's client will wait indefinitely for the objective to complete. To accomplish this requires changing the wire protocol. Therefore, that's kept as future work to narrow the scope of this PR.

**How Has This Been Tested?**

I added a new integration test which checks that:
1. Bob's rejected objective completes, and the rejected objective is stored in the store
2. Alice's objective does not complete3. 

(2) should be changed after the wire protocol accommodates communicating rejected objectives.


client/engine/engine.go
```diff
@@ -394,7 +403,7 @@ func (e Engine) spawnConsensusChannelIfDirectFundObjective(crankedObjective prot
 	return nil
 }
 
-// getOrCreateObjective creates the objective if the supplied message is a proposal. Otherwise, it attempts to get the objective from the store.
+// getOrCreateObjective retrieves the objective from the store. if the objective does not exist, it creates the objective if the supplied signed state, and stores it in the store
```
lalexgap, line 406:
```suggestion
// getOrCreateObjective retrieves the objective from the store. if the objective does not exist, it creates the objective using the supplied signed state, and stores it in the store
```

client/engine/engine.go
```diff
@@ -424,28 +433,26 @@ func (e *Engine) constructObjectiveFromMessage(id protocols.ObjectiveId, ss stat
 
 	switch {
 	case directfund.IsDirectFundObjective(id):
-		dfo, err := directfund.ConstructFromState(true, ss.State(), *e.store.GetAddress())
+		dfo, err := directfund.ConstructFromState(false, ss.State(), *e.store.GetAddress())
```
lalexgap, line 436: Now that we're introducing approval/rejection could we just remove the `preApprove` flag from `ConstructFromState` ?

lalexgap (approved):

client/engine/engine.go
andrewgordstewart: Actually, I briefly went down this path, and the size of the change grew quite large.

I ran into a couple of issues. One is:
- `preapprove=false` is probably the right default value
- however, when creating an objective due to an API event, we need to set it to `true`
- setting it to `true` via the `Approve` function is awkward, because the return value has a different type. many places that use the objective are expecting a specific objective struct, but `Approve` returns a `protocols.Objective` interface

PR #740, opened 2022-06-03
Record performance metrics

lalexgap:
The engine now records metrics about various events and stats.
It records metrics such as:
- amount of proposals/states in an outgoing message
- how long objectives takes to complete
- how long crank, handleMessage.. functions take to complete

The client now accepts a `MetricsAPI` that is used to record metrics. This API based around the testground [metrics API](https://github.com/testground/sdk-go/blob/master/runtime/metrics_api.go) so we can easily pass in the testground metric recorder. The testground metrics api is wrapper around  [go-metrics](https://github.com/rcrowley/go-metrics) which is a performant and light-weight metrics library.  If a nil pointer is passed in for the MetricsAPI we use a `NoOpMetricsAPI` which doesn't record anything. 

The `MetricsRecorder` is a helper struct that wraps the `MetricsAPI`. It exposes functions to record metrics about objectives,  queue lengths and messages.  It provides functions to record the duration of objectives, statistics about messages and a function to time the length of functions.

I originally looked at recording metrics outside of the `client` however by adding metrics inside the client we're able to gather performance metrics about various function calls and get a better view of where slow downs are.

These graphs are based on these capture metrics after running the test via testground:
image


client/engine/metrics.go
```diff
@@ -0,0 +1,145 @@
...
+	//
+	// The format of the metric name is a comma-separated list, where the first
+	// element is the metric name, and optionally, an unbounded list of
+	// key-value pairs. Example:
+	//
+	//   requests_received,tag1=value1,tag2=value2,tag3=value3
+	Gauge(name string) metrics.GaugeFloat64
+}
+
+// NewNoOpMetrics returns a MetricsApi that does nothing.
```
geoknee, line 59: nit: is "returns" the correct terminology here?

geoknee (approved):
Looks good! 

My only point of criticism is that we are building quite a bit of functionality here, the advantages of which are not really visible in this repo. In fact, we don't really get to see the metrics being recorded at all. Would it be worth adding a small test which uses a non-nil `MetricsAPI` to _this_ repo?

lalexgap:
> Looks good!
> 
> My only point of criticism is that we are building quite a bit of functionality here, the advantages of which are not really visible in this repo. In fact, we don't really get to see the metrics being recorded at all. Would it be worth adding a small test which uses a non-nil `MetricsAPI` to _this_ repo?

That's a good point! I think the only problem is that [go-metrics](https://github.com/rcrowley/go-metrics) requires a `influxdb` backend, so we'd either have to add more configuration/setup or implement our own metrics recorder.

I'm going to kick this can down the road a little bit and defer to this to a separate issue for now, since I think it would be non-trivial to implement a `MetricsAPI` in this repo.

PR #741, opened 2022-06-06
reimplement MockChain transaction re-broadcast

lalexgap:
To support `testground` tests  the MockChain previously allowed a `chan` to be passed in that receives any submitted transactions. This lets us listen for transactions a `MockChain` receives and replay them on other clients `MockChain`s.

This reimplements that behaviour after the recent chain service refactor. This should only be temporary until we're using a shared ganache/geth instance.

### DIFF
--- a/client/engine/chainservice/mockchain.go
+++ b/client/engine/chainservice/mockchain.go
@@ -12,8 +12,9 @@ import (
 type MockChain struct {
 	ChainServiceBase
 
-	holdings map[types.Destination]types.Funds // holdings tracks funds for each channel
-	blockNum *uint64                           // MockChain is often passed around by value. The pointer allows for shared state.
+	holdings   map[types.Destination]types.Funds // holdings tracks funds for each channel
+	blockNum   *uint64                           // MockChain is often passed around by value. The pointer allows for shared state.
+	txListener chan protocols.ChainTransaction   // this is used to broadcast transactions that have been received
 }
 
 // NewMockChain returns a new MockChain.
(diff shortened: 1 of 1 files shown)


client/engine/chainservice/mockchain.go
```diff
@@ -50,6 +59,13 @@ func (mc *MockChain) SubscribeToEvents(a types.Address) <-chan Event {
 // SendTransaction responds to the given tx.
 func (mc *MockChain) SendTransaction(tx protocols.ChainTransaction) {
 	*mc.blockNum++
+	if mc.txListener != nil {
+		// Send to txListener and ignore if the chan is full
```
kerzhner, line 63: Should we ignore the "channel full" error? Seems like a full channel might lead to behavior that is tricky to debug.

kerzhner (approved):
This seems reasonable.

Do the customers of the MockChain run in separate processes? Or is there another reason different customers are not able to share the same MockChain object? If customers are able to share the same MockChain object, we would not need to re-submit the same transaction to multiple MockChains. We can subscribe many listeners to an event stream.

lalexgap:
> This seems reasonable.
> 
> Do the customers of the MockChain run in separate processes? Or is there another reason different customers are not able to share the same MockChain object? If customers are able to share the same MockChain object, we would not need to re-submit the same transaction to multiple MockChains. We can subscribe many listeners to an event stream.

In the testground integration test each client is run in its own process (or docker image) so they can't share the same `MockChain` object unfortunately.

Issue #461, opened 2022-03-31
Commit an ADR on how to migrate from using "union types"


andrewgordstewart:
# Context
Many types of proposals need to be serialized & deserialized, in order to be sent across the wire.

```
p1 = Add{guarantee}
p2 = Remove{guarantee}
p3 = TopUp{...}
```

A ConsensusChannel may modify its outcome by receiving these proposal:

```
sp_leader = leader.Propose(p1)

follower.Receive(sp1_leader)

_ = 1follower.SignNextProposal()

sp3_leader = leader.Propose(p3)

follower.Receive(sp3_leader)

sp3_follower = follower.SignNextProposal())

leader.UpdateConsensus(sp_follower)
```

A `message` might include any such type of proposal.
```
m1 = {
  id: "ledgerTopUp-143",
  proposal: sp3_follower
}

m2 = {
  id: "virtualfund-0xf00",
  proposal: sp_leader
}
```

Coming from typescript, we handled this type of situation with a combination of JSON schema definitions and generics.

## Problem

It is unclear how proposals, messages etc should be passed around as values.

Complications arise especially when serializing/deserializing these values: when `m2` is received, how does one determine which `UnmarshalJSON` function to call?

### 1. Empty interface
A `SignedProposal`, `Message` etc could hold an empty interface
```
type SignedProposal struct {
  Proposal interface{}
  Sig state.Signature
}

type Message struct {
  ObjectiveId string
  Data interface{}
}
```
In this case, `Proposal` & `Data` can be any type, and some combination of reflection, type inspection, or data inspection.

### 2. Poor man's sum type
A flexible type could pretend to be a sum type
```
type Proposal {
  Add Add
  Remove Remove
  TopUp TopUp
}

type SignedProposal struct {
  Proposal
  Sig state.Signature
}
```

In this case, something that uses a SignedProposal should probably check that exactly one of `Add`, `Remove` and `TopUp` are present.

### 3. Generics
Generics were recently added and may provide another option

# Action

If possible, explore how other codebases solve this problem. Commit an ADR with our solution.

andrewgordstewart:
https://github.com/statechannels/go-nitro/pull/454 is related

andrewgordstewart:
go-perun uses Option 1 (more or less) in their `wire` package:
[...]

PR #745, opened 2022-06-08
Contain the MessagePayload type to protocols package

andrewgordstewart:
This PR came about as I was trying to add more data to a MessagePayload. The existing type is awkward: it is a struct that **can** hold multiple values, but **should only** hold one value at a time. This has a few consequences:
1. when serialized into JSON, empty structs are serialized and sent across the wire. This hinders debugging somewhat, and is somewhat wasteful.
2. A package using the Message type can access a MessagePayload. The compiler does not prevent these packages from accessing zero-valued SignedState on a payload that contains a SignedProposal. 

This PR addressees these two issues by
i) Making `Message.payload` and `messagePayload` internal to the `protocols` package. Other packages must use the `SignedProposals()` or `SignedStates()` API, which return type-safe data structures.
ii)  implementing a custom `MarshalJSON` function for `messagePayload` type, shrinking serialized payloads to only contain non-zero values

This PR does not intend to change the technique used to implement a form of "sum types", because I wanted to minimize the amount of changes needed to comfortably support more payload types. We may wish to consider a different approach!


protocols/messages.go
```diff
@@ -121,12 +124,26 @@ func (p *messagePayload) MarshalJSON() ([]byte, error) {
...
+		if p.hasProposal() {
+			numPresent += 1
+		}
+		if p.hasState() {
+			numPresent += 1
+		}
+		if numPresent != 1 {
+			return Message{}, ErrInvalidPayload
+		}
+	}
```
geoknee, line 145: Is there an opportunity to DRY things out a bit by reusing the `.Type()`? method?

protocols/messages.go
```diff
@@ -47,13 +47,12 @@ func (p messagePayload) hasProposal() bool {
...
+	case p.hasProposal():
 		return SignedProposalPayload
-	case !p.hasProposal() && p.hasState():
+	case p.hasState():
 		return SignedStatePayload
-	case p.hasProposal() && p.hasState():
-		panic("payload has both state and proposal %v")
 	default:
+		// This package guarantees that payloads have exactly one state or proposal
 		panic("payload has neither state nor proposal")
```
geoknee, line 56: I'm not sure this is a good change. We're lumping the "neither" and "both" cases together and panicking with a message of "neither". The panic "should never happen" but if it does, it would be better to have the extra information (and not something potentially misleading). 

There's a small chance the panic will be triggered in future (after a misguided refactor, say). If we were 100% sure it would never be triggered, then we wouldn't even need the default case in the switch at all.

geoknee (approved):
A definite improvement. 

I also agree about us probably needing to have a different approach altogether. Relevant to https://github.com/statechannels/go-nitro/issues/461.

protocols/messages.go
andrewgordstewart: Unfortunately not, since the correctness of `.Type()` depends on this validation passing.

This change is actually shifting the responsibility of validating data:
- before, data was validated every time `Type()` was called. Adding new types makes this validation awkward (imagine `hasProposal() && !hasState() && !hasPoke() && !hasRejection() && !hasAddressChange()` on every line)
- now, data is validated when the message is constructed. Messages are constructed in two ways: i) a constructor and ii) by deserializing incoming messages

Alice's wallet will create **new** messages via constructors. We can assume these messages are safe. Alice may receive an unsafe message from Eve, so we should validate when deserializing, but from that point on we can assume the messages are safe. (This is of course implicitly assuming that payloads are immutable once created.)

All in all, I think that the pattern used at the present is risky if we care about "valid" union types (ie. this Payload is **exactly one of** a SignedStatePayload and a SignedProposalPayload)

Another idea is to put a static `type` property on the payload. On a payload of type `SignedState`, external packages would not see any data other than the SignedState. I'd probably prefer that to the status quo, but would do that in a different PR. (This PR was about removing Payload from the `protocols` package API)

protocols/messages.go
andrewgordstewart: Are you suggesting that `Type()` should continue to check:
- if there's a signed state, there's no other data
- if there's a signed proposal, there's no other data
?

My goal is to allow a Payload to contain more types of data. If there are `n` data types, then this sort of validation would perform up to `n^2` checks in the worst case. I am not worried about performance, but it is definitely not ideal.

geoknee: I agree that the full validation here does not scale well and is not necessary if full validation is done at construction time.

andrewgordstewart: Right. In that case, `Type()` can assume that validation has been performed, which means that "both" is not possible. (Also, notice that if "both" were present, `Type()` would return `SignedPayloadType` instead of panicking.)

Given that, do you still see an issue with it?

geoknee: If I were to nitpick, then yes. By the same argument "neither" is not possible, but yet we are handling that with a panic.

It would be less confusing to me if we had:

```go
if p.hasState() {
   return SignedStatePayload
} 
return SignedProposalPayload
```

andrewgordstewart: I am happy with doing it that way, as it is functionally equivalent. On the other hand, it has the downside of returning the wrong value when a new Type is introduced, rather than panicking. This increases the chance of a mistake slipping past the test suite.

fd1e42dffbe1afc5e103a5c3bc4f732a4efc0701

PR #749, opened 2022-06-17
Add channel defunding to EthChainService

kerzhner:
The primary goal of this PR is to use the EthChainService instead of the MockChain in the `TestDirectDefund` client test. To accomplish that:
- Conclude transaction is added to the EthChainService.
- Concluded and AllocationUpdated events are added to the EthChainService.
- A test is added for the above logic.
- ChannelNonce and wait-for-withdrawal bug fixes are implemented.

### DIFF
--- a/client/engine/chainservice/eth_chainservice.go
+++ b/client/engine/chainservice/eth_chainservice.go
@@ -65,7 +65,19 @@ func (ecs *EthChainService) SendTransaction(tx protocols.ChainTransaction) {
 				panic(err)
 			}
 		}
-	// TODO handle other transaction types
+	case protocols.WithdrawAllTransaction:
+		withdrawTx := tx.(protocols.WithdrawAllTransaction)
+
+		state := withdrawTx.SignedState.State()
+		signatures := withdrawTx.SignedState.Signatures()
+		nitroFixedPart := NitroAdjudicator.IForceMoveFixedPart(state.FixedPart())
+		nitroVariablePart := NitroAdjudicator.ConvertVariablePart(state.VariablePart())
+		nitroSignatures := []NitroAdjudicator.IForceMoveSignature{NitroAdjudicator.ConvertSignature(signatures[0]), NitroAdjudicator.ConvertSignature(signatures[1])}
+
+		_, err := ecs.na.ConcludeAndTransferAllAssets(&txOpt, nitroFixedPart, nitroVariablePart, 1, []uint8{0, 0}, nitroSignatures)
+		if err != nil {
+			panic(err)
+		}
 	default:
 		panic("unexpected chain transaction")
 	}
--- a/client/engine/chainservice/eth_chainservice.go
+++ b/client/engine/chainservice/eth_chainservice.go
@@ -16,8 +16,9 @@ import (
 	"github.com/statechannels/go-nitro/types"
 )
 
-var depositedTopic = crypto.Keccak256Hash([]byte("Deposited(bytes32,address,uint256,uint256)"))
 var allocationUpdatedTopic = crypto.Keccak256Hash([]byte("AllocationUpdated(bytes32,uint256,uint256)"))
+var concludedTopic = crypto.Keccak256Hash([]byte("Concluded(bytes32,uint48)"))
+var depositedTopic = crypto.Keccak256Hash([]byte("Deposited(bytes32,address,uint256,uint256)"))
 
 type eventSource interface {
 	SubscribeFilterLogs(ctx context.Context, query ethereum.FilterQuery, ch chan<- ethTypes.Log) (ethereum.Subscription, error)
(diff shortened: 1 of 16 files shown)


client/engine/chainservice/eth_chainservice.go
```diff
@@ -43,22 +43,23 @@ func NewEthChainService(na *NitroAdjudicator.NitroAdjudicator, naAddress common.
...
+		From:     ecs.txSigner.From,
+		Nonce:    ecs.txSigner.Nonce,
+		Signer:   ecs.txSigner.Signer,
+		GasPrice: big.NewInt(10000000000),
+	}
+	switch tx.(type) {
+	case protocols.DepositTransaction:
+		depositTx := tx.(protocols.DepositTransaction)
+		for tokenAddress, amount := range depositTx.Deposit {
+			txOpt.Value = amount
```
geoknee, line 57: I think we want to do this only when the tokenAddress indicates the native token (e.g ETH)?

protocols/interfaces.go
```diff
@@ -9,24 +9,39 @@ import (
...
-// ChainTransaction is an object to be sent to a blockchain provider.
-type ChainTransaction struct {
-	// TODO support other transaction types (deposit, challenge, respond, conclude)
-	Type      TransactionType
-	ChannelId types.Destination
-	Deposit   types.Funds
+// ChainTransaction defines the interface that every transaction must implement
+type ChainTransaction interface {
+	ChannelId() types.Destination
+}
```
geoknee, line 19: I understand this is one of the ways we can build a sort of union type in Go. I don't have a concrete suggestion for something better right now... but I just want to flag the awkwardness of the fact that now `state.FixedPart` and `state.State` implement `protocols.ChainTransaction`.

client_test/directfund_test.go
```diff
@@ -31,7 +31,7 @@ func directlyFundALedgerChannel(t *testing.T, alpha client.Client, beta client.C
 		AppDefinition:     types.Address{},
 		AppData:           types.Bytes{},
 		ChallengeDuration: big.NewInt(0),
-		Nonce:             rand.Int63(),
```
geoknee, line 34: Am I right in thinking the issue is that (with very high probability) the nonce was simply too large to be expressed as a 48 bit integer? Presumably the data type itself isn't the real issue, since after the fix we are generating a smaller random number and then casting to a 64 bit int anyway?

protocols/directdefund/directdefund.go
```diff
@@ -210,6 +210,8 @@ func (o *Objective) UpdateWithChainEvent(event chainservice.Event) (protocols.Ob
 				updated.C.OnChainFunding = e.Holdings.Clone()
 			}
 		}
+	case chainservice.ConcludedEvent:
+		break
 	default:
 		return &updated, fmt.Errorf("objective %+v cannot handle event %+v", updated, event)
```
geoknee, line 216: Would it be more scalable to have the `default` case simply `break`? 

Here it looks as though we are expecting `ConcludedEvents` to be "routed" to the `directdefund` objective. Is that a sign of a programming error or not? What about some future event type we have yet to cook up and integrate?

client/engine/chainservice/eth_chainservice.go
```diff
@@ -124,6 +123,7 @@ func (ecs *EthChainService) listenForLogEvents(na *NitroAdjudicator.NitroAdjudic
 					panic(err)
 				}
 
+				// TODO set Holdings based on chain event
```
geoknee, line 126: Shall we make this a full-blown issue?

client/engine/chainservice/simulated_backend_chainservice_test.go
```diff
@@ -1,16 +1,75 @@
...
+
+// actors namespaces the actors exported for test consumption
+type actors struct {
+	Alice actor
+	Bob   actor
+}
+
+// Actors is the endpoint for tests to consume constructed statechannel
+// network participants (public-key secret-key pairs)
+var Actors actors = actors{
```
geoknee, line 30: I think we have some of this boilerplate in the codebase already, which we could reuse here.

geoknee (approved):

client/engine/chainservice/eth_chainservice.go
kerzhner: Good call, filed https://github.com/statechannels/go-nitro/issues/755

protocols/interfaces.go
kerzhner: That's a good point. Down the line, the `ChainTransaction` interface will likely define more methods that are specific to chain transactions (as opposed to a generic `ChannelId()` methods).

protocols/directdefund/directdefund.go
kerzhner: That's a great question. In this particular case, the `directdefund` protocol might want to observe and record the concluded event. If a client needs to withdraw funds in the unhappy path, it is helpful for the client to know if the channel has been concluded. Though a client can always read the chain state.

In general, if a protocol receives an unexpected event, the protocol should log the event. At the moment, if a protocol receives an unexpected event, this is likely a sign of either a bug or a missing implementation.

client/engine/chainservice/eth_chainservice.go
kerzhner: That would be good, though this work is currently in progress.

client/engine/chainservice/simulated_backend_chainservice_test.go
kerzhner: Much obliged!

PR #754, opened 2022-06-21
Add ConsensusApp.sol

geoknee:
We are changing the API for both the on-chain `challenge` method, and also the behaviour around signatures.

This PR 
* updates existing test which integrate a `go-nitro` with the on chain code to use the new API. **Note:** I believe there are now more tests like these -- the base branch has moved on somewhat. 
* adds a simple `ConsensusApp`, allowing for unanimous consensus (`n` signatures on 1 state) to support a state. 
* Uses that app in the test.


generate-adjudicator-bindings.sh
```diff
@@ -1,6 +1,13 @@
 set -e
 cd nitro-protocol
+# NitroAdjudicator
 solc --base-path $(pwd) @statechannels/exit-format/=node_modules/@statechannels/exit-format/ @openzeppelin/contracts/=node_modules/@openzeppelin/contracts/ contracts/NitroAdjudicator.sol --optimize --bin --abi -o tmp-build
 abigen --abi=$(pwd)/tmp-build/NitroAdjudicator.abi --bin=$(pwd)/tmp-build/NitroAdjudicator.bin --pkg=NitroAdjudicator --out=$(pwd)/../client/engine/chainservice/adjudicator/NitroAdjudicator.go
+
+# TrivialApp
```
geoknee, line 7:
```suggestion
# ConsensusApp
```

nitro-protocol/test/contracts/ConsensusApp/latestSupportedState.test.ts
```diff
@@ -0,0 +1,113 @@
...
+const chainId = process.env.CHAIN_NETWORK_ID;
+const challengeDuration = 0x100;
+
+// Populate wallets and participants array
+for (let i = 0; i < 3; i++) {
+  wallets[i] = Wallet.createRandom();
+  participants[i] = wallets[i].address;
+}
+
+const channel: Channel = {chainId, channelNonce: 8, participants};
```
nksazonov, line 34: Why does `channelNonce` is equal to 8?

nksazonov (approved):

nitro-protocol/test/contracts/ConsensusApp/latestSupportedState.test.ts
geoknee: I just chose a random value, it shouldn't affect the test.

PR #756, opened 2022-06-22
Rejected objectives are communicated across the wire

andrewgordstewart:
Previously, when a go-nitro client rejects an objective, the objective stalls for other participants of the objective.

This PR changes that, by:
1. including a `Rejected` flag on message payloads
2. changing the signature of `Reject()` to return side effects, and forwarding "rejection notices"
3. rejecting objectives when a rejection notice is received

(2) seemed to be the path of least resistance, but 2+3 have some shortcomings:
i) in case of a lost message, the objective will stall
ii) since rejection notices are not signed, incoming messages require authentication. otherwise, anyone can send a rejection notice for any objective and cause objectives to stall

(i) is a problem elsewhere anyway, and would likely be resolved with some general objective synchronization code.

The behaviour is tested by adding two new integration tests:
- when direct funding, if one peer rejects the objective, then the other peer's objective should also complete (in a rejected state)
- when virtual de funding, if one peer rejects the objective, then the other peer's objective should also complete (in a rejected state) and gracefully handle happy-path messages in that rejected state


geoknee:
This seems pretty sensible, but I wanted to just offer an alternative view to spark a discussion.

Without the authentication part, this change is enlarging the attack surface of the `go-nitro` client. There is a new griefing vector -- adversaries can prevent me from making progress on objectives which I would otherwise be able to make progress on. Of course the authentication would not be difficult to add (as you suggest, a signature ought to do it). 

Inactivity is perhaps as likely as an explicit rejection from a peer. In that case, my client should have a timeout for objective progress (it could be set via a new policy, say). Some periodic process would detect stalls, and transform objectives into rejected objectives after the appropriate time has elapsed with no progress. 

If we take that as the default, explicit rejection messages are really just a courtesy offered to my peers to allow them to garbage collect channels more rapidly, or engage fallback behaviour more quickly. So still worth having, but probably only with the authentication in place. Otherwise, we are trading a courtesy or performance gain for a griefing vector, which seems like a bad trade?

Thoughts?

andrewgordstewart:
That's an interesting question!

The first question that pops into my mind is: What should happen in case Bob goes offline momentarily, mid-objective? Should the system be designed to enable recovery, or should it be designed so that Alice treats the objective as rejected or cancelled?

I would imagine that the system should enable recovery in this case. Let's say it does. Then, Alice can't obviously distinguish between "Bob is offline" and "Bob has rejected the objective". Presumably, Alice might want to poke Bob after some timeout, and if Bob has rejected the objective, then Bob ought to respond, right?

What I'm wondering here is:
- it seems like we would need to communicate rejected objectives anyway. In that case, is it better to communicated rejected objectives right away?
- is it advantageous to design the protocol to not depend on authentication? or is it better to depend on authentication at a high level, so that the protocol can assume that messages have been authenticated?

andrewgordstewart:
> > * is it advantageous to design the protocol to not depend on authentication? or is it better to depend on authentication at a high level, so that the protocol can assume that messages have been authenticated?
> 
> I think it would be best to do the auth inside each protocol. Although this will likely lead to some duplication, I think it fits with the existing design philosophy of:
> 
> * having as much of the complexity inside the (pure) protocol implementations as possible, keeping the engine as thin as possible
> * doing signature recovery _inside_ protocols, rather than outside.
> 
> We could make an issue to add authentication as a follow-up to this PR?

To be clear, you're suggesting that authentication is performed by having a participant sign a message with the same key they use inside the channel? That sounds like a reasonable approach to me, though I don't think it is a typical authentication strategy.

I am curious, though, about why it makes sense to do signature recover inside protocols. Normally, one would handle authentication as close to the entry point to a system as possible, rather than in the business logic. In our current protocols, the signature verification is (arguably) business logic -- a state update is valid only when accompanied with a correct signature. In general, however, message authentication may happen via some auth token that's periodically refreshed, rather than via signatures on messages.

geoknee:
I take your points about authentication. I suppose what you have in mind  is not really part of each protocol _per se_, but something which (as you say) can cut out malicious or irrelevant messages as early as possible to prevent/reduce unnecessary work and ultimately DoS attacks. 

Having said that, doing this "properly" with rotating tokens and so on seems like quite a bit of work. That's why I was drawn to a quick-and-dirty solution which can reuse a lot of code (requiring signatures by counterparty signing keys on rejection messages, and recovering / verifying those signatures in-protocol).

Another way forward is to merge this PR as is (I haven't reviewed it yet, but will), and file an issue so that we can circle back and close off the DoS vector when that becomes important.

client_test/directfund_test.go
```diff
@@ -81,12 +78,12 @@ func TestWhenObjectiveIsRejected(t *testing.T) {
 
 	response := clientA.CreateDirectChannel(request)
 
-	waitTimeForCompletedObjectiveIds(t, &clientB, time.Second, response.Id)
+	waitTimeForCompletedObjectiveIds(t, &clientA, time.Second, response.Id)
 
 	obj, _ := storeA.GetObjectiveById(response.Id)
 
-	if obj.GetStatus() != protocols.Approved {
```
geoknee, line 88: It looks like this test was asserting incorrect behaviour, and passing because the client was exhibiting the same incorrect behaviour? Is that right?

I suppose in an ideal world the test would have asserted correct behaviour, and been skipped. And then you could have just unskipped it here?

protocols/interfaces.go
```diff
@@ -88,6 +88,11 @@ type Objective interface {
 // ObjectiveId is a unique identifier for an Objective.
 type ObjectiveId string
 
+// SortInfo needs to be implemented so that RejectionNotices can be received.
+func (id ObjectiveId) SortInfo() (channelID types.Destination, turnNum uint64) {
+	return
+}
```
geoknee, line 94: I'm a bit confused about this. Can you explain is some more?

protocols/messages.go
```diff
@@ -176,7 +185,7 @@ func (se *SideEffects) Merge(other SideEffects) {
 // PayloadValue is a type constraint that specifies a payload is either a SignedProposal or SignedState.
 // It includes functions to get basic info to allow sorting.
 type PayloadValue interface {
-	state.SignedState | consensus_channel.SignedProposal
+	state.SignedState | consensus_channel.SignedProposal | ObjectiveId
```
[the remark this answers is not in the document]
geoknee: Probably also some godoc to update on L65, too.

protocols/messages.go
```diff
@@ -100,6 +100,20 @@ func (m Message) SignedProposals() []ObjectivePayload[consensus_channel.SignedPr
 	return signedProposals
 }
 
+// SignedProposals returns a slice of signed proposals with their objectiveId that were contained in the message.
+// The proposals are sorted by ledger id then turnNum.
```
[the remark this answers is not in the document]
geoknee: Looks like the whole godoc is wrong, maybe a copy-paste error.

protocols/messages.go
```diff
@@ -91,6 +100,20 @@ func (m Message) SignedProposals() []ObjectivePayload[consensus_channel.SignedPr
...
+func (m Message) RejectedObjectives() []ObjectivePayload[ObjectiveId] {
+	rejectedObjectives := make([]ObjectivePayload[ObjectiveId], 0)
+	for _, p := range m.payloads {
+		if p.Type() == RejectionNoticePayload {
+			entry := ObjectivePayload[ObjectiveId]{p.ObjectiveId, p.ObjectiveId}
```
geoknee, line 109: This looks a bit smelly. See my other comment about duplicating `p.ObjectiveId`.

geoknee (approved):
This is looking pretty good, but I do have a few points of confusion which I would like clearing up 🙏 .

client_test/directfund_test.go
andrewgordstewart: Skipping the intended test seems totally reasonable to me. I think that the old behaviour was somewhat intentional though?

client/engine/engine.go
```diff
@@ -263,6 +266,28 @@ func (e *Engine) handleMessage(message protocols.Message) (ObjectiveChangeEvent,
...
+		if err != nil {
+			return ObjectiveChangeEvent{}, err
+		}
+		if objective.GetStatus() == protocols.Rejected {
+			e.logger.Printf("Ignoring payload for rejected objective  %s", objective.Id())
+			continue
+		}
+
+		// We assume the peer who rejected the objective, so we ignore the side effects.
+		// (We could also forward side effects at this point, introducing some redundant messages.)
```
andrewgordstewart, line 282:
```suggestion
		// We assume the peer who rejected the objective has already messaged the other peers
		// , so we ignore the side effects.
		// (We could also forward side effects at this point, introducing some redundant messages.)
```

client/engine/engine.go
```diff
@@ -264,6 +262,28 @@ func (e *Engine) handleMessage(message protocols.Message) (ObjectiveChangeEvent,
...
```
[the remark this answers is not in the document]
andrewgordstewart: Yes, exactly. I finished what I intended to say below, but your suggestion is a bit more clear IMO

client/engine/engine.go
andrewgordstewart:
```suggestion
		// we are rejecting due to a counterparty message notifying us of their rejection. We 		
		// do not need to send a message back to that counterparty, and furthermore we assume that
		// counterparty has already notified all other interested parties. We can therefore ignore the side effects
```

protocols/messages.go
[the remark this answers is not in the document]
andrewgordstewart: Yes, L65 needed to be updated. I tried to clarify this in 8127ead437b133b8914a5f14b5c55eb3ea3710c5.

protocols/messages.go
andrewgordstewart: Resolved in 8127ead437b133b8914a5f14b5c55eb3ea3710c5

protocols/messages.go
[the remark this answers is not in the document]
andrewgordstewart: Resolved in 8127ead437b133b8914a5f14b5c55eb3ea3710c5

protocols/interfaces.go
andrewgordstewart: Does 8127ead437b133b8914a5f14b5c55eb3ea3710c5 clarify things?

client/engine/engine.go
```diff
@@ -264,6 +262,28 @@ func (e *Engine) handleMessage(message protocols.Message) (ObjectiveChangeEvent,
...
 	}
+
+	for _, entry := range message.RejectedObjectives() {
+		objective, err := e.store.GetObjectiveById(entry.ObjectiveId)
+
+		if err != nil {
+			return ObjectiveChangeEvent{}, err
+		}
+		if objective.GetStatus() == protocols.Rejected {
+			e.logger.Printf("Ignoring payload for rejected objective  %s", objective.Id())
```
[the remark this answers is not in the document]
andrewgordstewart: What context is missing?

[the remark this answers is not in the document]
geoknee: I don't know 😆 . This seems find to me on a second read.

PR #757, opened 2022-06-23
AllocationUpdated event includes holding based on chain state

kerzhner:
Fixes https://github.com/statechannels/go-nitro/blob/a70eef3220cb4fd0cfecd727969a1a6a57bdaf75/client/engine/chainservice/eth_chainservice.go#L126

A chain AllocationUpdated event is transformed by the chain service into a go-nitro AllocationUpdated event. This PR modifies the go-nitro events. 

AllocationUpdated events generated by the EthChainService now set the holdings based on state of the contract after the transaction which generated the chain event. A new behavior is that an AllocationUpdated event only contains assets which have were updated by the transaction. This behavior is introduced since:
- A separate AllocationUpdated is emitted by the chain for each asset.
- We are not able to iterate over all assets for a channel to determine which assets a channel holds.


protocols/directdefund/directdefund_test.go
```diff
@@ -270,7 +270,7 @@ func TestCrankBob(t *testing.T) {
 	}
 
 	// The third crank. Bob is expected to enter the terminal state of the defunding protocol.
-	updated, err = updated.(*Objective).UpdateWithChainEvent(chainservice.AllocationUpdatedEvent{Holdings: types.Funds{}})
```
geoknee, line 273: Why was this necessary?

client/engine/chainservice/eth_chainservice.go
```diff
@@ -123,8 +124,17 @@ func (ecs *EthChainService) listenForLogEvents(na *NitroAdjudicator.NitroAdjudic
...
-				event := AllocationUpdatedEvent{CommonEvent: CommonEvent{channelID: au.ChannelId, BlockNum: chainEvent.BlockNumber}, Holdings: types.Funds{}}
+				tx, pending, err := chain.TransactionByHash(context.Background(), chainEvent.TxHash)
+				if pending || err != nil {
+					panic("Expected transacion to be part of the chain")
+				}
+
+				assetAddress, amount, err := getChainHolding(na, tx, au)
+				if err != nil {
+					panic(err)
+				}
```
geoknee, line 135: Do we want to panic here? What kind of errors might we expect?

client/engine/chainservice/eth_chainservice.go
```diff
@@ -123,8 +124,17 @@ func (ecs *EthChainService) listenForLogEvents(na *NitroAdjudicator.NitroAdjudic
 					panic(err)
 				}
 
-				// TODO set Holdings based on chain event
-				event := AllocationUpdatedEvent{CommonEvent: CommonEvent{channelID: au.ChannelId, BlockNum: chainEvent.BlockNumber}, Holdings: types.Funds{}}
+				tx, pending, err := chain.TransactionByHash(context.Background(), chainEvent.TxHash)
+				if pending || err != nil {
+					panic("Expected transacion to be part of the chain")
```
geoknee, line 129: I think we want to avoid panicking here. I suppose the condition could be triggered by a re-org? In any case, we don't want to crash the whole wallet. It would be better to log the problem and carry on (for now, until we come up with a better error handling strategy).

client/engine/chainservice/eth_chain_helpers.go
```diff
@@ -0,0 +1,71 @@
...
+		AppData []uint8  "json:\"appData\""
+		TurnNum *big.Int "json:\"turnNum\""
+		IsFinal bool     "json:\"isFinal\""
+	})
+	return variablePart.Outcome[index.Int64()].Asset, nil
+
+}
+
+func decodeTxParams(abi *abi.ABI, data []byte) (map[string]interface{}, error) {
+	v := map[string]interface{}{}
```
geoknee, line 60: Super nit: you could wait to initialise this until you know there is no error on L61. Would improve readability a touch.

client/engine/chainservice/eth_chain_helpers.go
```diff
@@ -0,0 +1,71 @@
...
+		return assetAddress, amount, err
+	}
+	return assetAddress, amount, nil
+}
+
+// assetAddressForIndex uses the input parameters of a transaction to map an asset index to an asset address
+func assetAddressForIndex(na *NitroAdjudicator.NitroAdjudicator, tx *types.Transaction, index *big.Int) (common.Address, error) {
+	abi, err := NitroAdjudicator.NitroAdjudicatorMetaData.GetAbi()
+	if err != nil {
+		return common.Address{}, errors.New("unable to get ABI for a bound contract")
```
geoknee, line 31: This seems like a strange error. Can you explain it some more? Why not wrap or simply re-use the `err` form L29?

client/engine/chainservice/eth_chain_helpers.go
```diff
@@ -0,0 +1,71 @@
...
+			Allocations []struct {
+				Destination    [32]uint8 "json:\"destination\""
+				Amount         *big.Int  "json:\"amount\""
+				AllocationType uint8     "json:\"allocationType\""
+				Metadata       []uint8   "json:\"metadata\""
+			} "json:\"allocations\""
+		} "json:\"outcome\""
+		AppData []uint8  "json:\"appData\""
+		TurnNum *big.Int "json:\"turnNum\""
+		IsFinal bool     "json:\"isFinal\""
```
geoknee, line 53: Do you need all this if all you want to reference is `Outcome[index].Asset`?

geoknee (commented):
Could we add a test to this PR to show off the new functionality? The code is quite complex so it is difficult to have full confidence that it works (or even precisely what bug it will fix?).

Our contract events system does seem to make it painful to extract this simple information. We should probably take another look and ensure the gas savings are worth it. Put another way, we could add an event to make the off chain code simpler, but that will increase gas consumption.

client/engine/chainservice/eth_chainservice.go
kerzhner: Do you have suggestion on what the code should do instead? An error here means that we attempted to read the amount of tokens held for a channel in the adjudicator and failed to do so.

At the moment, the chain service is panic-happy. Down the line, a better approach might be for the chain service to emit errors via a channel and allow other components to handle the errors.

client/engine/chainservice/eth_chainservice.go
kerzhner: You are pointing out a larger problem here. We don't have a strategy for chain service error handling. Perhaps we should address this in a future PR instead of starting to mix logs vs panics for errors?

client/engine/chainservice/eth_chain_helpers.go
kerzhner: Good point. Changed in 78dc149a.

client/engine/chainservice/eth_chain_helpers.go
kerzhner: It seems that golang requires the exact struct type to be specified to convert an interface to that struct. Otherwise, we see a panic along the lines of "panic: interface conversion: interface {} is _exact struct_ not _incomplete struct_

protocols/directdefund/directdefund_test.go
geoknee: I think we are in agreement about the change. I was just wondering about the difference between: `Holdings: types.Funds{}` and `Holdings: types.Funds{types.Address{}: new(big.Int).SetInt64(0)}`. But I think I figured it out -- you've replaced an empty funds mapping with a nonempty mapping with an explicit `0: 0` entry. 

The thrust of this PR implies that `Holdings` must be iterated over: that is, a missing key is no longer the same as a key mapped to `0`. So the test needs to feed in more explicit data (to match the way the chain service will do it). 

Anyway, TLDR: this change makes sense and your suggested change to `AllocationUpdatedEvent` looks good to me.

geoknee (approved):
I think that amended test does the job nicely 👍

PR #765, opened 2022-06-29
Prevent directdefund on active ledger channels

### LINKED ISSUE #762: Specify behaviour of directdefund where ledger channel has existing guarantees
As far as I can tell, a ledger channel can currently be defunded while it has guarantees funding live virtual channels. As part of its `Crank`, a **WithdrawAll** transaction side effect is spawned. 

Options:

- **Allowing this sort of thing**: the `WithdrawAll` transaction side-effect could "pay out" the guarantees on-chain to the specified destinations, converting the virtual channels into directly funded channels.
- **Disallowing this sort of thing**: the `WithdrawAll` transaction could revert if the channel outcome contains guarantees. This should be accompanied by some client-side checks on the directdefund protocol for an empty guarantee list.

NiloCK:
Closes #762, in the direction of "disallow". This seems preferable because it makes for a better ratio of `[offchain work] / [onchain work]` for the protocol at large.

**Changes**

PR adds:
- [a simple check](https://github.com/statechannels/go-nitro/blob/2ca592a3053a326eb3956e34b0a58d6750179364/protocols/directdefund/directdefund.go#L76) on the consensus channel targeted by a `directdefund` objective.
- a pipeline for communication of failed objectives from the engine to the client

A test has been added to `client_test`, which
- sets up a ledger channel
- funds a virtual channel
- attempts to close the ledger channel
- asserts on the returned objective failure

---

#### Code quality

#### Project management

### DIFF
--- a/channel/consensus_channel/consensus_channel.go
+++ b/channel/consensus_channel/consensus_channel.go
@@ -217,6 +217,11 @@ func (c *ConsensusChannel) Follower() common.Address {
 	return c.fp.Participants[Follower]
 }
 
+// FundeingTargets returns a list of channels funded by the ConsensusChannel
+func (c *ConsensusChannel) FundingTargets() []types.Destination {
+	return c.current.Outcome.fundingTargets()
+}
+
 func (c *ConsensusChannel) Accept(p SignedProposal) error {
 	panic("UNIMPLEMENTED")
 }
@@ -499,6 +504,17 @@ func (o *LedgerOutcome) AsOutcome() outcome.Exit {
 			Allocations: allocations,
 		},
 	}
+}
+
+// fundingTargets returns a list of channels funded by the LedgerOutcome
+func (o *LedgerOutcome) fundingTargets() []types.Destination {
+	targets := []types.Destination{}
+
+	for dest := range o.guarantees {
+		targets = append(targets, dest)
+	}
+
+	return targets
 }
 
 // Vars stores the turn number and outcome for a state in a consensus channel.
--- a/client/client.go
+++ b/client/client.go
@@ -59,7 +59,7 @@ func (c *Client) handleEngineEvents() {
 
 		}
 
-		for _, erred := range update.ErredObjectives {
+		for _, erred := range update.FailedObjectives {
 			c.failedObjectives <- erred
 		}
 
--- a/client/client.go
+++ b/client/client.go
@@ -22,6 +22,7 @@ type Client struct {
 	engine              engine.Engine // The core business logic of the client
 	Address             *types.Address
 	completedObjectives chan protocols.ObjectiveId
+	failedObjectives    chan protocols.ObjectiveId
 }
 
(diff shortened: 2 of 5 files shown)


channel/consensus_channel/consensus_channel.go
```diff
@@ -217,6 +217,11 @@ func (c *ConsensusChannel) Follower() common.Address {
 	return c.fp.Participants[Follower]
 }
 
+// FundeingTargets returns a list of channels funded by the ConsensusChannel
```
geoknee, line 220:
```suggestion
// FundingTargets returns a list of channels funded by the ConsensusChannel
```

client/client.go
```diff
@@ -67,6 +73,11 @@ func (c *Client) CompletedObjectives() <-chan protocols.ObjectiveId {
 	return c.completedObjectives
 }
 
+// FailedObjectives returns a chan that receives a objective id whenever that objective has failed
```
geoknee, line 76:
```suggestion
// FailedObjectives returns a chan that receives an objective id whenever that objective has failed
```

client/engine/engine.go
```diff
@@ -60,6 +60,8 @@ type APIEvent struct {
 type ObjectiveChangeEvent struct {
 	// These are objectives that are now completed
 	CompletedObjectives []protocols.Objective
+	// These are objectives that have failed
+	ErredObjectives []protocols.ObjectiveId
```
geoknee, line 64: Shall we just call this `FailedObjectives`?

client/engine/engine.go
```diff
@@ -331,7 +339,7 @@ func (e *Engine) handleAPIEvent(apiEvent APIEvent) (ObjectiveChangeEvent, error)
 			e.metrics.RecordObjectiveStarted(request.Id(*e.store.GetAddress()))
 			ddfo, err := directdefund.NewObjective(request, true, e.store.GetConsensusChannelById)
 			if err != nil {
-				return ObjectiveChangeEvent{}, fmt.Errorf("handleAPIEvent: Could not create objective for %+v: %w", request, err)
+				return ObjectiveChangeEvent{ErredObjectives: []protocols.ObjectiveId{request.Id(e.metrics.me)}}, fmt.Errorf("handleAPIEvent: Could not create objective for %+v: %w", request, err)
```
geoknee, line 342:
```suggestion
				return ObjectiveChangeEvent{ErredObjectives: []protocols.ObjectiveId{request.Id(*e.store.GetAddress())}}, fmt.Errorf("handleAPIEvent: Could not create objective for %+v: %w", request, err)
```

1. Is this equivalent to what you have?
2. Is it better this way, to not mix production and metrics code?

geoknee (approved):
Looks very reasonable to me.

client/engine/engine.go
NiloCK: 1. in fact, anything you pass into this function would be equivalent - the argument isn't used by the function! (todo)
2. the `metrics` property was new to me, but it seems generally hard-wired into production methods. See `engine.Run`

In any case, your suggestion is an improvement.

Issue #753, opened 2022-06-20
Implement metrics API in local test


lalexgap:
> Looks good!
> 
> My only point of criticism is that we are building quite a bit of functionality here, the advantages of which are not really visible in this repo. In fact, we don't really get to see the metrics being recorded at all. Would it be worth adding a small test which uses a non-nil `MetricsAPI` to _this_ repo?

That's a good point! I think the only problem is that [go-metrics](https://github.com/rcrowley/go-metrics) requires a `influxdb` backend, so we'd either have to add more configuration/setup or implement our own metrics recorder.

I'm going to kick this can down the road a little bit and defer to this to a separate issue for now, since I think it would be non-trivial to implement a `MetricsAPI` in this repo.

_Originally posted by @lalexgap in https://github.com/statechannels/go-nitro/issues/740#issuecomment-1160669609_

lalexgap:
We've migrated away from testground so this is no longer relevant.

PR #782, opened 2022-07-07
Add more metrics to the wallet

### LINKED ISSUE #781: Capture more metrics from the wallet
We should capture more metrics about various function calls in the wallet (like executeSideEffects) so we can see where any bottlenecks.

lalexgap:
Fixes #781 

Adds some additional metric capturing to the wallet so we can have better insight into what parts of the engine are fast / slow.

Unfortunately since I haven't completed #753 these metrics are only visible via the testground test.

### DIFF
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -12,6 +12,7 @@ import (
 	"github.com/statechannels/go-nitro/client/engine/chainservice"
 	"github.com/statechannels/go-nitro/client/engine/messageservice"
 	"github.com/statechannels/go-nitro/client/engine/store"
+
 	"github.com/statechannels/go-nitro/protocols"
 	"github.com/statechannels/go-nitro/protocols/directdefund"
 	"github.com/statechannels/go-nitro/protocols/directfund"
@@ -221,7 +222,11 @@ func (e *Engine) handleMessage(message protocols.Message) (ObjectiveChangeEvent,
 			SignedProposal: consensus_channel.SignedProposal{},
 			SignedState:    entry.Payload,
 		}
-		updatedObjective, err := objective.Update(event)
+		var updatedObjective protocols.Objective
+
+		e.metrics.RecordDuration("objective_update", func() {
+			updatedObjective, err = objective.Update(event)
+		})
 		if err != nil {
 			return ObjectiveChangeEvent{}, err
 		}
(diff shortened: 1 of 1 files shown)


kerzhner:
Have we explored any alternatives to the current `RecordDuration` API? I find that a level of nesting required to create a function for the second parameter of `RecordDuration` results in less readable code. Some alternatives (which have their own flaws are):
1. A start timer call and an end timer call. The resulting code would be:
``` golang
timer.start('funcA + funcB')
funcA()
funcB()
timer.end('funcA + funcB')
```
2. If a single function is timed, we could leverage `defer`:
``` golang
func funcA() {
  defer endTimer('funcA')()
  // funcA implementation
}
```

lalexgap:
> Have we explored any alternatives to the current `RecordDuration` API? I find that a level of nesting required to create a function for the second parameter of `RecordDuration` results in:
> 
> * less readable code.
> * a deceiving git history + larger than necessary PR diffs.

Fair point! Originally I was avoiding seperate calls for `start` and `end` to avoid having to worry about remembering the correct timer name to pass into `end`.

However I think you're right that the nesting for `RecordDuration` ends up being less clear so I'd be happy to switch over to `start/end`

Another possible option would be having the `start` function that returns some struct that exposes an `end` function:
 ```go
 myTimer:= timer.start('funcA + funcB')
 funcA()
 funcB()
myTimer.end()
```
> 2. If a single function is timed, we could leverage `defer`:
> 
> ```go
> func funcA() {
>   defer endTimer('funcA')()
>   // funcA implementation
> }
> ```
For this option I'm assuming we'd still need to start the timer (either in or before `funcA`). I think making use of `defer` where we can make sense!

**Comment by kerzhner** [comment]
> For this option I'm assuming we'd still need to start the timer (either in or before `funcA`)? I think making use of `defer` where we can make sense!

If the duration of an entire function is timed, it should be possible to do so with a single line. [This reference](https://stackoverflow.com/a/45766707/13280122) describes how.

**Comment by lalexgap** [comment]
```
timer.start('funcA + funcB')
funcA()
funcB()
timer.end('funcA + funcB')
```
One downside to the approach of using a `start` and `end` function is that the book keeping can be slightly painful. When `start` is called we need to record the start time somewhere that we can access later in the call to `end` to calculate the duration.

We could store the start time in a `map` and use the timer `name` as a key for it. However this relies on the assumption that timer names are unique and you'll never have two timers with the same name running at the same time. This assumption probably holds in our current engine as the run loop runs in one goroutine  but would probably break down if we introduce more concurrent processing.

We could just allow the calling code to handle capturing the start time.
```
start:= time.Now()
funcA()
funcB()
timer.RecordDuration('funcA + funcB', time.Since(start))
```

or we could hide that behind a struct returned by the `start` function
```
myTimer:= timer.start('funcA + funcB')
funcA()
funcB()
myTimer.end()
```
However I'm starting to lean towards the defer option you outlined (which is pretty clever!). 
```
func funcA() {
  defer endTimer('funcA')()
  // funcA implementation
}
```
 We'd lose a bit granularity as we 'd only record the duration of complete function calls, but I think that's an ok trade off for how clean it is.

lalexgap:
Closing in favour of #815

PR #783, opened 2022-07-07
Clean up chainservice events

kerzhner:
This PR introduces:
- `asset` struct for the address + amount field pair.
- Uses the struct in Deposited and AllocationUpdated events.
- Converts a public `CommonEvent` to a private `commonEvent`

### DIFF
--- a/client/engine/chainservice/chainservice.go
+++ b/client/engine/chainservice/chainservice.go
@@ -34,9 +34,8 @@ type asset struct {
 // DepositedEvent is an internal representation of the deposited blockchain event
 type DepositedEvent struct {
 	CommonEvent
-	Asset           common.Address
-	AmountDeposited *big.Int
-	NowHeld         *big.Int
+	asset
+	NowHeld *big.Int
 }
 
 // AllocationUpdated is an internal representation of the AllocatonUpdated blockchain event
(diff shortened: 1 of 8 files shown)


client/engine/chainservice/chainservice.go
```diff
@@ -16,13 +16,13 @@ type Event interface {
 	ChannelID() types.Destination
 }
 
-// CommonEvent declares fields shared by all chain events
-type CommonEvent struct {
```
geoknee, line 20: 👍

client/engine/chainservice/chainservice.go
```diff
@@ -76,33 +76,33 @@ type ChainService interface {
 	SendTransaction(protocols.ChainTransaction)
 }
 
-type ChainServiceBase struct {
+type chainServiceBase struct {
```
geoknee, line 79: 👍

client/engine/chainservice/chainservice.go
```diff
@@ -16,35 +16,46 @@ type Event interface {
...
 
-func (ce CommonEvent) ChannelID() types.Destination {
+func (ce commonEvent) ChannelID() types.Destination {
 	return ce.channelID
 }
 
+type asset struct {
+	AssetAddress common.Address
+	AssetAmount  *big.Int
+}
```
geoknee, line 32: Nit: Elsewhere we have used `asset` as a synonym for `assetAddress`...

Might this be better?
```go
type assetAmount struct {
    Asset common.Address
    Amount *big.Int
}
```
(or even `assetAndAmount`)

Also, I'm unsure about how much cleaner it is to bundle these fields into a struct. I think you've saved most lines you saved were due to using a constructor (which anyway accepts the two fields separately)?

geoknee (approved):

Issue #770, opened 2022-07-05
chainservice: error handling


kerzhner:
The chain service is very panic-happy. On encountering an error, the chain service usually panics. A more nuanced approach is needed. Questions to answer are:
- Should the chainservice just log some types of errors as they are not consequential to the core operations of the wallet?
- Are there any errors that should result in a panic and terminate the entire wallet?
- Should chainservice have an error feed that allows another wallet component to subscribe to and handle errors?

PR #788, opened 2022-07-11
Remove panics from ChainService

kerzhner:
Fixes https://github.com/statechannels/go-nitro/issues/770.

One option is to allow a state channel wallet to continue to operate even if its chain service fails. This allows the wallet to continue to run application channels. This PR enables this mode of operation and includes the following changes:
- Calls to the chainservice APIs (such as initialization and transaction submission) now return errors. The caller can decide how to handle a returned error.
- Errors in event subscription logic are logged.

Note that a wallet should not operate for too long without a working chainservice. The duration of time that a wallet can run without a chainservice (assuming opening new directly-funded channels is not urgent) depends on challenge timeout duration.

### DIFF
--- a/client/engine/chainservice/chainservice.go
+++ b/client/engine/chainservice/chainservice.go
@@ -73,7 +73,7 @@ type ChainService interface {
 	// SubscribeToEvents creates and returs a subscription channel.
 	SubscribeToEvents(types.Address) <-chan Event
 	// SendTransaction is for sending transactions with the chain service
-	SendTransaction(protocols.ChainTransaction)
+	SendTransaction(protocols.ChainTransaction) error
 	// GetConsensusAppAddress returns the address of a deployed ConsensusApp (for ledger channels)
 	GetConsensusAppAddress() types.Address
 }
--- a/client/engine/chainservice/eth_chainservice.go
+++ b/client/engine/chainservice/eth_chainservice.go
@@ -73,21 +73,21 @@ func (ecs *EthChainService) SendTransaction(tx protocols.ChainTransaction) error
 			} else {
 				tokenTransactor, err := Token.NewTokenTransactor(tokenAddress, ecs.chain)
 				if err != nil {
-					panic(err)
+					return err
 				}
 				_, err = tokenTransactor.Approve(ecs.defaultTxOpts(), ecs.naAddress, amount)
 				if err != nil {
-					panic(err)
+					return err
 				}
 			}
 			holdings, err := ecs.na.Holdings(&bind.CallOpts{}, tokenAddress, tx.ChannelId())
 			if err != nil {
-				panic(err)
+				return err
 			}
 
 			ethTx, err := ecs.na.Deposit(txOpts, tokenAddress, tx.ChannelId(), holdings, amount)
 			if err != nil {
-				panic(err)
+				return err
 			}
 			ethTxs = append(ethTxs, ethTx)
 		}
@@ -105,7 +105,7 @@ func (ecs *EthChainService) SendTransaction(tx protocols.ChainTransaction) error
 		}}
 		_, err := ecs.na.ConcludeAndTransferAllAssets(ecs.defaultTxOpts(), nitroFixedPart, nitroSignedVariableParts)
 		if err != nil {
-			panic(err)
+			return err
 		}
 		return nil
 
--- a/client/engine/chainservice/eth_chainservice.go
+++ b/client/engine/chainservice/eth_chainservice.go
@@ -3,6 +3,7 @@ package chainservice
 import (
 	"context"
 	"fmt"
+	"io"
 	"log"
 	"math/big"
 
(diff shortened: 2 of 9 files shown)


client/engine/chainservice/chainservice.go
```diff
@@ -73,7 +73,7 @@ type ChainService interface {
 	// SubscribeToEvents creates and returs a subscription channel.
 	SubscribeToEvents(types.Address) <-chan Event
 	// SendTransaction is for sending transactions with the chain service
-	SendTransaction(protocols.ChainTransaction)
```
geoknee, line 76: The `EthChainService` implementation of this method was previously returning a `[]*ethTypes.Transaction`. Why is this not a type error? Are we allowed to return whatever we like if nothing is specified here?

client/engine/chainservice/mockchain_test.go
```diff
@@ -38,13 +38,20 @@ func TestDeposit(t *testing.T) {
 	testTx := protocols.NewDepositTransaction(types.Destination(common.HexToHash(`4ebd366d014a173765ba1e50f284c179ade31f20441bec41664712aac6cc461d`)), testDeposit)
 
 	// Send one transaction and receive one event from it.
-	chain.SendTransaction(testTx)
+	err = chain.SendTransaction(testTx)
+	if err != nil {
+		t.Error(err)
```
geoknee, line 43: You could use `t.Fatal` here #387.

geoknee (approved):
If I've understood correctly, the behaviour is currently:

> if encountering an error in the chain service, panic and crash the whole wallet and parent process

and you're changing it to:

> if encountering certain errors in the chain service, log them and carry on. if encountering certain other errors, stop executing the current side effect and return an error to the engine. The engine will then bubble up the error to an event handler and ultimately up to the run loop where it can be dealt with globally.*

This seems to me to be a huge improvement.

Nit: I don't think the chain service stops working at all? It just stops trying to finish executing the current side effect (this may cause other side effects declared in the same crank to not be executed). 

---
** currently, this results in a panic anyway, so we are back to the previous behaviour. But it is a TODO to fix:

https://github.com/statechannels/go-nitro/blob/5807f3d1304445eb0179b37cf69ec617f14e66ee/client/engine/engine.go#L143-L148

client/engine/chainservice/mockchain_test.go
kerzhner: Changed in 6e3c47f1.

client/engine/chainservice/eth_chainservice.go
```diff
@@ -100,7 +100,7 @@ func (ecs *EthChainService) SendTransaction(tx protocols.ChainTransaction) error
 
 		_, err := ecs.na.ConcludeAndTransferAllAssets(ecs.defaultTxOpts(), nitroFixedPart, nitroVariablePart, 1, []uint8{0, 0}, nitroSignatures)
 		if err != nil {
-			panic(err)
+			return err
 		}
 		return nil
```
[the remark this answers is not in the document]
kerzhner: Good call, done in fb362d50.

client/engine/engine.go
```diff
@@ -365,11 +365,15 @@ func (e *Engine) executeSideEffects(sideEffects protocols.SideEffects) {
 	}
 	for _, tx := range sideEffects.TransactionsToSubmit {
 		e.logger.Printf("Sending chain transaction for channel %s", tx.ChannelId())
-		e.chain.SendTransaction(tx)
+		err := e.chain.SendTransaction(tx)
+		if err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
kerzhner: Good point. We should probably have a more nuanced approach down the line.

client/engine/chainservice/chainservice.go
kerzhner: At the moment, only `SimulatedBackendChainService` and `MockChainService` are used by the tests and the clients. So `EthChainService` was never used where a `ChainService` interface is required.

`EthChainService` will be used in production environments. So this PR modifies `EthChainService` to adhere to the `ChainService` interface.

kerzhner:
> Nit: I don't think the chain service stops working at all? It just stops trying to finish executing the current side effect (this may cause other side effects declared in the same crank to not be executed).

That's correct!

The cases I am thinking about are: let's say that the ethereum chain service fails to read adjudicator state. This could be an indicator that there is a faulty network link between the statechannel wallet and an ethereum node or between an ethereum node used by the wallet and the rest of the ethereum network. With this PR, instead of trying to resolve this outage in some manner, the state channel wallet will continue to operate. Perhaps the outage resolves itself and the state channel wallet is able to read chain state and submit transactions in some near future. But it could be the case that some component (the state channel wallet, the ethereum node, etc) should be restarted.

client/engine/chainservice/eth_chainservice.go
```diff
@@ -46,7 +46,7 @@ func NewEthChainService(chain ethChain, na *NitroAdjudicator.NitroAdjudicator, n
 
 	go ecs.listenForLogEvents()
 
-	return &ecs
+	return &ecs, nil
```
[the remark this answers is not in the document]
geoknee: Yeah apologies I made this comment on an intermediate commit and then realised it didn't apply. I thought I'd deleted it! Disregard.

PR #789, opened 2022-07-12
Integrate merge conflicts

geoknee:
WIP trying to sort merge conflicts


nksazonov (approved):

client/client.go
```diff
@@ -94,6 +110,10 @@ func (c *Client) CloseVirtualChannel(channelId types.Destination, paidToBob *big
 // CreateDirectChannel creates a directly funded channel with the given counterparty
 func (c *Client) CreateDirectChannel(objectiveRequest directfund.ObjectiveRequest) directfund.ObjectiveResponse {
 
+	// This next line overwrites the requested application address with the consensus app address.
+	// TODO remove this behaviour
```
geoknee, line 114: This is very surprising behaviour, which is bad.

Potential improvements:
1. detect a zero `appAddress` and only in that condition overwrite with the consensus app.
2. change all the tests to pass in a valid (deployed, where necessary) consensus app address 
3. Strip out the `appAddress` from the parameters of this function, so that we don't even ask for it

PR #796, opened 2022-07-13
fix nitro-protocol/package.json so it can be used in imports

willemolding:
**Title** Fix nitro-protocol package.json to allow imports

**Body** 

The `package.json` in the nitro-protocol package directory did not list the files it exports making it impossible to import contracts when added as a dependency via npm.

With the changes it is now possible to use [gitpkg](https://gitpkg.vercel.app/) to add the latest nitro-protocol as a dependency in Solidity projects and import the contracts  (e.g. `import '@statechannels/nitro-protocol/contracts/NitroAdjudicator.sol'`)

Even if the protocol is not ready for release yet it is handy being able to add a dependency this way to experiment and test

### DIFF
--- a/nitro-protocol/package.json
+++ b/nitro-protocol/package.json
@@ -1,8 +1,11 @@
 {
-  "name": "contracts",
-  "version": "1.0.0",
+  "name": "@statechannels/nitro-protocol",
+  "version": "2.0.0-alpha.0",
   "description": "",
   "main": "index.js",
+  "files": [
+    "/contracts/**/*.sol"
+  ],
   "scripts": {
     "contracts:deploy-localhost": "npx hardhat deploy --network localhost --export-all addresses.json && node ./scripts/postdeploy.ts",
     "contracts:node": "npx hardhat node --no-deploy",


nitro-protocol/package.json
```diff
@@ -1,8 +1,11 @@
 {
-  "name": "contracts",
-  "version": "1.0.0",
+  "name": "@statechannels/nitro-protocol",
+  "version": "2.0.0-alpha.0",
   "description": "",
```
geoknee, line 4:
```suggestion
  "version": "2.0.0-alpha.0",
  "description": "Smart contracts and typescript libraries for nitro v2",
```

nitro-protocol/package.json
```diff
@@ -1,8 +1,11 @@
...
   "main": "index.js",
+  "files": [
+    "/contracts/**/*.sol"
+  ],
```
geoknee, line 8: We may wish to include some other stuff here, as well:
* typechain smart contract typings
* `solc` output (i.e. precompiled contracts ready for deployment)
* other typescript library code

We don't need to add those now, though.

geoknee (approved):
Thanks for this contribution @willemolding! Hopefully it unblocks you.

nitro-protocol/package.json
willemolding: This will require deploying to NPM with a deployment script I think unless you want to add the build output to the repo

geoknee: Yes, that's right.

Issue #760, opened 2022-06-24
Test supplied vs generated `signedBy` implementation, extract signature checking logic from DApp to `ForceMove`


nksazonov:
Current Nitro protocol has `signedBy` bitfield supplied alongside with `sigs` and its correctness checks (whether a participant, who claimed in `signedBy` they have signed the state, indeed signed the state) inside DApp.

It would be more wise to verify `signedBy` inside of `ForceMove`, and pass only `signedBy` (and not the raw signatures) to the application. That way we remove the possibility someone forgot to verify signatures and saves gas on calldata.

Another approach would be not to supply `signedBy` field at all and generate it in `ForceMove`. In such case, we would not need to verify it, because it was generated out of supplied signatures. DApp would be supplied with states and `signedBy` like in a previous option.

Both these approaches have their upsides and downsides in speed and gas consumption, so to understand which one is better, we need to implement both of them and compare gas benchmarks.

PR #797, opened 2022-07-14
Move `signedBy` from adjudicator API to `IForceMoveApp`API

geoknee:
Towards #760 .

The approach taken here is to generate `signedBy` on chain.

The other approach we want to consider is to supply `signedBy` and have it verified on chain.


nitro-protocol/contracts/interfaces/INitroTypes.sol
```diff
@@ -21,6 +21,10 @@ interface INitroTypes {
     struct SignedVariablePart {
         VariablePart variablePart;
         Signature[] sigs;
+    }
+
+    struct RecoveredVariablePart {
+        VariablePart variablePart;
         uint256 signedBy; // bitmask
     }
```
geoknee, line 29: This is the change in a nutshell. We still used `SignedVariableParts` off chain, but the adjudicator will transform these into  `RecoveredVariableParts` before passing them to the application. The application can then trust the `signedBy` bitmask and does not need to do any signature recovery.

nitro-protocol/gas-benchmarks/gas.ts
```diff
@@ -58,26 +58,26 @@ export const gasRequiredTo: GasRequiredTo = {
...
     // initially                 ⬛ ->  X  -> 👩
     // challenge + timeout       ⬛ -> (X) -> 👩
     // transferAllAssets         ⬛ --------> 👩
     satp: {
-      challenge: 114_012,
-      transferAllAssets: 110_049,
-      total: 224_061,
+      challenge: 132_381,
+      transferAllAssets: 110_030,
+      total: 242_411,
```
geoknee, line 80: This is probably a good number to compare with the other approach.

PR #798, opened 2022-07-14
Remove SubscribeToEvents from the ChainService interface

kerzhner:
Part of the work needed for https://github.com/statechannels/go-nitro/issues/790.

An implementation of the `ChainService` interface is used by the engine to submit transactions to a chain and receive events from the chain.

Previously, the `ChainService` included a subscription API to accommodate multiple engines using the same `ChainService`. This has been changed to:
- A single engine/client uses a single ChainService instance that is passed to the engine/client constructor. The instance should not be used by multiple engines.
- A `ChainService` has a reference to a shared state chain. In the case of `EthChainService`, the chain is an Ethereum chain (which can be a simulated chain, a testnet, or the mainnet). In the case of `MockChainService`, the chain is a `MockChain` (introduced in this PR).

Since this PR updates the `MockChainService` constructor, [go-nitro-testground](https://github.com/statechannels/go-nitro-testground/blob/0c6d670d05a1aacefc99e67dfbb4ec73e35cc94f/chain/syncer.go#L113) should be updated after this PR is merged.

### DIFF
--- a/client/engine/chainservice/mock_chain.go
+++ b/client/engine/chainservice/mock_chain.go
@@ -1,0 +1,74 @@
+package chainservice
+
+import (
+	"fmt"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/statechannels/go-nitro/client/engine/store/safesync"
+	"github.com/statechannels/go-nitro/protocols"
+	"github.com/statechannels/go-nitro/types"
+)
+
+// MockChain is an in-memory "chain" that accepts transactions and broadcasts events
+type MockChain interface {
+	SubmitTransaction(protocols.ChainTransaction) error // unlike an ethereum blockchain, Mockhain accepts go-nitro protocols.ChainTransaction
+	SubscribeToEvents(a types.Address) <-chan Event     // returns a channel that produces all chain Events
+}
+
+// MockChainImpl mimicks the Ethereum blockchain by keeping track of block numbers and account balances.
+type MockChainImpl struct {
+	blockNum uint64
+	// holdings tracks funds for each channel.
+	holdings map[types.Destination]types.Funds
+	// out maps addresses to an Event channel. Given that MockChainServices only subscribe
+	// (and never unsubscribe) to events, this can be converted to a list.
+	out safesync.Map[chan Event]
+}
+
+// NewMockChainImpl creates a new MockChainImpl
+func NewMockChainImpl() *MockChainImpl {
+	chain := MockChainImpl{}
+	chain.blockNum = 1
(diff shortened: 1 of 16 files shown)


client/engine/chainservice/mock_chain.go
```diff
@@ -0,0 +1,74 @@
...
+
+// SubmitTransaction updates internal state and brodcasts events
+func (mc *MockChainImpl) SubmitTransaction(tx protocols.ChainTransaction) error {
+	mc.blockNum++
+	switch tx := tx.(type) {
+	case protocols.DepositTransaction:
+		if tx.Deposit.IsNonZero() {
+			mc.holdings[tx.ChannelId()] = mc.holdings[tx.ChannelId()].Add(tx.Deposit)
+		}
+		for address, amount := range tx.Deposit {
```
andrewgordstewart, line 45: Is `address` a token address or participant address? I think it's a token address -- renaming it to `assetAddress` might be helpful.

In either case, it's surprising to me that one deposit could generate multiple events. Is this consistent with `nitro-protocol`?

client/engine/chainservice/mock_chain.go
```diff
@@ -0,0 +1,74 @@
...
+	"fmt"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/statechannels/go-nitro/client/engine/store/safesync"
+	"github.com/statechannels/go-nitro/protocols"
+	"github.com/statechannels/go-nitro/types"
+)
+
+// MockChain is an in-memory "chain" that accepts transactions and broadcasts events
+type MockChain interface {
```
andrewgordstewart, line 13: nit: is this interface necessary? It seems like there'd only ever be one implementation of it -- would it be cleaner to delete this interface and rename `MockChainImp` to `MockChain`?

andrewgordstewart (approved):
LGTM

client/engine/chainservice/mock_chain.go
kerzhner: Good call, done in 37d7c771.

client/engine/chainservice/mock_chain.go
kerzhner: `address` is renamed to `assetAddress` in d744a989.

My understanding is: for a channel that contains `n` token types, `n` deposit transactions are submitted to the blockchain. This will result in `n` chain events. This code mimics that behavior. Do you think I am misinterpreting nitro-protocol?

andrewgordstewart: ah yes, you're totally correct!

PR #800, opened 2022-07-14
implement Payment/Receipt managers

andrewgordstewart:
This includes a minimal implementation of a payment and receipt manager, making the assumption that some other component is responsible for sending vouchers across the wire.

A voucher is a simple struct containing a channel ID and an amount. This voucher is signed by the sender in a virtual channel, and sent to the receiver in the channel.

If the sender or hub do not cooperatively close the virtual channel, any voucher signed by the sender can be redeemed by the receiver through a forced transition on-chain, pending application rules implementation.

A go-nitro client can now initialize a payment and/or receipt manager in order to implement a payments API.


payments/payment_manager.go
```diff
@@ -0,0 +1,95 @@
...
+	Balance struct {
+		Remaining *big.Int
+		Paid      *big.Int
+	}
+
+	// PaymentManager can be used to make a payment for a given channel, issuing a new, signed voucher to be sent to the receiver
+	PaymentManager struct {
+		signer   common.Address
+		channels map[types.Destination]*Balance
+	}
```
geoknee, line 22: Could this struct be private (non exported)? Perhaps an interface should be exported, instead of a concrete type.

payments/vouchers.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+type (
+	// A Voucher signed by Alice can be used by Bob to redeem payments in case of
+	// a misbehaving Alice.
+	//
+	// During normal operation, Alice & Bob would terminate the channel with an
+	// outcome reflecting the largest amount signed by Alice. For instance,
+	// - if the channel started with balances {alice: 100, bob: 0}
+	// - and the biggest voucher signed by alice had amount = 20
```
geoknee, line 22: So vouchers are "replace by incentive", right? As in, Bob can only redeem one voucher per channel?

payments/receipt_manager.go
```diff
@@ -0,0 +1,91 @@
...
+		channels map[types.Destination]*PaymentStatus
+	}
+)
+
+func NewReceiptManager() *ReceiptManager {
+	channels := make(map[types.Destination]*PaymentStatus)
+	return &ReceiptManager{channels}
+}
+
+// Register registers a channel for use, given the
```
geoknee, line 30: godoc trails off

payments/payment_manager.go
```diff
@@ -0,0 +1,95 @@
...
+
+// Balance returns the balance of the channel
+func (pm *PaymentManager) Balance(channelId types.Destination) (Balance, error) {
+	balance := Balance{&big.Int{}, &big.Int{}}
+	stored, ok := pm.channels[channelId]
+	if !ok {
+		return balance, fmt.Errorf("channel not found")
+	}
+	balance.Paid.Set(stored.Paid)
+	balance.Remaining.Set(stored.Remaining)
```
geoknee, line 92: Nit: is this slightly better? 
```suggestion
	stored, ok := pm.channels[channelId]
	if !ok {
		return balance, fmt.Errorf("channel not found")
	}
	return Balance(big.NewInt(0).Set(stored.Paid),big.NewInt(0).Set(stored.Remaining)
```

payments/receipt_manager.go
```diff
@@ -0,0 +1,91 @@
...
+// Remove deletes the channel's status
+func (pm *ReceiptManager) Remove(channelId types.Destination) {
+	delete(pm.channels, channelId)
+}
+
+// Receive validates the incoming voucher, and returns the total amount received so far
+func (rm *ReceiptManager) Receive(voucher Voucher) (*big.Int, error) {
+	status, ok := rm.channels[voucher.channelId]
+	if !ok {
+		return &big.Int{}, fmt.Errorf("channel not registered")
```
geoknee, line 54: What if the channel was actually not yet / np longer open or not yet / no longer funded? Shouldn't we return an error in that case?

payments/payment_manager_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+	"github.com/statechannels/go-nitro/internal/testactors"
+	. "github.com/statechannels/go-nitro/internal/testhelpers"
+	"github.com/statechannels/go-nitro/types"
+)
+
+// manager lets us implement a getBalancer helper to make test assertions a little neater
+type manager interface {
+	Balance(chanId types.Destination) (Balance, error)
+}
```
geoknee, line 15: Could be worth extracting this into a (non-test) interface (see other comment).

payments/payment_manager_test.go
```diff
@@ -0,0 +1,130 @@
+package payments
```
geoknee, line 1: Could the name of this file reflect that both payment and receipt managers are being covered?

payments/payment_manager_test.go
```diff
@@ -0,0 +1,130 @@
...
+	received, err = receiptMgr.Receive(firstVoucher)
+	Ok(t, err)
+	Equals(t, doublePayment, received)
+	Equals(t, twoPaymentsMade, getBalance(receiptMgr))
+
+	// Only the signer can sign vouchers
+	_, err = paymentMgr.Pay(channelId, triplePayment, testactors.Bob.PrivateKey)
+	Assert(t, err != nil, "only Alice can sign vouchers")
+
+	// Receiving a voucher for an unknown channel fails
```
geoknee, line 114: It seems like this will lead to race conditions. Quite likely that I receive a voucher before I register the channel?

geoknee (approved):
Thanks @andrewgordstewart! 

Overall I'm a little uneasy about the loose coupling between the `go-nitro` store and these managers. It seems a bit awkward to me to have the consuming application orchestrate between them -- for example telling the `PaymentManager` about the existence or funding status of a channel (which the existing `client` has in its store), and then later telling the `client` about the largest voucher sent or received. I feel like it might be nicer to have that orchestration work done _inside_ the `client` and not having API methods such as `Register` and `Remove`. 

That said, I think it is worth merging this code so that we can experiment with integrating it into the main `client` and then into a test script. I've left a bunch of comments / questions.

andrewgordstewart:
> Thanks @andrewgordstewart!
> 
> Overall I'm a little uneasy about the loose coupling between the `go-nitro` store and these managers. It seems a bit awkward to me to have the consuming application orchestrate between them -- for example telling the `PaymentManager` about the existence or funding status of a channel (which the existing `client` has in its store), and then later telling the `client` about the largest voucher sent or received. I feel like it might be nicer to have that orchestration work done _inside_ the `client` and not having API methods such as `Register` and `Remove`.
> 
> That said, I think it is worth merging this code so that we can experiment with integrating it into the main `client` and then into a test script. I've left a bunch of comments / questions.

A `PaymentManager` is simply the logic that manages vouchers. It can be used by a client, so all of the necessary  orchestration can still be done by the client! (Including, for instance, sending vouchers across the wire.)

payments/payment_manager.go
andrewgordstewart: What's the benefit of exporting an interface over a concrete type?

payments/vouchers.go
andrewgordstewart: That's correct

payments/payment_manager.go
andrewgordstewart: I personally find `Balance(big.NewInt(0).Set(stored.Paid),big.NewInt(0).Set(stored.Remaining)` to be hard to read, but that's a personal choice I think?

payments/receipt_manager.go
andrewgordstewart: Would all those cases be covered by "this channel is not registered"?

My mental model is: a channel is either in a state where it can receive vouchers, or it is not in that state. It might not be in that state for many reasons (it's a made up channel, it's being de-funded, etc.). But the receipt manager doesn't care about that -- it only cares about managing vouchers.

payments/payment_manager_test.go
andrewgordstewart: That's a good point. How would you suggest handling that case?

One idea is to optimistically store vouchers even for unregistered channels. I prefer to avoid that, since that gives someone else the permission to fill up my store with junk vouchers.

payments/payment_manager_test.go
andrewgordstewart: I'm not sure how useful that would be; if it becomes useful, could we export it as a public interface then?

I understood the (a?) rule of thumb in go is to define interfaces where types are _consumed_, not where types are _defined_. So, someone who wants to use either a PaymentManager or a ReceiptManager, just to call the `Balance` function, would define a `Manager` interface.

payments/payment_manager_test.go
andrewgordstewart: dd4c4721

payments/receipt_manager.go
andrewgordstewart: 9cd90718

payments/payment_manager.go
geoknee: My thinking was that not exporting the concrete type forces consumers to use the constructor, rather than doing things manually. Nice to have in general, but perhaps not that important here. 

If we were not exporting the concrete type, it might be useful for there to be an interface exported, so that consumers know how to interact with e.g. a `PaymentManager`. Probably also not super important right now.

payments/payment_manager.go
geoknee: I agree it is hard to read. The other advantage of my suggestion is that we aren't constructing a `Balance` until we know that the check on `pm.channels` is ok. Not a huge deal, though.

payments/receipt_manager.go
geoknee: > Would all those cases be covered by "this channel is not registered"?

I think in principle, yes. I'm just wondering about keeping the register of channels in sync with the client's store. But your response to my other question addresses that.

geoknee:
> A `PaymentManager` is simply the logic that manages vouchers. It can be used by a client, so all of the necessary orchestration can still be done by the client! (Including, for instance, sending vouchers across the wire.)

I see. We would still end up with some state replication in that case, and the `client.store` and `manager.channels` may get out of sync -- thinking about how we have designed the client to be "restartable by default" (we sometimes call this crash tolerance), this doesn't obviously fit into that model in my mind (yet). 

In the case of a client restart, we may lose our payment/receipt manager's state (list of registered channels and their balances etc). So, if recovering from that is important to us, we would either need a mechanism to rebuild thats state, or to integrate the payment/receipt manager's more tightly into the client so they can read and write to its store as the source of truth and benefit from the existing crash tolerance properties?

payments/payment_manager_test.go
geoknee: I think we can avoid race conditions by ensuring that we register the channel before we send the prefund signature. If clients _won't send a voucher before they have received the prefund signature,_ we shouldn't get a race condition. If clients depart from this behaviour, I think it is OK for their voucher to be dropped. Someone else can still spam me with many prefund setup states / channel proposals -- but that is an existing problem which we likely need to solve independently.

payments/payment_manager.go
andrewgordstewart: What you're saying makes sense, but I understood that to be an anti-pattern in the go ecosystem -- that we are encouraged to use interfaces where types are _used_, not where types are _defined_.  That said, I'm struggling to find a reference/reason for **not** defining interfaces where we define types.

It's easy to see why we'd be encouraged to use interfaces to define what we need out of a function argument. It also seems commonplace (and fine) to allow people to construct structs explicitly, especially when the zero value is useful! (Unfortunately, the zero value is not useful here.)

@NiloCK do you remember what I'm thinking of? Unless I can be convinced otherwise, I'm happy to export an interface.

andrewgordstewart:
> > A `PaymentManager` is simply the logic that manages vouchers. It can be used by a client, so all of the necessary orchestration can still be done by the client! (Including, for instance, sending vouchers across the wire.)
> 
> I see. We would still end up with some state replication in that case, and the `client.store` and `manager.channels` may get out of sync -- thinking about how we have designed the client to be "restartable by default" (we sometimes call this crash tolerance), this doesn't obviously fit into that model in my mind (yet).
> 
> In the case of a client restart, we may lose our payment/receipt manager's state (list of registered channels and their balances etc). So, if recovering from that is important to us, we would either need a mechanism to rebuild thats state, or to integrate the payment/receipt manager's more tightly into the client so they can read and write to its store as the source of truth and benefit from the existing crash tolerance properties?

So presumably, down the road, a payment or receipt manager will persist their state in some DB. I would expect this to happen by allowing a Payment/Receipt manager to accept some kind of `PaymentStore`/`ReceiptStore` interface. This would give users the flexibility of providing their own store implementations: since payments are on the critical path of state channel **users**, the payments API can be very sensitive to performance, being able to optimize (or ignore!) storage calls might be very important!

payments/payment_manager.go
andrewgordstewart: a5bc911c avoids constructing a `Balance` early

payments/payment_manager.go
andrewgordstewart: 73deee843ade73762c5090c9df9e550b1bfc589e

PR #801, opened 2022-07-18
upgrade golangci-lint

geoknee:
To upgrade locally, use

brew upgrade golangci-lint


geoknee:
I'm seeing some linting errors on #797 that I can't reproduce locally. Perhaps upgrading will help.

nksazonov (approved):

PR #802, opened 2022-07-18
Automate gas benchmarks

### LINKED ISSUE #546: Automate updating gas benchmark results

nksazonov:
Closes #546 .


nksazonov:
There are two points need clarifying:
1. Do we want gas benchmarks to remain as a **test** or to do it as a **script**?
2. **Where should gas benchmark results be written?** Currently they are written into `gasResults.json` file, but this creates some inconvenience when trying to understand obtained results, as to do this the one has to be reading `gas.ts` simultaneously to see the comments deciphering each of gas benchmarks properties.

What do you think, @geoknee ?

Another thing that has to be included in this PR is a change to GH action, which now should be comparing the file with new gas benchmarks results with the old ones.

geoknee:
Did you mean to change `package-lock.json`?

nksazonov:
> Did you mean to change `package-lock.json`?

I just reinstalled `npm modules` to prevent any issues and decided to make a separate commit regardless updated `package-lock`.

geoknee:
> 1. Do we want gas benchmarks to remain as a **test** or to do it as a **script**?

I think a script would probably be fine. I think it changes the mental model a bit, but this is fine --especially if we explain what is going on with a readme. The important features as I see them:
a. we can see "at a glance" what the gas consumption of our system is. This means the numbers are documented (what do they mean), and checked into source code
b. we make it impossible for PRs to be merged which change the implementation and forget to update the gas numbers

I think (you can confirm) that your solution here preserves those features? There is also 

c. we get a nice output showing us how changes in the implementation propagate to gas consumption

Maybe we have lost this? Previously we got a color-coded, percentage increase printed out to screen. This is less important. 



> 2. **Where should gas benchmark results be written?** Currently they are written into `gasResults.json` file, but this creates some inconvenience when trying to understand obtained results, as to do this the one has to be reading `gas.ts` simultaneously to see the comments deciphering each of gas benchmarks properties.
> 
> What do you think, @geoknee ?
> 
> Another thing that has to be included in this PR is a change to GH action, which now should be comparing the file with new gas benchmarks results with the old ones.

I would recommend putting the "documentation" for the numbers (currently in `gas.ts`) in the (test) script itself, and then delete `gas.ts`. I guess we can't put comments in `json` so this is the only real option.

nksazonov:
> I think (you can confirm) that your solution here preserves those features?

Yes, it surely does.

> c. we get a nice output showing us how changes in the implementation propagate to gas consumption
>
> Maybe we have lost this? Previously we got a color-coded, percentage increase printed out to screen. This is less important.

Yes, unfortunately, it is not possible now, although it was pretty convenient to have all the changes color-coded in the terminal. I think, it is worth keeping this feature. Maybe we can do the following:
- keep the logic in `benchmark.test.ts` as it was, but let it take the figures from `gasResults.json`.
- introduce a script called `exportBenchmarkResults.ts`, which will update `gasResults.json`.

Having done that we separate _checking_ benchmark changes (color-coded ones) from actually _updating_ them (which surely will make it impossible to see benchmark diff). The logical continuation of this work will be to create the following `npm` scripts:
- `benchmark:diff` - the same as `test:gas` now
- `benchmark:update` - runs `exportBenchmarkResults.ts`, updating `gasResults.json` and making it impossible to the the diff
- `benchmark` - runs both in the aforementioned order

geoknee:
Interesting proposal. I think for now I would rather just lose the colour coding. We can put that back in later if we really want it. That's because this work is only to improve DevEx, not part of our core roadmap. And I think we will have a sweeter experience already with these changes, even if we have lost a feature.

---

When we look to reinstate that feature, here's some other ideas: we could have the comparison between old and new done in a github action / write a small script to compare the two `json` files on two different git branches?

geoknee (approved):
Thanks for this work @nksazonov!

It seems to work pretty well and definitely solves the problem we wanted to solve. I pushed a couple of commits tidying things up.

However, I think one downside is now we have quite a bit of duplication:
* `gas.ts` contains similar information to `gasResults.json`, only _with_ comments and _without_ meaningful numbers. There is potential for confusion here. Could you simply parse the existing `gasResults.json` instead of using an `emptyGasResults`? You may not even need to zero everything out, then (but you could).
*  `exportBenchmarkResults.ts` duplicates the comments from `gas.ts`, and is also duplicated with respect to `benchmark.test.ts`. 


When we want to add a new benchmark, it seems as if there are now quite a lot of files to update.

One way to improve this is to drop the `benchmark:diff` side of things. I imagine we won't use it so much -- and even in CI we can simply run `npm run benchmark:update` and let `git diff --exit-code` do the "check" on the gas numbers. What do you think?

If you like we can merge this as is, and return to the issue if/when we run into it again?

nksazonov:
You are right, we have quite a bit of duplication. Actually, the reason for both points you describe is to keep color-coded comparison results, which I believe is a powerful feature as it is very convenient to see the difference already in console, not needing to do math in your head.

It is also worth mentioning that `exportBenchmarkResults.ts` takes zeroed `gas.ts`, but not `gasResults.json`, as the latest may not exist.

So to keep the color-coded diff feature, we definitely need to keep `benchmark.test.ts` (otherwise we would need to explicitly print colorful messages to console ourselves). Thus, I can see several options how we can remove the duplication:

1. Create a static `Benchmark` object, which will describe both benchmark result structure interface and benchmark method to invoke for that result.

``` javascript
{
    title: 'directlyFundAChannelWithETHSecond.satp',
    setup: async () => {
        const setupTX = nitroAdjudicator.deposit(MAGIC_ADDRESS_INDICATING_ETH, X.channelId, 0, 5, {
            value: 5,
        });
        await (await setupTX).wait();
    },
    method: async () => {
        await nitroAdjudicator.deposit(MAGIC_ADDRESS_INDICATING_ETH, X.channelId, 5, 5, {value: 5})
    }
}
``` 
Both test and script files will just traverse through all the field, invoking the `setup`, benchmarking the `method` and either testing for equality with or saving the results to `title`.

This way, the only place we would need to make a change to (if a new benchmark is introduced / old changed) is this `Benchmark` object.

2. Give up the script and proceed only with the test. Also introduce a flag (`--write=false/true`), which will control whether the test is updating gas results after it runs or not. To remove zeroed `gas.ts`, we could introduce a function to create an empty gas results object, which will be changed during benchmark test with `--write=true`.

Which do you think is the best suitable option for us, @geoknee?

Also indeed we can save this conversation for the work in future and merge this PR as is.

geoknee:
Thanks for all these suggestions. I think it is best to merge this PR as is, and return to this problem some time in the future.

PR #804, opened 2022-07-19
`client.CreateDirectChannel` does not accept an app definition and is not called `client.CreateLedgerChannel`

### LINKED ISSUE #791: Do not silently overwrite `appAddress` in `client.CreateDirectChannel`  API method
This is very surprising behaviour, which is bad.

Potential improvements:
1. detect a zero `appAddress` and only in that condition overwrite with the consensus app.
2. change all the tests to pass in a valid (deployed, where necessary) consensus app address 
3. Strip out the `appAddress` from the parameters of this function, so that we don't even ask for it

_Originally posted by @geoknee in https://github.com/statechannels/go-nitro/pull/789#discussion_r919058368_

geoknee:
Closes #791

### DIFF
--- a/protocols/directfund/directfund.go
+++ b/protocols/directfund/directfund.go
@@ -406,11 +406,10 @@ func IsDirectFundObjective(id protocols.ObjectiveId) bool {
 	return strings.HasPrefix(string(id), ObjectivePrefix)
 }
 
-// ObjectiveRequestWithoutAppDefinition represents a request to create a new direct funding objective.
+// ObjectiveRequestForConsensusApp represents a request to create a new direct funding objective.
 // There is no AppDefinition, since we currently only support full consensus rules for direct channels.
-type ObjectiveRequestWithoutAppDefinition struct {
+type ObjectiveRequestForConsensusApp struct {
 	CounterParty      types.Address
-	AppData           types.Bytes
 	ChallengeDuration *types.Uint256
 	Outcome           outcome.Exit
 	Nonce             int64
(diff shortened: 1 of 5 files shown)


protocols/directfund/directfund.go
```diff
@@ -404,18 +404,25 @@ func (o *Objective) clone() Objective {
...
+
 }
 
 // ObjectiveRequest represents a request to create a new direct funding objective.
-type ObjectiveRequest struct {
+// There is no AppDefinition, since we currently only support full consensus rules for direct channels.
+type ObjectiveRequestWithoutAppDefinition struct {
 	CounterParty      types.Address
-	AppDefinition     types.Address
 	AppData           types.Bytes
```
NiloCK, line 414: Should a `NullApp` request contain an `AppData` field? May be better to include `AppData` alongside `AppDefinition` in `ObjectiveRequest`.

protocols/directfund/directfund.go
```diff
@@ -404,18 +404,25 @@ func (o *Objective) clone() Objective {
 // IsDirectFundObjective inspects a objective id and returns true if the objective id is for a direct fund objective.
 func IsDirectFundObjective(id protocols.ObjectiveId) bool {
 	return strings.HasPrefix(string(id), ObjectivePrefix)
+
```
NiloCK, line 407:
```suggestion
```

NiloCK (approved):
This succeeds in removing a potentially confusing gotcha from the client API :+1:. The result remains a little awkward because this protocol retains a very general name (directfund) but in practice does a more specific thing (creates ledger channels).

Maybe worth considering what we prefer on the client API - is `client.CreateLedgerChannel` more communicative of what the function does?

protocols/directfund/directfund.go
geoknee: Resolved in [4909c90](https://github.com/statechannels/go-nitro/pull/804/commits/4909c9011f9480be1754ebf20634acd1fbd66230)

geoknee:
> Maybe worth considering what we prefer on the client API - is `client.CreateLedgerChannel` more communicative of what the function does?

I think it is. I have made this change in [0077bb0](https://github.com/statechannels/go-nitro/pull/804/commits/0077bb0c897848ccc5a790ac6a2f4bf0091146d9)

PR #805, opened 2022-07-19
Run prettier on `.sol` files in CI. 

### LINKED ISSUE #578: Run prettier check on contracts during CI
If our linting was configured well, `npm run prettier:write` should not change any files. 

But it does!

I think the issue is that we run prettier checks via an eslint plugin, but only on files in the `src` directory. Therefore contracts are excluded.

geoknee:
Closes #578.


nitro-protocol/gas-benchmarks/gas.ts
```diff
@@ -28,7 +28,7 @@ type Path =
 export const gasRequiredTo: GasRequiredTo = {
   deployInfrastructureContracts: {
     satp: {
-      NitroAdjudicator: 3_577_883, // Singleton
+      NitroAdjudicator: 3_577_871, // Singleton
```
geoknee, line 31: Interesting that running prettier would reduce gas consumption 🤔 ...

NiloCK (approved):
:+1:

nitro-protocol/gas-benchmarks/gas.ts
geoknee: I believe this is explained by two facts:

1. That the bytecode deployed to chain includes a hash of the contract metadata, which in turn depends on the hash of the source code file itself (including filename and path, and presumably without any "normalization" being performed). So changing the source code in anyway, including semantically unimportant changes, may change the hash and therefore _the amount of nonzero bytes_ deployed to chain. 
2. The gas payable depends (in part) on the amount of nonzero bytes sent in the transaction.

geoknee: Sources:

https://docs.soliditylang.org/en/v0.8.13/metadata.html#encoding-of-the-metadata-hash-in-the-bytecode

https://ethereum.github.io/yellowpaper/paper.pdf

PR #810, opened 2022-07-22
Update `(re)claim` to follow SATP spec

### LINKED ISSUE #777: Update contracts to understand nitro-v2 guarantees
This is a case where the off chain code:

https://github.com/statechannels/go-nitro/blob/1ca62a9207dc07c05a3da453c2ed5c508c2cfe37/channel/state/outcome/guarantee.go#L8-L14

is ahead of the on chain code:

https://github.com/statechannels/go-nitro/blob/227c86117a2edeecbd7e1f39d0064267753a0cec/nitro-protocol/contracts/MultiAssetHolder.sol#L646-L648

1. On chain, a guarantee needs to change from being an array of destinations to being a struct with a left destination and right destination. 
2. We need to rewrite `claim` on chain to handle the new data type. 
3. We may like to alter the behaviour so that `claim` **only** changes outcomes on chain and doesn't actually pay anything out. This would be closer to how it is described in https://www.overleaf.com/project/618a7979d3893c93c6f7a931 . There may be higher gas costs overall (for exiting a virtual channel down the sad path) but it may reduce the chance of having bugs / attacks.

geoknee:
- Rename `claim` => `reclaim`
- Guarantees are encoded as an explicit struct with `left` and `right` members, not just an array
- There are no  longer any "interactions" with external contracts or payouts of any kind when calling `reclaim`
- There is **only** an update to the ledger channel, which depends on the outcome of a target channel registered on chain.
- This means the attack surface has gotten smaller. There is now only one way to get money out of the adjudicator (`transfer`). Reclaim just rewrites some outcomes on chain. 
- Correspondingly, there will be a higher overall gas cost to recover money from a virtual channel. Peers may need to `reclaim` and _then_ `transfer` to fully liquidate their funds. 
- There are _no_ updates to the target channel, either. This is because more guarantees may need to be reclaimed. There is potential for some gas savings by deleting the storage on chain for the target channel, but it is not obvious how to do this safely, i.e. how to safely detect  when there are no more guarantees pointing at the target channel. 

Closes #777 .

TODO:
- [x] decide which events should be emitted


nitro-protocol/contracts/MultiAssetHolder.sol
```diff
@@ -358,128 +328,74 @@ contract MultiAssetHolder is IMultiAssetHolder, StatusManager {
...
+
+        require(foundTarget, 'could not find target');
+        require(foundLeft, 'could not find left');
+        require(foundRight, 'could not find right');
+
+        return newSourceAllocations;
     }
 
     /**
      * @dev Applies precomputed side effects for claim. Updates the holdings of the source channel. Updates the fingerprint of the outcome for the source and the target channel. Emits an event for each channel.
```
nksazonov, line 387:
```suggestion
     * @dev Updates the fingerprint of the outcome for the source channel and emit an event for it.
```

nitro-protocol/contracts/MultiAssetHolder.sol
```diff
@@ -488,36 +404,10 @@ contract MultiAssetHolder is IMultiAssetHolder, StatusManager {
...
-        targetOutcome[targetAssetIndex].allocations = newTargetAllocations;
-        _updateFingerprint(
-            targetChannelId,
-            claimArgs.targetStateHash,
-            keccak256(abi.encode(targetOutcome))
-        );
-
         // emit the information needed to compute the new source outcome stored in the fingerprint
-        emit AllocationUpdated(sourceChannelId, sourceAssetIndex, initialHoldings);
+        // emit AllocationUpdated(sourceChannelId, sourceAssetIndex); // TODO what do we really need here?
```
nksazonov, line 408: Maybe we can add a separate event `Reclaimed` and use here for an intermediary to listen to.

nitro-protocol/contracts/MultiAssetHolder.sol
```diff
@@ -643,7 +533,13 @@ contract MultiAssetHolder is IMultiAssetHolder, StatusManager {
         return a > b ? b : a;
     }
 
-    function decodeGuaranteeData(bytes memory data) internal pure returns (bytes32[] memory) {
-        return abi.decode(data, (bytes32[]));
+    struct Guarantee {
+        // TODO absorb into exit format
```
nksazonov, line 537: Do we want to do some specific work prior to adding this into exit format?

nitro-protocol/contracts/MultiAssetHolder.sol
```diff
@@ -358,128 +328,74 @@ contract MultiAssetHolder is IMultiAssetHolder, StatusManager {
...
-            Outcome.Allocation[] memory newTargetAllocations,
-            Outcome.Allocation[] memory exitAllocations,
-            uint256 totalPayouts
-        )
-    {
-        // `targetAllocationIndicesToPayout == []` means "pay out to all"
+        uint256 indexOfTargetInSource
+    ) public pure returns (Outcome.Allocation[] memory) {
+        Outcome.Allocation[] memory newSourceAllocations = new Outcome.Allocation[](
+            sourceAllocations.length - 1
```
nksazonov, line 339:
```suggestion
            sourceAllocations.length - 1 // is one slot shorter as we remove the guarantee
```

nitro-protocol/src/contract/multi-asset-holder.ts
```diff
@@ -64,126 +64,65 @@ export function convertAddressToBytes32(address: string): string {
...
-  newSourceAllocations: ExitFormat.Allocation[];
-  newTargetAllocations: ExitFormat.Allocation[];
-  exitAllocations: ExitFormat.Allocation[];
-  totalPayouts: string;
-} {
-  let totalPayouts = BigNumber.from(0);
-  let k = 0;
+  indexOfTargetInSource: number
+): ExitFormat.Allocation[] {
+  const newSourceAllocations: ExitFormat.Allocation[] = []; // will be one slot shorter than sourceAllocations
```
nksazonov, line 72: It is a useful comment. I have copied it to `MultiAssetHolder.compute_reclaim_effects`.

nitro-protocol/test/contracts/MultiAssetHolder/computeReclaimEffects.test.ts
```diff
@@ -122,51 +93,32 @@ const testcase1: TestCase = {
...
       },
     ],
-    totalPayouts: '0x07',
   },
 };
 
 const testCases: TestCase[][] = [[testcase1]];
 
-describe('computeClaimEffectsAndInteractions', () => {
+describe('computeReClaimEffects', () => {
```
nksazonov, line 101:
```suggestion
describe('computeReclaimEffects', () => {
```

nitro-protocol/test/contracts/MultiAssetHolder/reclaim.test.ts
```diff
@@ -0,0 +1,162 @@
...
+
+const testNitroAdjudicator: TESTNitroAdjudicator & Contract = setupContract(
+  provider,
+  TESTNitroAdjudicatorArtifact,
+  process.env.TEST_NITRO_ADJUDICATOR_ADDRESS
+) as unknown as TESTNitroAdjudicator & Contract;
+
+// Amounts are valueString representations of wei
+describe('reclaim', () => {
+  it('handles a simpe case as expected', async () => {
```
nksazonov, line 27:
```suggestion
  it('handles a simple case as expected', async () => {
```

nitro-protocol/test/contracts/MultiAssetHolder/reclaim.test.ts
```diff
@@ -0,0 +1,162 @@
...
+const provider = getTestProvider();
+
+const testNitroAdjudicator: TESTNitroAdjudicator & Contract = setupContract(
+  provider,
+  TESTNitroAdjudicatorArtifact,
+  process.env.TEST_NITRO_ADJUDICATOR_ADDRESS
+) as unknown as TESTNitroAdjudicator & Contract;
+
+// Amounts are valueString representations of wei
+describe('reclaim', () => {
```
nksazonov, line 26:
```suggestion
describe('reclaim', () => {
// TODO: add a test case to show off a multihop reclaim, where we have Alice, Irene, Ivan and Bob. 
```

nksazonov (approved):

nitro-protocol/contracts/interfaces/IMultiAssetHolder.sol
```diff
@@ -60,15 +60,14 @@ interface IMultiAssetHolder {
         bytes32 targetStateHash;
         bytes targetOutcomeBytes;
         uint256 targetAssetIndex;
-        uint256[] targetAllocationIndicesToPayout;
```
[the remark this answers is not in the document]
geoknee: Good spot: [30fbcfd](https://github.com/statechannels/go-nitro/pull/810/commits/30fbcfd4bad37c34755adeb17722a667a1e00eaf)

nitro-protocol/contracts/MultiAssetHolder.sol
```diff
@@ -358,128 +328,74 @@ contract MultiAssetHolder is IMultiAssetHolder, StatusManager {
...
-            );
-            sourceSurplus -= affordsForDestination;
-        }
+        require(
+            guarantee.allocationType == uint8(Outcome.AllocationType.guarantee),
+            'not a guarantee'
+        );
 
-        uint256 targetSurplus = min(sourceSurplus, sourceAllocations[indexOfTargetInSource].amount);
+        Guarantee memory guaranteeData = decodeGuaranteeData(guarantee.metadata); // This is [left, right] TODO make this explicit using a struct
```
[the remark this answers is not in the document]
geoknee: Yes! [8adc2b1](https://github.com/statechannels/go-nitro/pull/810/commits/8adc2b142b1b7c112e93ad344eb2a72b6484eccf)

nitro-protocol/contracts/MultiAssetHolder.sol
```diff
@@ -358,128 +328,74 @@ contract MultiAssetHolder is IMultiAssetHolder, StatusManager {
...
-            uint256 affordsForDestination = min(
-                sourceAllocations[sourceAllocationIndex].amount,
-                sourceSurplus
-            );
-            sourceSurplus -= affordsForDestination;
-        }
+        require(
+            guarantee.allocationType == uint8(Outcome.AllocationType.guarantee),
+            'not a guarantee'
+        );
```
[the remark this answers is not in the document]
geoknee: We don't need to do the check twice [8adc2b1](https://github.com/statechannels/go-nitro/pull/810/commits/8adc2b142b1b7c112e93ad344eb2a72b6484eccf)

nitro-protocol/contracts/MultiAssetHolder.sol
geoknee: Thinking about this some more, I am not sure it belongs in the exit format. So I removed the TODO [b3045f3](https://github.com/statechannels/go-nitro/pull/810/commits/b3045f3a4dbfbc31b3bed8780938955e83b99597).

nksazonov: Actually, you are right. Guarantee concept is more nitro-related.

nitro-protocol/contracts/MultiAssetHolder.sol
geoknee: Good idea [38c1ccb](https://github.com/statechannels/go-nitro/pull/810/commits/38c1ccb2a49d63a7b128b0d024b75b4421f762ed). I think it is important for both parties in that ledger channel to listen to, because they won't be able to call claim a second time without computing the change in outcome themselves off chain (it is stored as a hash).

PR #814, opened 2022-07-27
Add check for `isFinal` during `conclude`.

### LINKED ISSUE #813: BUG: no `isFinal` check during conclude.
The [_conclude  function in ForceMove](https://github.com/statechannels/go-nitro/blob/d9bc78c4aeceb7fa221d657f8be1d0f5e5a29ef2/nitro-protocol/contracts/ForceMove.sol#L133) specifies in the docstring that all of the provided states in the support proof must be final (which makes perfect sense) but there are no checks to enforce this?

---
originally posted by @willemolding in slack

geoknee:
Closes #813


nitro-protocol/contracts/ForceMove.sol
```diff
@@ -139,6 +139,7 @@ contract ForceMove is IForceMove, StatusManager {
 
         // checks
         _requireStateSupportedBy(fixedPart, signedVariableParts, channelId);
+        _requireEachStateFinal(signedVariableParts);
```
geoknee, line 142: We should consider the following, streamlined version of this:
1. require a single state
2. require that state to have `isFInal=true`
3. require `nParticipants` signatures
4. do not defer to the application rules at all

As it is, we have a slightly confusing soup of core logic vs application defined logic dictating whether `conclude` will succeed.

geoknee: [84f8376](https://github.com/statechannels/go-nitro/pull/814/commits/84f8376cd4adc3f2b79cec59d0940da1ca92a7c5)

nksazonov: I have left a [comment](https://github.com/statechannels/go-nitro/issues/813#issuecomment-1201157704) describing what I believe we also have to bear in mind, while making these changes.

lalexgap (approved):
This looks good to me! It's currently based on https://github.com/statechannels/go-nitro/pull/817 so @geoknee I'll let you figure out how you want to merge the PRs.

lalexgap (approved):
Makes sense to me!

PR #815, opened 2022-07-27
Small refactor of metrics

lalexgap:
Fixes https://github.com/statechannels/go-nitro/issues/781

Based on some feedback from @kerzhner I've made a few small changes to how we record metrics in the wallet:
- Instead of wrapping a function with `e.metrics.RecordDuration` to record the duration `e.metrics.RecordFunctionDuration()()` can now be called at the start of the function. See https://stackoverflow.com/a/45766707/13280122
- I've corrected some of the queue length metrics being recorded and put them in one place so they're all reported with the same frequency
- I've removed `Counters` as they didn't function as I had thought. It looks like the counters get reset (I believe when the counter metric info is recorded to influxdb) so the active/completed objective counts didn't make sense

### DIFF
--- a/client/engine/metrics.go
+++ b/client/engine/metrics.go
@@ -92,23 +92,25 @@ func NewMetricsRecorder(me types.Address, metrics MetricsApi) *MetricsRecorder {
 	}
 }
 
+// RecordMessageReceived records the duration of the function
+// It should be called at the start of the function like so  `defer e.metrics.RecordFunctionDuration()()`
 func (o *MetricsRecorder) RecordFunctionDuration() func() {
 	start := time.Now()
 	return func() {
 
 		elapsed := time.Since(start)
 
-		// Skip this function, and fetch the PC and file for its parent.
+		// Skip this function, and fetch the PC  for its parent.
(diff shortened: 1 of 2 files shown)


client/engine/metrics.go
```diff
@@ -90,23 +78,32 @@ func NewMetricsRecorder(me types.Address, metrics MetricsApi) *MetricsRecorder {
 	}
 }
 
-// RecordDuration records the duration of the given function for the metric specified by name
-func (o *MetricsRecorder) RecordDuration(name string, funcToTime func()) {
+// RecordMessageReceived records the duration of the function
+// It should be called at the start of the function like so  `defer e.metrics.RecordFunctionDuration()()`
+func (o *MetricsRecorder) RecordFunctionDuration() func() {
```
kerzhner, line 83: Is it worth it to write a unit test for this function?

kerzhner (approved):

client/engine/metrics.go
lalexgap: I think a test would be a good idea but I'm going to defer it to get this PR merged in.

Issue #768, opened 2022-07-04
Add tests for `StrictTurnTaking` and `Consensus`


nksazonov:
Originated from #676 .

PR #817, opened 2022-07-29
Separately supply proof and candidate

### LINKED ISSUE #812: Separately supply `proof` and `candidate`
The result of #764.

My proposal would be to have the following API:

``` solidity
// In the application:
function requireStateSupported(
        FixedPart calldata fixedPart,
        SignedVariablePart calldata candidate
        SignedVariablePart[] calldata proof
    ) external pure returns () 
    
// In the adjudicator:
checkTurnNumberIncreased(candidate)
IForceMoveApp(fixedPart.appDefinition).requireStateSupportedBy(fixedPart, candidate, proof);
setStatus(fixedPart, candidate)
```
       
1. It's easier for application developers to understand which state they are trying to construct a proof for.
2. It is harder for them to make a mistake by messing up indices
3. The code in the adjudicator becomes simpler (also less bug prone).
4. Gas is saved by not copying and passing around structs.
5. We can fail quickly and more cheaply by checking the turn number before we call into application code.
6. (negative) It is a bit awkward that the `proof` doesn't contain the `candidate` state nor the signatures on it. We could play with some other names -- for example "support" (as in "supporting evidence").

_Originally posted by @geoknee_

---

The idea of `latestSupportedState` not returning anything is great (and thus the new naming).

I really like an explicit `candidate`, although it doubtfully is more gas efficient as to separate it from the rest of the proof supplied to `ForceMove`, we would need to copy the whole proof:
```
proof = SignedVariableParts[:-1]
candidate = SignedVariableParts[-1]
```
Another approach would be to extend the `requireStateSupported` API to `ForceMove`, such so `challenge`, `checkpoint` and `conclude` would also require an explicit `candidate` to be supplied. We need a decision on this.

Moreover, having a separate `candidate` would allow DApp developers to have some peculiar logic, where `candidate` is not a successor of `proof` (by passing `SignedVariableParts` to `ForceMove` in a special order).

It is also worth mentioning that having a separate `proof` and its successor `candidate` somewhat complicates the simplicity of traversing the whole proof in `ForceMoveApp`. Nevertheless, this small downside is fully justified by resolving the naming ambiguity (last vs latest) and removing a weak point of the `candidate` index (you mention it as 2) by explicitly requiring to supply the `candidate`. 

_Originally posted by @nksazonov_

---

Perfect. Then the changes we are to implement are the following:
1. Rename `latestSupportedState` to `requireStateSupported`
2. Supply `proof` and `candidate` separately to both `requireStateSupported` and `ForceMove` external functions
3. Change `StrictTurnTaking` accordingly
4. Change the logic in `ForceMove` to what you have described

nksazonov:
Closes #812.


geoknee:
Tests fail due to #819. When that is resolved, we should re-run the CI on this branch.

nitro-protocol/contracts/NitroAdjudicator.sol
```diff
@@ -93,19 +95,21 @@ contract NitroAdjudicator is ForceMove, MultiAssetHolder {
...
-    function latestSupportedState(
+    function requireStateSupported(
         FixedPart calldata fixedPart,
-        SignedVariablePart[] calldata signedVariableParts
+        SignedVariablePart[] calldata proof,
+        SignedVariablePart calldata candidate
     ) external pure returns (VariablePart memory) {
         // To avoid `Stack to deep` error, signedVariableParts are copied to `memory` array explicitly
-        SignedVariablePart[] memory _signedVariableParts = signedVariableParts;
+        SignedVariablePart[] memory _proof = proof;
```
geoknee: Out of scope for this PR: I tried removing this manual step and everything still worked :-)

nitro-protocol/contracts/libraries/signature-logic/StrictTurnTaking.sol
```diff
@@ -11,21 +11,28 @@ library StrictTurnTaking {
...
+     * @param candidate Recovered variable part the proof was supplied for. The candidate state is supported by proof states.
      */
     function requireValidTurnTaking(
         INitroTypes.FixedPart memory fixedPart,
-        INitroTypes.RecoveredVariablePart[] memory recoveredVariableParts
+        INitroTypes.RecoveredVariablePart[] memory proof,
+        INitroTypes.RecoveredVariablePart memory candidate
     ) internal pure {
-        _requireValidInput(fixedPart.participants.length, recoveredVariableParts.length);
+        _requireValidInput(fixedPart.participants.length, proof.length + 1);
```
geoknee, line 22: It's a bit surprising to pass `proof.length + 1`. I suggest we just remove the check on the proof length. Having `candidate` as a separate parameter is giving us this check for free implicitly.

nitro-protocol/test/contracts/ForceMove/challenge.test.ts
```diff
@@ -235,15 +246,22 @@ describe('challenge', () => {
...
-        expect(
-          parseOutcomeEventResult(
-            eventSignedVariableParts[eventSignedVariableParts.length - 1].variablePart.outcome
-          )
-        ).toEqual(signedVariableParts[signedVariableParts.length - 1].variablePart.outcome);
-        expect(
-          eventSignedVariableParts[eventSignedVariableParts.length - 1].variablePart.appData
-        ).toEqual(signedVariableParts[signedVariableParts.length - 1].variablePart.appData);
+
+        if (proof.length > 0) {
```
geoknee, line 250: Should we add some expectations for the candidate part (which is never empty?)

nitro-protocol/test/contracts/examples/SingleAssetPayments/requireStateSupported.test.ts
```diff
@@ -62,8 +65,8 @@ const reason2 = 'not a simple allocation';
 const reason3 = 'Total allocated cannot change';
 const reason4 = 'outcome: Only one asset allowed';
 
-describe('validTransition', () => {
```
geoknee, line 65: 👍

nitro-protocol/test/contracts/TrivialApp/requireStateSupported.test.ts
```diff
@@ -98,10 +95,14 @@ describe('requireStateSupported', () => {
     const from: RecoveredVariablePart = mockSigs(getVariablePart(fromState));
     const to: RecoveredVariablePart = mockSigs(getVariablePart(toState));
 
-    const tx = trivialApp.requireStateSupported(getFixedPart(fromState), [from], to);
+    const requireStateSupported = await trivialApp.requireStateSupported(
```
geoknee, line 98: Perhaps a better name for this variable? `result`?

nitro-protocol/test/contracts/NitroAdjudicator/concludeAndTransferAllAssets.test.ts
```diff
@@ -225,8 +226,10 @@ describe('concludeAndTransferAllAssets', () => {
       // Form transaction
       const tx = testNitroAdjudicator.concludeAndTransferAllAssets(
         getFixedPart(states[0]),
-        proof,
-        candidate,
+        proof as SignedVariablePart[],
+        candidate as SignedVariablePart,
```
geoknee: I've pushed a commit which removes the need for these type assertions, by making `separateProofAndCandidate` take a generic type.

geoknee (commented):
I think this looks good and achieves the aim of simplifying and clarifying the API for state channel applications.

It does seem more optimal for "everyone signs a single state" than round-robin type applications, which have to treat the separate `candidate` state specially every time. 

We will discuss this change in our standup to gather more opinions. At this point I am inclined to approve and merge this, but probably after cleaning up the history a bit (we could squash the commits, for example).

nitro-protocol/contracts/NitroAdjudicator.sol
geoknee: [bfb398d](https://github.com/statechannels/go-nitro/pull/817/commits/bfb398deefec7988c0a7c0665dac902475f3743d)

nitro-protocol/test/contracts/NitroAdjudicator/concludeAndTransferAllAssets.test.ts
geoknee: [a7aa42e](https://github.com/statechannels/go-nitro/pull/817/commits/a7aa42e958858d873c21ce978355c54ebfe30eef)

nitro-protocol/contracts/ForceMove.sol
```diff
@@ -40,16 +40,18 @@ contract ForceMove is IForceMove, StatusManager {
...
     function challenge(
         FixedPart memory fixedPart,
-        SignedVariablePart[] memory signedVariableParts,
+        SignedVariablePart[] memory proof,
+        SignedVariablePart memory candidate,
         Signature memory challengerSig
     ) external override {
         bytes32 channelId = NitroUtils.getChannelId(fixedPart);
-        uint48 largestTurnNum = _lastVariablePart(signedVariableParts).turnNum;
+        uint48 largestTurnNum = candidate.variablePart.turnNum;
```
geoknee, line 54: Can we be _sure_ that this turn number is "largest"? Let's just use `candidateTurnNum` or something, and let L56-63 handle turn number comparison logic?

geoknee (approved):
Approving following positive feedback during today's standup.

nitro-protocol/contracts/ForceMove.sol
nksazonov: Previously, in Nitro V1, we had `largestTurnNum` supplied and it was used to compose the state, treating the last variable part supplied as the one having the `largestTurnNum`.

Now, in V2, we _may_ imply that `candidate` is the state with `largestTurnNum`, thus resembling V1 logic.

On the other hand, that would make it impossible to create DApps with some peculiar logic expecting `candidate` not to have `largestTurnNum`, but that seems weird.

If participants specify the `candidate`, they expect State Channel state to "upgrade" specifiacally to it.

Moreover, if we change the API to expect a separate `largestTurnNum`, how would its check look like? Will we iterate through all supplied `variableParts`, checking if any has `largestVariablePart` in it?

geoknee: I think since we are handing off almost 100% of the responsibility about which state is supported to the application, we should not give this variable a name like "largestTurnNum` **unless** we have some mechanism in the core protocol to ensure that this is an appropriate name. I don't think we should put that mechanism in. Most applications (perhaps all) will do the requisite checks on the turn number (see e.g. `StrictTurnTaking`). But "most likely" is not a strong enough guarantee to use this variable name. 

I don't think we need to add a separate `largestTurnNum` to the API. I propose that we allow for (but discourage) the "weird" logic where an application can allow a later state to form the support proof for an earlier state. 

In Nitro v2 the application and the core protocol must be audited together by a suspicious user. So such weird logic would be rejected/avoided in most cases.

nksazonov: Renaming `largestTurnNum` to `candidateTurnNum` then.

nitro-protocol/test/contracts/NitroAdjudicator/concludeAndTransferAllAssets.test.ts
nksazonov: Nice, great thanks!

nitro-protocol/test/contracts/TrivialApp/requireStateSupported.test.ts
nksazonov: Agree with you. Also have noticed, the comment on why we are checking the length of `result` is not 100% clear, so I will tidy them also.

nitro-protocol/test/contracts/ForceMove/challenge.test.ts
```diff
@@ -235,15 +246,22 @@ describe('challenge', () => {
...
+          ).toEqual(proof[proof.length - 1].variablePart.outcome);
+          expect(eventProof[eventProof.length - 1].variablePart.appData).toEqual(
+            proof[proof.length - 1].variablePart.appData
+          );
+        }
+
+        expect(parseOutcomeEventResult(eventCandidate.variablePart.outcome)).toEqual(
+          candidate.variablePart.outcome
+        );
+        expect(eventCandidate.variablePart.appData).toEqual(candidate.variablePart.appData);
```
nksazonov, line 262: Candidate event check here.

nitro-protocol/test/contracts/ForceMove/challenge.test.ts
nksazonov: It is right below, see [261-264](https://github.com/statechannels/go-nitro/pull/817/files/c2c0264c775401a3b1edac3bc49c03c92c3edaf7..bf8e9e33e346e587597721ee306955a9af086a49#r950047050).

geoknee: Sure it is! Not sure how I missed it!

nitro-protocol/contracts/libraries/signature-logic/StrictTurnTaking.sol
nksazonov: Do you suggest we remove checks on proof length entirely (from `_requireValidInput` method)?
I see no other simple way, which can back up this check.

nitro-protocol/contracts/NitroAdjudicator.sol
nksazonov: Wow, solidity surprises me every now and then!

nitro-protocol/contracts/examples/SingleAssetPayments.sol
```diff
@@ -10,35 +10,45 @@ import '../interfaces/IForceMoveApp.sol';
...
+                    proof[i].variablePart
                 );
             }
         }
 
-        return recoveredVariableParts[recoveredVariableParts.length - 1].variablePart;
+        _requireValidOutcome(
+            fixedPart.participants.length,
+            candidate.variablePart.outcome
+        );
```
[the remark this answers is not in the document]
nksazonov: Somehow I forgot solidity has ternary operator, this makes code much clearer.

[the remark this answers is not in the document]
nksazonov: The same change could be applied to `StrictTurnTaking` and `ShortcuttingTurnTaking`.

nitro-protocol/contracts/libraries/signature-logic/StrictTurnTaking.sol
geoknee: That is what I was suggesting yes. However, now I am not so sure :-P . For now I'm happy to leave this as is, actually.

nksazonov: Nevertheless, I am still sure `proof.length + 1` is a bad practise, so small refactor will be implemented 😃

nitro-protocol/contracts/interfaces/IForceMoveApp.sol
```diff
@@ -5,18 +5,19 @@ pragma experimental ABIEncoderV2;
...
 interface IForceMoveApp is INitroTypes {
     /**
-     * @notice Encodes application-specific rules for a particular ForceMove-compliant state channel.
-     * @dev Encodes application-specific rules for a particular ForceMove-compliant state channel.
+     * @notice Encodes application-specific rules for a particular ForceMove-compliant state channel. Must revert when invalid support proof and a candidate are supplied.
+     * @dev Encodes application-specific rules for a particular ForceMove-compliant state channel. Must revert when invalid support proof and a candidate are supplied.
      * @param fixedPart Fixed part of the state channel.
-     * @param recoveredVariableParts Array of recovered variable parts to find the latest of.
-     * @return VariablePart Latest supported by application variable part from supplied array.
+     * @param proof Array of recovered variable parts which constitutes a support proof for the candidate.
```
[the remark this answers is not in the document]
nksazonov: Nice catch!

nitro-protocol/contracts/interfaces/IForceMove.sol
```diff
@@ -12,32 +12,42 @@ interface IForceMove is INitroTypes {
...
+        SignedVariablePart[] memory proof,
+        SignedVariablePart memory candidate,
         Signature memory challengerSig
     ) external;
 
     /**
-     * @notice Overwrites the `turnNumRecord` stored against a channel by providing a state with higher turn number, supported by a signature from each participant.
-     * @dev Overwrites the `turnNumRecord` stored against a channel by providing a state with higher turn number, supported by a signature from each participant.
+     * @notice Overwrites the `turnNumRecord` stored against a channel by providing a proof with higher turn number.
+     * @dev Overwrites the `turnNumRecord` stored against a channel by providing a proof with higher turn number.
```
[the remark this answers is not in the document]
nksazonov: A `support proof` must have been there, but I like your option more

nitro-protocol/contracts/libraries/signature-logic/StrictTurnTaking.sol
```diff
@@ -23,16 +23,11 @@ library StrictTurnTaking {
...
-            requireHasTurnNum(proof[i].variablePart, turnNum);
+            isSignedByMover(fixedPart, (i < proof.length ? proof[i] : candidate));
+            requireHasTurnNum((i < proof.length ? proof[i].variablePart : candidate.variablePart), turnNum);
             turnNum++;
         }
-
-        // validating the candidate
-        isSignedByMover(fixedPart, candidate);
-        requireHasTurnNum(candidate.variablePart, turnNum);
     }
```
nksazonov, line 34: @geoknee , do you think this change make the code clearer?
For sure, it follows the practice of treating `proof` and `candidate` equally, but does it improves readability?

geoknee: 1. I don't think it makes a huge difference. Probably on balance it is slightly **less** readable IMO.
2. I'm not sure it is actually correct? The condition on the loop is for `i<proof.length`, meaning that `candidate` will never be selected by the ternary operator? I think you need to have the loop condition `i<=proof.length` or `i<proof.length+1`.

nksazonov: Wow!
How did the tests not see 2)?

geoknee: https://github.com/statechannels/go-nitro/issues/768

nksazonov: Treating `proof` and `candidate` in the same way (using ternary operator to specify which one to use) is doubtfully increasing the readability, so I propose we give up on it and left the code as it was.

PR #821, opened 2022-08-15
Upgrade to go 1.19

### LINKED ISSUE #819: `golanglint-ci` does not start properly
See this run https://github.com/statechannels/go-nitro/runs/7840888001?check_suite_focus=true. 

This issue seems related:
https://github.com/golangci/golangci-lint/issues/2374

lalexgap:
Fixes #819 

Upgrades `go-nitro` to 1.19. 

Previously our CI could use go 1.19 (since we specify version range `^1.18.0`) but the linter version we used (`v1.46.2`) did not support go 1.19. 

This PR updates go-nitro to use go 1.19 and the linter to use a version that supports go 1.19.

### DIFF
--- a/protocols/directfund/serde.go
+++ b/protocols/directfund/serde.go
@@ -22,10 +22,8 @@ type jsonObjective struct {
 }
 
 // MarshalJSON returns a JSON representation of the DirectFundObjective
-//
 // NOTE: Marshal -> Unmarshal is a lossy process. All channel data
-//
-//	(other than Id) from the field C is discarded
+// (other than Id) from the field C is discarded
 func (o Objective) MarshalJSON() ([]byte, error) {
 	jsonDFO := jsonObjective{
 		o.Status,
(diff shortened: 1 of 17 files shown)


protocols/directfund/serde.go
```diff
@@ -24,7 +24,8 @@ type jsonObjective struct {
 // MarshalJSON returns a JSON representation of the DirectFundObjective
 //
 // NOTE: Marshal -> Unmarshal is a lossy process. All channel data
-//       (other than Id) from the field C is discarded
+//
+//	(other than Id) from the field C is discarded
```
geoknee, line 28: These changes were automatic, but they highlight something that seems off. Probably this line should never have been indented?

geoknee (approved):

protocols/directfund/serde.go
lalexgap: Good catch! I've cleaned up the spacing in [094f166](https://github.com/statechannels/go-nitro/pull/821/commits/094f166feab9267d2520b9f8ca55206c8ae9e799)

PR #824, opened 2022-08-16
Add vouchers to message format

### LINKED ISSUE #823: Add support for vouchers to messaging service
This assumes that we'll want to use our existing message service to handle sending and receiving vouchers. 

We should update the message service interface (and basic implementations) to support sending/receiving vouchers.

lalexgap:
Fixes #823 

This adds support for transmitting vouchers across the wire using our existing message format. 

- Adds basic serde and test for `payment.Voucher` 
- Adds `payment.Voucher`s to the message format and extends the message test.

### DIFF
--- a/payments/payments_test.go
+++ b/payments/payments_test.go
@@ -1,11 +1,15 @@
 package payments
 
 import (
+	"fmt"
 	"math/big"
+	"path/filepath"
+	"reflect"
+	"runtime"
 	"testing"
 
 	"github.com/statechannels/go-nitro/internal/testactors"
-	. "github.com/statechannels/go-nitro/internal/testhelpers"
+
 	"github.com/statechannels/go-nitro/types"
 )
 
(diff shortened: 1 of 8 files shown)


payments/payments_test.go
```diff
@@ -128,3 +132,43 @@ func TestPaymentManager(t *testing.T) {
 	Assert(t, err != nil, "expected an error")
 	Equals(t, twoPaymentsMade, getBalance(receiptMgr))
 }
+
+// TODO: This is a copy of the test helpers from github.com/statechannels/go-nitro/internal/testactors
```
lalexgap, line 136: This PR means that the `protocols` package (where the message code is) now references the `payment` package.

This means using the `testhelper` package in the `payment` package would introduce an import cycle (`protocols`-> `payment` -> `testhelper`->`protocols`).

To get around this I just copied the test helpers into the `payment` package for now. However we might want to do some package reorganization to avoid running into this.

payments/vouchers.go
```diff
@@ -46,3 +47,17 @@ func (v Voucher) recoverSigner() (types.Address, error) {
 	}
 	return nitroCrypto.RecoverEthereumMessageSigner(h[:], v.signature)
 }
+
+// Equal returns true if the two vouchers have the same channel id, amount and signaturesß
```
geoknee, line 51: 1. Typo on this line
2. Should this (and all of the other methods) have a pointer receiver rather than a value receiver?

geoknee (approved):

payments/vouchers.go
lalexgap: We probably should be consistent and use pointer receivers. My natural instinct is to reserve pointer receivers for functions that modify the receiver but best practice is being [consistent](https://go.dev/doc/faq#methods_on_values_or_pointers:~:text=Next%20is%20consistency.%20If%20some%20of%20the%20methods%20of%20the%20type%20must%20have%20pointer%20receivers%2C%20the%20rest%20should%20too%2C%20so%20the%20method%20set%20is%20consistent%20regardless%20of%20how%20the%20type%20is%20used.%20See%20the%20section%20on%20method%20sets%20for%20details.)

lalexgap: [2b78cca](https://github.com/statechannels/go-nitro/pull/824/commits/2b78ccaf18ecff12e7c66a88298a3f4409e4ea4d)

payments/payments_test.go
geoknee: Is there an issue tracking this?

PR #829, opened 2022-08-17
feat: add `VirtualPaymentApp.sol`

### LINKED ISSUE #559: Writing a `VirtualChannel` application conforming to the new `ForceMoveApp` interface
See https://www.overleaf.com/project/618a7979d3893c93c6f7a931 or https://github.com/statechannels/go-nitro/discussions/764#discussioncomment-3053372.

The work should include:
* A typescript test in the `./nitro-protocol/contracts/tests` sub dir.
* Checked-in go-bindings for this new contract which are available for the chain service to use. See for example:

https://github.com/statechannels/go-nitro/pull/676/files?file-filters%5B%5D=.sh&show-deleted-files=true&show-viewed-files=true#diff-3ac5afd992608bf76f556ef1f83e3394e068f4a2a59d11af8f7c95894639ce41

We can reuse / replace `EmbeddedApp.sol`.

geoknee:
Closes #559.


nitro-protocol/contracts/VirtualPaymentApp.sol
```diff
@@ -0,0 +1,108 @@
...
+            require(
+                NitroUtils.getClaimedSignersNum(candidate.signedBy) ==
+                    fixedPart.participants.length,
+                '!unanimous; |proof|=0'
+            );
+            if (candidate.variablePart.turnNum == 0) return; // prefund
+            if (candidate.variablePart.turnNum == 1) return; // postfund
+            if (candidate.variablePart.turnNum == 3) {
+                // final (note: there is a core protocol escape hatch for this, too, so it could be removed)
+                require(candidate.variablePart.isFinal, '!final; turnNum=3 && |proof|=0');
```
lalexgap, line 49: If the `redemption` state was also a `final` state would that allow Bob to unilaterally clear the challenge? It would be nice if Bob submitting the voucher just closed the channel, since there's nothing Alice or Irene can do after that to change the outcome (without Bob's support).

Or does the `ForceMove` framework require everyone to sign the concluded state? [Comments](https://github.com/statechannels/go-nitro/blob/6678d15d9ce091b415883bb13e266f5b91379cce/nitro-protocol/contracts/ForceMove.sol#L118) seem to indicate that conclude requires final states but I don't actually see where that's enforced. 🤔

lalexgap: It looks like this is already being discussed in https://github.com/statechannels/go-nitro/issues/813

geoknee: This is a good data point to feed into that discussion 👍

nitro-protocol/contracts/VirtualPaymentApp.sol
```diff
@@ -0,0 +1,139 @@
...
+        // 2 redemption
+        // 3 final
+
+        // states 0,1,3 can be supported via unanimous consensus:
+
+        if (proof.length == 0) {
+            require(
+                NitroUtils.getClaimedSignersNum(candidate.signedBy) ==
+                    fixedPart.participants.length,
+                '!unanimous; |proof|=0'
```
lalexgap, line 48: Should we make these `require` strings easier to parse? Or is this a notation we use all over the place?

nitro-protocol/contracts/VirtualPaymentApp.sol
```diff
@@ -0,0 +1,139 @@
...
+        uint256 numParticipants
+    ) internal pure {
+        require(rVP.variablePart.turnNum == 1, 'bad proof[0].turnNum; |proof|=1');
+        require(
+            NitroUtils.getClaimedSignersNum(rVP.signedBy) == numParticipants,
+            'postfund !unanimous; |proof|=1'
+        );
+    }
+
+    function requireValidVoucher(bytes memory appData, FixedPart memory fixedPart)
```
lalexgap, line 96: nit but most of the other `requireFoo` functions don't return anything only check assertions. Would it make sense to rename this to `getVoucherAmount`?

nitro-protocol/contracts/VirtualPaymentApp.sol
```diff
@@ -0,0 +1,139 @@
...
+        pure
+        returns (uint256)
+    {
+        VoucherAmountAndSignature memory voucher = abi.decode(appData, (VoucherAmountAndSignature));
+
+        address signer = NitroUtils.recoverSigner(
+            keccak256(abi.encode(NitroUtils.getChannelId(fixedPart), voucher.amount)),
+            voucher.signature
+        );
+        require(signer == fixedPart.participants[0], 'irrelevant voucher'); // could be incorrect channelId or incorrect signature
```
lalexgap, line 107: Nit but I think `irrelevant voucher` could be slightly confusing.
```suggestion
        require(signer == fixedPart.participants[0], 'invalid signature for voucher'); // could be incorrect channelId or incorrect signature
```

nitro-protocol/contracts/VirtualPaymentApp.sol
```diff
@@ -0,0 +1,139 @@
...
+                newOutcome.length == 1 &&
+                oldOutcome[0].asset == address(0) &&
+                newOutcome[0].asset == address(0),
+            'only native asset allowed'
+        );
+
+        require(
+            voucherAmount <= oldOutcome[0].allocations[uint256(AllocationIndices.Alice)].amount,
+            'underflow'
+        ); // TODO this check may be unecessary in a future version of solidity
```
lalexgap, line 127: [This is handled in solidity >= 0.8](https://solidity-by-example.org/hacks/overflow/)

lalexgap (approved):
This looks good to me! It's currently based on https://github.com/statechannels/go-nitro/pull/817 so @geoknee I'll let you figure out how you want to merge the PRs.

nitro-protocol/contracts/VirtualPaymentApp.sol
geoknee: The reason they are a bit difficult to parse is because I am trying to keep them short to save gas. I need to refresh my memory about whether the gas costs are only significant for deployment, though.

Probably we should be adopting a more modern approach to errors https://docs.soliditylang.org/en/latest/contracts.html#errors-and-the-revert-statement .

Issue #809, opened 2022-07-22
Remove duplicated code from `gas-benchmarks`


nksazonov:
However, I think one downside is now we have quite a bit of duplication:

- `gas.ts` contains similar information to `gasResults.json`, only _with_ comments and _without_ meaningful numbers. There is potential for confusion here. Could you simply parse the existing `gasResults.json` instead of using an `emptyGasResults`? You may not even need to zero everything out, then (but you could).
- `exportBenchmarkResults.ts` duplicates the comments from `gas.ts`, and is also duplicated with respect to `benchmark.test.ts`.
When we want to add a new benchmark, it seems as if there are now quite a lot of files to update.

One way to improve this is to drop the `benchmark:diff` side of things. I imagine we won't use it so much -- and even in CI we can simply run `npm run benchmark:update` and let `git diff --exit-code` do the "check" on the gas numbers. What do you think?

_Originally posted by @geoknee in [#802](https://github.com/statechannels/go-nitro/pull/802#pullrequestreview-1046151404)._

---

You are right, we have quite a bit of duplication. Actually, the reason for both points you describe is to keep color-coded comparison results, which I believe is a powerful feature as it is very convenient to see the difference already in console, not needing to do math in your head.

It is also worth mentioning that `exportBenchmarkResults.ts` takes zeroed `gas.ts`, but not gasResults.json, as the latest may not exist.

So to keep the color-coded diff feature, we definitely need to keep `benchmark.test.ts` (otherwise we would need to explicitly print colorful messages to console ourselves). Thus, I can see several options how we can remove the duplication:

1. Create a static `Benchmark` object, which will describe both benchmark result structure interface and benchmark method to invoke for that result.
``` javascript
{
    title: 'directlyFundAChannelWithETHSecond.satp',
    setup: async () => {
        const setupTX = nitroAdjudicator.deposit(MAGIC_ADDRESS_INDICATING_ETH, X.channelId, 0, 5, {
            value: 5,
        });
        await (await setupTX).wait();
    },
    method: async () => {
        await nitroAdjudicator.deposit(MAGIC_ADDRESS_INDICATING_ETH, X.channelId, 5, 5, {value: 5})
    }
}
```
Both test and script files will just traverse through all the field, invoking the setup, benchmarking the method and either testing for equality with or saving the results to title.

This way, the only place we would need to make a change to (if a new benchmark is introduced / old changed) is this `Benchmark` object.

[...]

PR #831, opened 2022-08-19
Add "sadly exiting a virtual channel" gas benchmark

geoknee:
This is an important test to unksip, and represents a vital on-chain flow protecting end users.

To get the test running and passing again, I had to make quite a lot of changes. I hope this are clear from the commit history and messages. 

Unfortunately there is a lot of replication in the existing gas benchmarking boilerplate code. We are tracking this issue here #809 . That issue has made this PR a bit bloated. 

The existing `fixtures.ts` is potentially also in need of some clean up. I would be happy to take that on in a new PR.


nitro-protocol/gas-benchmarks/benchmark.test.ts
```diff
@@ -199,72 +205,100 @@ describe('Consumes the expected gas for sad-path exits', () => {
     ).toEqual(gasRequiredTo.ETHexitSadLedgerFunded.satp.total);
   });
 
-  // TODO unskip this test when the contracts are satp compatible
-  it.skip(`when exiting a virtual funded (with ETH) channel`, async () => {
+  it(`when exiting a virtual funded (with ETH) channel`, async () => {
```
geoknee, line 208: This is a good place to start reviewing. From here to the end of the file you can see the operations Bob is doing to get his money.

nitro-protocol/gas-benchmarks/gasResults.json
```diff
@@ -23,28 +23,34 @@
     }
   },
   "ETHexit": {
-    "satp": 141446
+    "satp": 109734
```
kerzhner, line 26: With the current gas prices, these transactions are fairly cheap!

0.000000001 eth/gwei * 1,500 $/eth  * 10 gwei/gas * 109,734 gas = $1.65

nitro-protocol/gas-benchmarks/benchmark.test.ts
```diff
@@ -199,72 +205,100 @@ describe('Consumes the expected gas for sad-path exits', () => {
...
+        0,
+        amountForAliceAndBob,
+        {
+          value: amountForAliceAndBob,
+        }
+      )
     ).wait();
     // end setup
-    // initially                   ⬛ ->  L  ->  J  ->  X  -> 👩
+    // initially                   ⬛ ->  L  ->  V  -> 👨
```
kerzhner, line 223: Not entirely sure what the face emoji in the comment means. Does the virtual channel only allocate to one participant?

nitro-protocol/gas-benchmarks/benchmark.test.ts
```diff
@@ -199,72 +205,100 @@ describe('Consumes the expected gas for sad-path exits', () => {
...
     );
+
+    expect(vGasUsed).toEqual(gasRequiredTo.ETHexitSadVirtualFunded.satp.challengeV);
+
     // begin wait
-    await waitForChallengesToTimeOut([ledgerFinalizesAt, jointChannelFinalizesAt, finalizesAt]);
+    await waitForChallengesToTimeOut([ledgerFinalizesAt, vFinalizesAt]);
     // end wait
-    // challenge L,J,X + timeout   ⬛ -> (L) -> (J) -> (X) -> 👩
+    // challenge L,V   + timeout   ⬛ -> (L) -> (V) -> 👨
```
kerzhner, line 250: Is it explained anywhere what parentheses in (L) stand for?

kerzhner (approved):

nitro-protocol/gas-benchmarks/exportBenchmarkResults.ts
```diff
@@ -1,5 +1,9 @@
 import {writeFileSync} from 'fs';
```
kerzhner, line 1: What is the difference between `benchmark.test.ts` and `exportBenchmarkResults.ts`?

nitro-protocol/src/contract/multi-asset-holder.ts
```diff
@@ -96,13 +96,13 @@ export function computeReclaimEffects(
     };
 
     // copy each element except the indexOfTargetInSource element
-    if (sourceAllocations[i].destination == left) {
+    if (sourceAllocations[i].destination.toLowerCase() == left.toLowerCase()) {
```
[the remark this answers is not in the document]
geoknee: Perhaps... how could we achieve that?

In my view, as long as we are representing bytes as hex strings we need to be careful when doing comparisons like this. I think it would be difficult / dangerous to try and keep them e.g. lowercase everywhere and then rely on a simple `==` check. 

We could introduce a helper `areEqualDestinations()`, but I fear these kinds of bugs will likely still crop up.

nksazonov (approved):

nitro-protocol/gas-benchmarks/benchmark.test.ts
geoknee: [c5ede49](https://github.com/statechannels/go-nitro/pull/831/commits/c5ede49843ed7e5c000c08b4f55c50897004bf90)

nitro-protocol/gas-benchmarks/benchmark.test.ts
geoknee: I added some explanation here [c5ede49](https://github.com/statechannels/go-nitro/pull/831/commits/c5ede49843ed7e5c000c08b4f55c50897004bf90). Hopefully it answers your question, LMK if not!

nitro-protocol/gas-benchmarks/exportBenchmarkResults.ts
geoknee: [Explain difference between files](https://github.com/statechannels/go-nitro/pull/831/commits/d3f8627fa5cee6f130860d9542a7b35c7728dfc5)

nitro-protocol/gas-benchmarks/fixtures.ts
```diff
@@ -200,37 +214,50 @@ export const LforX = new TestChannel(
...
       allocationType: AllocationType.simple,
     },
-  ]
+  ],
+  virtualPaymentAppAddress
 );
 
-/** Ledger channel between Alice and Ingid, with Guarantee targeting joint channel J */
-export const LforJ = new TestChannel(
+/** Ledger channel between Alice and Ingrid, with Guarantee targeting virtual channel V */
```
geoknee, line 238:
```suggestion
/** Ledger channel between Bob and Ingrid, with Guarantee targeting virtual channel V */
```

nitro-protocol/gas-benchmarks/benchmark.test.ts
kerzhner: The explanation is great!

PR #834, opened 2022-08-22
Add `Consensus` signatures logic library, move `ShortcuttingTurnTaking` to `contracts/test`

### LINKED ISSUE #771: Add `Consensus` signatures logic library, move `ShortcuttingTurnTaking` to `contracts/test`
I think it would be great to add a specific signature library (alongside `StrictTurnTaking` and `ShortcuttingTurnTaking`) for the Consensus app (that is "everyone signs one state" mode). This will resemble the fact that there are 3 different signature modes (the ones I just mentioned).

I don't believe that we should simplify the code in `ShortcuttingTurnTaking` if we introduce `Consensus`. On the contrary, it is great that we have some example code that can be used by developers. Maintenance of this code should not take much time as well as writing tests for it.

*Originally posted by @nksazonov in [#676](https://github.com/statechannels/go-nitro/pull/676#discussion_r913153448).*

This should also undo changes (use higher turn number for pre- and post-fund states) to `challenge_test.go` required for `ShortcuttingTurnTaking` to accept S0 and S1 as pre-, post-fund states.

nksazonov:
Closes #771 .

Main alongside changes:
- `CountingApp` and `SingleAssetPayments` have switched to `StrictTurnTaking`, which has broken some tests that expected `ShortcuttingTurnTaking` logic to be present


nitro-protocol/test/contracts/examples/SingleAssetPayments/requireStateSupported.test.ts
```diff
@@ -69,18 +67,16 @@ describe('requireStateSupported', () => {
...
-    ${true}  | ${[1, 1]} | ${[true, true]}   | ${{A: 1, B: 1, C: 1}} | ${4}     | ${{A: 1, B: 0, C: 2}} | ${whoSignedWhatB} | ${undefined} | ${'B pays C 1 wei'}
-    ${true}  | ${[1, 1]} | ${[true, true]}   | ${{A: 1, B: 1, C: 1}} | ${5}     | ${{A: 1, B: 2, C: 0}} | ${whoSignedWhatC} | ${undefined} | ${'C pays B 1 wei'}
-    ${false} | ${[1, 1]} | ${[true, true]}   | ${{A: 1, B: 1, C: 1}} | ${5}     | ${{A: 0, B: 2, C: 1}} | ${whoSignedWhatA} | ${reason1}   | ${'A pays B 1 wei (not their move)'}
-    ${false} | ${[1, 1]} | ${[false, false]} | ${{A: 1, B: 1, C: 1}} | ${3}     | ${{A: 0, B: 2, C: 1}} | ${whoSignedWhatA} | ${reason2}   | ${'Guarantee'}
-    ${false} | ${[1, 1]} | ${[true, true]}   | ${{A: 1, B: 1, C: 1}} | ${3}     | ${{A: 1, B: 2, C: 1}} | ${whoSignedWhatA} | ${reason3}   | ${'Total amounts increase'}
-    ${false} | ${[2, 2]} | ${[true, true]}   | ${{A: 1, B: 1, C: 1}} | ${3}     | ${{A: 2, B: 0, C: 1}} | ${whoSignedWhatA} | ${reason4}   | ${'More than one asset'}
+    numAssets | isAllocation      | balancesA       | turnNumB | balancesB       | whoSignedWhat     | reason       | description
```
geoknee, line 70: Out of scope nit: `turnNumB` is arguably a legacy parameter. It is no longer a part of the `ForceMove` API, so it is a bit confusing to have test cases defined using it in this way. 

I think it would potentially be more natural to have something like `turnNums: [number,number]`.

geoknee (approved):
👍

nitro-protocol/test/contracts/examples/SingleAssetPayments/requireStateSupported.test.ts
nksazonov: 0dbdc45

nksazonov: Agree with you. Applied this suggestion in the commit above.

Issue #795, opened 2022-07-12
Add the ability to settle vouchers and close the virtual channel


andrewgordstewart:
We will need to change the `virtualdefund` objective so that:
    
1. Alice & Bob agree on the “largest voucher”
1. Considering that a voucher might be in-flight, maybe Bob would expect Alice to provide the largest voucher within the objective’s business logic, and we program Alice to politely send the largest voucher (even though she’s not obligated to).
2. Once Bob receives the largest voucher, he constructs and signs a final state, and sends to Irene
3. Once Irene has a final state signed by Bob, she signs it and sends to Alice & Bob
4. Once everyone has a supported final state, they remove the guarantees from their ledger channels

There are two options for orchestrating the virtual channel to close:

A. Alice, Bob & Irene sign “forcing states” that are not final, but include the information necessary to construct the final state. This is compatible with the current message format
B. Bob & Irene send messages to Alice saying something like “Alice, I’d like to close the virtual channel. Please provide the largest voucher to Bob.” This is not compatible with the current message format.

**Question:** If we choose B, should objectives be responsible for interpreting their own payloads? If so, would there be some concern about spreading out message-related responsibilities? duplicated code?

lalexgap:
Since we've decided to include vouchers in the message format (see #824) I think approach B would make sense.

lalexgap:
See the [ADR](https://github.com/statechannels/go-nitro/pull/863) for the approach we will follow.

lalexgap:
This was fixed by #866

Issue #775, opened 2022-07-06
Integrate the payment and receipt managers into the go-nitro client.


geoknee:
(If following a TDD approach this follows #773 and is then all about making the failing "payment integration" test pass)

We’ve implemented a PaymentManager and a ReceiptManager  #800 .

The “user story” would look something like

- [ ] https://github.com/statechannels/go-nitro/issues/794 Alice, Bob and Irene create a virtual channel. Once opened, Alice creates a PaymentManager and Bob creates a ReceiptManager.
- [x] https://github.com/statechannels/go-nitro/issues/793: : Alice sends a stream of vouchers to Bob
- [ ] https://github.com/statechannels/go-nitro/issues/795 Alice, Bob or Irene trigger the virtual channel to close. The channel is defunded and ledgers rebalanced according to the "largest" voucher signed by Alice.
- [x] https://github.com/statechannels/go-nitro/issues/772
- [ ] https://github.com/statechannels/go-nitro/issues/822
- [x] https://github.com/statechannels/go-nitro/issues/823
- [ ] #827 
- [x] #837

lalexgap:
I'm closing this as the bulk of the work is done. There are still some remaining things to be done but they can be tracked under the payments label.

PR #836, opened 2022-08-22
Voucher integration into engine

### LINKED ISSUE #772: Add a minimal payment API to the go-nitro client
See [this design doc.](https://www.notion.so/pl-strflt/Open-Grant-Proposal-go-nitro-Productization-f2b935ab26324545ac6f06f1974a518e#558bb00ca1c44a2795ec51ed421a5081).

Depends on https://github.com/statechannels/go-nitro/issues/793

lalexgap:
This is work towards #775.

Fixes #772 

It integrates the existing `PaymentManager` and `ReceiptManager` into the `engine` and exposes two new functions on the client API:
- `Pay`: This signs a voucher and sends it to fellow participants in the channel.
- `ReceivedVouchers`: This is a `chan` that alerts the client they have a received a payment from another participant.

Calling `pay` will sign a voucher and send that signed voucher to the payee. `virtual-defund` doesn't [use the vouchers yet but that will be tackled in a separate PR](#795 ).

## Shortcoming / caveats
- #837 The `PaymentManager/ReceiptManager` API was designed to be more externally facing making it slightly awkward to use in the engine.  Longer term we may want to refactor into a singular `VoucherManager` that simply manages storing vouchers.
- Some of the logic around payments has leaked [into the engine](https://github.com/statechannels/go-nitro/blob/143903faa83594c18b901506eb47c92d3c664e3d/client/engine/engine.go#L328). Probably ok for now but we should consider moving that into the `VoucherManager` if we do that refactor
- `payments.Voucher`s are exposed directly from the client API. We should probably use an interface or a richer helper struct
- There are still some TODOs lurking


client/engine/engine.go
```diff
@@ -346,43 +384,73 @@ func (e *Engine) handleAPIEvent(apiEvent APIEvent) (ObjectiveChangeEvent, error)
...
-			return ObjectiveChangeEvent{}, fmt.Errorf("handleAPIEvent: Unknown objective type %T", request)
+			return EngineChangeEvent{}, fmt.Errorf("handleAPIEvent: Unknown objective type %T", request)
 		}
 
 	}
 
-	return ObjectiveChangeEvent{}, nil
+	zeroAddress := types.Destination{}
+	// TODO: Should this live in the payment manager?
+	if cId := apiEvent.PaymentToMake.ChannelId; cId != zeroAddress {
```
geoknee, line 430: Somewhat out of scope -- this zero value check seems a bit awkward and is probably a consequence of us approximating union types with an intersection type. We have several api methods, a **single** type for all of them, and then several handlers for that type. That single type therefore has to have a bunch of zero values on it. 

What if we had a single type per method and handler? It would make the master engine "run loop" more verbose, but might be more efficient and less messy overall.

client/engine/engine.go
```diff
@@ -453,6 +521,24 @@ func (e *Engine) attemptProgress(objective protocols.Objective) (outgoing Object
 	return
 }
 
+func (e Engine) registerVirtualChannelWithManagers(vfo virtualfund.Objective) error {
+	prefund := vfo.V.PreFundState()
+	startingBalance := big.NewInt(0)
+	// TODO: Assumes one asset for now
+	startingBalance.Set(prefund.Outcome[0].Allocations[0].Amount)
+
+	switch vfo.MyRole {
```
geoknee, line 530: I'm not sure this is correct:

client/engine/engine.go
```diff
@@ -453,6 +521,24 @@ func (e *Engine) attemptProgress(objective protocols.Objective) (outgoing Object
...
+	case 0:
+		return e.pm.Register(vfo.V.Id, startingBalance)
+	case 2:
+		return e.rm.Register(vfo.V.Id, prefund.Participants[0], startingBalance)
+	default:
+		// The intermediary does not need to use the payment or receipt manager
```
geoknee, line 536: Nit: intermediaries

client/engine/engine.go
```diff
@@ -453,6 +521,24 @@ func (e *Engine) attemptProgress(objective protocols.Objective) (outgoing Object
 	return
 }
 
+func (e Engine) registerVirtualChannelWithManagers(vfo virtualfund.Objective) error {
+	prefund := vfo.V.PreFundState()
```
geoknee, line 525: Nit: would the post fund make more sense? They should have the same outcome, of course...

client/readme.md
```diff
@@ -11,15 +11,16 @@ The flow of data through the client is shown in this diagram:
    - the `message` service
    - the `chain` service.
 2. The engine reads channels and objectives from the `store`, and computes updates and side effects.
-3. The updates arre committed to the `store`.
-4. The side effects are sent on go channels to:
+3. The engine reads payments the `payment manager`, and computes updates and side effects.
```
geoknee, line 14: This sentence doesn't seem to scan right?

client_test/payment_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+	directlyFundALedgerChannel(t, clientA, clientI)
+	directlyFundALedgerChannel(t, clientI, clientB)
+	outcome := td.Outcomes.Create(alice.Address(), bob.Address(), 100, 100)
+	request := virtualfund.ObjectiveRequest{
+
+		CounterParty:      bob.Address(),
+		Intermediary:      irene.Address(),
+		Outcome:           outcome,
+		AppDefinition:     types.Address{},
```
geoknee, line 48: We should make an issue to deal with using the correct `AppDefinition` here. A decent short term solution is to remove it from the API, and have the engine or virtualffund protocol fill in the correct app definition automatically.

client_test/payment_test.go
```diff
@@ -0,0 +1,70 @@
...
+	r := clientA.CreateVirtualChannel(request)
+
+	ids := []protocols.ObjectiveId{r.Id}
+
+	waitTimeForCompletedObjectiveIds(t, &clientA, defaultTimeout, ids...)
+	waitTimeForCompletedObjectiveIds(t, &clientB, defaultTimeout, ids...)
+	waitTimeForCompletedObjectiveIds(t, &clientI, defaultTimeout, ids...)
+	clientA.Pay(r.ChannelId, big.NewInt(5))
+
+	payment := <-clientB.ReceivedVouchers()
```
geoknee, line 63: Nit: if we broke the payment behaviour, the test might stall here. I think one approach to improve the test behaviour is to start a race between a timer and this `chan`. If we don't get the payment before 1s elapses, fail the test. There is something similar here:

payments/vouchers.go
```diff
@@ -11,6 +11,16 @@ import (
...
+// ChannelId returns the channel id of the voucher
+func (v Voucher) ChannelId() types.Destination {
+	return v.channelId
+}
+
+// Amount returns the amount of the voucher
+func (v Voucher) Amount() *big.Int {
+	return v.amount
+}
+
```
geoknee, line 23: Is this a hint that these fields should just be exported / public?

geoknee (approved):

client/engine/engine.go
```diff
@@ -346,43 +384,73 @@ func (e *Engine) handleAPIEvent(apiEvent APIEvent) (ObjectiveChangeEvent, error)
...
+			apiEvent.PaymentToMake.Amount,
+			*e.store.GetChannelSecretKey())
+		if err != nil {
+			return EngineChangeEvent{}, fmt.Errorf("handleAPIEvent: Error making payment: %w", err)
+		}
+		c, ok := e.store.GetChannelById(cId)
+		if !ok {
+			return EngineChangeEvent{}, fmt.Errorf("handleAPIEvent: Could not get channel from the store %s", cId)
+		}
+		sender, recipient := c.Participants[0], c.Participants[2]
```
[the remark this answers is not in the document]
geoknee: I think we also want to check that the channel is ready before sending the voucher. See https://github.com/statechannels/go-nitro/pull/800#discussion_r931904360 . I'm not sure the best place for that check to be, but I believe it is important in preventing race conditions -- essentially we want to be sure that the recipient has registered the channel in their receipt manager.

client/engine/engine.go
lalexgap: I think we should look into that approach. I agree that it would probably be less messy, checking for zero values is a pain.

client/engine/engine.go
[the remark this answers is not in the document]
lalexgap: > I think we also want to check that the channel is ready before sending the voucher. See [#800 (comment)](https://github.com/statechannels/go-nitro/pull/800#discussion_r931904360) . I'm not sure the best place for that check to be, but I believe it is important in preventing race conditions -- essentially we want to be sure that the recipient has registered the channel in their receipt manager.

Right now we use a crude solution to avoid race conditions, we register the virtual channel with the payment/receipt manager when we start creating the virtual channel. This means that we can rely on both participants having registered the virtual channel before any vouchers are sent.

This does mean that our payment/receipt manager will accept vouchers before the virtual channel is ready so we probably need to add a safeguard there.

client/engine/engine.go
```diff
@@ -50,19 +52,31 @@ type Engine struct {
...
 // APIEvent is an internal representation of an API call
 type APIEvent struct {
 	ObjectiveToSpawn protocols.ObjectiveRequest
+	PaymentToMake    PaymentRequest
 }
 
-// ObjectiveChangeEvent is a struct that contains a list of changes caused by handling a message/chain event/api event
-type ObjectiveChangeEvent struct {
+// EngineChangeEvent is a struct that contains a list of changes caused by handling a message/chain event/api event
+type EngineChangeEvent struct {
```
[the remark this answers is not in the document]
lalexgap: `EngineEvent` makes sense to me.

client/engine/engine.go
```diff
@@ -295,11 +312,31 @@ func (e *Engine) handleMessage(message protocols.Message) (ObjectiveChangeEvent,
...
+		if recipient != *e.store.GetAddress() {
+			return EngineChangeEvent{}, fmt.Errorf("not the recipient in channel %s", c.Id)
+		}
+		// TODO: return the amount we paid?
+		_, err := e.rm.Receive(voucher)
+
+		allCompleted.ReceivedVouchers = append(allCompleted.ReceivedVouchers, voucher)
+		if err != nil {
+			return EngineChangeEvent{}, fmt.Errorf("error accepting payment voucher: %w", err)
+		}
```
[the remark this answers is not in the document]
lalexgap: That's a good point! I think right now in `handleMessage` we [generally stop and return an error](https://github.com/statechannels/go-nitro/blob/1edd6b56f293a599570d05c3c5088549dbae6545/client/engine/engine.go#L275) but we should probably consider handling a mix of valid/invalid messages.

I think it might make sense to tackle this work in a separate scope?

client/readme.md
lalexgap: I think I'm missing a `from`
```suggestion
3. The engine reads payments from the `payment manager`, and computes updates and side effects.
```

client/readme.md
```diff
@@ -11,15 +11,16 @@ The flow of data through the client is shown in this diagram:
...
+4. The updates are committed to the `store`.
+5. The side effects are sent on go channels to:
    - the `message` service
    - the `chain` service
    - _back_ to the `engine` (e.g. when an update declares further progress can be made)
```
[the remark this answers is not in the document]
lalexgap: Oops! That was not intentional😐

payments/vouchers.go
lalexgap: I think so! Especially since `Voucher` are exchanged in messages and on chain.

client_test/payment_test.go
lalexgap: That's a good point!

client/engine/engine.go
geoknee: I think when you are trying to detect the case where you are Bob, you want to match against `vfo.myRole` being `n`? `n=2` only if there is a single intermediary.

client/engine/engine.go
```diff
@@ -346,43 +384,73 @@ func (e *Engine) handleAPIEvent(apiEvent APIEvent) (ObjectiveChangeEvent, error)
...
 		default:
-			return ObjectiveChangeEvent{}, fmt.Errorf("handleAPIEvent: Unknown objective type %T", request)
+			return EngineChangeEvent{}, fmt.Errorf("handleAPIEvent: Unknown objective type %T", request)
 		}
 
 	}
 
-	return ObjectiveChangeEvent{}, nil
+	zeroAddress := types.Destination{}
+	// TODO: Should this live in the payment manager?
```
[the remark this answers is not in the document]
lalexgap: I think it would make sense to move this into a manager when we refactor the # [current payment and receipt managers]( https://github.com/statechannels/go-nitro/issues/837) .

I'm thinking that we can have one `VoucherManager` who can handle this logic internally. The `engine` just passes off the voucher to the `VoucherManager` and the `VoucherManager` handles all the rest.

client/engine/engine.go
```diff
@@ -295,11 +312,31 @@ func (e *Engine) handleMessage(message protocols.Message) (ObjectiveChangeEvent,
...
 		allCompleted.CompletedObjectives = append(allCompleted.CompletedObjectives, objective)
 	}
+
+	for _, voucher := range message.Vouchers() {
+
+		c, ok := e.store.GetChannelById(voucher.ChannelId())
+		if !ok {
+			return EngineChangeEvent{}, fmt.Errorf("could not get channel from the store %s", c.Id)
+		}
+		recipient := c.Participants[2]
```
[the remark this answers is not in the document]
lalexgap: We could also expose this as a function `getVirtualPaymentAppRecipient(participants)` so we don't have to deal with the index.

client/engine/engine.go
lalexgap: Ah good point. I think there's probably some other assumptions that `n=2` in our code but I might as well clean this one up!

client/engine/engine.go
lalexgap: updated in fb08bac

client/engine/engine.go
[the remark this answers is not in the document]
lalexgap: I've switched to using a helper function in [5f5c7d0](https://github.com/statechannels/go-nitro/pull/836/commits/5f5c7d092cdef2198ed23068cea3388a1f0ae3de)

client/engine/engine.go
```diff
@@ -346,43 +384,73 @@ func (e *Engine) handleAPIEvent(apiEvent APIEvent) (ObjectiveChangeEvent, error)
 			e.metrics.RecordObjectiveStarted(request.Id(*e.store.GetAddress()))
 			vfo, err := virtualfund.NewObjective(request, true, *e.store.GetAddress(), e.store.GetConsensusChannel)
 			if err != nil {
-				return ObjectiveChangeEvent{}, fmt.Errorf("handleAPIEvent: Could not create objective for %+v: %w", request, err)
+				return EngineChangeEvent{}, fmt.Errorf("handleAPIEvent: Could not create objective for %+v: %w", request, err)
+			}
+
+			err = e.registerVirtualChannelWithManagers(vfo)
```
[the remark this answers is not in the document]
lalexgap: Done in [c542c04](https://github.com/statechannels/go-nitro/pull/836/commits/c542c0480cb72ab3dd83223e93fdfea5b96a0806)

client/engine/engine.go
lalexgap: Done in [0490cd7](https://github.com/statechannels/go-nitro/pull/836/commits/0490cd7480d27e8c67ad81e7f18c9cb7eb093bd5)

client/engine/engine.go
lalexgap: Done in [0490cd7](https://github.com/statechannels/go-nitro/pull/836/commits/0490cd7480d27e8c67ad81e7f18c9cb7eb093bd5)

client/engine/engine.go
```diff
@@ -453,6 +521,24 @@ func (e *Engine) attemptProgress(objective protocols.Objective) (outgoing Object
...
+	prefund := vfo.V.PreFundState()
+	startingBalance := big.NewInt(0)
+	// TODO: Assumes one asset for now
+	startingBalance.Set(prefund.Outcome[0].Allocations[0].Amount)
+
+	switch vfo.MyRole {
+	case 0:
+		return e.pm.Register(vfo.V.Id, startingBalance)
+	case 2:
+		return e.rm.Register(vfo.V.Id, prefund.Participants[0], startingBalance)
```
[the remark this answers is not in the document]
lalexgap: [5f5c7d0](https://github.com/statechannels/go-nitro/pull/836/commits/5f5c7d092cdef2198ed23068cea3388a1f0ae3de)

client/engine/engine.go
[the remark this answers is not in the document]
lalexgap: [e36361e](https://github.com/statechannels/go-nitro/pull/836/commits/e36361e967b182f6811848ea5d88dc33de1e50d7)

client_test/payment_test.go
lalexgap: #839

payments/vouchers.go
lalexgap: I'm going to defer to a separate smaller PR.

client/engine/engine.go
lalexgap: https://github.com/statechannels/go-nitro/issues/842

client/engine/engine.go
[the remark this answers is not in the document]
lalexgap: https://github.com/statechannels/go-nitro/issues/840

client/readme.md
lalexgap: Fixed in https://github.com/statechannels/go-nitro/pull/836/commits/1032e29cb979d9f897f934771fb7e72cfbd32bd0

client/readme.md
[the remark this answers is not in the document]
lalexgap: Fixed in https://github.com/statechannels/go-nitro/pull/836/commits/1032e29cb979d9f897f934771fb7e72cfbd32bd0

client_test/payment_test.go
lalexgap: Done in https://github.com/statechannels/go-nitro/pull/836/commits/8ae50daf4396da777cb2e968edbe47aa29e90af9. There's probably a cleaner way to do this using generics but for now I've mostly duplicated `waitTimeForCompletedObjectiveIds`

PR #844, opened 2022-08-23
Voucher Manager refactor

### LINKED ISSUE #837: Refactor `Payment/Receipt Manager` into one `VoucherManager`
Currently the API for the Payment and Receipt Manager are written for external use. However it makes it slightly awkward to use in the engine(we have to decide which manager to call by inspecting the channel participants).

It would be nice if we could simplify the two managers into one `VoucherManager` that is responsible for managing all received vouchers. This would simplify the wallet and remove some duplicate code.

lalexgap:
Fixes #837 

Refactors the `PaymentManager` and `ReceiptManager` into one `VoucherManager`. 

This simplifies how the engine deals with voucher management as it no longer has to keep track of senders/receivers.

I decided to [completely remove the manager interface](https://github.com/statechannels/go-nitro/pull/844/commits/2b81d027db588f49e8593682ae9ec6f734ffa56a) and just directly reference the `VoucherManager` struct in the `engine` for simplicity. Happy to revert that change if anyone disagrees.


payments/voucher_manager.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// Remove deletes the channel's status
+func (vm *VoucherManager) Remove(channelId types.Destination) {
+	delete(vm.channels, channelId)
+}
+
+// Pay will deduct amount from balance and add it to paid, returning a signed voucher for the
+// total amount paid.
+func (vm *VoucherManager) Pay(channelId types.Destination, amount *big.Int, pk []byte) (Voucher, error) {
```
lalexgap, line 54: It would be nice if this could return `protocol.SideEffects`. Unfortunately since the `payments` package is referenced by the `protocols` package we can't import from the `protocols` package here or we'll get an import cycle.

payments/voucher_manager.go
```diff
@@ -0,0 +1,136 @@
...
+	"fmt"
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/statechannels/go-nitro/types"
+)
+
+// paymentStatus stores the status of payments for a given payment channel.
+type paymentStatus struct {
+	channelSender   common.Address
```
geoknee, line 13: Nit: I would prefer payer / payee over sender / receiver?

payments/voucher_manager.go
```diff
@@ -0,0 +1,136 @@
...
+	me       common.Address
+}
+
+// NewVoucherManager creates a new voucher manager
+func NewVoucherManager(me types.Address) *VoucherManager {
+	channels := make(map[types.Destination]*paymentStatus)
+	return &VoucherManager{channels, me}
+}
+
+// Register registers a channel for use, given the sender and starting balance of the channel
```
geoknee, line 32: This comment needs updating to reflect the inclusion of the receiver in the arguments.

payments/voucher_manager.go
```diff
@@ -0,0 +1,136 @@
...
+	if !ok {
+		return voucher, fmt.Errorf("channel not found")
+	}
+
+	if types.Gt(amount, pStatus.currentBalance.Remaining) {
+		return voucher, fmt.Errorf("unable to pay amount: insufficient funds")
+	}
+
+	pStatus.currentBalance.Remaining.Sub(pStatus.currentBalance.Remaining, amount)
+	pStatus.currentBalance.Paid.Add(pStatus.currentBalance.Paid, amount)
```
geoknee, line 67: Do we need to update `pStatus.voucher` here?

payments/voucher_manager.go
```diff
@@ -0,0 +1,136 @@
...
+
+	voucher.Amount.Set(pStatus.currentBalance.Paid)
+	voucher.ChannelId = channelId
+
+	if err := voucher.Sign(pk); err != nil {
+		return voucher, err
+	}
+
+	// question: is there a more efficient way to validate the signature against the purported signer?
```
geoknee, line 76: How about we just check that the payer/sender of the channel (according to `pStatus` ) is the `me` of the voucher manager? I don't think we need to protect against the wrong private key being used, except perhaps when debugging. It's seems wrong to do this cryptographic operation on every payment, just in case we made a programming mistake?

payments/voucher_manager.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// Pay will deduct amount from balance and add it to paid, returning a signed voucher for the
+// total amount paid.
+func (vm *VoucherManager) Pay(channelId types.Destination, amount *big.Int, pk []byte) (Voucher, error) {
+	pStatus, ok := vm.channels[channelId]
+
+	voucher := Voucher{Amount: &big.Int{}}
+	if !ok {
+		return voucher, fmt.Errorf("channel not found")
```
geoknee, line 59: Personally I prefer to see the anonymous literal zero type being returned here, rather than `voucher`. I
When reading the code I have to do a bit more mental work to understand what is happening. Nit.

geoknee (approved):
LGTM happy for the interface to go.

payments/voucher_manager.go
lalexgap: Good catch!

payments/voucher_manager.go
lalexgap: I think that makes sense. This comment and code was copied over from the initial `ReceiptManager` implementation

payments/voucher_manager.go
lalexgap: Done in https://github.com/statechannels/go-nitro/pull/844/commits/659ea22d1d8624973eb4e08cc0f41515b2744c78

payments/voucher_manager.go
lalexgap: Done in https://github.com/statechannels/go-nitro/pull/844/commits/659ea22d1d8624973eb4e08cc0f41515b2744c78

payments/voucher_manager.go
lalexgap: Done in [1d20e2c](https://github.com/statechannels/go-nitro/pull/844/commits/1d20e2cf1d573c92c4d2328c50890416f4181961)

payments/voucher_manager.go
lalexgap: Fixed in https://github.com/statechannels/go-nitro/pull/844/commits/c76786eec9851a27b5642c92e79468a072a3fce3

payments/voucher_manager.go
lalexgap: Done in https://github.com/statechannels/go-nitro/pull/844/commits/4a60e94ee60f75fdd7c2472725b66f0cc6aefd88

PR #845, opened 2022-08-25
split `fromAPI` chan into `ObjectiveRequests` and `Payments` chans

### LINKED ISSUE #842: Replace intersection type with several api methods
Somewhat out of scope -- this zero value check seems a bit awkward and is probably a consequence of us approximating union types with an intersection type. We have several api methods, a **single** type for all of them, and then several handlers for that type. That single type therefore has to have a bunch of zero values on it. 

What if we had a single type per method and handler? It would make the master engine "run loop" more verbose, but might be more efficient and less messy overall.

_Originally posted by @geoknee in https://github.com/statechannels/go-nitro/pull/836#discussion_r952520633_

geoknee:
- the client API methods become even simpler  due to less indirection / the removal of APIEvent
- the engine.handleAPI method is split into two, increasing modularity and clarity. The payment handler does not need to return nul EngineEvents
- NOTE we now panic if we encounter nil requests (this should now never happen, since we no longer use intersection types)

Closes #842 . I decided against splitting into one `chan` per API method, since this is currently handled with an interface and type switch -- a pattern that seems OK to me.

### DIFF
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -416,7 +416,7 @@ func (e *Engine) handleObjectiveRequest(or protocols.ObjectiveRequest) (EngineEv
 
 }
 
-// handleObjectiveRequest handles an PaymentRequest (triggered by a client API call).
+// handlePaymentRequest handles an PaymentRequest (triggered by a client API call).
 // It prepares and dispatches a payment message to the counterparty.
 func (e *Engine) handlePaymentRequest(request PaymentRequest) error {
 	if (request == PaymentRequest{}) {
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -365,10 +365,6 @@ func (e *Engine) handleChainEvent(chainEvent chainservice.Event) (EngineEvent, e
 func (e *Engine) handleObjectiveRequest(or protocols.ObjectiveRequest) (EngineEvent, error) {
 	defer e.metrics.RecordFunctionDuration()()
 
-	if or == nil {
-		panic("tried to handle nil objective request")
-	}
-
 	switch request := or.(type) {
 
 	case virtualfund.ObjectiveRequest:
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -36,7 +36,11 @@ func (uce *ErrUnhandledChainEvent) Error() string {
 // Engine is the imperative part of the core business logic of a go-nitro Client
 type Engine struct {
 	// inbound go channels
-	FromAPI    chan APIEvent // This one is exported so that the Client can send API calls
+
+	// From API
+	ObjectiveRequestsFromAPI chan protocols.ObjectiveRequest
+	PaymentRequestsFromAPI   chan PaymentRequest
+
 	fromChain  <-chan chainservice.Event
 	fromMsg    <-chan protocols.Message
 	fromLedger chan consensus_channel.Proposal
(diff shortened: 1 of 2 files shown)


client/engine/engine.go
```diff
@@ -36,7 +36,11 @@ func (uce *ErrUnhandledChainEvent) Error() string {
 // Engine is the imperative part of the core business logic of a go-nitro Client
 type Engine struct {
 	// inbound go channels
-	FromAPI    chan APIEvent // This one is exported so that the Client can send API calls
+
+	// From API
```
lalexgap, line 40: Outside the scope of this PR but are we using `chan` to communicate with the client API to allow us to easily scale to more engines?

client/engine/engine.go
```diff
@@ -358,92 +360,89 @@ func (e *Engine) handleChainEvent(chainEvent chainservice.Event) (EngineEvent, e
...
 
+	default:
+		return EngineEvent{}, fmt.Errorf("handleAPIEvent: Unknown objective type %T", request)
 	}
-	return EngineEvent{}, nil
 
 }
 
-// executeSideEffects executes the SideEffects declared by cranking an Objective
+// handleObjectiveRequest handles an PaymentRequest (triggered by a client API call).
```
lalexgap, line 419:
```suggestion
// handlePaymentRequest handles an PaymentRequest (triggered by a client API call).
```

client/engine/engine.go
```diff
@@ -358,92 +360,89 @@ func (e *Engine) handleChainEvent(chainEvent chainservice.Event) (EngineEvent, e
...
-		se := protocols.SideEffects{MessagesToSend: protocols.CreateVoucherMessage(voucher, payee)}
-		err = e.executeSideEffects(se)
+		return e.attemptProgress(&dfo)
+
+	case directdefund.ObjectiveRequest:
+		e.metrics.RecordObjectiveStarted(request.Id(*e.store.GetAddress()))
+		ddfo, err := directdefund.NewObjective(request, true, e.store.GetConsensusChannelById)
 		if err != nil {
-			return EngineEvent{}, fmt.Errorf("handleAPIEvent: Error sending payment voucher: %w", err)
+			return EngineEvent{FailedObjectives: []protocols.ObjectiveId{request.Id(*e.store.GetAddress())}}, fmt.Errorf("handleAPIEvent: Could not create objective for %+v: %w", request, err)
```
lalexgap, line 407: nit but we could break this up into multiple lines for visibility

lalexgap (approved):
🚀

client/engine/engine.go
```diff
@@ -358,92 +360,89 @@ func (e *Engine) handleChainEvent(chainEvent chainservice.Event) (EngineEvent, e
...
-			return e.attemptProgress(&vdfo)
-
-		case directfund.ObjectiveRequest:
-			e.metrics.RecordObjectiveStarted(request.Id(*e.store.GetAddress()))
-			dfo, err := directfund.NewObjective(request, true, *e.store.GetAddress(), e.store.GetChannelsByParticipant, e.store.GetConsensusChannel)
-			if err != nil {
-				return EngineEvent{}, fmt.Errorf("handleAPIEvent: Could not create objective for %+v: %w", request, err)
-			}
-			return e.attemptProgress(&dfo)
+	if or == nil {
```
lalexgap, line 368: I would be in favour of just removing this if we don't expect it to happen. If `or` is `nil` the standard nil pointer error would still be enough to figure out the problem.

client/engine/engine.go
geoknee: I think the pattern will allow us to handle concurrency easily. I don't know if that means more engines, necessarily -- perhaps more request handlers (running in goroutines). 

If we exposed the engine's handler functions directly and had the client call those synchronously, the client (and therefore the consuming application) would block until that handler had completed. Using a `chan` means the request can be dispatched asynchronously.

client/engine/engine.go
geoknee: I'm going to tackle this in a small follow-up PR.

geoknee: #850

PR #846, opened 2022-08-25
experiment with splitting `engine.go` into smaller files

geoknee:
This is just an experiment in organisation. I'm not sure the marginal improvements (if any) are worth it? Opinions welcome.


lalexgap:
I think I'm in favour of splitting up code into seperate files but I think 8 files might be too many?

I tend to navigate by using the vscode command palette to open a file by filename and then search the file for the code I'm interested in. So I think that biases me towards larger files with filenames that I can remember 😛 

That said I don't feel too strongly about it, so if others prefer this I'm happy to go along with it. 👍

lalexgap:
I think this is also a sign that the `engine` is doing too much/has too much responsibility.

I wonder if we could delegate some of the work to the `VoucherManager` and maybe some kind of `ObjectiveManager` package?

geoknee:
If we want fewer files, I would suggest we keep one file per _handler_ and then having `engine.go` contain everything else. That gives us 6 files, and the five "additional" files nicely mirrors the five inbound `chans` to the engine. 

This kind of refactor does mean that you may need to switch from `engine.go` to `handleFoo.go` and back to `engine.go` to understand the execution flow. 

Let's see if we can get some more opinions from @kerzhner @NiloCK .

kerzhner:
No strong opinions from my side. I tend to use VSCode Outline to wrap my head around and navigate larger files
Screen Shot 2022-08-26 at 3 51 35 PM

geoknee:
I think we should probably default to not doing this change unless someone feels strongly in favour.

PR #848, opened 2022-08-25
Add off-chain mirror code for `computeReclaimEffects`

### LINKED ISSUE #832: add off-chain mirror code for `computeReclaimEffects` (in Go)
We have a typescript version here https://github.com/statechannels/go-nitro/blob/9aa563ce9ca03a05d3be5a722de2059c64adb7af/nitro-protocol/src/contract/multi-asset-holder.ts#L60-L126 

We should transpile it to Go so that a `go-nitro` client can track on-chain changes to a ledger channel as `reclaim` is called.

geoknee:
Closes #832


geoknee:
I think we haven't considered that. It's something we can look into for sure -- I'm not sure how possible / easy that would be. One approach would be to use a geth `SimulatedBackEnd`, but that seems overly clunky.

kerzhner (approved):
LGTM

client/engine/chainservice/adjudicator/reclaim.go
```diff
@@ -0,0 +1,59 @@
...
+			continue
+		}
+		newSourceAllocations[k] = outcome.Allocation{
+			Destination:    sourceAllocations[i].Destination,
+			Amount:         sourceAllocations[i].Amount,
+			AllocationType: sourceAllocations[i].AllocationType,
+			Metadata:       sourceAllocations[i].Metadata,
+		}
+		if sourceAllocations[i].Destination == guaranteeData.Left {
+			newSourceAllocations[k].Amount.Add(newSourceAllocations[k].Amount, targetAllocations[0].Amount)
```
kerzhner, line 38: Does this work if destinations are repeated in `sourceAllocation`?

client/engine/chainservice/adjudicator/reclaim.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"fmt"
+
+	"github.com/statechannels/go-nitro/channel/state/outcome"
+)
+
+// computeReclaimEffects mirrors on chain code.
+// It computes side effects for the reclaim function. Returns updated allocations for the source, computed by finding the guarantee in the source for the target, and moving money out of the guarantee and back into the ledger channel as regular allocations for the participants.
+func computeReclaimEffects(sourceAllocations []outcome.Allocation, targetAllocations []outcome.Allocation, indexOfTargetInSource uint) ([]outcome.Allocation, error) {
```
kerzhner, line 11: It seems that the amount a source channel allocates to a target channel is never checked. Is this a concern for underfunded channels?

geoknee: Good catch! I think this check is missing from the satp paper (which we are using as a spec for this work). 

I think we need to add that check!

client/engine/chainservice/adjudicator/reclaim.go
geoknee: If destinations are repeated, this could lead to a problem here (multiple entries will be incremented). 
We can 
1. add client checks about not having repeated destinations
2. add a `continue` statement on L39 and another on L43
3. revert the transaction if the total amount reclaimed is too high (see your other comment thread)

geoknee: 2. EDIT `continue` statements won't do it. But we could check `!foundLeft` on L37 etc.

Issue #855, opened 2022-09-05
Create ts helper to expect a transaction to succeed


nksazonov:
While testing contracts using jest it is a common thing to expect a transaction, which does not return anything, to succeed. Thus, a helper, alike `expectRevert` is needed.

Some possible implementations:
- checking returned object/array to be empty
``` ts
async function expectSucceed(fn: () => void) {
  const txResult = (await fn()) as any;
  
  // As 'requireStateSupported' method is constant (view or pure), if it succeeds, it returns an object with returned values
  // which in this case should be empty
  expect(txResult.length).toBe(0);
}
```

- `try/catch/finally`
``` ts
async function expectSucceed(fn: () => void) {
  try {
    await fn();
  } catch (e) {
    // ?
  } finally {
    assert(true);
  }
}
```

nksazonov:
`expectRevert` is defined in `@statechannels/devtools`, thus implementing `expectSucceed` there seems logical.

geoknee:
Nit: I think in these tests we are actually not executing a _transaction_. If we were, we could inspect the transaction receipt for clues about whether it was successful. 

In fact we are actually using the `eth_call` endpoint, which generally returns a value instead of a transaction receipt. In this case as you say, the returned value is empty because in solidity we do not return anything from the function. 

Another approach you could use, which we have used in the past, is to force the execution of an actual transaction to test the method.

geoknee:
Yet another way, which might be simplest, would be to just call the method without using any special helpers. And add a comment each time to explain that should the call `revert`, an error will be thrown and the test will be failed.

nksazonov:
> Yet another way, which might be simplest, would be to just call the method without using any special helpers. And add a comment each time to explain that should the call `revert`, an error will be thrown and the test will be failed.

Yes, I agree this method is the simplest. Isn't not executing any assertion explicitly in the test and relying on thrown error if anything goes wrong a bad practice?

geoknee:
Perhaps... in that case, 

> checking returned object/array to be empty

seems the better of your two proposals. I'm not sure it is worth adding this to @statechannels/devtools right now.

nksazonov:
Okay will add the method `expectSucceed` to `go-nitro/nitro-protocol/src` then.

Issue #849, opened 2022-08-26
Extract `participants` and `wallets` generation from tests to a helper function


nksazonov:
Almost all nitro protocol contract test files duplicate the same code:

``` typescript
const participants = ['', '', ''];
const wallets = new Array(3);

// Populate wallets and participants array
for (let i = 0; i < 3; i++) {
  wallets[i] = new Wallet(keys[i]);
  participants[i] = wallets[i].address;
}
```

It would be logical to extract it into a helper function.

PR #859, opened 2022-09-06
Add tests for `StrictTurnTaking.sol` and `Consensus.sol`

### LINKED ISSUE #768: Add tests for `StrictTurnTaking` and `Consensus`

nksazonov:
Closes #768, #849 and #855.


nitro-protocol/src/contract/transaction-creators/revert-reasons.ts
```diff
@@ -19,5 +19,10 @@ export const WRONG_REFUTATION_SIGNATURE = 'Refutation state not signed by challe
 export const NONFINAL_STATE = 'State must be final';
 export const MUST_SUBMIT_EXACTLY_1_STATE = 'Must submit exactly 1 state';
 
+// Application-specifis messages
+export const COUNTING_APP_INVALID_TRANSITION = 'Counter must be incremented';
+
 // Turn Taking
 export const SIGNED_BY_NON_MOVER = 'Invalid signedBy';
+export const INVALID_NUMBER_OF_PROOF = 'Insufficient or excess proof states';
```
geoknee: `INVALID_NUMBER_OF_PROOF_STATES`?

nitro-protocol/test/contracts/libraries/StrictTurnTaking/StrictTurnTaking.test.ts
```diff
@@ -0,0 +1,172 @@
...
+});
+
+let channelNonce = getRandomNonce('StrictTurnTaking');
+beforeEach(() => (channelNonce += 1));
+
+describe('isSignedByMover', () => {
+  const accepts1 = 'should not revert when signed only by mover';
+
+  const reverts1 = 'should revert when not signed by mover';
+  const INVALID_SIGNED_BY = SIGNED_BY_NON_MOVER;
```
geoknee: Why introduce this alias?

nitro-protocol/test/contracts/libraries/StrictTurnTaking/StrictTurnTaking.test.ts
```diff
@@ -0,0 +1,172 @@
...
+    }) => {
+      expect(await StrictTurnTaking.moverAddress(participants, turnNum)).toEqual(
+        wallets[expectedParticipantIdx].address
+      );
+    }
+  );
+});
+
+describe('requireValidInput', () => {
+  const accepts1 = 'accept when all rules are preserved';
```
geoknee, line 123: all rules are obeyed / satisfied?

nitro-protocol/contracts/libraries/signature-logic/StrictTurnTaking.sol
```diff
@@ -86,10 +86,13 @@ library StrictTurnTaking {
...
-     * @param numStates Number of states submitted.
+     * @param numProofStates Number of proof states submitted.
      */
-    function _requireValidInput(uint256 numParticipants, uint256 numStates) internal pure {
-        require((numParticipants >= numStates) && (numStates > 1), 'Insufficient or excess states');
+    function _requireValidInput(uint256 numParticipants, uint256 numProofStates) internal pure {
+        require(
+            (numProofStates > 0) && (numParticipants == numProofStates + 1),
+            'Insufficient or excess proof'
+        );
```
nksazonov, line 95: Do we ever need to check that there are more than one participant in the channel?
Is there a way to abuse creating channels with one participant?

nitro-protocol/src/signatures.ts
```diff
@@ -59,6 +62,91 @@ export async function sign(wallet: Wallet, msgHash: string | Uint8Array): Promis
   return utils.splitSignature(await wallet.signMessage(utils.arrayify(msgHash)));
 }
 
+// Towards #761 https://github.com/statechannels/go-nitro/issues/761
+//
+export type ShortenedVariablePart =
```
nksazonov, line 67: Maybe better naming is possible...
Options were `ShortVariablePart`, `VariablePartCore`, etc.

nitro-protocol/src/signatures.ts
```diff
@@ -59,6 +62,91 @@ export async function sign(wallet: Wallet, msgHash: string | Uint8Array): Promis
...
+  | {
+      signerIndices: number[];
+      appData?: Bytes;
+      outcome?: Outcome;
+      isFinal?: boolean;
+    }
+  | number[];
+
+// Map<turnNum => ShortenedVariablePart> | Map<turnNum => signerIndices>
+export type TurnNumToShortenedVariablePart = Map<number, ShortenedVariablePart>;
```
nksazonov, line 77: Here naming becomes even worse... If to use the current `ShortenedVariablePart` name, it can be `ShortenedVariablePartShortHand` which is too confusing.
Any suggestions are hugely appreciated 😄

nitro-protocol/test/test-helpers.ts
```diff
@@ -58,6 +58,26 @@ export function getCountingAppContractAddress(): string {
 
 export const nonParticipant = ethers.Wallet.createRandom();
 
+/**
+ * Generate `n` wallets and return them alongside with array of their addresses.
+ * @param n Number of wallets to create.
+ * @returns N wallets with their addresses.
+ */
+export function generateParticipants(n: number): {
```
geoknee, line 66: 👍

nitro-protocol/test/expect-succeed.ts
```diff
@@ -0,0 +1,7 @@
+export async function expectSucceed(fn: () => void) {
+  const txResult = (await fn()) as any;
+
+  // As 'requireStateSupported' method is constant (view or pure), if it succeedes, it returns an object/array with returned values
+  // which in this case should be empty
+  expect(txResult.length).toBe(0);
+}
```
geoknee, line 7: 👍 Good idea to encapsulate this -- if the ethers.js API / behaviour changes, we only need to update this one file.

nitro-protocol/src/contract/transaction-creators/revert-reasons.ts
```diff
@@ -26,3 +26,5 @@ export const INVALID_SIGNED_BY = 'Invalid signedBy';
 export const INVALID_NUMBER_OF_PROOF = 'Insufficient or excess proof states';
 export const TOO_MANY_PARTICIPANTS = 'Too many participants';
 export const WRONG_TURN_NUM = 'Wrong variablePart.turnNum';
+export const NOT_UNANIMUS = '!unanimous';
```
geoknee, line 29: `NOT_UNANIMOUS`

nitro-protocol/test/contracts/libraries/Consensus/Consensus.test.ts
```diff
@@ -0,0 +1,103 @@
...
+import {separateProofAndCandidate} from '../../../../src/contract/state';
+import {expectSucceed} from '../../../expect-succeed';
+const provider = getTestProvider();
+let Consensus: Contract & TESTConsensus;
+
+const chainId = process.env.CHAIN_NETWORK_ID;
+const challengeDuration = 0x1000;
+const asset = Wallet.createRandom().address;
+const defaultOutcome: Outcome = [{asset, allocations: [], metadata: '0x'}];
+const appDefinition = getCountingAppContractAddress();
```
geoknee, line 35: Do we want to use the counting app when checking the consensus app?

geoknee (approved):
Thanks @nksazonov ! This was a long PR but I think there's some good cleanup work here. left a few comments that need addressing.

nitro-protocol/contracts/test/TESTStrictTurnTaking.sol
```diff
@@ -0,0 +1,64 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.7.6;
+pragma experimental ABIEncoderV2;
+
+import '../interfaces/INitroTypes.sol';
+import {StrictTurnTaking} from '../libraries/signature-logic/StrictTurnTaking.sol';
+
+/**
+ * @dev This contract uses the StrictTurnTaking library to make it more easily unit-tested. It exposes public or external functions call into internal functions. It should not be deployed in a production environment.
```
geoknee, line 9:
```suggestion
 * @dev This contract uses the StrictTurnTaking library to make it more easily unit-tested. It exposes public or external functions which call into internal functions. It should not be deployed in a production environment.
```

and similarly for the other `TEST*.sol` files.

nitro-protocol/contracts/libraries/signature-logic/StrictTurnTaking.sol
geoknee: I can't think of any attacks that exploit that. It is unexpected, but I don't think we want to expend gas checking for it unless we know of a problem?

nitro-protocol/src/signatures.ts
geoknee: I think what you have is OK, actually (can't think of anything better just now!)

nitro-protocol/contracts/libraries/signature-logic/StrictTurnTaking.sol
nksazonov: Can't think of such a problem. I agree we don't want to spend gas for unnecessary checks, will proceed without them then.

nitro-protocol/src/bitfield-utils.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+
+export function getSignersIndices(signedByStr: Uint256): number[] {
+  let signedBy = parseFloat(signedByStr);
+
+  const signerIndices: number[] = [];
+  let signerNum = 0;
+  let acceptedSigners = 0;
+
+  while (signedBy > 0) {
```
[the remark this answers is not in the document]
nksazonov: No reason, just copy paste from `NitroUtils.sol` 🙈 
Good remark to follow the same code style, though.
Will replace `while` with `for` in both `bitfield-utils.ts` and `NitroUtils.sol`.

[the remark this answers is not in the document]
nksazonov: 4b2d0bfb66096525333d9fe42ae935a1ef5af0f8

nitro-protocol/test/contracts/libraries/StrictTurnTaking/StrictTurnTaking.test.ts
nksazonov: It was removed in the following commits.

nitro-protocol/test/contracts/libraries/StrictTurnTaking/StrictTurnTaking.test.ts
nksazonov: b383266b70f1e6200d313aa0e651b1d148a02ff9

nitro-protocol/src/contract/transaction-creators/revert-reasons.ts
nksazonov: f2d4e54c542b2e4c2cbb59c457053162f8424afa

nitro-protocol/src/contract/transaction-creators/revert-reasons.ts
nksazonov: Nice catch! I hate ytpos 😛

nitro-protocol/test/contracts/libraries/Consensus/Consensus.test.ts
nksazonov: The address does not play any role here, except for creating the `FixedPart`. It seems more logical to use consensus app address here though.

nitro-protocol/src/contract/transaction-creators/revert-reasons.ts
nksazonov: 86f58b8b97047b304243717b604ee8d6b043097e

nitro-protocol/test/contracts/libraries/Consensus/Consensus.test.ts
nksazonov: 36912f830aed69c22bd8f7c23c9c0405eb88f2b9

nitro-protocol/contracts/test/TESTStrictTurnTaking.sol
nksazonov: d3aeefb544fa6f65d978f5b14e4c0b01ec4568b0

PR #826, opened 2022-08-16
Exchange Vouchers in Virtual Defunding

lalexgap:
Fixes #795

This PR introduces vouchers into to the virtual defunding objective.

In a previous [PR](https://github.com/statechannels/go-nitro/pull/836) we enabled the exchange of vouchers for payments. This work builds on that to use the stored voucher to figure out the amount to close the payment channel with. 

The virtual defunding protocol has been updated to handle exchanging the latest voucher so every participant has the voucher and can calculate the outcome. Since Alice is the one signing vouchers the virtual defunding protocol now requires Alice to send out the latest voucher to everyone before moving on to signing the state. This does mean that virtual defunding assumes Alice has the latest voucher (since she is the one authoring them).

Since messages can arrive out of order we also have to handle the case where we receive a signed state for Alice before we receive the latest voucher from her. This means we have to support creating an objective from both a signed state or a voucher.

I think this work is at a point where it can be merged in and things like additional testing can be tackled in separate PRs.

## Testing

Currently this relies on existing tests and a new [payment integration test](https://github.com/statechannels/go-nitro/blob/engine-pm-integration/client_test/payment_test.go). This integration test is pretty simple it:
- Funds and opens ledger channels
- Creates one virtual channel
- Alice calls `Pay` twice using the virtual channel
- Alice calls close to close the virtual channel
- We inspect the ledger channel outcomes to make sure they're updated correctly.

**EDIT** The test has been updated to do this for multiple virtual channels

We could also update some of the existing `virtual defund` integration tests to call `Pay`.

## TODO
- [ ] More thorough testing
- [x]  Update payment integration test to check closing the channel
- [x] clean up commit history

PR #863, opened 2022-09-07
Add architectural decision record for the virtual defund protocol

geoknee:
This describes our latest thoughts following discussions raised by #826, and describes a simpler solution we hope to implement soon.


lalexgap:
Thanks for doing this! Based on our discussion in the notion doc this looks good to me!

.adr/0007-virtualdefund-protocol.md
```diff
@@ -0,0 +1,69 @@
...
+
+Bob’s client will get this message and trigger a verification procedure. All he cares about is that the adjusted outcome is **as good or better as** he expects given _his_ latest voucher. So if the special message comes in ahead of the actual latest voucher, he will be pleasantly surprised and continue with the protocol. If a malicious message comes in (Alice knowingly sending a low voucher), he will ignore it.
+
+At this point Alice and Bob have agreed on a final state with an appropriate adjustment. Irene did not need to be involved in that. Bob sends his signature and Alice’s signature to Irene. When she has both signatures, she adds her own and broadcasts to Alice and Bob.
```
lalexgap, line 40: Could Irene optimistically sign the final state before Bob validates the latest amount? It seems more efficient to have Alice broadcast the signed final state to both Bob and Irene at the same time. Irene can blindly sign it and then wait on Bob to validate and sign it as well.

If Alice is malicious I don't think it matters if Irene signs the state,  as we can rely on Bob to verify the state and only sign it if the amount is big enough.

geoknee: Interesting question. This got me thinking about what the purpose of Irene's signature on the final state is. 

Irene cares about the following things:
1. She must be able to force the channel to close, since here money "points at it" in two ledger channels, and she may need to `reclaim` it.
2. The channel must close with a **reclaimable** outcome. That is, one which allocates the same total funds to the same destinations as the postfund state. if there is no money allocated, no money can be `reclaimed`. If the expected destinations aren't in the outcome, `reclaim` will revert:
https://github.com/statechannels/go-nitro/blob/c5e051a95ded01ce51017809bd84a5192391259f/nitro-protocol/contracts/MultiAssetHolder.sol#L377
3. When the channel closes, she must be certain about which outcome will "win" on chain. Otherwise, she might update one ledger channel based on a voucher of `x`, and then see the other ledger channel challenged as well as `V` challenged with a voucher of `y`. She can lose money this way. Once she gets a full set of signatures on a final state, she has this certainty since final states trump all other states. That's true both when challenging:
https://github.com/statechannels/go-nitro/blob/c5e051a95ded01ce51017809bd84a5192391259f/nitro-protocol/contracts/VirtualPaymentApp.sol#L37-L40
(since the final turnnum is greatest), and when shortcutting with `conclude`:
https://github.com/statechannels/go-nitro/blob/c5e051a95ded01ce51017809bd84a5192391259f/nitro-protocol/contracts/ForceMove.sol#L146
Therefore she can happily proceed to defund off chain with voucher `x`, since she knows she can force this on chain. 

So I think Irene definitely needs to check for reclaim-ability before signing `V.isFinal`. And one she has signed `V.isFinal`, she should never sign it again!

I think none of this prevents your idea of optimistic signing. I've also spotted that Alice never receives Bob's `V.isFinal` signature. I'll update the ADR.

geoknee: [e0a336b](https://github.com/statechannels/go-nitro/pull/863/commits/e0a336bf031d450edf4d7f6f010df8b11fbb61fc)

geoknee: We save an entire round of network comms using your suggestion.

lalexgap (approved):

PR #866, opened 2022-09-08
Implement virtual defunding per ADR

lalexgap:
This roughly implements the virtual defunding protocol as specified in the [ADR](https://github.com/statechannels/go-nitro/pull/863).  When Bob or Irene triggers virtual defunding the protocol will send a `FinalStateRequestPayload` message to Alice requesting the final state.

Note that this implementation differs slightly from the ADR around signature communication. In the ADR Alice sends her signed state to Bob and then Bob forwards to Alice. In this implementation Alice simply broadcasts her signed state to all participants.  See https://github.com/statechannels/go-nitro/pull/863#discussion_r966444821 

The virtual defunding objective struct has two new fields on it:
```go

	// FinalOutcome is the final outcome of the virtual channel from Alice
	FinalOutcome outcome.SingleAssetExit

	// MinimumPaymentAmount is the latest payment amount we have received from Alice before starting defunding.
	// This is set by Bob so he can ensure he receives the latest amount from any vouchers he's received.
	// If this is not set then virtual defunding will accept any final outcome from Alice.
	MinimumPaymentAmount *big.Int

``` 

`FinalOutcome` is the outcome of the final state signed by Alice. For Bob and Irene this starts as `outcome.SingleAssetExit{}` and gets populated when Alice sends her signed final state.

`MinimumPaymentAmount` is the minimum payment amount Bob expects to receive. This gets set to the latest voucher Bob has whenever he starts virtual defunding. When Bob receives a final state from Alice he makes sure the payment amount is at least `MinimumPaymentAmount`.

The virtual defund protocol has a new `waitingFor` state `	WaitingForFinalStateFromAlice`. Bob and Irene enter this if before they have received the final state from Alice.

The virtual defunding integration test has been modified with two important changes:
- It calls `clientA.Pay` to actually send a payment to Bob.
- It tests all participants calling `client*.Close` not just Alice

### DIFF
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -372,7 +372,11 @@ func (e *Engine) handleObjectiveRequest(or protocols.ObjectiveRequest) (EngineEv
 		return e.attemptProgress(&vfo)
 
 	case virtualdefund.ObjectiveRequest:
-		vdfo, err := virtualdefund.NewObjective(request, true, myAddress, e.store.GetChannelById, e.store.GetConsensusChannel)
+		bal, err := e.vm.Balance(request.ChannelId)
+		if err != nil {
+			return EngineEvent{}, fmt.Errorf("handleAPIEvent: Failed trying to get latest voucher balance %+v: %w", request, err)
+		}
+		vdfo, err := virtualdefund.NewObjective(request, true, myAddress, bal.Paid, e.store.GetChannelById, e.store.GetConsensusChannel)
 		if err != nil {
 			return EngineEvent{}, fmt.Errorf("handleAPIEvent: Could not create objective for %+v: %w", request, err)
 		}
@@ -573,7 +577,15 @@ func (e *Engine) constructObjectiveFromMessage(id protocols.ObjectiveId, p proto
 		}
 		return &vfo, nil
 	case virtualdefund.IsVirtualDefundObjective(id):
-		vdfo, err := virtualdefund.ConstructObjectiveFromPayload(p, false, *e.store.GetAddress(), e.store.GetChannelById, e.store.GetConsensusChannel)
+		vId, err := virtualdefund.GetVirtualChannelFromId(id)
+		if err != nil {
+			return &virtualdefund.Objective{}, fmt.Errorf("could not determine virtual channel id: %w", err)
+		}
+		bal, err := e.vm.Balance(vId)
+		if err != nil {
+			return &virtualdefund.Objective{}, fmt.Errorf("could determine voucher balance: %w", err)
+		}
+		vdfo, err := virtualdefund.ConstructObjectiveFromPayload(p, false, *e.store.GetAddress(), e.store.GetChannelById, e.store.GetConsensusChannel, bal.Paid)
 		if err != nil {
 			return &virtualfund.Objective{}, fmt.Errorf("could not create virtual fund objective from message: %w", err)
 		}
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -577,7 +577,7 @@ func (e *Engine) constructObjectiveFromMessage(id protocols.ObjectiveId, p proto
 		}
 		return &vfo, nil
 	case virtualdefund.IsVirtualDefundObjective(id):
-		vId, err := virtualdefund.GetVirtualChannelFromId(id)
+		vId, err := virtualdefund.GetVirtualChannelFromObjectiveId(id)
 		if err != nil {
 			return &virtualdefund.Objective{}, fmt.Errorf("could not determine virtual channel id: %w", err)
 		}
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -583,7 +583,7 @@ func (e *Engine) constructObjectiveFromMessage(id protocols.ObjectiveId, p proto
 		}
 		bal, err := e.vm.Balance(vId)
 		if err != nil {
-			return &virtualdefund.Objective{}, fmt.Errorf("could determine voucher balance: %w", err)
+			return &virtualdefund.Objective{}, fmt.Errorf("could not determine voucher balance: %w", err)
 		}
 		vdfo, err := virtualdefund.ConstructObjectiveFromPayload(p, false, *e.store.GetAddress(), e.store.GetChannelById, e.store.GetConsensusChannel, bal.Paid)
 		if err != nil {
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -365,18 +365,26 @@ func (e *Engine) handleObjectiveRequest(or protocols.ObjectiveRequest) (EngineEv
 		if err != nil {
 			return EngineEvent{}, fmt.Errorf("handleAPIEvent: Could not create objective for %+v: %w", request, err)
 		}
-		err = e.registerPaymentChannel(vfo)
+		// Only Alice or Bob care about registering the objective and keeping track of vouchers
+		if vfo.MyRole == payments.PAYEE_INDEX || vfo.MyRole == payments.PAYER_INDEX {
+			err = e.registerPaymentChannel(vfo)
+			if err != nil {
+				return EngineEvent{}, fmt.Errorf("could not register channel with payment/receipt manager: %w", err)
+			}
+		}
+
 		if err != nil {
 			return EngineEvent{}, fmt.Errorf("could not register channel with payment/receipt manager: %w", err)
 		}
 		return e.attemptProgress(&vfo)
 
 	case virtualdefund.ObjectiveRequest:
-		bal, err := e.vm.Balance(request.ChannelId)
-		if err != nil {
-			return EngineEvent{}, fmt.Errorf("handleAPIEvent: Failed trying to get latest voucher balance %+v: %w", request, err)
+		minAmount := big.NewInt(0)
+		if e.vm.ChannelRegistered(request.ChannelId) {
+			bal, _ := e.vm.Balance(request.ChannelId)
+			minAmount = bal.Paid
 		}
-		vdfo, err := virtualdefund.NewObjective(request, true, myAddress, bal.Paid, e.store.GetChannelById, e.store.GetConsensusChannel)
+		vdfo, err := virtualdefund.NewObjective(request, true, myAddress, minAmount, e.store.GetChannelById, e.store.GetConsensusChannel)
 		if err != nil {
 			return EngineEvent{}, fmt.Errorf("handleAPIEvent: Could not create objective for %+v: %w", request, err)
 		}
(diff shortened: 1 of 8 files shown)


protocols/virtualdefund/virtualdefund.go
```diff
@@ -71,6 +77,7 @@ type GetTwoPartyConsensusLedgerFunction func(counterparty types.Address) (ledger
 func NewObjective(request ObjectiveRequest,
 	preApprove bool,
 	myAddress types.Address,
+	largestPaymentAmount *big.Int,
```
lalexgap, line 80: I think there's an underlying assumption here that by the time we've started `virtualDefund` we are done issuing or receiving vouchers. Since we set `latestVoucherAmount` at objective creation, the objective is going to ignore any new vouchers added to the `voucher manager`.

I think that's ok, but probably something to be aware of.

protocols/virtualdefund/virtualdefund.go
```diff
@@ -17,14 +17,15 @@ import (
...
+	WaitingForFinalStateFromAlice     protocols.WaitingFor = "WaitingForFinalStateFromAlice"
+	WaitingForSignedFinal             protocols.WaitingFor = "WaitingForSignedFinal"             // Round 1
 	WaitingForCompleteLedgerDefunding protocols.WaitingFor = "WaitingForCompleteLedgerDefunding" // Round 2
 	WaitingForNothing                 protocols.WaitingFor = "WaitingForNothing"                 // Finished
 )
 
 const (
-	SignedStatePayload protocols.PayloadType = "SignedStatePayload"
-	InitPayload        protocols.PayloadType = "InitPayload"
+	SignedStatePayload       protocols.PayloadType = "SignedStatePayload"
```
lalexgap, line 27: We could spit `SignedStatePayload` into two seperate payloads: `SignaturesPayload` and `FinalOutcomePayload`. The `FinalOutcomePayload` could contain the latest voucher amount and be sent out by Alice. The `SignaturesPayload` would be used by all participants to share their signature on the final state.

client_test/virtualdefund_test.go
```diff
@@ -29,8 +30,12 @@ func TestVirtualDefundIntegration(t *testing.T) {
 	logFile := "test_virtual_defund.log"
 	truncateLog(logFile)
 	logDestination := newLogWriter(logFile)
-	runVirtualDefundIntegrationTest(t, 0, defaultTimeout, logDestination)
+	for _, closer := range []testactors.Actor{alice, irene, bob} {
```
geoknee, line 33: 🌋

protocols/virtualdefund/virtualdefund.go
geoknee: I agree that it's OK and good to be aware.

lalexgap:
As pointed out by @geoknee Irene also needs to perform validation to prevent Alice and Bob colluding against her. This PR needs to be updated with Irene's check

protocols/virtualdefund/virtualdefund.go
geoknee: That would be a nice use of the new payload format 👍 Would `FinalOutcomePayload` include Alice's signature? Or would / could she send two payloads at once?

protocols/virtualdefund/virtualdefund.go
```diff
@@ -607,10 +689,18 @@ func isZero(sig state.Signature) bool {
...
-	PaidToBob *big.Int
 }
 
 // Id returns the objective id for the request.
 func (r ObjectiveRequest) Id(types.Address) protocols.ObjectiveId {
 	return protocols.ObjectiveId(ObjectivePrefix + r.ChannelId.String())
 }
+
+// GetVirtualChannelFromId gets the virtual channel id from the objective id.
+func GetVirtualChannelFromId(id protocols.ObjectiveId) (types.Destination, error) {
```
geoknee, line 700:
```suggestion
// GetVirtualChannelIdFromObjectvieId gets the virtual channel id from the objective id.
func GetVirtualChannelIdFromObjectvieId(id protocols.ObjectiveId) (types.Destination, error) {
```

client/engine/engine.go
```diff
@@ -372,7 +372,11 @@ func (e *Engine) handleObjectiveRequest(or protocols.ObjectiveRequest) (EngineEv
 		return e.attemptProgress(&vfo)
 
 	case virtualdefund.ObjectiveRequest:
-		vdfo, err := virtualdefund.NewObjective(request, true, myAddress, e.store.GetChannelById, e.store.GetConsensusChannel)
+		bal, err := e.vm.Balance(request.ChannelId)
+		if err != nil {
```
geoknee, line 376: Won't we get an error if we call this as Irene?

geoknee: I guess not since your `client_test` covers this. I think the reason is that Irene _does_ register the channel with her voucher manager. Is that what we want to happen? It seems a bit strange for intermediaries to do that -- they might not even want to have a voucher manager at all, I would have thought.

client/engine/engine.go
```diff
@@ -573,7 +577,15 @@ func (e *Engine) constructObjectiveFromMessage(id protocols.ObjectiveId, p proto
...
 		return &vfo, nil
 	case virtualdefund.IsVirtualDefundObjective(id):
-		vdfo, err := virtualdefund.ConstructObjectiveFromPayload(p, false, *e.store.GetAddress(), e.store.GetChannelById, e.store.GetConsensusChannel)
+		vId, err := virtualdefund.GetVirtualChannelFromId(id)
+		if err != nil {
+			return &virtualdefund.Objective{}, fmt.Errorf("could not determine virtual channel id: %w", err)
+		}
+		bal, err := e.vm.Balance(vId)
+		if err != nil {
+			return &virtualdefund.Objective{}, fmt.Errorf("could determine voucher balance: %w", err)
```
geoknee, line 586:
```suggestion
			return &virtualdefund.Objective{}, fmt.Errorf("could not determine voucher balance: %w", err)
```

protocols/virtualdefund/virtualdefund.go
```diff
@@ -17,14 +17,15 @@ import (
...
+	WaitingForSignedFinal             protocols.WaitingFor = "WaitingForSignedFinal"             // Round 1
 	WaitingForCompleteLedgerDefunding protocols.WaitingFor = "WaitingForCompleteLedgerDefunding" // Round 2
 	WaitingForNothing                 protocols.WaitingFor = "WaitingForNothing"                 // Finished
 )
 
 const (
-	SignedStatePayload protocols.PayloadType = "SignedStatePayload"
-	InitPayload        protocols.PayloadType = "InitPayload"
+	SignedStatePayload       protocols.PayloadType = "SignedStatePayload"
+	FinalStateRequestPayload protocols.PayloadType = "FinalStateRequestPayload"
```
geoknee, line 28: Might be worth some godoc on these types?

protocols/virtualdefund/virtualdefund.go
```diff
@@ -121,33 +128,41 @@ func NewObjective(request ObjectiveRequest,
...
 
 	}
+
+	if largestPaymentAmount == nil {
+
+		largestPaymentAmount = big.NewInt(0)
+	}
+
+	finalOutcome := outcome.SingleAssetExit{}
+	// Since is Alice is responsible for issuing vouchers she always has the largest payment amount
```
geoknee, line 138:
```suggestion
	// Since Alice is responsible for issuing vouchers she always has the largest payment amount
```

protocols/virtualdefund/virtualdefund.go
```diff
@@ -121,33 +128,41 @@ func NewObjective(request ObjectiveRequest,
...
+func calculatePaidToBob(initialOutcome outcome.Exit, finalOutcome outcome.Exit) (*big.Int, error) {
 
-	if !found {
-		return big.NewInt(0), fmt.Errorf("could not find channel %s", cId)
-	}
-	initialBobAmount := pf.Outcome[0].Allocations[1].Amount
-	finalBobAmount := proposedFinalState.Outcome[0].Allocations[1].Amount
+	initialBobAmount := initialOutcome[0].Allocations[1].Amount
+	finalBobAmount := finalOutcome[0].Allocations[1].Amount
 	return big.NewInt(0).Sub(finalBobAmount, initialBobAmount), nil
```
geoknee, line 166: Does Bob need to check Alice's slot, too? What if she increased her balance and Bob's balance? Then the channel would be underfunded, and someone other than Alice will end up losing out! I don't think we have an on chain check for this, so we had better have an off chain check!

protocols/virtualdefund/virtualdefund.go
geoknee: Is `RequestFinalStatePayload` better?

protocols/virtualdefund/virtualdefund.go
```diff
@@ -510,6 +564,16 @@ func getSignedStatePayload(b []byte) (state.SignedState, error) {
 	return ss, nil
 }
 
+// getRequestDefundPayload takes in a serialized channel id payload and returns the deserialized channel id.
+func getRequestDefundPayload(b []byte) (types.Destination, error) {
```
geoknee, line 568: The name here should match the types at the top of the file. See my other comment about a good name (I personally prefer  `RequestFooPayload` to `FooRequestPayload`).

geoknee (approved):
Looking good @lalexgap !

Before merging, can we ensure that this marries up with the ADR #863 in terms of messages sent and validation checks performed? I will leave that unmerged in case you want to make changes in either direction.

client/engine/engine.go
lalexgap: I think you're correct that this doesn't fail because Irene does register the channel as well.

>It seems a bit strange for intermediaries to do that -- they might not even want to have a voucher manager at all, I would have thought.

I think you're right, I'll update so Irene doesn't store vouchers.

protocols/virtualdefund/virtualdefund.go
lalexgap: Good point!

protocols/virtualdefund/virtualdefund.go
lalexgap: I was thinking Alice could send two payloads at once. We might need to tweak some of the message helpers to allow creating a message with more than one payload.

protocols/virtualdefund/virtualdefund.go
lalexgap: I've updated this in https://github.com/statechannels/go-nitro/pull/866/commits/bd0ec246ad4631d0d97d5337f6bdfbae3e4f31a6.

Now when receiving the final signed state from Alice we check:
- The Destinations in the outcome are correct
- The amount paid to bob is added to bob's allocation
- The amount paid to bob is subtracted from Alice's allocation
- If we're Bob that the payment amount is greater than the latest voucher we have

protocols/virtualdefund/virtualdefund.go
lalexgap: Added some godoc and renamed in [c0aaaee](https://github.com/statechannels/go-nitro/pull/866/commits/c0aaaee3408ae1e49ccb6acd088549047b452013)

client/engine/engine.go
lalexgap: I've updated this in [1f5f8a2](https://github.com/statechannels/go-nitro/pull/866/commits/1f5f8a2a7575d17290bd8da6d1c3d3f775d43614) so Irene no longer stores vouchers or attempts to call `e.vm.Balance`

protocols/virtualdefund/virtualdefund.go
lalexgap: I will probably tackle this in a seperate PR.

protocols/virtualdefund/virtualdefund.go
lalexgap: @geoknee since this is security related it might be worth you taking a look at https://github.com/statechannels/go-nitro/commit/bd0ec246ad4631d0d97d5337f6bdfbae3e4f31a6 and making sure those checks are sufficient/correct.

geoknee: They look good to me, but it would be nice to have some tests that try and break them :-)

PR #872, opened 2022-09-12
WIP Add a docs website

geoknee:
Work in progress


kerzhner:
Do we know why the CI is failing?

geoknee:
> Do we know why the CI is failing?

Looks like a testground problem (surely unrelated to this change set). Looks like we may have filled up the storage on the VM or something ?

kerzhner:
> The site uses https://squidfunk.github.io/mkdocs-material/. I've really enjoyed using this framework, the site is very rich to view, but composed of just a handful of markdown files.

The doc site looks and feels excellent!

docs/index.md
```diff
@@ -0,0 +1,16 @@
...
+
+The network benefits from significant advantages, however -- in a fraction of a second, a direct, private connection can be forged with anyone else in the network. Then, payments can be streamed incredibly fast: up to **hundreds of times per second**. What's more, there is no per-payment fee to cover (just a small fee per unit time for maintaining the connection).
+
+This is in sharp contrast to the experience of transacting publicly on a "layer 1" blockchain or even on a rollup chain. There, fees and latency apply to each and every action.
+
+## What's on offer
+
+These docs cover the open source code at [https://github.com/statechannels/go-nitro](https://github.com/statechannels/go-nitro). This includes
+
+- :fontawesome-brands-npm: [@satechannels/nitro-protocol](https://www.npmjs.com/package/@statechannels/nitro-protocol): smart contracts :simple-solidity: and utilities :simple-typescript:.
```
lalexgap, line 15:
```suggestion
- :fontawesome-brands-npm: [@statechannels/nitro-protocol](https://www.npmjs.com/package/@statechannels/nitro-protocol): smart contracts :simple-solidity: and utilities :simple-typescript:.
```

docs/faqs.md
```diff
@@ -0,0 +1,43 @@
...
+
+Nitro protocol is a state of the art state channel protocol which is focussed on security, performance and extensibility. It has been developed over several [years of research](./research-papers.md).
+
+One of the key features of Nitro are [_virtual_ channels](./protocol-tutorial/0060-funding-a-channel.md#fund-virtually), where peers can setup a direct connection with each other entirely off-chain.
+
+## How is it implemented?
+
+The on-chain components of Nitro protocol are implemented in solidity, and are published alongside lightweight off-chain support in Typescript in the npm package [@statechannels/nitro-protocol](https://www.npmjs.com/package/@statechannels/nitro-protocol).
+
+The off-chain component of the protocol are implemented in [`go-nitro`](https://pkg.go.dev/github.com/statechannels/go-nitro), a reference client for Nitro Protocol.
```
lalexgap, line 23: It might be worth breaking this into a "What is go-nitro" question?

docs/index.md
```diff
@@ -0,0 +1,16 @@
...
+
+They work by allowing users to deposit their funds into a "layer 2" network which sits above a blockchain. This layer 2 network inherits many of the properties of the underlying chain -- for example, **permisionlessness** and the lack of a need to place **trust** in any counterparty.
+
+The network benefits from significant advantages, however -- in a fraction of a second, a direct, private connection can be forged with anyone else in the network. Then, payments can be streamed incredibly fast: up to **hundreds of times per second**. What's more, there is no per-payment fee to cover (just a small fee per unit time for maintaining the connection).
+
+This is in sharp contrast to the experience of transacting publicly on a "layer 1" blockchain or even on a rollup chain. There, fees and latency apply to each and every action.
+
+## What's on offer
+
+These docs cover the open source code at [https://github.com/statechannels/go-nitro](https://github.com/statechannels/go-nitro). This includes
```
lalexgap, line 13: How about documentation generated by `godoc`: https://pkg.go.dev/github.com/statechannels/go-nitro ?

lalexgap (approved):
The tutorials are great!

docs/faqs.md
geoknee: Yep I think that's a good shout.  I started out on this but ended up in a bit of rabbit hole thinking about the go-`nitro.Client` API. I'm going to tackle it in a new PR since there are many questions about it. #936

docs/index.md
geoknee: [fd78f8f](https://github.com/statechannels/go-nitro/pull/872/commits/fd78f8f1ad772969134c0505d972b1ae59103e15)

docs/protocol-tutorial/0010-states-channels.md
```diff
@@ -0,0 +1,305 @@
...
+
+Let's take each property in turn:
+
+### Chain id
+
+This needs to match the id of the chain where assets are to be locked (i.e. the 'root' of the funding graph for this channel). In the event of a mismatch, the channel cannot be concluded and funds cannot be unlocked.
+
+### Participants
+
+This is a dynamic array of Ethereum addresses, each derived from an ECDSA private key in the usual manner. Each address represents a participant in the state channel who is able to commit to state updates and thereby cause the channel to finalize on chain.
```
NiloCK, line 104: Dynamic?

docs/protocol-tutorial/0010-states-channels.md
```diff
@@ -0,0 +1,305 @@
...
+
+!!! warning
+
+    Before joining a state channel, you (or your off-chain software) should check that it has length at least 2, but no more than 255, and include a public key (account) that you control. Each entry should be a nonzero ethereum address.
+
+### ChannelNonce
+
+This is a unique number used to differentiate channels with an otherwise identical `FixedPart`. For example, if the same participants want to run the same kind of channel on the same chain as a previous channel, they can choose a new `ChannelNonce` to prevent state updates from from the existing channel being replayed.
```
NiloCK, line 112:
```suggestion
This is a unique number used to differentiate channels with an otherwise identical `FixedPart`. For example, if the same participants want to run the same kind of channel on the same chain as a previous channel, they can choose a new `ChannelNonce` to prevent state updates for the original channel from being replayed on the new one.
```

docs/protocol-tutorial/0010-states-channels.md
```diff
@@ -0,0 +1,305 @@
...
+
+!!! warning
+
+    You should never join a channel which re-uses a channel nonce.
+
+### AppDefinition
+
+This is an Ethereum address where a Nitro application has been deployed. This is a contract conforming to the `ForceMoveApp` and defining [application rules](./0020-execution-rules.md#application-rules).
```
NiloCK, line 120:
```suggestion
This is an Ethereum address where a Nitro application has been deployed. This is a contract conforming to the `ForceMoveApp` and defining [application rules](./0020-execution-rules.md#application-rules). The zero-address is a valid `AppDefinition` value, and it is interpreted by the protocol as the full-consensus app - states are valid if and only if they are signed by every member of `FixedPart.participants`.
```

docs/protocol-tutorial/0010-states-channels.md
```diff
@@ -0,0 +1,305 @@
...
+
+!!! warning
+
+    You should have confidence that the application is not malicious or suffering from security flaws. You should inspect the source code (which should be publically available and verifiable) or appeal to a trusted authority to do this.
+
+### ChallengeDuration
+
+This is duration (in seconds) of the challenge-response window. If a challenge is raised on chain at time `t`, the channel will finalize at `t + ChallengeDuration` unless cleared by a subqsequent on-chain transaction.
```
NiloCK, line 128:
```suggestion
This is duration (in seconds) of the challenge-response window. If a challenge is raised on chain at time `t`, the channel will be finalizable at `t + ChallengeDuration` unless cleared by a subqsequent on-chain transaction.
```

(marginal)

NiloCK, line 128:
```suggestion
This is duration (in seconds) of the challenge-response window. If a challenge is raised on chain at time `t`, the channel will finalize at `t + ChallengeDuration` unless cleared by a subsequent on-chain transaction.
```

docs/protocol-tutorial/0010-states-channels.md
```diff
@@ -0,0 +1,305 @@
...
+
+This is a boolean flag which allows the [channel execution rules](#execution-rules) to be bypassed and for the channel to be finalized "instantly" without waiting for the challenge-response window to lapse.
+
+!!! warning
+
+    As soon as an `isFinal=true` state is _enabled_ (that is to say, you cannot prevent it from becoming supported) it is not safe to continue executing the state channel. It should be finalized immediately.
+
+## Channel IDs
+
+Channels are identified by the hash of the `FixedPart`` of the state (those parts that may _not_ vary):
```
NiloCK, line 160:
```suggestion
Channels are identified by the hash of the `FixedPart` of the state (those parts that may _not_ vary):
```

docs/protocol-tutorial/0010-states-channels.md
```diff
@@ -0,0 +1,305 @@
...
+ bytes32 stateHash = keccak256(abi.encode(
+        channelId,
+        variablePart.appData,
+        variablePart.outcome,
+        variablePart.turnNum,
+        variablePart.isFinal
+    ));
+```
+
+and this hash is signed using an _ephemeral_ Ethereum private key. _Ephemeral_ in this context means a dedicated private key, generated solely for the purpose of executing the state channel. The signature has the following type:
```
NiloCK, line 190: I am not aware of any ephemeral keygen stuff happening in go-nitro. I even wrote a [notion doc](https://www.notion.so/statechannels/Hub-Hot-Wallet-Risk-ed5ed921c2fe4c03aac70c0966e3e825) about it!

Am I misunderstanding something here?

docs/protocol-tutorial/0010-states-channels.md
```diff
@@ -0,0 +1,305 @@
...
+        // INitroTypesSignedVariablePart is an auto generated low-level Go binding around an user-defined struct.
+        type INitroTypesSignedVariablePart struct {
+            VariablePart INitroTypesVariablePart
+            Sigs         []INitroTypesSignature
+        }
+    ```
+
+### Support proofs
+
+Support proofs consist of just enough information for the chain to verify that a state has been supported. They usually consist of [`FixedPart`](#states), plus a singular [`SignedVariablePart`](#signedvariableparts) named `candidate`, plus an array of [`SignedVariableParts`](#signedvariableparts) named `proof`.
```
NiloCK, line 273:
```suggestion
A support proof is any bundle of information sufficient for the chain to verify that a given channel state is legitimate. They usually consist of [`FixedPart`](#states), plus a singular [`SignedVariablePart`](#signedvariableparts) named `candidate`, plus an array of [`SignedVariableParts`](#signedvariableparts) named `proof`.

The trivial support proof is a state with `IsFinal: true` signed by every participant. For an intuition around more complicated support proofs, see [Putting the 'state' in state channels](https://blog.statechannels.org/putting-the-state-in-state-channels/).
```

docs/protocol-tutorial/0010-states-channels.md
```diff
@@ -0,0 +1,305 @@
...
+
+### `RecoveredVariableParts`
+
+The adjudicator smart contract will recover the signer from each signature on a `SignedVariablePart`, and convert the resulting list into a `signedBy` bitmask indicating which participant has signed that particular state. The bitfield is bundled with the `VariablePart` into a `RecoveredVariablePaet`:
```
NiloCK, line 277:
```suggestion
The adjudicator smart contract will recover the signer from each signature on a `SignedVariablePart`, and convert the resulting list into a `signedBy` bitmask indicating which participant has signed that particular state. The bitfield is bundled with the `VariablePart` into a `RecoveredVariablePart`:
```

NiloCK, line 277: This is the first usage of the word adjudicator. Maybe it should be on top of page where the L1 lockup is mentioned (eg: locked up and managed by a contract we call the Adjudicator).

docs/research-papers.md
```diff
@@ -0,0 +1,16 @@
...
+
+## (2018) [:fontawesome-solid-file-pdf:](https://magmo.com/force-move-games.pdf) ForceMove
+
+An n-party state channel protocol.
+
+## (2019) [:fontawesome-solid-file-pdf:](https://magmo.com/nitro-protocol.pdf) Nitro
+
+A protocol for state channel networks.
+
+## (2022) [:fontawesome-solid-file-pdf:](https://statechannels.github.io/satp_paper/satp.pdf) Stateful Asset Transfer Protocol
```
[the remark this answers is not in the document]
geoknee: Strange, it works for me?

docs/protocol-tutorial/0010-states-channels.md
geoknee: dynamically sized array? https://docs.soliditylang.org/en/develop/types.html#arrays

docs/protocol-tutorial/0010-states-channels.md
geoknee: This might be describing the way things ought to be more than the way things are. 

Right now we have injected dependencies, so I believe it is possible to instantiate a client where the same key is used by the chain service and by the store/protocols for signing state updates. But at least the current system allows for these to be different (granted, it does not enforce it). But technically we are outsourcing the responsibility for it to be ephemeral, and we should probably bring that in house eventually. And yes your notion doc has some ideas which we might want to use. 

Does that clarify things?

docs/research-papers.md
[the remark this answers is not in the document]
NiloCK: Uhh, works for me now as well. Maybe it was a github pages outage more globally at the time of viewing?

docs/protocol-tutorial/0010-states-channels.md
NiloCK: That description is solidity specific (go would refer to a slice, TS just an array). Given that the type definitions are immediately above in each language, we can probably drop that level of technical description in favour of "A collection of Ethereum addresses, ...".

docs/protocol-tutorial/0010-states-channels.md
NiloCK: Medium clarity. I'm not sure how close go-nitro is to supporting ephemeral keys. Currently we take a chainService and a store as injections. Store has interface methods 
```go
GetChannelSecretKey() *[]byte // Get a pointer to a secret key for signing channel updates
GetAddress() *types.Address   // Get the (Ethereum) address associated with the ChannelSecretKey
```
which don't seem well situated for handling a key / address per channel.

The supplied chain service can be applying a different key to transactions, but the protocols lean on the store's reported `myAddress` in order to fill out values in the participants list. Does NitroAdjudicator make any inferences based on the transaction's `msg.sender`? (I think not?) If it does, the transaction signer would have to match the address listed in `participants`.

An aside: If we design with the expectation of ephemeral keys per channel, we probably don't require nonces, because ephemeral keys map to ephemeral addresses (burner address), and there is a requirement for state signatures to be from the addresses of `Participants`.

Overall I think this is too far removed from current implementation to be included in the docs... ?

docs/protocol-tutorial/0010-states-channels.md
geoknee, line 104:
```suggestion
This is a list of Ethereum addresses, each derived from an ECDSA private key in the usual manner. Each address represents a participant in the state channel who is able to commit to state updates and thereby cause the channel to finalize on chain.
```

docs/protocol-tutorial/0010-states-channels.md
geoknee: This is actually no longer true :-P. We now only support consensus apps via an explicit set of rules deployed on chain.

docs/protocol-tutorial/0010-states-channels.md
geoknee: [502b7fe](https://github.com/statechannels/go-nitro/pull/872/commits/502b7fe7bb01a71e0e1c8b07a5a3ef40a6f2eaa3)

docs/protocol-tutorial/0010-states-channels.md
geoknee: > Does NitroAdjudicator make any inferences based on the transaction's msg.sender? (I think not?)

Correct, it does not.

>  don't seem well situated for handling a key / address per channel.

Yes we are far from supporting a key per channel -- and this is probably the best definition of what an "ephemeral key" ought to be. I've removed "ephemeral" from the docs for now [a715bb3](https://github.com/statechannels/go-nitro/pull/872/commits/a715bb31dfdb1b97fc61ef056d8572a459ab2850).

I think what i) we have typically meant by using "ephemeral" and ii) what is actually the most important thing from a security point of view is simply using a special key for channel updates which does not directly control any funds on chain. In hindsight this may not be a good usage of the term, as the key may actually end up being long-lived. 

See also https://github.com/statechannels/go-nitro/blob/502b7fe7bb01a71e0e1c8b07a5a3ef40a6f2eaa3/.adr/0002-signature-scheme.md#L13 where we could also remove the word "ephemeral".

PR #874, opened 2022-09-14
Convert solidity FixedPart.channelNonce to uint64

kerzhner:
`uint64` is a type supported by golang (as well as many other languages). `uint48` is not a golang type. Using a solidity type that is not supported in golang can result in mismatch between data provided to an ethereum function call parameter and the expectation for that data.

Specifically for channelNonce, with uint48, some of the calls to solidity functions supplied integers that overflowed the type. The error message is a cryptic "transaction reverted without a reason string". One way to avoid the error is to match golang types for variables with solidity types.

Increasing uint48 to uint64 has minimal gas implications (and in fact results in certain gas improvements) since channelNonce is not persisted to storage.


nitro-protocol/config/jest/jest.config.js
```diff
@@ -18,7 +18,7 @@ module.exports = {
   rootDir: root,
   collectCoverageFrom: ['**/*.{js,jsx,ts,tsx}'],
   reporters: ['default'],
-  testMatch: ['<rootDir>/test/src/**/?(*.)test.ts?(x)'],
+  testMatch: ['<rootDir>/test/src/**/?(*.)test.ts?(x)', '<rootDir>/test//?(*.)test.ts?(x)'],
```
kerzhner, line 21: Do you know why the double `/` is needed in `'<rootDir>/test//?(*.)test.ts?(x)'`?

kerzhner:
@geoknee Thanks for updating the Typescript code. The changes look good!

nitro-protocol/config/jest/jest.config.js
geoknee: It is not needed, it is a typo :-P. It turns out that it still works, though -- you can try it out here https://www.digitalocean.com/community/tools/glob .

PR #877, opened 2022-09-15
Convert FixedPart.ChallengeDuration to uint32

kerzhner:
In Solidity contracts, ChallengeDuration type is `uint48`. This PR changes the type of ChallengeDuration from `big.Int` to `uint64` on golang. While golang `uint64` will not guard against overflowing the Solidity type, we should favor primitive types over structs when a struct provides no other significant advantages.

### DIFF
--- a/channel/channel_test.go
+++ b/channel/channel_test.go
@@ -210,7 +210,7 @@ func TestChannel(t *testing.T) {
 			},
 			ChannelNonce:      37140676581,
 			AppDefinition:     common.HexToAddress(`0x5e29E5Ab8EF33F050c7cc10B5a0456D975C5F88d`),
-			ChallengeDuration: big.NewInt(60),
+			ChallengeDuration: 60,
 			AppData:           []byte{},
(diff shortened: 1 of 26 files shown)


geoknee:
@kerzhner let's ensure this is consistent with the proposed ADR #876 . Could you review that PR, and then we need to decide whether 32 or 64 bits is best. I think a `uint32` might make the most sense.

geoknee:
> @kerzhner let's ensure this is consistent with the proposed ADR #876 . Could you review that PR, and then we need to decide whether 32 or 64 bits is best. I think a `uint32` might make the most sense.

Decision from standup today: we will go with 32 bytes, which is

kerzhner:
> Decision from standup today: we will go with 32 bytes, which is about 136 yrs.

[2446a39](https://github.com/statechannels/go-nitro/pull/877/commits/2446a39718bd0191a9a31afcc9cbb40e6e869f3d) changes challenge duration to uint32.

geoknee (approved):

PR #878, opened 2022-09-19
save some gas computing channel ids

geoknee:
This is a minor clean up which I spotted when working on #872


kerzhner (approved):
Nice catch!

PR #880, opened 2022-09-20
add test case for trying to transfer a guarantee

### LINKED ISSUE #879: Audit use of `executeSingleAssetExit`
I see two issues with the current implementation:

https://github.com/statechannels/go-nitro/blob/332f2eabdca278399f8d44c0eca1dfdab9475e3f/nitro-protocol/contracts/MultiAssetHolder.sol#L416-L427

1. We should use `executeExit` from `@statechannels/exit-format` rather than reimplementing here. It supports withdraw helpers.
2. We should be checking the allocation type before paying funds out (we should not be able to pay out a guarantee, for example).

geoknee:
Closes #879

### DIFF
--- a/nitro-protocol/test/contracts/MultiAssetHolder/transfer.test.ts
+++ b/nitro-protocol/test/contracts/MultiAssetHolder/transfer.test.ts
@@ -36,31 +36,33 @@ const addresses = {
 const reason0 = 'Channel not finalized';
 const reason1 = 'Indices must be sorted';
 const reason2 = 'incorrect fingerprint';
+const reason3 = 'cannot transfer a guarantee';
 
 // c is the channel we are transferring from.
 describe('transfer', () => {
   it.each`
-    name                                   | heldBefore | setOutcome            | indices      | newOutcome            | heldAfter       | payouts         | reason
-    ${' 0. channel not finalized        '} | ${{c: 1}}  | ${{}}                 | ${[0]}       | ${{}}                 | ${{}}           | ${{A: 1}}       | ${reason0}
-    ${' 1. funded          -> 1 EOA'}      | ${{c: 1}}  | ${{A: 1}}             | ${[0]}       | ${{A: 0}}             | ${{}}           | ${{A: 1}}       | ${undefined}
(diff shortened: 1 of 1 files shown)


nitro-protocol/test/contracts/MultiAssetHolder/transfer.test.ts
```diff
@@ -36,31 +36,33 @@ const addresses = {
...
-    ${'13. -> 2 indices'}                  | ${{c: 3}}  | ${{C: 2, X: 2}}       | ${[0, 1]}    | ${{C: 0, X: 1}}       | ${{c: 0, X: 1}} | ${{C: 2}}       | ${undefined}
-    ${'14. -> 3 indices'}                  | ${{c: 5}}  | ${{A: 1, C: 2, X: 2}} | ${[0, 1, 2]} | ${{A: 0, C: 0, X: 0}} | ${{c: 0, X: 2}} | ${{A: 1, C: 2}} | ${undefined}
-    ${'15. -> reverse order (see 13)'}     | ${{c: 3}}  | ${{C: 2, X: 2}}       | ${[1, 0]}    | ${{C: 2, X: 1}}       | ${{c: 2, X: 1}} | ${{}}           | ${reason1}
-    ${'16. incorrect fingerprint        '} | ${{c: 1}}  | ${{}}                 | ${[0]}       | ${{}}                 | ${{}}           | ${{A: 1}}       | ${reason2}
+    name                                   | heldBefore | isSimple | setOutcome            | indices      | newOutcome            | heldAfter       | payouts         | reason
+    ${' 0. channel not finalized        '} | ${{c: 1}}  | ${true}  | ${{}}                 | ${[0]}       | ${{}}                 | ${{}}           | ${{A: 1}}       | ${reason0}
```
geoknee, line 45: it is difficult to parse this diff. An additional column `isSimple` has been added. Existing rows have this field set to `true`. An additional row has been added, which is the same as case 1 but with `isSimple: false`.

kerzhner:
Does this PR fix https://github.com/statechannels/go-nitro/issues/879? That issue is about Solidity implementation, whereas this code adds a test.

kerzhner (approved):

geoknee:
It resolves #879 by showing that there is "no case to answer". Our current implementation passes the new test.

Issue #839, opened 2022-08-23
Use correct App definition for virtual channels


lalexgap:
We should make an issue to deal with using the correct `AppDefinition` here. A decent short term solution is to remove it from the API, and have the engine or virtualffund protocol fill in the correct app definition automatically.

_Originally posted by @geoknee in https://github.com/statechannels/go-nitro/pull/836#discussion_r952535303_

PR #881, opened 2022-09-20
Make the `client` responsible for nonce management

geoknee:
(Recommend commit-by-commit review)

This is a significant change to the `go-nitro` client APi. 

It is no longer necessary / possible to provide an `AppDefinition`, `AppData` or `Nonce` for a channel. 

`AppDefinition` -- this is set to the deployed contract address for ledger channels, and shortly something similar will happen for virtual channels #839.

`AppData` - neither channel type currently supported has any app data.

`Nonce` - this is now being generated in the client, avoiding the need for the consuming application to worry about it.  We are using random nonces for now, and aren't doing any checks against previously-used nonces at the current time.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -4,6 +4,7 @@ package client // import "github.com/statechannels/go-nitro/client"
 import (
 	"io"
 	"math/big"
+	"math/rand"
 
 	"github.com/statechannels/go-nitro/client/engine"
 	"github.com/statechannels/go-nitro/client/engine/chainservice"
@@ -92,7 +93,13 @@ func (c *Client) ReceivedVouchers() <-chan payments.Voucher {
 }
 
 // CreateVirtualChannel creates a virtual channel with the counterParty using ledger channels with the intermediary.
-func (c *Client) CreateVirtualChannel(objectiveRequest virtualfund.ObjectiveRequest) virtualfund.ObjectiveResponse {
+func (c *Client) CreateVirtualChannel(request virtualfund.ObjectiveRequestForVirtualPaymentApp) virtualfund.ObjectiveResponse {
+
+	objectiveRequest := virtualfund.ObjectiveRequest{
+		ObjectiveRequestForVirtualPaymentApp: request,
+		Nonce:                                rand.Uint64(),
+		// AppDefinition implicitly zero TODO https://github.com/statechannels/go-nitro/issues/839
+	}
 
 	// Send the event to the engine
 	c.engine.ObjectiveRequestsFromAPI <- objectiveRequest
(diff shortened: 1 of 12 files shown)


protocols/virtualfund/virtualfund.go
```diff
@@ -699,15 +698,21 @@ func (o *Objective) updateLedgerWithGuarantee(ledgerConnection Connection, sk *[
 	return sideEffects, nil
 }
 
-// ObjectiveRequest represents a request to create a new virtual funding objective.
-type ObjectiveRequest struct {
+// ObjectiveRequestForVirtualPaymentApp represents a request to create a new virtual funding objective.
+// There is no AppDefinition, since we currently only support full consensus rules for direct channels.
+// There is no Nonce, since the client fills this in.
+type ObjectiveRequestForVirtualPaymentApp struct {
```
lalexgap, line 704: It seems slightly strange to have a `ObjectiveRequestForVirtualPaymentApp` in `virtualfund` when that struct is only used in the `client` package.

Maybe `ObjectiveRequestForVirtualPaymentApp`(and other structs exposed by API) should live in the `client` package instead of `virtualfund`?

client/client.go
```diff
@@ -92,7 +93,13 @@ func (c *Client) ReceivedVouchers() <-chan payments.Voucher {
 }
 
 // CreateVirtualChannel creates a virtual channel with the counterParty using ledger channels with the intermediary.
-func (c *Client) CreateVirtualChannel(objectiveRequest virtualfund.ObjectiveRequest) virtualfund.ObjectiveResponse {
+func (c *Client) CreateVirtualChannel(request virtualfund.ObjectiveRequestForVirtualPaymentApp) virtualfund.ObjectiveResponse {
```
lalexgap, line 96: Maybe we should rename `CreateVirtualChannel` to `CreateVirtualPaymentChannel` to keep it in line with `ObjectiveRequestForVirtualPaymentApp`?

protocols/virtualfund/virtualfund.go
lalexgap: We could also just update the API to accept parameters instead of a request struct: `func (c *Client) CreateVirtualChannel(counterparty, intermediary types.Address ....`

lalexgap (approved):

protocols/virtualfund/virtualfund.go
geoknee: I ended up going with the parameters approach you suggested (also for the ledger channel API method). Saved a lot of lines! [5284054](https://github.com/statechannels/go-nitro/pull/881/commits/528405467e497575796b3780399ec5398f3b60cc) and [3b340dc](https://github.com/statechannels/go-nitro/pull/881/commits/3b340dc4da702638bc786f38785b80ad0375d803)

client/client.go
geoknee: [5284054](https://github.com/statechannels/go-nitro/pull/881/commits/528405467e497575796b3780399ec5398f3b60cc)

Issue #378, opened 2022-03-22
Investigate Engine Concurrency


lalexgap:
Look into allowing the engine to handle things concurrently instead of having a single threaded run loop. 

This has been explored in this draft [PR](https://github.com/statechannels/go-nitro/pull/374)

PR #884, opened 2022-09-21
Use a `safesync.Map` in VoucherManager

geoknee:
While experimenting with #378, I ran tests and saw panics around concurrent reads from the channels map in the voucher manager. 

This PR follows a pattern we have used elsewhere (in the chain service, in the store) and upgrades the use of a vanilla `map` into a concurrency safe `map`, which prevents the aforementioned panics. 

This will allow future work to explore various concurrency models in the engine.

### DIFF
--- a/payments/voucher_manager.go
+++ b/payments/voucher_manager.go
@@ -5,6 +5,7 @@ import (
 	"math/big"
 
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/statechannels/go-nitro/client/engine/store/safesync"
 	"github.com/statechannels/go-nitro/types"
 )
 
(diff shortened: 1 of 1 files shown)


payments/voucher_manager.go
```diff
@@ -5,6 +5,7 @@ import (
 	"math/big"
 
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/statechannels/go-nitro/client/engine/store/safesync"
```
lalexgap, line 8: Probably beyond the scope of this PR but the `safesync` package could probably live outside the `store` now that it's being used beyond the `store` package.

That might prevent some annoying import cycles in the future.

lalexgap (approved):

PR #886, opened 2022-09-21
Log message summaries

lalexgap:
This PR re-introduces a summarize function so we can output a message summary to the log file. We use this to log message summaries for incoming and outgoing messages.

I added this locally to aid in debugging some testground stuff and figured the changes were helpful enough that they should be rolled into `main`.

### DIFF
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -357,6 +357,7 @@ func (e *Engine) handleObjectiveRequest(or protocols.ObjectiveRequest) (EngineEv
 	defer e.metrics.RecordFunctionDuration()()
 	myAddress := *e.store.GetAddress()
 	objectiveId := or.Id(myAddress)
+	e.logger.Printf("handling new objective request for %s", objectiveId)
 	e.metrics.RecordObjectiveStarted(objectiveId)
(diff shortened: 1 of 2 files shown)


geoknee (approved):
Longer term question: do you think we would ever be worried about the performance impact of functions like these?

If we ever were, I suppose one solution might be to dump raw data and have a separate program process into something nice to read after-the-fact.

lalexgap:
> Longer term question: do you think we would ever be worried about the performance impact of functions like these?
> 
> If we ever were, I suppose one solution might be to dump raw data and have a separate program process into something nice to read after-the-fact.

That's a good thought! Another option is that we handle logging in its own `go-routine`. The `engine` would make a non blocking request to log something. The go-routine could handle summarizing and printing out to the log when it picks up this request.

However right now I suspect the performance impact of `Summarize` is pretty minimal in the grand scheme of things, but it's something to keep an eye on.

lalexgap:
In [070f09a](https://github.com/statechannels/go-nitro/pull/886/commits/070f09a179e768a1d79953d5d0602dabc91a7c67) we are now capturing metrics for how long it takes to summarize a message.

PR #888, opened 2022-09-22
Commit go bindings for `VirtualAppPaymentChannel` and add `GetVirtualAppPaymentChannelAddress()` to `ChainService` interface.

### LINKED ISSUE #839: Use correct App definition for virtual channels

geoknee:
Closes #839.


NiloCK (approved):
LGTM

PR #891, opened 2022-09-22
Trigger testground run from CI

lalexgap:
Adds a job to our github actions that runs a short testground run using the given commit. The test case is run remotely on our testground runner cloud vm, so the load on the CI shouldn't be too bad.

We run a 60 second test with 6 instances:
```bash
 testground --endpoint=http://34.168.92.245:8042  run s --wait -p=go-nitro-testground -t=virtual-payment -b=docker:go -r=local:docker -tp=numOfHubs=2 -tp=numOfPayers=2 -tp=numOfPayees=2  -i=6 -tp=paymentTestDuration=60  -tp=concurrentPaymentJobs=1
```
We use the `--wait` flag so the the github action blocks and only returns success when the test run is done.

Even though the test case is handled in the cloud vm we still need a local install of `testground` to send the test case to the cloud vm. This means we need to build testground (which is just a docker image underneath the hood). To help speed that up we use the `docker-layer-caching` action.

## TODO
- [ ] It would be nice if we could output a link to the dashboards 
- [ ] Decide how long / large test run to perform
- [ ] Decide when we should trigger this GH action

### DIFF
--- a/.github/workflows/testground.yml
+++ b/.github/workflows/testground.yml
@@ -18,7 +18,7 @@ jobs:
         with:
           repository: 'statechannels/go-nitro-testground'
           path: "code/go-nitro-testground"
-          ref: update-to-main #TODO Remove when https://github.com/statechannels/go-nitro-testground/pull/103 is merged
+          ref: main
       
       # Set up caching for docker (which testground uses) and go modules
       - uses: actions/cache@v3
--- a/.github/workflows/testground.yml
+++ b/.github/workflows/testground.yml
@@ -51,6 +51,9 @@ jobs:
          -p=go-nitro-testground -t=virtual-payment \
          -b=docker:go -r=local:docker \
          -tp=numOfHubs=2 -tp=numOfPayers=2 -tp=numOfPayees=2  -i=6 \
-         -tp=paymentTestDuration=60  -tp=concurrentPaymentJobs=1
+         -tp=paymentTestDuration=60  -tp=concurrentPaymentJobs=1 \
+          --metadata-repo "${{github.repository}}" \
+          --metadata-branch "${{github.ref_name}}" \
+          --metadata-commit "${{github.sha}}" 
 
       
--- a/.github/workflows/testground.yml
+++ b/.github/workflows/testground.yml
@@ -1,7 +1,7 @@
 name: Run Testground
 
 on:
-  push:
+  pull_request:
 jobs:
   run-testground:
       runs-on: ubuntu-latest
@@ -38,7 +38,7 @@ jobs:
         working-directory: "code/testground"
       # Update our test plan so it uses the version of go-nitro from this workflow
       - name: Update Test Dependency
-        run: go get github.com/statechannels/go-nitro@${{github.sha}}
+        run: go get github.com/statechannels/go-nitro@${{github.event.pull_request.head.sha}}
         working-directory: "code/go-nitro-testground"
 
       - name: Import Test 
@@ -53,7 +53,7 @@ jobs:
          -tp=numOfHubs=2 -tp=numOfPayers=2 -tp=numOfPayees=2  -i=6 \
          -tp=paymentTestDuration=60  -tp=concurrentPaymentJobs=1 \
           --metadata-repo "${{github.repository}}" \
-          --metadata-branch "${{github.ref_name}}" \
+          --metadata-branch "${{github.ref}}" \
           --metadata-commit "${{github.sha}}" 
 
       
--- a/.github/workflows/testground.yml
+++ b/.github/workflows/testground.yml
@@ -51,7 +51,7 @@ jobs:
          -p=go-nitro-testground -t=virtual-payment \
          -b=docker:go -r=local:docker \
          -tp=numOfHubs=2 -tp=numOfPayers=2 -tp=numOfPayees=2  -i=6 \
-         -tp=paymentTestDuration=60  -tp=concurrentPaymentJobs=1 \
+         -tp=paymentTestDuration=30  -tp=concurrentPaymentJobs=1 \
           --metadata-repo "${{github.repository}}" \
           --metadata-branch "${{github.ref}}" \
           --metadata-commit "${{github.sha}}" 
--- a/.github/workflows/testground.yml
+++ b/.github/workflows/testground.yml
@@ -1,0 +1,53 @@
+name: Run Testground
+
+on:
+  push:
+    branches: ['**'] # TODO: Restrict this to main
+jobs:
+  run-testground:
+      runs-on: ubuntu-latest
+      steps:
+      - uses: actions/setup-go@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/testground.yml
```diff
@@ -0,0 +1,53 @@
...
+      # Get testground and our test plan code
+      - uses: actions/checkout@v3
+        with:
+         repository: 'statechannels/testground'
+         path: "code/testground"
+      - uses: actions/checkout@v3
+        with:
+          repository: 'statechannels/go-nitro-testground'
+          path: "code/go-nitro-testground"
+          ref: update-to-main #TODO Remove when https://github.com/statechannels/go-nitro-testground/pull/103 is merged
```
lalexgap, line 22: This is required because `main` on `go-nitro-testground` hasn't been updated to work with the latest `go-nitro` (see https://github.com/statechannels/go-nitro-testground/pull/103)

.github/workflows/testground.yml
```diff
@@ -0,0 +1,53 @@
+name: Run Testground
+
+on:
+  push:
```
lalexgap, line 4: We use `push` instead of `pull request` because pull request events [use a different value for](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request ) the`github.sha` var.

lalexgap:
I just discovered that there is an existing [testground github action](https://github.com/testground/testground-github-action). I think that action only works when dealing with `compositions` though.

.github/workflows/testground.yml
```diff
@@ -0,0 +1,59 @@
...
+        working-directory: "code"
+      # Run a short test using the wait flag so we block until it completes
+      - name: Run Test 
+        run: | 
+         testground --endpoint=http://34.168.92.245:8042  run s --wait \
+         -p=go-nitro-testground -t=virtual-payment \
+         -b=docker:go -r=local:docker \
+         -tp=numOfHubs=2 -tp=numOfPayers=2 -tp=numOfPayees=2  -i=6 \
+         -tp=paymentTestDuration=60  -tp=concurrentPaymentJobs=1 \
+          --metadata-repo "${{github.repository}}" \
```
lalexgap, line 55: These metadata fields are used to create a link to the commit from the [testground task page](http://34.168.92.245:8042/tasks)

.github/workflows/testground.yml
```diff
@@ -0,0 +1,59 @@
...
+        run: go get github.com/statechannels/go-nitro@${{github.sha}}
+        working-directory: "code/go-nitro-testground"
+
+      - name: Import Test 
+        run:    testground plan import --from ./go-nitro-testground
+        working-directory: "code"
+      # Run a short test using the wait flag so we block until it completes
+      - name: Run Test 
+        run: | 
+         testground --endpoint=http://34.168.92.245:8042  run s --wait \
```
geoknee, line 50: Will the github action fail if the testground task fails? I am guessing it will?

Trying it out locally, the `--wait` flag causes some logs to be streamed back to the console, which should be useful. And if the task fails to launch, we definitely get a nonzero exit code.

.github/workflows/testground.yml
```diff
@@ -0,0 +1,59 @@
...
+         -p=go-nitro-testground -t=virtual-payment \
+         -b=docker:go -r=local:docker \
+         -tp=numOfHubs=2 -tp=numOfPayers=2 -tp=numOfPayees=2  -i=6 \
+         -tp=paymentTestDuration=60  -tp=concurrentPaymentJobs=1 \
```
geoknee, line 54: I think this duration is probably ok for starters. I feel like we get a good idea of performance even over 30s. I expect we will have a minute or two of setup time before we get to run the test, anyway?

geoknee (approved):

geoknee:
This is going to be great! I expect us to refine this a lot in future but can't wait to try it out. 

It occurs to me that if a PR changes the go-nitro API, the testground job will likely fail, and we will need to go through the rigamarole of updating the other repo's test script. Is there an argument for pulling that test script (at least) into this repo to make it easier to keep things in sync and prevent breaking changes?

.github/workflows/testground.yml
```diff
@@ -0,0 +1,59 @@
...
+      # Get testground and our test plan code
+      - uses: actions/checkout@v3
+        with:
+         repository: 'statechannels/testground'
+         path: "code/testground"
+      - uses: actions/checkout@v3
+        with:
+          repository: 'statechannels/go-nitro-testground'
+          path: "code/go-nitro-testground"
+          ref: update-to-main #TODO Remove when https://github.com/statechannels/go-nitro-testground/pull/103 is merged
```
geoknee, line 21:
```suggestion
          ref: main
```

That PR is now merged.

.github/workflows/testground.yml
geoknee: Screenshot 2022-09-23 at 10 33 07

.github/workflows/testground.yml
geoknee: In standup today, we decided we would like to use `pull request` if possible.

.github/workflows/testground.yml
lalexgap: Yup, if the testground task fails the test will fail as well.

> This is great!
> 
> It occurs to me that if a PR changes the go-nitro API, the testground job will likely fail, and we will need to go through the rigamarole of updating the other repo's test script. Is there an argument for pulling that test script (at least) into this repo to make it easier to keep things in sync and prevent breaking changes?

Yup, changing the API is slightly painful as you'll have to update  `go-nitro-testground` before the job passes. I think it would be nice to pull in `go-nitro-testground` into `go-nitro`. The only catch is that `go-nitro-testground` is it's own module, which would mean we'd hav two go modules in one repo if we move it to `go-nitro`. Maybe something like [workspaces](https://go.dev/ref/mod#workspaces) could help there?

lalexgap:
> This is great!
> 
> It occurs to me that if a PR changes the go-nitro API, the testground job will likely fail, and we will need to go through the rigamarole of updating the other repo's test script. Is there an argument for pulling that test script (at least) into this repo to make it easier to keep things in sync and prevent breaking changes?

Yup, changing the API is slightly painful as you'll have to update  `go-nitro-testground` before the job passes. I think it would be nice to pull in `go-nitro-testground` into `go-nitro`. The only catch is that `go-nitro-testground` is it's own module, which would mean we'd hav two go modules in one repo if we move it to `go-nitro`. Maybe something like [workspaces](https://go.dev/ref/mod#workspaces) could help there?

.github/workflows/testground.yml
lalexgap: Updated in https://github.com/statechannels/go-nitro/pull/891/commits/8725fde15389ace1e164dab7dead432d7ea4738b

PR #893, opened 2022-09-23
Experiment with running handlers in go routines

geoknee:
This PR is just an experiment with greater concurrency inside the engine. Theory being, the engine can get to work on the next request without blocking on the current one completing.

So far I didn't see any improvements. In fact we probably expect things to go wrong more frequently with this approach.
Doing a similar refactor with the message handler might be more likely to enable a busy hub to deal with things more efficiently.

Sharing so we can road test https://github.com/statechannels/go-nitro/pull/891 with it.


geoknee:
Closing this as we don't want to enable any concurrency until we have the proper locking system in place in the store.

PR #895, opened 2022-09-23
Update CI to post link to dashboards

### LINKED ISSUE #894: Update CI to post a comment with grafana link
Update our CI script to post a comment to the pr with a link to the grafana dashboard.

We should be able to use an action like [this](https://github.com/marketplace/actions/create-or-update-comment) to post the comment.  The testground action has some logic around pulling out the[ test run id](https://github.com/testground/testground-github-action/blob/2e4df1e5ac0c574ceed7bd387fab00bca73747c5/entrypoint.sh#L39).

With the id it's pretty easy to format a url to grafana: `http://34.168.92.245:3000/d/miulKz7Vk/message-stats?orgId=1&var-runId=ccmv7dgnr2gqfovjdiig&from=now-5m&to=now`.

lalexgap:
Fixes #894

Updates our CI to post a comment with links to grafana dashboards and logs for the testground run kicked off by the CI.

Right now I have it configured so CI will only use one comment on the PR. Each testground run is appended to the comment.

An example of the [comment](https://github.com/statechannels/go-nitro/pull/895#issuecomment-1256729528) can be found on this PR. **Note** during testing I've manually edited this comment which is why it doesn't show the test runs for every commit.

We capture timestamps before and after we run the `testground run` command which we use for the time ranges in the dashboard links.

### DIFF
--- a/.github/workflows/testground.yml
+++ b/.github/workflows/testground.yml
@@ -31,6 +31,10 @@ jobs:
             ${{ runner.os }}-go-
       - uses: satackey/action-docker-layer-caching@v0.0.11
         continue-on-error: true
+        with:
+          key: docker-cache-{hash}
+          restore-keys: |
+           docker-cache-
 
       # Build the testground daemon
       - name: Build
--- a/.github/workflows/testground.yml
+++ b/.github/workflows/testground.yml
@@ -52,6 +52,8 @@ jobs:
       # Set the start time for the dashboard time range
       - name: Set Start Time
         id: set_start_time
+        # We add 40 seconds to the start time so the dashboard ignores startup when nothing is happening
+        # We multiply by 1000 to get the timestamp in MS which grafana expects
         run: | 
          echo "::set-output name=START_TIME_OFFSET::$(((`date '+%s'`+40)*1000))"
       # Run a short test using the wait flag so we block until it completes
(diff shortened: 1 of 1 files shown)


lalexgap:
I just realized that timestamp links are using the relative `now`. That means the time range will be wrong if you're viewing the dashboards later on.

.github/workflows/testground.yml
```diff
@@ -52,8 +64,38 @@ jobs:
...
+         comment-author: 'github-actions[bot]'
+         body-includes: Testground Run
+      - name: Update dashboard links comment
+        uses: peter-evans/create-or-update-comment@v2
+        with:
+          comment-id: ${{ steps.find-comment.outputs.comment-id }}
+          issue-number: ${{ github.event.pull_request.number }}
+          body: |
+            # 🧪 Testground Run for ${{github.event.pull_request.head.sha}} 
+            - [Message Stats Dashboard](http://34.168.92.245:3000/d/miulKz7Vk/message-stats?orgId=1&var-runId=${{steps.set_run_id.outputs.RUN_ID}}&from=${{steps.set_start_time.outputs.START_TIME_OFFSET}}&to=${{steps.set_done_time.outputs.DONE_TIME}})
```
geoknee, line 97: Do you think it is worth defining the IP address and port(s) as constants somewhere (perhaps even github secrets)? Might help us to refactor in future.

geoknee (approved):
🚀

.github/workflows/testground.yml
lalexgap: Updated in https://github.com/statechannels/go-nitro/pull/895/commits/b859f86b1b4c432b81178667c08adb77088ea287

PR #897, opened 2022-09-23
WIP: Capture screenshots of dashboards

lalexgap:
Takes a screenshot of the dashboards and posts them as a comment to the pull request. 

This relies on anonymous authentication being turned on for grafana. This means anyone can see our dashboards 👀 

I'm using the `comment-webpage-screenshot` action but it takes a while to build and doesn't offer much in the way of configuration. Ideally it'd be nice to embed a screenshot in the comment with dashboard links.

### DIFF
--- a/.github/workflows/testground.yml
+++ b/.github/workflows/testground.yml
@@ -31,6 +31,10 @@ jobs:
             ${{ runner.os }}-go-
       - uses: satackey/action-docker-layer-caching@v0.0.11
         continue-on-error: true
+        with:
+          key: docker-cache-{hash}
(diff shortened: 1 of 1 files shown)


lalexgap:
I'm not sure if this is worth the extra complication/time and it also seems a bit clunky (ie: we don't see all the graphs).

We may want to look at the [image rendering plugin ](https://grafana.com/docs/grafana/latest/setup-grafana/image-rendering/) which renders panels directly to an image.

For now I think the existing functionality of #895 is enough and we can close this.

PR #901, opened 2022-09-27
Run Benchmark Scenario on CI

lalexgap:
Updates the testground job we kick off in CI to follow the scenario we agreed upon in standup: https://github.com/statechannels/go-nitro-testground/issues/106#issuecomment-1259875461


geoknee (approved):

geoknee:
Should we commit some information about how we chose these parameters somewhere?

lalexgap:
> Should we commit some information about how we chose these parameters somewhere?

There's some links tracing back to the original discussion in notion but should we make an ADR?

geoknee:
> > Should we commit some information about how we chose these parameters somewhere?
> 
> There's some links tracing back to the original discussion in notion but should we make an ADR?

Perhaps no need just now as it is not really an architectural decision.

PR #903, opened 2022-09-27
Use public docker image to avoid building testground on CI

lalexgap:
It turns out you can just configure a github workflow job to run using a container. This means we can specify the public [testground image](https://hub.docker.com/r/iptestground/testground) instead of building `testground` ourselves and fooling around with docker caching.

Based on this PR I've seeing a reduction from [4 minutes](https://github.com/statechannels/go-nitro/actions/runs/3138567121/jobs/5098038931) to[ 2 minutes](https://github.com/statechannels/go-nitro/actions/runs/3138593404/jobs/5098094332) for the duration of the testground action. We also get to remove some docker caching and simplify our action 🎉

### DIFF
--- a/.github/workflows/testground.yml
+++ b/.github/workflows/testground.yml
@@ -21,7 +21,7 @@ jobs:
           path: "code/go-nitro-testground"
           ref: main
       
-      # Set up caching for docker (which testground uses) and go modules
+      # Set up caching for go modules
       - uses: actions/cache@v3
         with:
           path: |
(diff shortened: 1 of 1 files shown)


.github/workflows/testground.yml
```diff
@@ -20,7 +21,7 @@ jobs:
           path: "code/go-nitro-testground"
           ref: main
       
-      # Set up caching for docker (which testground uses) and go modules
+      # Set up caching for go modules
```
geoknee, line 24: How important is this / what is it doing exactly?

geoknee (approved):
👏

.github/workflows/testground.yml
lalexgap: The docker layer caching action was providing caching when we triggered `docker build` (by calling `make install` to build testground). Now that we're not manually triggering `docker build` I don't think the action provides any caching for us.

geoknee: The comment implies there is also some caching for `go modules`? How about that?

lalexgap: https://github.com/statechannels/go-nitro/pull/903/commits/11ebb7246f7c2f768662d0fb0ad8bc79a57c4e54

Issue #186, opened 2022-01-28
Consider a new documentation website for nitro v2 protocol


geoknee:
We have docs.statechannels.org, which I think is quite good, but has the following drawbacks:
* it documents our existing solidity code from `nitro-protocol` npm package. The on-chain code for `go-nitro` may be different (it does not yet exist)
* it explains typical user flows and protocol details based on the original nitro protocol. We are now using a v2 protocol (SATP) in `go-nitro`. 
* it could do better at explaining the big picture:
    * a generalised state channel protocol
    * optimised for retrieval market-style applications
    * focus on performance / security etc
    * relation to rollups (a different set of tradeoffs)
 
* and answering the following FAQs:
    * how many `Adjudicator` contracts will there be?
    * how many custom application contracts will there be?

PR #909, opened 2022-09-30
Delete some dead code (i.e. `State` struct)

geoknee:
While working on #186 , I started to document this bit of solidity source. Then I got confused. Then I realised it was dead code 🤷


lalexgap (approved):
👍

PR #910, opened 2022-09-30
Remove `Channel` type (typescript)

geoknee:
This type:

- has not counterpart in on chain code
- just adds indirection
- is a hangover from when the channel id only depended on this type (it now depends on the full fixedpart)

Removing it makes things easier to understand.

### DIFF
--- a/nitro-protocol/src/contract/state.ts
+++ b/nitro-protocol/src/contract/state.ts
@@ -2,22 +2,11 @@ import {utils} from 'ethers';
 import {Signature} from '@ethersproject/bytes';
 import {ParamType} from 'ethers/lib/utils';
 
-import {Channel, getChannelId} from './channel';
+import {getChannelId} from './channel';
 import {Outcome} from './outcome';
 import {Address, Bytes, Bytes32, Uint256, Uint48, Uint64} from './types';
 
-/**
- * Holds all of the data defining the state of a channel
- */
-export interface State {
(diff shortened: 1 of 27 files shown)


nitro-protocol/src/contract/state.ts
```diff
@@ -2,22 +2,11 @@ import {utils} from 'ethers';
...
-export interface State {
-  turnNum: number; // TODO: This should maybe be a string b/c it is uint256 in solidity
-  isFinal: boolean;
-  channel: Channel;
-  challengeDuration: number;
-  outcome: Outcome;
-  appDefinition: string;
-  appData: string;
-}
+export type State = FixedPart & VariablePart;
```
NiloCK, line 9: :+1:

nitro-protocol/src/index.ts
```diff
@@ -38,12 +38,13 @@
...
  * ```typescript
- * import {Channel} from '@statechannels/nitro-protocol';
+ * import {State} from '@statechannels/nitro-protocol';
  *
- * const channel: Channel = {
+ * const state: State = {
  *   chainId: '0x1',
  *   channelNonce: 0,
  *   participants: ['0xalice...', '0xbob...'],
+  * // etc
```
NiloCK, line 47:
```suggestion
 *   // etc
```

NiloCK (approved):
LGTM. Agreed on justification.

PR #912, opened 2022-09-30
Adds the lib-p2p message service 

lalexgap:
This ports the message service written in the [go-nitro-testground repo](https://github.com/statechannels/go-nitro-testground/blob/main/messaging/service.go) into this repo.

I've removed the `simpletcp` message service and replaced it with a lib p2p message service. I've also removed some redundant tests since `payment_with_p2p_ms_test.go` will also implicitly direct and virtual funding.

Once this is merged in the `go-nitro-testground` repo can be updated to use this message service and it's version can be removed. That means that the testground runs on this PR are still using the **testground** libp2p version.


client/engine/messageservice/p2p-message-service/service.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+// NewMessageService returns a running P2PMessageService listening on the given ip and port
+func NewMessageService(ip string, port int, pk []byte) *P2PMessageService {
+	// We generate a random key using the hash of the pk as a seed
+	// This should mean that the message key is deterministic
+	// TODO: Ideally we would use the pk directly, but I haven't figured out of this is possible with lib p2p
+	hash := sha256.Sum256(pk)
+	seed := big.NewInt(0).SetBytes(hash[:]).Int64()
+	messageKey, _, err := p2pcrypto.GenerateECDSAKeyPair(rand.New(rand.NewSource(seed)))
```
lalexgap, line 81: I believe the key for lib p2p uses a different curve than ours, and isn't directly compatible but I'm not really sure.

client/engine/messageservice/p2p-message-service/service.go
```diff
@@ -0,0 +1,194 @@
...
+			ms.checkError(err)
+
+			writer.Flush()
+			s.Close()
+
+			return
+		}
+
+		// TODO: Hook up to a logger
+		fmt.Printf("attempt %d: could not open stream to %s, retrying in %s\n", i, msg.To.String(), RETRY_SLEEP_DURATION.String())
```
lalexgap, line 165: It's actually quite annoying to hook up the message service to the engine logger/metrics. The engine takes an instantiated message service, so the message service is created before the engine creates its logger/metricsRecorder.

We could expose some kind of `AttachEngine` function on the message service interface that the `engine` can call once it's started to share it's `logger and metric recorder but that's kind of clunky

NiloCK:
FYI I have a review-in-progress for this that I'll complete before @lalexgap's workday begins tomorrow.

client/engine/messageservice/p2p-message-service/service.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+// Id returns the libp2p peer ID of the message service
+func (ms *P2PMessageService) Id() peer.ID {
+	id, _ := peer.IDFromPrivateKey(ms.key)
+	return id
+}
+
+// AddPeers adds the peers to the message service
+// We ignore peers that are ourselves
```
NiloCK, line 54:
```suggestion
// AddPeers adds the peers to the message service.
// We ignore peers that are ourselves.
```

client/engine/messageservice/p2p-message-service/service.go
```diff
@@ -0,0 +1,196 @@
...
+		info, _ := peer.AddrInfoFromP2pAddr(multi)
+		// Add the destination's peer multiaddress in the peerstore.
+		// This will be used during connection and stream creation by libp2p.
+		ms.p2pHost.Peerstore().AddAddrs(info.ID, info.Addrs, peerstore.PermanentAddrTTL)
+
+		ms.peers.Store(p.Address.String(), info.ID)
+	}
+}
+
+// NewMessageService returns a running P2PMessageService listening on the given ip and port
```
NiloCK, line 74:
```suggestion
// NewMessageService returns a running P2PMessageService listening on the given ip and port.
```

client/engine/messageservice/p2p-message-service/service.go
```diff
@@ -0,0 +1,196 @@
...
+	// This should mean that the message key is deterministic
+	// TODO: Ideally we would use the pk directly, but I haven't figured out of this is possible with lib p2p
+	hash := sha256.Sum256(pk)
+	seed := big.NewInt(0).SetBytes(hash[:]).Int64()
+	messageKey, _, err := p2pcrypto.GenerateECDSAKeyPair(rand.New(rand.NewSource(seed)))
+	if err != nil {
+		panic(err)
+	}
+	options := []libp2p.Option{libp2p.Identity(messageKey),
+		libp2p.ListenAddrStrings(fmt.Sprintf("/ip4/%s/tcp/%d", ip, port)),
```
NiloCK, line 86: I'd be interested to see real-world usage where the public facing IP address is determined - I see that this currently only receives 127.0.0.1

client/engine/messageservice/p2p-message-service/service.go
```diff
@@ -0,0 +1,196 @@
...
+	h := &P2PMessageService{
+		out:     make(chan protocols.Message, BUFFER_SIZE),
+		peers:   &safePeers,
+		p2pHost: host,
+		quit:    make(chan struct{}),
+		key:     messageKey,
+		me:      crypto.GetAddressFromSecretKeyBytes(pk),
+	}
+
+	h.p2pHost.SetStreamHandler(MESSAGE_ADDRESS, func(stream network.Stream) {
```
NiloCK, line 106: If I'm reading this correctly
- each go-nitro protocol message will initiate a new stream
- the message service here reads from the stream until encountering a `\n`ewline
- the read message is parsed & returned to the engine

Are there potential performance impacts here? If Alice and Bob are interacting in a VC, do we expect much performance gain with something like:

``` go
for raw, err := reader.ReadString(DELIMITER); err == nil; {
	m, err := protocols.DeserializeMessage(raw)

	h.checkError(err)
	h.out <- m
}
```

which reads sequential messages on the same stream?

(I've seen a little lib-p2p documentation that refers separately to `streams` and `connections`, so maybe the stream creation / destruction is cheap, given an existing connection?)

client/engine/messageservice/p2p-message-service/service.go
```diff
@@ -0,0 +1,196 @@
...
+			stream.Close()
+		}
+
+	})
+
+	return h
+
+}
+
+// Send sends messages to other participants
```
NiloCK, line 137:
```suggestion
// Send sends messages to other participants.
```

client/engine/messageservice/p2p-message-service/service.go
```diff
@@ -0,0 +1,196 @@
...
+	raw, err := msg.Serialize()
+	ms.checkError(err)
+
+	id, ok := ms.peers.Load(msg.To.String())
+	if !ok {
+		panic(fmt.Errorf("could not load peer %s", msg.To.String()))
+	}
+
+	for i := 0; i < NUM_CONNECT_ATTEMPTS; i++ {
+		s, err := ms.p2pHost.NewStream(context.Background(), id, MESSAGE_ADDRESS)
```
NiloCK, line 150: Similar on the streamHandler. Are we meant to be "looking up" existing streams rather than constructing new ones?

NiloCK:
Meta Q, but is it worthwhile to leave the simpleTCP service as an in-test reference implementation?

eg, unit tests should produce the same results independent of injected message service, and the redundancy is maybe a good baseline for the evolving libp2p based service?

client/engine/engine.go
```diff
@@ -447,6 +447,7 @@ func (e *Engine) executeSideEffects(sideEffects protocols.SideEffects) error {
 
 	for _, message := range sideEffects.MessagesToSend {
 		e.logMessage(message, Outgoing)
+		e.recordMessageMetrics(message)
```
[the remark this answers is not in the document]
lalexgap: > Related: do we have any specific plan for producing go-nitro builds that exclude the metrics & timing logging functions?

If a `nil` metrics is passed into the [client constructor](https://github.com/statechannels/go-nitro/blob/db9a1d158d37b166759d304e51bcf13300e3e29d/client/client.go#L38) we use `NoOpMetrics`. `NoOpMetrics` doesn't record or time anything, so it's essentially excluding metrics/logging.

client/engine/messageservice/p2p-message-service/service.go
lalexgap: I think we use publish-ish IPs in the [testground test](http://34.168.92.245:8042/logs?task_id=cctj4j8nr2gg1ntf7310)
```
I am {PeerInfo:{PeerInfo:{SNIP IpAddress:16.14.0.38} Role:2} SNIP}}
```

client/engine/messageservice/p2p-message-service/service.go
lalexgap: > which reads sequential messages on the same stream?

Originally the message service would open a `stream` per peer at the start and then reuse that `stream` whenever we want to send messages.

Unfortunately I kept running into `stream reset` errors. It seemed like clients struggled to keep all the streams open (maybe when under load) for the duration of the test. To avoid this I just switched to creating the stream each time we want to send a message.

>(I've seen a little lib-p2p documentation that refers separately to streams and connections, so maybe the stream creation / destruction is cheap, given an existing connection?)

I believe `connections` are seperate than `stream`s. ([Docs](https://github.com/libp2p/specs/blob/master/connections/README.md)). A connection is automatically established when we open our first stream and we then re-use that connection to open streams (I think?). I'm not sure how "cheap" a stream really is.

Longer term maybe our message service could reuse streams but gracefully handle reset errors when they occur?

I should also look at benchmarking the reuse vs new stream scenarios.

lalexgap:
> Meta Q, but is it worthwhile to leave the simpleTCP service as an in-test reference implementation?
> 
> eg, unit tests should produce the same results independent of injected message service, and the redundancy is maybe a good baseline for the evolving libp2p based service?

It would be nice to test the engine with multiple message services! However the `simpleTCP` and libp2p message services  do share a lot of overlap, as they both send messages directly over TCP. We also have the in-memory message service we use for most of our unit tests.

I also want to promote the use of the libp2p message service so it can be tested as thoroughly as possible. By removing `SimpleTCP` there's no confusion on which message service to use when writing a new test.

That being said I'd be happy to keep the `simpleTCP` message service if others feel like it makes sense to!

lalexgap:
I'm going to merge this is in now and we can tackle changes related to the [open](https://github.com/statechannels/go-nitro/pull/912#issuecomment-1267351875) [discussions](https://github.com/statechannels/go-nitro/pull/912#discussion_r98702042) in another PR.

PR #913, opened 2022-10-03
refactor: Funded -> IsFundingTheTarget...

NiloCK:
`Funded` as an adjective describes the desired final state of the target virtual channel, not a state of the funding ledger channels. This tripped me up last week.

(Notable: `Funded` could also be read as "did fund", ie, "has performed its funding duties", which would be correct, but I think the refactor is less ambiguous)

### DIFF
--- a/protocols/virtualfund/virtualfund.go
+++ b/protocols/virtualfund/virtualfund.go
@@ -92,8 +92,8 @@ func (c *Connection) handleProposal(sp consensus_channel.SignedProposal) error {
 	return nil
 }
 
-// Funded computes whether the ledger channel on the receiver funds the guarantee expected by this connection
-func (c *Connection) Funded() bool {
+// IsFundingTheTarget computes whether the ledger channel on the receiver funds the guarantee expected by this connection
+func (c *Connection) IsFundingTheTarget() bool {
 	g := c.getExpectedGuarantee()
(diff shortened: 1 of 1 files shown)


protocols/virtualfund/virtualfund.go
```diff
@@ -92,8 +92,8 @@ func (c *Connection) handleProposal(sp consensus_channel.SignedProposal) error {
 	return nil
 }
 
-// Funded computes whether the ledger channel on the receiver funds the guarantee expected by this connection
-func (c *Connection) Funded() bool {
+// IsFundingTheTarget computes whether the ledger channel on the receiver funds the guarantee expected by this connection
+func (c *Connection) IsFundingTheTarget() bool {
 	g := c.getExpectedGuarantee()
 	return c.Channel.Includes(g)
 }
```
geoknee, line 99: How about  `includesExpectedGuarantee`?

geoknee (approved):

protocols/virtualfund/virtualfund.go
NiloCK: I'm going to leave this as is because it:
- is a more "active voice"
- asks less of the reader: someone can tell that the connection has "done its job" with respect to the virtualfund objective without knowing anything about the internals of a ledger channel (including what a guarantee is, for example)

PR #917, opened 2022-10-03
Setup Nightly Testground Run

lalexgap:
Towards https://github.com/statechannels/go-nitro-testground/issues/107

This PR adds a scheduled github workflow, that starts a testground run nightly at 5AM UTC.   Since this is only run nightly we run it for ~5~3 minutes (shortened to 3 min to avoid this [issue](https://github.com/statechannels/go-nitro-testground/issues/116)). 

To test the workflow I set the [nightly workflow to run on PR](https://github.com/statechannels/go-nitro/actions/runs/3177472243/jobs/5177949776) before switching to `schedule`. Only scheduled workflows in `main` are picked up, so we'll have to merge this to test the scheduling bit.

This PR also updates both testground workflows to set the `isCI`/`isNightly` flags in anticipation of https://github.com/statechannels/go-nitro-testground/pull/119. This will allow us to track stats of our nightly/ci testground runs over time. `testground` will ignore the parameters if they're not defined so this is safe to merge in before [the PR](https://github.com/statechannels/go-nitro-testground/pull/119).

### DIFF
--- a/.github/workflows/testground-nightly.yml
+++ b/.github/workflows/testground-nightly.yml
@@ -1,0 +1,47 @@
+name: Run Testground Nightly
+# TODO: Change this to run on a schedule
+on:
+  pull_request:
+# on:
+#   schedule:
+#     - cron:  '13 20 * * *'
+jobs:
+  run-testground-nightly:
+      runs-on: ubuntu-latest
+      container: iptestground/testground:edge
+      steps:
+      - uses: actions/setup-go@v3
+        with:
+          go-version: "^1.19.0"
+      # Get testground and our test plan code
+      - uses: actions/checkout@v3
+        with:
+         repository: 'statechannels/testground'
+         path: "code/testground"
+      - uses: actions/checkout@v3
+        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/testground-nightly.yml
```diff
@@ -0,0 +1,45 @@
...
+        working-directory: "code/go-nitro-testground"
+
+      - name: Import Test 
+        run:    testground plan import --from ./go-nitro-testground
+        working-directory: "code"
+
+      - name: Run 3 min Test 
+        run: | 
+         testground --endpoint=${{secrets.TG_SERVER_URL}}  run s --wait \
+         -tp=isNightly=true -p=go-nitro-testground -t=virtual-payment \
```
geoknee, line 37: Not super important but in your opinion does this nightly run count as "CI"?

lalexgap: Since this test has a longer duration I thought it might make sense to keep it out of the CI metric data. That way we're comparing like to like. I don't feel too strongly about it though!

PR #920, opened 2022-10-04
Add create2 deployer to nitro-protocol and abi to go-nitro

kerzhner:
Create2Deployer has no Nitro specific functionality. This utility could be pulled out to a seperate npm module and go package. However, it is convenient to include the utility in `go-nitro` (at least for now).

PR #921, opened 2022-10-04
Refactor generate-adjudicator-bindings.sh

kerzhner:
- Use `runAbigen` function instead of copy/pasting `abigen` command.
- Break up long line into multiple lines.

### DIFF
--- a/generate-adjudicator-bindings.sh
+++ b/generate-adjudicator-bindings.sh
@@ -1,19 +1,25 @@
 set -e
 cd nitro-protocol
 
-solc --base-path $(pwd) @statechannels/exit-format/=node_modules/@statechannels/exit-format/ @openzeppelin/contracts/=node_modules/@openzeppelin/contracts/ contracts/NitroAdjudicator.sol \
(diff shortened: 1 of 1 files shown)


geoknee (approved):
Looks good but I think there are some unintended files checked in? Also you have some conflicts.

kerzhner:
> Looks good but I think there are some unintended files checked in? Also you have some conflicts.

The PR diff did not make sense due to https://github.com/statechannels/go-nitro/pull/920 getting merged in. This feature branch has been rebased on the latest `main`.

geoknee:
I think we may have merged some broken code, here. 

- The "check bindings" action did not run on this PR because we made no changes to the contracts:
https://github.com/statechannels/go-nitro/blob/d9f17f0d93c09514f6f5cc113dee7f6882c040b8/.github/workflows/bindings-check.yml#L6-L7
(We should run the action when the shell script is modified, too)
- the "check bindings" action does not run on `main`, so we haven't yet detected this problem (but soon will on a pull request)

PR #922, opened 2022-10-04
Periodically resubscribe to events

lalexgap:
It looks like `geth` has an[ open issue](https://github.com/ethereum/go-ethereum/issues/23845) around `SubscribeFilterLogs` failing to report events after it's been running for some time. This is similar to the behaviour we see in our [long running test failure](https://github.com/statechannels/go-nitro-testground/issues/116).

Based on some code snippets on the [issue](https://github.com/ethereum/go-ethereum/issues/23845#issuecomment-1035800486) this PR resubscribes to the event log every 60 seconds. This avoids the problem by ensuring that we don't have long running subscription.


kerzhner:
Do you think our issue and https://github.com/ethereum/go-ethereum/issues/23845 are related? The geth issue description talks about missing events after days of running a script, not minutes...

In general, having to resubscribe to events every minute seems a bit wild. Also, is it possible to miss events while a subscription is inactive (after cancellation but before resubscription)?

client/engine/chainservice/eth_chainservice.go
```diff
@@ -111,25 +116,41 @@ func (ecs *EthChainService) SendTransaction(tx protocols.ChainTransaction) error
...
-
-func (ecs *EthChainService) listenForLogEvents(sub ethereum.Subscription, logs chan ethTypes.Log) {
 	for {
 		select {
 		case err := <-sub.Err():
-			// TODO should we try resubscribing to chain events
-			ecs.logger.Printf("event subscription error: %v", err)
+			// If the error is nil then the subscription was closed and we need to re-subscribe.
+			// This is a workaround for https://github.com/ethereum/go-ethereum/issues/23845
+			if err == nil {
```
kerzhner, line 139: A level of nesting can be avoided if this is refectored to:
``` golang
if err != nil {
  panic(err)
}
//...remaining logic
```

client/engine/chainservice/eth_chainservice.go
```diff
@@ -111,25 +116,41 @@ func (ecs *EthChainService) SendTransaction(tx protocols.ChainTransaction) error
 }
 
 func (ecs *EthChainService) subcribeToEvents() error {
+
+	go ecs.listenForLogEvents()
+	return nil
+}
+
+func (ecs *EthChainService) listenForLogEvents() {
```
kerzhner, line 124: I am probably missing something, but why is the below logic moved to `listenForLogEvents`?

lalexgap: > Do you think our issue and [ethereum/go-ethereum#23845](https://github.com/ethereum/go-ethereum/issues/23845) are related? The geth issue description talks about missing events after days of running a script, not minutes...

Based on our test run behaviour I'm pretty sure that our issue is being caused by the [geth issue](https://github.com/ethereum/go-ethereum/issues/23845) as:
- Without this change (with a long running test) we see the withdraw transaction submitted but no chain events in the logs
- Without this change tests will fail if they run over ~3 minutes


>In general, having to resubscribe to events every minute seems a bit wild.

Agreed 😔. 

>Also, is it possible to miss events while a subscription is inactive (after cancellation but before resubscription)?

That's a good point! I haven't run into that problem when running testground runs. I think we may be ok but I'll investigate.

kerzhner: > Based on our test run behaviour I'm pretty sure that our issue is being caused by the [geth issue](https://github.com/ethereum/go-ethereum/issues/23845) as:
> 
> * Without this change (with a long running test) we see the withdraw transaction submitted but no chain events in the logs
> * Without this change tests will fail if they run over ~3 minutes

For a test with a long runtime, does the test open/close more on-chain channels than a short running test? Or do longer running tests open/use/close just more virtual channels?

We use a websocket for connecting to an [Ethereum JSON rpc endpoint](https://github.com/statechannels/go-nitro-testground/blob/a0a5a200487d36fb1e62e057ef9212f55e67c79b/chain/chain-service.go#L22). I wonder if there is some timeout issues with that websocket...

lalexgap: >For a test with a long runtime, does the test open/close more on-chain channels than a short running test? Or do longer running tests open/use/close just more virtual channels?

Nope, the long running test would create the same amount of ledger channels, so we should have the same amount of on chain channels (every participant creates a ledger channel with every hub).

client/engine/chainservice/eth_chainservice.go
lalexgap: Updated in https://github.com/statechannels/go-nitro/pull/922/commits/0c41f84611fb6ccd6ab86b0c48530d7c02096646

client/engine/chainservice/eth_chainservice.go
lalexgap: > I am probably missing something, but why is the below logic moved to `listenForLogEvents`?

Sorry I accidentally replied to this comment for another issue 🙃.  Since `listenForLogEvents` is now periodically recreating `sub` every 60 seconds I though it made more sense for `listenForLogEvents` to create the `sub` the first time`. This required moving the logic into `listenForLogEvents`. Otherwise `listenForLogEvents` would accept a `sub` that it overwrites in 60 seconds.

lalexgap: I did a little spelunking through `go-ethereum` and found [this](https://github.com/ethereum/go-ethereum/blob/e14164d516600e9ac66f9060892e078f5c076229/eth/filters/filter_system.go#L43):

```
// Config represents the configuration of the filter system.
type Config struct {
	LogCacheSize int           // maximum number of cached blocks (default: 32)
	Timeout      time.Duration // how long filters stay active (default: 5min)
}
```

5 minutes seems roughly in line with the failure's we've seen, so that might explain it.

kerzhner: Does it makes sense to change our resubscription period to something like 4 minutes?

client/engine/chainservice/eth_chainservice.go
```diff
@@ -111,25 +116,41 @@ func (ecs *EthChainService) SendTransaction(tx protocols.ChainTransaction) error
 }
 
 func (ecs *EthChainService) subcribeToEvents() error {
+
```
kerzhner, line 119:
```suggestion
```

client/engine/chainservice/eth_chainservice.go
lalexgap: > Does it makes sense to change our resubscription period to something like 4 minutes?

Yup, now that we know it's a 5 minute timeout it makes sense to use a larger value.

lalexgap:
> Approving this PR, but I am worried about:
> 
> * Missing events between unsubscribe and resubscribe.
> * Duplicate events if resubscribe has a lookback window for events.

As discussed in slack the plan is to merge this in and keep an eye out for any duplicate or missing events

PR #928, opened 2022-10-06
Fail on testground run failure

### LINKED ISSUE #925: Testground CI run can report success when it fails to build
Here is an [example](https://github.com/statechannels/go-nitro/actions/runs/3191648440/jobs/5208207313). I think we do get an error back when the test fails, but if building fails it seems to report success.

lalexgap:
Fixes #925 
Add a to step to the github workflow that checks that the testground run returns `success`. If not it calls `exit 1`.

### DIFF
--- a/.github/workflows/testground.yml
+++ b/.github/workflows/testground.yml
@@ -50,7 +50,11 @@ jobs:
          --metadata-branch "${{github.event.pull_request.head.ref}}" \
          --metadata-commit "${{github.event.pull_request.head.sha}}"  | tee run.out
           
-      # Parse the run id from the testground run output
+      # Parse the run outcome and id from the testground run output
+      - name: Set Run Outcome
+        id: set_run_outcome
+        run: | 
+         echo "::set-output name=RUN_OUTCOME::$(awk '/run finished with outcome/ {print $10}' <run.out)"
       - name: Set Run Id
         id: set_run_id
         run: | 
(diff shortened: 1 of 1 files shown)


.github/workflows/testground.yml
```diff
@@ -50,7 +50,11 @@ jobs:
          --metadata-branch "${{github.event.pull_request.head.ref}}" \
          --metadata-commit "${{github.event.pull_request.head.sha}}"  | tee run.out
           
-      # Parse the run id from the testground run output
+      # Parse the run outcome and id from the testground run output
+      - name: Set Run Outcome
+        id: set_run_outcome
+        run: | 
+         echo "::set-output name=RUN_OUTCOME::$(awk '/run finished with outcome/ {print $10}' <run.out)"
```
geoknee, line 57: In the case of a failing build, what will be the value of `RUN_OUTCOME` ? I guess it won't be `'success'`, so this will still cause the GHA to fail due to the check on L89 (which is what we want). Might be worth a comment to explain?

geoknee: According to #925, the problem is about catching (i) failing builds rather than (ii) successful builds that fail (due to timeout or something). I think this solution does fix the problem, it is just a bit unexpected because the solution is to attempt to extract part of a string which will never appear in case (i). And in case (ii) it will appear, but #925 suggests we don't have a problem in such cases anyway?

Sorry for nitpicking, I just want to check I understand this properly.

geoknee (approved):

.github/workflows/testground.yml
lalexgap: You're right, the issue is only about the building step but this fix is broader than that. I think I may have seen some non-build failures as well (even though I claimed in #925 I only saw it on build) I.figured I'd make a more robust check that will only pass if we know the testground run was successful. This resolves the builds failing and reporting success but it should also catch any other cases where testground returns success when the run doesn't fully succeed.

geoknee: Got you! 👍

Issue #904, opened 2022-09-28
Generalize to `n` hop virtual channels


geoknee:
This would be a stretch goal for Lisbon.

NiloCK:
[ ] #907 Add tests that hit modified `ConstructFromState`

geoknee:
@NiloCK what are the remaining sub goals we need to hit to close this issue?

PR #932, opened 2022-10-11
Test the n-hop virtual fund

NiloCK:
Towards #904 

This PR
- modifies `vitualfund_test.go` to create 0, 1, and 2 hop virtually funded channels
- modifies `virtualfund.NewObjective()` to allow 0-hop funding (ie, persons with a ledger channel can use that channel to fund a virtual channel between them)

---

#### Code quality

#### Project management

### DIFF
--- a/client_test/virtualfund_test.go
+++ b/client_test/virtualfund_test.go
@@ -47,11 +47,66 @@ func TestVirtualFundIntegration(t *testing.T) {
 	chainServiceA := chainservice.NewMockChainService(chain, alice.Address())
 	chainServiceB := chainservice.NewMockChainService(chain, bob.Address())
 	chainServiceI := chainservice.NewMockChainService(chain, irene.Address())
+	chainServiceBr := chainservice.NewMockChainService(chain, brian.Address())
 	broker := messageservice.NewBroker()
 
 	clientA, _ := setupClient(alice.PrivateKey, chainServiceA, broker, logDestination, 0)
+	irene, _ := setupClient(irene.PrivateKey, chainServiceI, broker, logDestination, 0)
+	ivan, _ := setupClient(brian.PrivateKey, chainServiceBr, broker, logDestination, 0)
 	clientB, _ := setupClient(bob.PrivateKey, chainServiceB, broker, logDestination, 0)
-	clientI, _ := setupClient(irene.PrivateKey, chainServiceI, broker, logDestination, 0)
 
-	openVirtualChannels(t, clientA, clientB, clientI, 1)
+	openN_HopVirtualChannels(t, []client.Client{clientA, irene, ivan, clientB}, 1)
+}
+
+// openN_HopVirtualChannels connects the n given participants in a line of ledger channels,
+// then uses these ledger connections to open channels between the first participant
+// and each other participant.
+//
+// This makes channels with 0, 1, ..., [len(participants) - 1] hops.
+func openN_HopVirtualChannels(t *testing.T, participants []client.Client, channelsPerCounterparty uint) {
+
+	// set a chain of ledger channels between incoming clients.
+	// network is a line: A <-> B <-> C <-> D ... <-> X
+	for i, participant := range participants {
+		if i+1 < len(participants) {
+			directlyFundALedgerChannel(t, participant, participants[i+1])
+		}
+	}
+
+	alice := participants[0] // alice initiates each virtualfund operation
+	counterparties := participants[1:]
+
+	for i := 0; i < int(channelsPerCounterparty); i++ {
+
+		// and funds channels with everyone else along the chain
+		// (including her immediate neighbor)
+		for j, bob := range counterparties {
+
+			intermediaries := counterparties[0:j]
+			intermediaryAddresses := clientsToAddresses(intermediaries)
+
+			outcome := td.Outcomes.Create(*alice.Address, *bob.Address, 1, 1)
+			response := alice.CreateVirtualPaymentChannel(
+				intermediaryAddresses,
+				*bob.Address,
+				0,
+				outcome,
+			)
+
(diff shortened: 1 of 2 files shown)


NiloCK:
> > * modifies `virtualfund.NewObjective()` to allow 0-hop funding (ie, persons with a ledger channel can use that channel to fund a virtual channel between them)
> 
> Does a 0 hop virtual channel make sense? Would the virtual channel be funded by a single guarantee in that case?

Yes, funded by a single guarantee. (If this sets off alarm bells, be assured that it is secure because both participating parties are members of the channel with that guarantee. 1-hop needs two funding guarantees so that each participant has a challenge path of channels **of which they are members** back to chain).

 I think it makes sense in at least two known ways:
- it provides a simple mechanism for two participants in a shared channel to implement 'lanes' based payments. ClientA and ProviderB can partition channel funds into separate file/sector/job-specific channels
- it's the mechanism for hubs to [virtualize their own ledger channels](https://www.notion.so/statechannels/Network-Design-Consideration-Virtual-Ledger-Channels-915c4a888f264206a9979cf5319a9f58)

geoknee:
I think we need to think carefully about how that sort of funding graph compares with one using a [simple allocation](http://localhost:8000/protocol-tutorial/0030-outcomes/#allocations) 

Also, a note on terminology here. Typically we have used "indirect funding" or "ledger funding" to refer to the use of a simple allocation to fund one channel from another:
http://localhost:8000/protocol-tutorial/0060-funding-a-channel/#fund-from-an-existing-channel, and "virtual funding" to fund one channel from two or more other channels.

It seems awkward for there to be two ways of doing "indirect funding" (one channel funding another channel). Can we tidy that up some how?

client_test/virtualfund_test.go
```diff
@@ -47,11 +47,66 @@ func TestVirtualFundIntegration(t *testing.T) {
...
+	}
+
+	alice := participants[0] // alice initiates each virtualfund operation
+	counterparties := participants[1:]
+
+	for i := 0; i < int(channelsPerCounterparty); i++ {
+
+		// and funds channels with everyone else along the chain
+		// (including her immediate neighbor)
+		for j, bob := range counterparties {
```
geoknee, line 83: Nit: potentially confusing to have the counterparty called "bob" each time, since we usually think of Bob as a fixed actor in the virtualfund protocol? Could we just use `counterparty`?

client_test/virtualfund_test.go
```diff
@@ -47,11 +47,66 @@ func TestVirtualFundIntegration(t *testing.T) {
...
-	openVirtualChannels(t, clientA, clientB, clientI, 1)
+	openN_HopVirtualChannels(t, []client.Client{clientA, irene, ivan, clientB}, 1)
+}
+
+// openN_HopVirtualChannels connects the n given participants in a line of ledger channels,
+// then uses these ledger connections to open channels between the first participant
+// and each other participant.
+//
+// This makes channels with 0, 1, ..., [len(participants) - 1] hops.
+func openN_HopVirtualChannels(t *testing.T, participants []client.Client, channelsPerCounterparty uint) {
```
geoknee, line 66: The `channelsperCounterparty` is not documented, and it is also not being used nontrivially here as far as I can tell? Maybe we should drop it until we need it?

protocols/virtualfund/virtualfund.go
```diff
@@ -137,7 +137,14 @@ type Objective struct {
 
 // NewObjective creates a new virtual funding objective from a given request.
 func NewObjective(request ObjectiveRequest, preApprove bool, myAddress types.Address, getTwoPartyConsensusLedger GetTwoPartyConsensusLedgerFunction) (Objective, error) {
-	rightCC, ok := getTwoPartyConsensusLedger(request.Intermediaries[0])
+	var rightCC *consensus_channel.ConsensusChannel
+	ok := false
+
+	if len(request.Intermediaries) > 0 {
+		rightCC, ok = getTwoPartyConsensusLedger(request.Intermediaries[0])
```
geoknee, line 144: It might make sense to separate out the support for 0-hop channels, since that is quite a major conceptual leap (even if it is one that seems to make good sense)?

Or, if we want this to go in now we can make some updates to documentation somewhere and point to them here?

client_test/virtualfund_test.go
geoknee, line 66: Does this function have canonical casing? Seems jarring to have an underscore in there when it is otherwise camel cased?

geoknee (approved):

client_test/virtualfund_test.go
NiloCK: I have renamed this to `concurrency` in some follow-up work, which is in line with usage on the testground repo.

Moving forward, I'd like to set up a simple script that shoots out graphs like the ones I had put in slack yesterday:

image

protocols/virtualfund/virtualfund.go
NiloCK: I'm in favour of putting it in - I am still in progress with the docs site review and should be able to suggest the appropriate changes quickly.

client_test/virtualfund_test.go
geoknee: You're right of course, I got confused. I think it is ok to have `bob` here, since each counterparty is in face playing the role of bob from the point of view of the `virtualfund` protocol.

client_test/virtualfund_test.go
NiloCK: We hyphenate n-hop in plaintext, but the hyphen isn't an allowed character in an identifier.

Also, the single-letter N wanted capitalization, and the word hop did as well. Putting both together loses the effect of capitals separating words.

In short, I tried a few different casing / spacing things, but all felt bad.

geoknee: OK. I don't feel strongly. But I think `openNHopVirtualChannels` is fine.

Issue #883, opened 2022-09-20
Closing a ledger channel too early silently fails


lalexgap:
Currently if we attempt to start a `DirectDefund` objective but there are running guarantees in the ledger channel we return an error.

However we swallow that [error in the engine](https://github.com/statechannels/go-nitro/blob/4ed5c8be2bfeb304f23b5b02184890bfa53f457c/client/engine/engine.go#L140). This means if a someone calls `client.CloseLedgerChannel` too early the objective silently fails. To the caller of `client.CloseLedgerChannel` it looks like the API succeeded (since the API returns an objective id) but in reality the objective doesn't even get created!

lalexgap:
This is similar to https://github.com/statechannels/go-nitro/issues/744 where we receive a message to start a new objective before we're ready. In this particular case we also swallowed the error so it wasn't apparent that the objective had failed.

lalexgap:
> What do you think is the correct behaviour of `go-nitro` when the "user" requests to directly defund a "non-empty" ledger channel?

For the short term (and maybe medium term?) solution I think the client reporting an error is sufficient enough. I think the client would only run into this if they're attempting to close a ledger channel before they're done using it so it's easy to avoid. In the testground test [we weren't properly waiting](https://github.com/statechannels/go-nitro-testground/pull/98) for virtual defund objectives so that explains why we ran into this.

Longer term I think something like your option 1 makes sense. It would be nice to apply this to virtual defunding as well to address #744. My preference would be create a new issue to implement option 1 and tackle that later.

> Ensure no new guarantees are added into the ledger channel (perhaps install a policy to reject any requests to do so)

I wonder the best way to tackle this. Maybe something similar to channel ownership we already do in the `store`.

geoknee:
Decision in standup today:

For now we will generate an error when this happens (we don't expect it often but we can't rule it out). 

We will save the longer -term solutions for a future time.

PR #934, opened 2022-10-12
Stop swallowing ErrNotEmpty Error

lalexgap:
Per our [discussion](https://github.com/statechannels/go-nitro/issues/883#issuecomment-1255272825) on #883 we decided to stop swallowing the `ErrNotEmpty`.

This PR implements that change.

### DIFF
--- a/client_test/directdefund_test.go
+++ b/client_test/directdefund_test.go
@@ -3,14 +3,11 @@ package client_test // import "github.com/statechannels/go-nitro/client_test"
 
 import (
 	"testing"
-	"time"
 
 	"github.com/statechannels/go-nitro/client"
 	"github.com/statechannels/go-nitro/client/engine/chainservice"
 	"github.com/statechannels/go-nitro/client/engine/messageservice"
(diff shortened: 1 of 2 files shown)


client_test/directdefund_test.go
```diff
@@ -83,27 +77,4 @@ func TestDirectDefund(t *testing.T) {
 		}
 	}
 
-	// test failure of teardown of a ledger channel currently funding a virtual channel
```
geoknee, line 86: Did you mean to remove this code? I don't quite see the motivation / connection to not swallowing the error?

lalexgap: Yup, although I should have made it clear why. Now that we're no longer swallowing the error, it bubbles up to engine where it [panics](https://github.com/statechannels/go-nitro/blob/f41a8b62e5da8bd5cc11f0e682dcc9149f79a71b/client/engine/engine.go#L151).

Since the engine run loop is in a different go routine, it's pretty hard to catch the panic in the test. So I just removed the test for now.

geoknee (approved):

PR #936, opened 2022-10-14
WIP add section on go-nitro client API to docs

geoknee:
~stacked on #872~

PR #937, opened 2022-10-14
Expose `Client.WaitForObjectivesToComplete()`

geoknee:
While working on #936, it occurred to me that we make heavy use of a test utility `waitTimeForCompletedObjectiveIds` in our integration tests. There is currently nothing like this available to consumers of go-nitro, placing a burden on those consumers to write their own boilerplate for this. 

This PR adds such a method, which:
* does _not_ accept  an explicit timeout parameter, but instead accepts a `context` which may be cancelled
* is used to refactor the existing test utility, using a cancellable `context`
* is used by example in one place (only) of our tests, reducing the boilerplate for consumers. We should either
  - roll this out across all tests
  - revert the change if we want tests to really wait a certain amount of time for such calls to resolve

NOTE 
We rolled something similar in out testground monitor called a completion monitor. We could consider absorbing that into the go-nitro client instead of using this PR -- let's discuss and maybe take the best of both. 
https://github.com/statechannels/go-nitro-testground/blob/3df3fa6b1ee1efbbfa1d5cc9fc69532cde3d9860/utils/monitor.go#L22

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -2,6 +2,7 @@
 package client // import "github.com/statechannels/go-nitro/client"
 
 import (
+	"context"
 	"io"
 	"math/big"
 	"math/rand"
@@ -164,4 +165,39 @@ func (c *Client) CloseLedgerChannel(channelId types.Destination) protocols.Objec
 func (c *Client) Pay(channelId types.Destination, amount *big.Int) {
 	// Send the event to the engine
 	c.engine.PaymentRequestsFromAPI <- engine.PaymentRequest{ChannelId: channelId, Amount: amount}
+}
+
+// WaitForObjectivesToComplete returns a mapping showing which of the supplied objective ids is completed, and a channel which will recieve an empty struct when all have done so.
+func (c *Client) WaitForObjectivesToComplete(ctx context.Context, ids ...protocols.ObjectiveId) (map[protocols.ObjectiveId]bool, chan struct{}) {
+
+	doneChan := make(chan struct{})
+
+	completed := make(map[protocols.ObjectiveId]bool)
+
+	handleCompletedObjectiveAndCheckAllCompleted := func(got protocols.ObjectiveId) bool {
+		allCompleted := true
+		for _, id := range ids {
+			if id == got {
+				completed[id] = true
+			} // TODO if the id is irrelevant to us, we could resend it on the CompletedObjective chan for others to read?
+			allCompleted = allCompleted && (completed[id])
+		}
+		return allCompleted
+	}
+
+	go func() {
+		for {
+			select {
+			case got := <-c.CompletedObjectives():
+				allCompleted := handleCompletedObjectiveAndCheckAllCompleted(got)
+				if allCompleted {
+					doneChan <- struct{}{}
+					return
+				}
+			case <-ctx.Done():
+				return
+			}
+		}
+	}()
+	return doneChan
 }
--- a/client/client.go
+++ b/client/client.go
@@ -167,20 +167,25 @@ func (c *Client) Pay(channelId types.Destination, amount *big.Int) {
 	c.engine.PaymentRequestsFromAPI <- engine.PaymentRequest{ChannelId: channelId, Amount: amount}
 }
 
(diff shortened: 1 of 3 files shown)


client_test/directfund_test.go
```diff
@@ -58,7 +59,7 @@ func TestWhenObjectiveIsRejected(t *testing.T) {
 	outcome := testdata.Outcomes.Create(alice.Address(), bob.Address(), ledgerChannelDeposit, ledgerChannelDeposit)
 	response := clientA.CreateLedgerChannel(bob.Address(), 0, outcome)
 
-	waitTimeForCompletedObjectiveIds(t, &clientA, time.Second, response.Id)
+	<-clientA.WaitForObjectivesToComplete(context.Background(), response.Id)
```
lalexgap, line 62: Outside of the scope of this PR but would it make sense for any of the other API functions to take in a context?

client/client.go
```diff
@@ -165,3 +166,44 @@ func (c *Client) Pay(channelId types.Destination, amount *big.Int) {
 	// Send the event to the engine
 	c.engine.PaymentRequestsFromAPI <- engine.PaymentRequest{ChannelId: channelId, Amount: amount}
 }
+
+// WaitForObjectivesToComplete returns a channel on which
+// a mapping showing which of the supplied objective ids is still incomplete
+// will be sent either when all objectives have been completed, or when the supplied context is cancelled.
+func (c *Client) WaitForObjectivesToComplete(ctx context.Context, ids ...protocols.ObjectiveId) chan map[protocols.ObjectiveId]bool {
```
lalexgap, line 173: I think this would really help the API!

lalexgap:
Have we thought about exposing API functions to allow clients to check on the status of the objective? That would allow clients to easily implement `WaitForObjectivesToComplete` themselves:

```
for{
   status:= client.GetObjectiveStatus(id)
   if status == Status.Complete {
      break;
   }
  time.sleep(someValue)
}
```

This would also allow clients a bit more flexibility on how they handle objective completion.

To support that we'd probably need to add some code to query the store on demand.

lalexgap:
Slightly related to my comment above (and also not really in scope of this PR) but I think we should start looking at exposing functions on the API to query channel/objective data. Right now the client is a bit of a blackbox. We call the API and get events when objectives succeed,  but there's no way to query things to see things like our current ledger balances.

It would be really nice if the client API had:
- A function to get ledger channels and their balances
- A function to get running payment channels and their balances
- A function to get the status of ledger/payment channel creation/closing

client/client.go
```diff
@@ -165,3 +166,44 @@ func (c *Client) Pay(channelId types.Destination, amount *big.Int) {
...
+			} // TODO if the id is irrelevant to us, we could resend it on the CompletedObjective chan for others to read?
+			allCompleted = allCompleted && !incomplete[id]
+		}
+		return allCompleted
+	}
+
+	go func() {
+		for {
+			select {
+			case got := <-c.CompletedObjectives():
```
lalexgap, line 196: I think this means that once `WaitForObjectivesToComplete` is called it will consume all completed objectives from `c.CompletedObjectives()`, even if the objectives are unrelated to the `ids` we passed in. This might be confusing for clients, ie if they call `WaitForObjectivesToComplete` for a long running objective it will appear like no other objectives complete while they're waiting.

 There could also be contention on the channel between `WaitForObjectivesToComplete` and some client code pulling from `c.CompletedObjectives()`?

geoknee: Agreed. Did you come up with a solution to these problems in your completion manager code?

Perhaps `WaitForObjectivesToComplete` could register an "event handler" (i.e. create a new `chan`) scoped to a specific objective id. The engine would then send on each registered event handler when that objective completes, which would eliminate contention and prevent consuming unrelated events.

geoknee:
> It would be really nice if the client API had:
> 
> * A function to get ledger channels and their balances
> * A function to get running payment channels and their balances
> * A function to get the status of ledger/payment channel creation/closing

I agree, although we might still prefer to do this in an "event callback" way as well, to dissuade consumers from polling / spamming methods which hit the store.

geoknee:
Closing until we resolve  https://www.notion.so/statechannels/RFC-23-Overhaul-go-nitro-API-2c83cd05f59946ebb7354296de09f30e

PR #947, opened 2022-10-18
Publish docs website using netlify

### LINKED ISSUE #945: point docs.statechannels.org at deployed docs website
This is currently broken. I believe this is because the CNAME file (that is added to the `gh-pages` branch when you add the domain in settings) is being lost when the github action force-pushes a new version of the site to that branch. 

Another route to getting this work would be to go back to netlify:
https://docs.netlify.com/integrations/frameworks/#mkdocs
Then we would have deploy previews and other nice features.

geoknee:
Closes #945


kerzhner (approved):
lgtm!

PR #950, opened 2022-10-19
ignore netlify builds unless docs change

geoknee:
See https://docs.netlify.com/configure-builds/ignore-builds/

This has been manually tested, and seems to work.


NiloCK (approved):
:+1:

PR #951, opened 2022-10-19
EXPERIMENTAL: allow hubs to exit virtualfund early

geoknee:
I wanted to try this out to see affects on performance.

I'm 80% sure that it is safe for intermediaries to exit the virtualfund protocol without having a full set of post fund signatures. To be discussed / confirmed. 

If you click through to the dashboards, this change drastically reduces the time taken for the hub to complete virtual funding (10s down to 2s, so an 80% reduction). The effects on the other participants (which are probably more important) are less dramatic. About 20% reduction in time to first payment.

### DIFF
--- a/protocols/virtualfund/virtualfund.go
+++ b/protocols/virtualfund/virtualfund.go
@@ -458,7 +458,11 @@ func (o *Objective) Crank(secretKey *[]byte) (protocols.Objective, protocols.Sid
 		sideEffects.MessagesToSend = append(sideEffects.MessagesToSend, messages...)
 	}
 
-	if !updated.V.PostFundComplete() {
+	// Alice and Bob need a complete post fund round to know V is fully funded.
+	// EXPERIMENTAL: Intermediaries do not require the complete post fund, so we allow them to finish the protocol early.
+	// If they need to recover funds, they can force V to close by challenging with the pre fund state.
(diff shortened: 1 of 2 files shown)


geoknee:
> This could be overkill but we could spec this out in an RFC?

I think that's a good call. In any case we need to commit some really clear documentation about how to maintain safety during channel opening, and then why (if) this optimisation is safe. 

I'll close this for now and make an RFC.

NiloCK:
> https://www.notion.so/statechannels/RFC-17-Hubs-exit-virtualfund-early-2e5cd47aeeb2458b904dfff7463f9f1c

This looks pretty fleshed out but is still marked draft. I'm guessing from the reopen here that you've got your head around these optimizations?

geoknee:
> > https://www.notion.so/statechannels/RFC-17-Hubs-exit-virtualfund-early-2e5cd47aeeb2458b904dfff7463f9f1c
> 
> This looks pretty fleshed out but is still marked draft. I'm guessing from the reopen here that you've got your head around these optimizations?

I think I have my head around the optimizations in theory, and the RFC will soon be taken out of draft. I'm reopening here to try and see how easy the implementations are in practice (I thought of a couple more while working on the RFC).

Spoiler alert: those extra optimizations don't work out of the box.

geoknee:
Closing in favour of #1022

PR #972, opened 2022-10-21
make femv-feat action run fevm specific test

NiloCK:
Points to test `fevm-virtual-fund`, which does not yet exist


NiloCK:
This isn't a separate file, but a temp alteration for this branch.

lalexgap (approved):
Nice one!

PR #974, opened 2022-10-21
FEVM feature branch

lalexgap:
Add basic support for the FEVM


lalexgap:
Now that FEVM supports events we can close this in favour of #993

PR #977, opened 2022-10-21
Send Messages in go routine

lalexgap:
This changes the engine to use a go routine for sending a message. This means we don't block on sending the message and realize some minor performance gains.

I've also refactored the code slightly so we can record the duration of the send message function.

### DIFF
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -437,15 +437,24 @@ func (e *Engine) handlePaymentRequest(request PaymentRequest) error {
 	return e.executeSideEffects(se)
 }
 
-// executeSideEffects executes the SideEffects declared by cranking an Objective or handling a payment request.
-func (e *Engine) executeSideEffects(sideEffects protocols.SideEffects) error {
+// sendMessages sends out the messages and records the metrics.
+func (e *Engine) sendMessages(msgs []protocols.Message) {
 	defer e.metrics.RecordFunctionDuration()()
 
-	for _, message := range sideEffects.MessagesToSend {
+	for _, message := range msgs {
 		e.logMessage(message, Outgoing)
 		e.recordMessageMetrics(message)
 		e.msg.Send(message)
 	}
+
+}
+
+// executeSideEffects executes the SideEffects declared by cranking an Objective or handling a payment request.
+func (e *Engine) executeSideEffects(sideEffects protocols.SideEffects) error {
+	defer e.metrics.RecordFunctionDuration()()
+
+	e.sendMessages(sideEffects.MessagesToSend)
+
 	for _, tx := range sideEffects.TransactionsToSubmit {
 		e.logger.Printf("Sending chain transaction for channel %s", tx.ChannelId())
 		err := e.chain.SendTransaction(tx)
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -453,7 +453,8 @@ func (e *Engine) sendMessages(msgs []protocols.Message) {
 func (e *Engine) executeSideEffects(sideEffects protocols.SideEffects) error {
 	defer e.metrics.RecordFunctionDuration()()
 
-	e.sendMessages(sideEffects.MessagesToSend)
+	// Send messages in a go routine so that we don't block on message delivery
+	go e.sendMessages(sideEffects.MessagesToSend)
 
 	for _, tx := range sideEffects.TransactionsToSubmit {
 		e.logger.Printf("Sending chain transaction for channel %s", tx.ChannelId())


client/engine/engine.go
```diff
@@ -436,15 +436,25 @@ func (e *Engine) handlePaymentRequest(request PaymentRequest) error {
...
-func (e *Engine) executeSideEffects(sideEffects protocols.SideEffects) error {
+// sendMessages sends out the messages and records the metrics.
+func (e *Engine) sendMessages(msgs []protocols.Message) {
 	defer e.metrics.RecordFunctionDuration()()
 
-	for _, message := range sideEffects.MessagesToSend {
+	for _, message := range msgs {
 		e.logMessage(message, Outgoing)
 		e.recordMessageMetrics(message)
 		e.msg.Send(message)
```
NiloCK, line 446: This still looks like a blocking sequence to me. EG, where there are 3 messages (Alice wants to send to Irene, Ivan, and Bob), each send operation blocks the next.

Maybe:

``` golang
go func() {
  e.recordMessageMetrics(message)
  e.msg.Send(message)
}()
```
so that the logging / metrics calls are closely bundled with the beginning of the send?

NiloCK (approved):
Approve in spirit - but check the comment as well!

client/engine/engine.go
lalexgap: You're right that within the `sendMessages` function that we're blocking on `e.msg.Send`. However I think that's ok as the main change in this PR to call `sendMessages` in a go-routine means the engine run loop no longer blocks on sending messages.

It's possible that calling `msg.Send` in a go-routine for each message might be slightly more efficient but I didn't notice that in the TTFP when I tried it in [62335ce](https://github.com/statechannels/go-nitro/pull/977/commits/62335ce507dd6c32afffd0b584c560fb18d63e8d). I think that makes sense as it's really the engine run loop that is the "hot path". By calling `sendMessages` in a go routine, we move sending messages off the "hot path", so it doesn't matter too much how fast `sendMessages` completes.

NiloCK: The queue of messages is off the hot path for this particular client at this particular instant (ie, the client machine can get on immediately with processing new inputs), but the protocol at large (virtual fund, virtual defund) can be held up here.

I'm not exactly clear on the nature of the blocking inside of `msg.send`, but imagine where Alice wants to send a postFund signature to Irene, Ivan and Bob. The send to Bob won't happen until the sends to Irene and Ivan have completed, but this broadcast could be done in parallel.

Highlighted in @geoknee's draft PR, Bob is waiting for that post-fund message to finish the VF protocol.

(this would be expected to show in the numbers more in a multi-hop, since there are bigger queues of outgoing messages created there)

lalexgap: > (this would be expected to show in the numbers more in a multi-hop, since there are bigger queues of outgoing messages created there) 

I've tried running a test with 5 hops and 20 concurrency.  The results seem similar between sending each message [in a go-routine ](http://34.168.92.245:3000/d/5OBBeW37k/time-to-first-payment?orgId=1&from=now-15m&to=now&var-runId=cdlioionr2gv4eib3akg&var-jobCount=20&var-testDuration=2m0s&var-hubs=5&var-payees=2&var-jitter=1&var-latency=10&var-payers=2&var-payeepayers=0&var-nitroVersion=v0.0.0-20221108220129-62335ce507dd) vs no [go-routine](http://34.168.92.245:3000/d/5OBBeW37k/time-to-first-payment?orgId=1&from=now-15m&to=now&var-runId=cdlimionr2gv4eib3ak0&var-jobCount=20&var-testDuration=2m0s&var-hubs=5&var-payees=2&var-jitter=1&var-latency=10&var-payers=2&var-payeepayers=0&var-nitroVersion=v0.0.0-20221107233712-45d9c84b5868) although that's far from an exhaustive test.

That said, it didn't seem to harm performance so I don't mind implementing that change!

lalexgap: Hmm it seems that this change causes the tests to [fail](https://github.com/statechannels/go-nitro/actions/runs/3425382054/jobs/5706156123) for some reason 😞

geoknee: The testground run seemed to work just fine. Which made me think there could be an issue with the test messaging service receiving concurrent requests to send messages. Hence #985

geoknee:
Should this change be made inside the messaging service rather than the engine?

(Originally suggested by @kerzhner during standup)

lalexgap:
> Should this change be made inside the messaging service rather than the engine?
> (Originally suggested by @kerzhner during standup)

This got me thinking about the current lib-p2p message service, how it works and how that may impact performance. Some ramblings below 😛

Right now our lib-p2p message service is pretty simple. Every call to `Send` attempts to open a new stream with a peer, and send a message. It has some very basic retry logic, that waits a little bit and then attempts to recreate the stream.

This means that high load on the message service could result in multiple concurrent streams being opened with a peer. I think streams are multiplexed on the same TCP connection, but I'm curious about the performance impact of that. Is it possible that all the go-routines end up blocking on access to the TCP connection, meaning they end up being executed serially? Is there a "sweet spot" for the amount of concurrent streams we should have with a peer?

I'm also unsure of the cost tradeoff of creating a new stream vs attempting to reuse an existing stream. In my first attempt at the message service I tried to use long running streams with peers. However I found they would fail with a `connection reset` error after some time, so I just switched to creating a new stream for every message. Is reusing streams more performant? Is there a reasonable "lifetime" we could expect for a stream?

We could implement a more robust architecture which queues outgoing messages and tries to send them as efficiently as possible to maximize output (like @kerzhner mentioned in stand up). The send message function be non-blocking and just queue the message for sending.

Another possible performance win could be switching the transport to [quic](https://connectivity.libp2p.io/#quic) (which should be faster than TCP?). I tried this briefly but ran into an issue around the [UDP buffer size](https://github.com/lucas-clemente/quic-go/wiki/UDP-Receive-Buffer-Size) on the testground docker containers. It's probably worth the time to get it working and see how that impacts performance.

I think it might make sense to scope a little performance investigation out of this (and make use of the power of `testground`  🦸 ). We could:
- Add some testground metrics to the message service, to capture message service throughput.
- Throw together a dashboard on grafana to report on that metrics.
- Get a testground run working with the QUIC transport and check the results
- Investigate some of the performance questions raised above
- Implement any quick wins, record issues for any longer term work

If others agree I'd be happy to create an issue for this work.

client/engine/engine.go
lalexgap: Based on standup discussion we're planning on just merging in with the original [change](https://github.com/statechannels/go-nitro/pull/977/commits/45d9c84b586810c8407e8d33d603fc2b1ed1db27). We will investigate this change and others when we look at performance in https://github.com/statechannels/go-nitro/issues/986

lalexgap: Hmm it looks like even after this reverting back to the original change, we're still seeing [test failures](https://github.com/statechannels/go-nitro/actions/runs/3430904846/jobs/5718429968). This may be [related to the test message service](https://github.com/statechannels/go-nitro/pull/977#discussion_r1017819461) that @geoknee pointed out?

I am also curious if the test is just slow or if it's outright failing. Will investigate!

lalexgap: I believe this is the [root cause](https://github.com/statechannels/go-nitro/pull/987) that is probably getting exposed by more concurrent message sending

lalexgap:
⚠️ This is now based on https://github.com/statechannels/go-nitro/pull/987

PR #983, opened 2022-10-31
replace filecoinnonce method with ethNonce

geoknee:
This is a small step on the way to making the fevm chain service a pure eth chain service. Wallaby test net now supports nonce queries using the ethereum API.


geoknee:
I think the test is not passing due to a message stuck in the mempool.

geoknee:
Closing this for now since CI is failing and we may be able to soon revert to a pure ethers / ethClient API as Wallaby gets upgraded.

PR #987, opened 2022-11-10
Fix some ledger proposal plumbing

lalexgap:
Right now a protocol can return a list of proposals to process in `sideEffects` This is used by a protocol to [queue up processing](https://github.com/statechannels/go-nitro/blob/e2df3ae7f32ce3c4c992a4fd11203e7a5031219f/protocols/virtualdefund/virtualdefund.go#L493) of other proposals that may now be unblocked because this protocol is done with the ledger.

However it looks like this isn't working correctly and the engine discards any`sideEffects.ProposalsToProcess`.

There were two problems:

1. `SideEffects.Merge` didn't merge `ProposalsToProcess`.  So when we attempted to merge the results we ended up dropping any `ProposalsToProcess`. I've added a small test for this and updated the merge function to include `ProposalsToProcess`.
2. We didn't instantiate a chan for `fromLedger` in the engine, which meant it was a nil chan. If we tried writing to it we'd end up blocking [forever](https://golangbyexample.com/send-receive-nil-channel-go/)

From running the integration tests it looks like they're is rarely a proposal queue which explains why this didn't cause test failures. (And even if a proposal is dropped from the queue its possible we'll receive a message from another peer that triggers a crank anyways)

I am slightly surprised this didn't show up testground runs more. Do we need to use a higher `jitter` value for more thorough message re-ordering?

### DIFF
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -88,6 +88,7 @@ func New(msg messageservice.MessageService, chain chainservice.ChainService, sto
 
 	e.store = store
 
+	e.fromLedger = make(chan consensus_channel.Proposal, 100)
 	// bind to inbound chans
 	e.ObjectiveRequestsFromAPI = make(chan protocols.ObjectiveRequest)
 	e.PaymentRequestsFromAPI = make(chan PaymentRequest)
--- a/protocols/messages.go
+++ b/protocols/messages.go
@@ -82,6 +82,7 @@ func (se *SideEffects) Merge(other SideEffects) {
 
 	se.MessagesToSend = append(se.MessagesToSend, other.MessagesToSend...)
 	se.TransactionsToSubmit = append(se.TransactionsToSubmit, other.TransactionsToSubmit...)
+	se.ProposalsToProcess = append(se.ProposalsToProcess, other.ProposalsToProcess...)
 
 }
 
--- a/protocols/messages_test.go
+++ b/protocols/messages_test.go
@@ -40,6 +40,30 @@ func toPayload(p interface{}) []byte {
 	return b
 }
 
+func TestSideEffectsMerge(t *testing.T) {
+	original := &SideEffects{
+		MessagesToSend:       []Message{{To: types.Address{'a'}}},
+		TransactionsToSubmit: []ChainTransaction{},
+		ProposalsToProcess:   []consensus_channel.Proposal{removeProposal().Proposal},
+	}
+	toMerge := SideEffects{
+		MessagesToSend:       []Message{{To: types.Address{'b'}}},
+		TransactionsToSubmit: []ChainTransaction{},
+		ProposalsToProcess:   []consensus_channel.Proposal{addProposal().Proposal},
+	}
+
+	original.Merge(toMerge)
+
+	expected := &SideEffects{
+		MessagesToSend:       []Message{{To: types.Address{'a'}}, {To: types.Address{'b'}}},
+		TransactionsToSubmit: []ChainTransaction{},
+		ProposalsToProcess:   []consensus_channel.Proposal{removeProposal().Proposal, addProposal().Proposal},
+	}
+	if !reflect.DeepEqual(original, expected) {
+		t.Errorf("incorrect merge: got:\n%v\nwanted:\n%v", original, expected)
+	}
+}
+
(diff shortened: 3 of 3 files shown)


lalexgap:
We should probably write some kind of test that would catch this scenario.  Maybe just an integration test with some particular message re-arranging to generate the problem? I think we can tackle that outside the scope of this PR.

client/engine/engine.go
```diff
@@ -88,6 +88,7 @@ func New(msg messageservice.MessageService, chain chainservice.ChainService, sto
 
 	e.store = store
 
+	e.fromLedger = make(chan consensus_channel.Proposal, 100)
```
geoknee, line 91: I'm kind of amazed that anything worked without this line -- I guess the bug is combining with the other problem so we didn't see a `panic` (which would be what I would have expected). Is that your understanding too?

protocols/messages.go
```diff
@@ -82,6 +82,7 @@ func (se *SideEffects) Merge(other SideEffects) {
 
 	se.MessagesToSend = append(se.MessagesToSend, other.MessagesToSend...)
 	se.TransactionsToSubmit = append(se.TransactionsToSubmit, other.TransactionsToSubmit...)
+	se.ProposalsToProcess = append(se.ProposalsToProcess, other.ProposalsToProcess...)
```
geoknee, line 85: 😲  This is a nasty one. Can you think of any better way of protecting against this other than good unit testing? Even that seems a bit brittle, say if we added a side effect?

geoknee (approved):
Nice catch! I think I'm still a bit confused as to why things worked at all given these bugs :-/

client/engine/engine.go
geoknee: Just re-read your PR description. It's not a panic but an infinite wait. Still bad.

geoknee:
> I am slightly surprised this didn't show up testground runs more. Do we need to use a higher `jitter` value for more thorough message re-ordering?

Slightly nuts idea: How about a we write a `TerribleMessagingService`(name subject to review) that we could use in certain tests (only). Its job would be to reorder messages. It could queue up incoming messages and only release them into the engine every second or so (or something much longer than network latency). It can then apply a random permutation to the message order while the messages are being held in the queue. This might be a much more reliable way of generating message reorder scenarios (than just relying on jitter).

client/engine/engine.go
lalexgap: > Just re-read your PR description. It's not a panic but an infinite wait. Still bad.

In general I think nil channels can be a huge footgun; it can be really easy to miss the infinite wait if that wait is happening in some go-routine we're not paying attention to.

geoknee: Yikes. We need a linting rule for this or something

lalexgap:
> Nice catch! I think I'm still a bit confused as to why things worked at all given these bugs :-/

The ledger proposal queue really boils down to a list of objectives that we want to recrank, as they may be unblocked now. So even if the queue is dropped, as long a messages comes in later for that objective from a peer, it'll trigger the recrank and things will be ok.

It also seems that we rarely have a proposal queue when running the integration tests, so it's rare a protocol tries to return any `ProposalsToProcess`.  I suspect that's because messages are sent and arrive roughly "in order" during our integration test.

protocols/messages.go
lalexgap: That's a good question. I think we also run into a similar problem with keeping serde logic up to date. Maybe we could write a test that uses reflection to automagically to check all properties get merged correctly? Not sure how feasible that is is in golang.

NiloCK:
> So even if the queue is dropped, as long a messages comes in later for that objective from a peer, it'll trigger the recrank and things will be ok.
 
This is a really astute observation. I was pretty shocked by this PR, and couldn't figure out at all how or why anything was working at all.

Related: @lalexgap and I had talked about some intermittent test failures in one of the last days at Lisbon. I believed at the time that the root cause may have been related to account generation and accidental trimming of leading zeros from some generated addresses or similar. I looked into that, and found it wasn't the problem.

Now I expect those intermittent failures (which failed with "objective X failed to complete for client Y in 10s") were likely rooted in this issue.

NiloCK (approved):

PR #988, opened 2022-11-14
remove redundant underflow check

geoknee:
This happens automatically since solidity `0.8`.


NiloCK:
For posterity:

https://docs.soliditylang.org/en/latest/080-breaking-changes.html

First bullet point:
> Arithmetic operations revert on underflow and overflow. You can use `unchecked { ... }` to use the previous wrapping behaviour.

>Checks for overflow are very common, so we made them the default to increase readability of code, even if it comes at a slight increase of gas costs.

nitro-protocol/test/contracts/VirtualPaymentApp/requireStateSupported.test.ts
```diff
@@ -145,7 +145,7 @@ describe('requireStateSupported (candidate plus single proof state route)', () =
     {...vVR, bobAdjustedCorrectly: false, reason: 'Bob not adjusted correctly'},
     {...vVR, nativeAsset: false, reason: 'only native asset allowed'},
     {...vVR, multipleAssets: true, reason: 'only native asset allowed'},
-    {...vVR, aliceUnderflow: true, reason: 'underflow'},
+    {...vVR, aliceUnderflow: true, reason: ' '}, // we expect transaction to revert without a reason string
```
NiloCK, line 148: Single space here is required in order to trigger the `expectRevert` clause, but doesn't `expectRevert` check the returned error against the given string? Shouldn't it fail here to read the error `" "`?

NiloCK (approved):
Underflow check removal looks good. Slightly confused about the test modification though!

nitro-protocol/test/contracts/VirtualPaymentApp/requireStateSupported.test.ts
geoknee: Fair question. The supplied string is used to pattern match (a regex, I think) the error message, which typically comes back as some header plus the string we choose to encode in solidity. Since that header has spaces in it, the pattern matching passes.

NiloCK: Somewhat hacky on both ends, but not too tricky to rediscover the logic, and probably not worth breaking out to a different style of check.

:+1:

PR #989, opened 2022-11-15
Experiment with `solidity-docgen`

geoknee:
WIP

https://deploy-preview-989--nitrodocs.netlify.app/on-chain-api/consensus/


geoknee:
This experiment showed that we can integrate this tool into our docs website pipeline fairly easily. It would need some more work to make it look good and organise the information in a logical way. 

Overall I am not sure of the overall utility of the generated documentation. What i really want is a tool that will give me the entire API of the NitroAdjudicator (i.e. the most derived, inheriting contract). As it is, we only get the bits defined in that contract, and we have to drill down into the ancestor contracts to learn about the rest of the API.

PR #996, opened 2022-11-19
For discussion purpose only

mod:
Copied the VirtualPaymentApp and annotated basic business logic

**CurrencySwapApp**

Alice and Bob are Maker and Taker, Maker has funded assets and the Taker would like to swap using a Voucher

- [x] @geoknee Please confirm my assumption regarding Signature of Voucher
- [x] Confirm the Adjustment logic incrementing and decrementing respective balance
- [ ] Fee must be collected according to a schedule of 100 bps, 80, 60, 40, 20
- [ ] Define the outcome in the form of a multi-currency balance sheet balance[address] => uint256

### DIFF
--- a/nitro-protocol/contracts/CurrencySwapApp.sol
+++ b/nitro-protocol/contracts/CurrencySwapApp.sol
@@ -8,19 +8,22 @@ import './interfaces/INitroTypes.sol';
 import {ExitFormat as Outcome} from '@statechannels/exit-format/contracts/ExitFormat.sol';
 
 /**
- * @dev The CurrencySwapApp contract complies with the ForceMoveApp interface and allows Alice and Bob (participants[0] to participants[n+1], where n is the number of intermediaries) to trade two assets atomically. 
+ * @dev The CurrencySwapApp contract complies with the ForceMoveApp interface and allows Alice and Bob (participants[0] to participants[n+1], where n is the number of intermediaries) to trade two assets atomically.
  */
 contract CurrencySwapApp is IForceMoveApp {
+    struct Order {
+        address asset;
+        uint256 amount;
+    }
 
-    struct VoucherSwap {
-        address takerAsset; // TODO: support multi-chain with struct {chainId, Address}
-        uint256 takerAmount;
-        address makerAsset;
-        uint256 makerAmount;
-        INitroTypes.Signature signature; // signature on abi.encode(channelId, amounts)
+    struct Voucher {
+        Order takerOrder;
+        Order makerOrder;
         // TODO: I assume both side need to confirm the amount
-        // INitroTypes.Signature takerSignature;
-        // INitroTypes.Signature makerSignature;
+        INitroTypes.Signature takerSignature;
+        INitroTypes.Signature makerSignature;
+        // TODO: to differentiate between Vouchers
+        uint256 number;
     }
 
     //TODO: Remove once logic is tested
@@ -31,8 +34,8 @@ contract CurrencySwapApp is IForceMoveApp {
 
     // TODO: To confirm indices are always 0 (Taker) and 1 (Maker)
     enum AllocationIndices {
-        Taker,  // Aggressor order
-        Maker   // Passive order was found in broker orderbook
+        Taker, // Aggressor order
+        Maker // Passive order was found in broker orderbook
     }
 
     /**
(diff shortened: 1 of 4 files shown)


nitro-protocol/package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "@statechannels/nitro-protocol",
-  "version": "2.0.0-alpha.0",
+  "version": "2.0.0-alpha.1",
```
mod, line 3: I don't know how this happened, I will probably remove the file from PR

mod:
image

nitro-protocol/contracts/CurrencySwapApp.sol
```diff
@@ -0,0 +1,156 @@
...
+    //TODO: Remove once logic is tested
+    struct VoucherAmountAndSignature {
+        uint256 amount;
+        INitroTypes.Signature signature; // signature on abi.encode(channelId,amount)
+    }
+
+    // TODO: To confirm indices are always 0 (Taker) and 1 (Maker)
+    enum AllocationIndices {
+        Taker,  // Aggressor order
+        Maker   // Passive order was found in broker orderbook
```
geoknee, line 35: You are the expert on this, but I would have thought the Maker proposes a trade, and the Taker fulfils it? What you have seems the other way around. 

In any case I think it is fine, we just have to be a) consistent and b) track down any assumptions of unidirectionality and remove them (see above comment about replacing vouchers).

nitro-protocol/contracts/CurrencySwapApp.sol
```diff
@@ -0,0 +1,156 @@
...
+                oldOutcome[0].asset == address(0) &&
+                newOutcome[0].asset == address(0),
+            'only native asset allowed'
+        );
+
+        // TODO: Adjustments as follows
+        // Outcome[0].allocations[Taker].amount - voucherTakerAmount
+        // Outcome[1].allocations[Taker].amount + voucherMakerAmount
+        // Outcome[0].allocations[Maker].amount - voucherMakerAmount
+        // Outcome[1].allocations[Maker].amount + voucherTakerAmount
```
geoknee, line 145: Looks good to me. I wonder if there the terminology may get confusing? _The (maker) taker amount and (maker) taker asset belong to the (maker) taker at the beginning when no swap has yet happened._ <- do you agree? If so, perhaps this terminology is OK.

nitro-protocol/test/contracts/CurrencySwapApp/requireStateSupported.test.ts
```diff
@@ -0,0 +1,239 @@
+import {expectRevert} from '@statechannels/devtools';
```
geoknee, line 1: Did you make any changes here, yet?

geoknee:
> * Define the outcome in the form of a multi-currency balance sheet balance[address] => uint256

The best we have so far is [the exit format](https://github.com/statechannels/exit-format). I don't think mappings are supported in solidity calldata, so for now we must use arrays (I believe)/

geoknee:
> image

In order to fund a virtual channel between Alice and Bob:

Alice needs to lock up 20 WETH in her ledger channel with BrokerI.

Bob needs to lock up 2 WBTC in her ledger channel with BrokerJ.

BrokerI needs to lock up 20 WETH in his channel with BrokerJ.
BrokerI needs to lock up 2 WBTC in his channel with Alice.

BrokerJ needs to lock up 20 WETH in his channel with Bob.
BrokerJ needs to lock up 2 WBTC in his channel with BrokerI.

nitro-protocol/contracts/CurrencySwapApp.sol
```diff
@@ -0,0 +1,156 @@
...
+import './libraries/NitroUtils.sol';
+import './interfaces/INitroTypes.sol';
+import {ExitFormat as Outcome} from '@statechannels/exit-format/contracts/ExitFormat.sol';
+
+/**
+ * @dev The CurrencySwapApp contract complies with the ForceMoveApp interface and allows payments to be made virtually from Alice to Bob (participants[0] to participants[n+1], where n is the number of intermediaries).
+ */
+contract CurrencySwapApp is IForceMoveApp {
+
+    struct VoucherSwap {
```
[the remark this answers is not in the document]
geoknee: > * b. Maintain the invariant (off chain) that only one voucher is ever doubly signed

The problem with this is that we can only do one swap per channel, which undermines the performance benefits of virtual channels with respect to an HTLC approach (we would need to involve the intermediaries in every swap).

nitro-protocol/contracts/CurrencySwapApp.sol
```diff
@@ -0,0 +1,156 @@
...
+        address takerAsset; // TODO: support multi-chain with struct {chainId, Address}
```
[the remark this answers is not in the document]
mod: Asset could be defined as {chainId, Address}

nitro-protocol/contracts/CurrencySwapApp.sol
mod: To be exact, Maker has an offer online, ready to be taken, but the initiator is what we call the Aggressing order, which is the one that will trigger the trade (the price is matching)

nitro-protocol/contracts/CurrencySwapApp.sol
```diff
@@ -0,0 +1,156 @@
...
+        pure
+        returns (uint256)
+    {
+        VoucherAmountAndSignature memory voucher = abi.decode(appData, (VoucherAmountAndSignature));
+
+        address signer = NitroUtils.recoverSigner(
+            keccak256(abi.encode(NitroUtils.getChannelId(fixedPart), voucher.amount)),
+            voucher.signature
+        );
+        //TODO: Verify amount, signature of Maker
```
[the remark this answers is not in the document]
mod: Noted and clear

nitro-protocol/contracts/CurrencySwapApp.sol
mod: I will improve the terminology maybe through a helper function swap, tested once and forget

nitro-protocol/contracts/CurrencySwapApp.sol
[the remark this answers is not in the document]
mod: 1. Noted
2. Aren't your payment vouchers single use ?
In our case amount can be partial, like matching half of the resting order, and more can come until fulfilment.

I don't fully understand the problem, but all Swaps are unique and must be enforced, we are likely going to add a timestamp, or incremental number to make sure participants applies all Swaps.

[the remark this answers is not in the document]
geoknee: I think an incremental number is fine, yes. 

To clarify the problem: with a unidirectional channel where the Bob gets more money monotonically: 

```
Voucher 0: (Give $1 to Bob, signed by Alice)
Voucher 1: (Give $2 to Bob, signed by Alice)
```
Voucher `1` is strictly better for Bob than Voucher `0`. Generally Voucher `i` will be better for Alice than Voucher `i-1`. So Alice and Bob can always know which is the "latest voucher" by understanding Bob's incentives. He will always try and use the "latest" (best for him) voucher. For this reason we don't even give the vouchers a number at all.

In a bidirectional channel (of which a currency swap is an example):
```
Voucher 0: (Bob => $1 => Alice && Bob <= £2 <= Alice, signed by Alice and Bob)
Voucher 1: (Bob => $2 => Alice && Bob <= £4 <= Alice, signed by Alice and Bob)
```

Voucher 1 is not obviously "better' for either party than Voucher 1*. Voucher 2 might represent "unswapping" the assets -- kind of natural rebalancing. And we still want to know that Voucher 2 supercedes Voucher 1. So we include the number, since we couldn't infer this just from inspecting the balances / voucher amounts. 

*Although perhaps with the introduction of fees, it would be.

nitro-protocol/contracts/CurrencySwapApp.sol
```diff
@@ -0,0 +1,175 @@
...
+import {ExitFormat as Outcome} from '@statechannels/exit-format/contracts/ExitFormat.sol';
+
+/**
+ * @dev The CurrencySwapApp contract complies with the ForceMoveApp interface and allows Alice and Bob (participants[0] to participants[n+1], where n is the number of intermediaries) to trade two assets atomically.
+ */
+contract CurrencySwapApp is IForceMoveApp {
+    struct Order {
+        address asset;
+        uint256 amount;
+    }
```
mod, line 17: No need orders, protocol would only understand MarginVoucher

nitro-protocol/contracts/CurrencySwapApp.sol
```diff
@@ -0,0 +1,175 @@
...
+
+    struct Voucher {
+        Order takerOrder;
+        Order makerOrder;
+        // TODO: I assume both side need to confirm the amount
+        INitroTypes.Signature takerSignature;
+        INitroTypes.Signature makerSignature;
+        // TODO: to differentiate between Vouchers
```
mod, line 25: We need a new convention, but we can try to use the Initiator / Receiver
We are now in single currency, but bi-directional,

which mean two Signatures.

nitro-protocol/contracts/CurrencySwapApp.sol
```diff
@@ -0,0 +1,175 @@
...
+    //TODO: Remove once logic is tested
+    struct VoucherAmountAndSignature {
+        uint256 amount;
+        INitroTypes.Signature signature; // signature on abi.encode(channelId,amount)
+    }
+
+    // TODO: To confirm indices are always 0 (Taker) and 1 (Maker)
+    enum AllocationIndices {
+        Taker, // Aggressor order
+        Maker // Passive order was found in broker orderbook
```
mod, line 38: Initiator,
Receiver

nitro-protocol/contracts/CurrencySwapApp.sol
```diff
@@ -0,0 +1,175 @@
...
+        // TODO: require Outcome.length == 2;
+        // && Outcome[0].asset == TakerAsset; Outcome[1].asset == MakerAsset
+        require(
+            oldOutcome.length == 1 &&
+                newOutcome.length == 1 &&
+                oldOutcome[0].asset == address(0) &&
+                newOutcome[0].asset == address(0),
+            'only native asset allowed'
+        );
+
```
mod, line 159: @nksazonov We can apply the respective amount of margin to the allocation amount
for simplicity, I suggest we express margin in amount - margin
the purpose is to cumulate margin calls,

amount - 10, -20, -30, -10.

mod (commented):
We can close this PR

nksazonov:
@mod , I will keep conversations unresolved for easier access later.

PR #999, opened 2022-11-23
Make interaction methods virtual

### LINKED ISSUE #994: Allow overriding `NitroAdjudicator` functions
## Problem

At Yellow we want to make our own `YellowAdjudicator`, which adds some logic to main adjudication methods (set not fully defined yet). However, that does not seem possible right now, as these methods are declared `external`, which means we can't invoke them in the inherited contract.

## Solution 1

The simplest and the most flexible solution would be to declare those methods `virtual` so that we can override them in our own Adjudicator.

## Solution 2

Another solution is to add hooks (`beforeDeposit`, `afterDeposit`, etc.) to each method we would like to add some behaviour to. Nevertheless, this looks not as flexible as the other solution, because it requires `NitroAdjudicator` to define those hooks and inheriting adjudicator to override them anyway. The only strong (and dubious) point of this solution seems to be the insurance the base `NitroAdjudicator` method is called and not forgotten.

geoknee:
Allowing inheriting contracts to override it.

Fixes #994

As an extension, we may want to make other functions in the API virtual also.


nksazonov:
Introduce more changes according to my [comment](https://github.com/statechannels/go-nitro/issues/994#issuecomment-1329259975).

nksazonov (approved):

PR #1000, opened 2022-11-23
add `RecoveredVariablePart` to exports

### LINKED ISSUE #997: `RecoveredVariablePart` missing from exports
When using package from npm,
I having found yet the repo for devtools, but I think it would be nice to have test-helpers in there to help App developers.

geoknee:
Closes #997


nksazonov (approved):

nksazonov (commented):
We would also need to export some test-helpers for Nitro App devs (edit: to close the #997), but I am not sure which ones.

Let me try to do this and we review the exported list afterwards.

Issue #1001, opened 2022-11-28
Enhance testing of directly funded channels to cover ERC20 tokens


geoknee:
We suspect there a bug in this flow somewhere and we are not exercising the code enough to see it currently.

geoknee:
In my first attempt to fix this, I made a test which uses the native token for ledger channels and then goes again with an ERC20 token. But this caused the error `a channel already exists with counterparty`. I think this means we will need to go deeper to get ourselves fully multi-asset compliant (i.e. we need to think about **routing** in multiple assets). 

In the short term, I will probably just aim for supporting one token or the other (exclusive).

PR #1005, opened 2022-11-30
Extend directfund test to cover ERC20 tokens

geoknee:
Changes:

* simulated backend distributed tokens to all accounts when it is setup
* the directfund test is repeated for native and then erc20 tokens. 
* we do not yet currently support mixing of assets, this test just shows things work if we choose an asset and stick with it.

The git history could be cleaned up, there are a couple of meanders. 

Towards #1001


geoknee:
In favour of #1011

PR #1011, opened 2022-12-02
Extend directfund test to cover ERC20 tokens

geoknee:
Changes:

* simulated backend distributed tokens to all accounts when it is setup
* the directfund test is repeated for native and then erc20 tokens.
* we do not yet currently support mixing of assets, this test just shows things work if we choose an asset and stick with it.

Towards https://github.com/statechannels/go-nitro/issues/1001

### DIFF
--- a/client/engine/chainservice/simulated_backend_chainservice.go
+++ b/client/engine/chainservice/simulated_backend_chainservice.go
@@ -25,7 +25,7 @@ type binding[T any] struct {
 	Contract *T
 }
 
-type bindings struct {
+type Bindings struct {
 	Adjudicator       binding[NitroAdjudicator.NitroAdjudicator]
 	Token             binding[Token.Token]
 	ConsensusApp      binding[ConsensusApp.ConsensusApp]
@@ -45,7 +45,7 @@ type SimulatedBackendChainService struct {
 
 // NewSimulatedBackendChainService constructs a chain service that submits transactions to a NitroAdjudicator
 // and listens to events from an eventSource
-func NewSimulatedBackendChainService(sim simulatedChain, bindings bindings,
+func NewSimulatedBackendChainService(sim simulatedChain, bindings Bindings,
 	txSigner *bind.TransactOpts, logDestination io.Writer) (ChainService, error) {
 	ethChainService, err := NewEthChainService(sim,
 		bindings.Adjudicator.Contract,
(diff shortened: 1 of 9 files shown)


client/engine/chainservice/simulated_backend_chainservice.go
```diff
@@ -25,7 +25,7 @@ type binding[T any] struct {
 	Contract *T
 }
 
-type bindings struct {
+type Bindings struct {
```
geoknee, line 28: This is a type accepted by `NewSimulatedBackendChainService` so it should probably be public? Keen for feedback here.

geoknee:
Hmm got a failing test here. If it passes when re-run, we should open an issue to cover the flicker.

client/engine/chainservice/simulated_backend_chainservice.go
NiloCK: There's a bit of a circle going here - the bindings are produced as output from `SetupSimulatedBackend`. Having it private means that external packages can't declare literals of this type, but it's not clear why they'd want or need to manually construct these bindings.

The member properties are all public and accessible from outside packages, although never accessed anywhere.

I don't think this change helps or hurts in any large amount.

NiloCK: Current usage all looks like `SetupSimulatedBackend()` being called, and its outputs being fed immediately into `NewSimulatedBackendChainService()`. The functions might be better as one function, which returns a chainservice for all accounts as well?

client_test/directfund_test.go
```diff
@@ -75,73 +81,85 @@ func TestWhenObjectiveIsRejected(t *testing.T) {
...
+
+				if !con.OnChainFunding.IsNonZero() {
+					t.Fatal("Expected nonzero on chain funding, but got zero")
+				}
+
+				if _, channelStillInStore := store.GetChannelById(con.Id); channelStillInStore {
+					t.Fatalf("Expected channel to have been destroyed in %v's store, but it was not", store.GetAddress())
+				}
+			}
 		}
```
NiloCK, line 142: Defining this in-line here feels a little off to me - if it's meant for re-use, should it be a package function? If not, are there benefits to the encapsulation as a function?

NiloCK (approved):
:+1:

client_test/directfund_test.go
geoknee: Good point. I think this is an artifact of a blind-alley I went up trying to extend the test. I think I will just flatten it back out for now, since it is a little too specific for re-use.

geoknee: [dda2478](https://github.com/statechannels/go-nitro/pull/1011/commits/dda2478eeeabec5e88ab48b250275902b52838fb)

PR #1012, opened 2022-12-03
Draft code structure

mod:
will add details


nitro-protocol/contracts/VirtualMarginApp.sol
```diff
@@ -0,0 +1,146 @@
...
+
+        require(
+            newOutcome[0].allocations[uint256(AllocationIndices.Leader)].amount == receipt.leaderAmount,
+            'Leader not adjusted correctly'
+        );
+        require(
+            newOutcome[0].allocations[uint256(AllocationIndices.Follower)].amount == receipt.followerAmount,
+            'Follower not adjusted correctly'
+        );
+        // TODO: Find a way to validate version number
```
nksazonov, line 144: There is no way to store margin version number in VirtualMarginApp as `requireStateSupported` must be `pure`.

nitro-protocol/contracts/VirtualMarginApp.sol
```diff
@@ -0,0 +1,146 @@
...
+        //      |
+        //      (1)_AIB   [fully signed postfund]
+
+        if (proof.length == 1) {
+            requireProofOfUnanimousConsensusOnPostFund(proof[0], fixedPart.participants.length);
+            require(candidate.variablePart.turnNum == 2, 'bad candidate turnNum; |proof|=1');
+            require(
+                NitroUtils.isClaimedSignedBy(candidate.signedBy, 2),
+                'redemption not signed by Bob'
+            );
```
nksazonov, line 77: This can be omitted. We need to check signatures on either `VariablePart` or `MarginState`.

nitro-protocol/contracts/VirtualMarginApp.sol
```diff
@@ -0,0 +1,149 @@
...
+        // newOutcome[1].asset == ASSET_COLLATERAL_ADDRESS,
+
+        // equal sums
+        uint256 oldAllocationSum;
+        uint256 newAllocationSum;
+        for (uint256 i = 0; i < nParticipants; i++) {
+            oldAllocationSum += oldOutcome[0].allocations[i].amount;
+            newAllocationSum += newOutcome[0].allocations[i].amount;
+        }
+        require(oldAllocationSum == newAllocationSum, 'total allocated cannot change');
```
chumaknadya, line 147: @nksazonov don't you think this is true for all channels and NitroAdjudicator should be checking this?

nksazonov: No, `NitroAdjudicator` does not check this.
They mention that participants must be careful about outcomes: https://docs.statechannels.org/protocol-tutorial/0060-funding-a-channel/#outcome-priority

nitro-protocol/contracts/VirtualMarginApp.sol
```diff
@@ -0,0 +1,178 @@
...
+     * @param fixedPart Fixed part of the state channel.
+     * @param proof Array of recovered variable parts which constitutes a support proof for the candidate.
+     * @param candidate Recovered variable part the proof was supplied for.
+     */
+    function requireStateSupported(
+        FixedPart calldata fixedPart,
+        RecoveredVariablePart[] calldata proof,
+        RecoveredVariablePart calldata candidate
+    ) external pure override {
+        // This channel has only 4 states which can be supported:
```
kerzhner, line 45: @mod This is super interesting. Thank you for putting together the PR. Do you see this contract evolving to be (roughly) similar to the outline [here](https://www.notion.so/statechannels/RFC-27-Nitro-Clearinghouse-App-ce88e4e127114d1abe5d7e6b19b502fe#a9c048e42af143719e10fdbfd7765a4f)? Or is there a different vision for how a Margin Virtual Channel app will function?

geoknee: This seems to be a bit of out date (base on call 12th Jan).

PR #1015, opened 2022-12-07
Handle proposals in sorted order

### LINKED ISSUE #885: The engine does not guarantee to process ledger proposals in a sorted order
We're not using a sorted list of proposals when [handling a message](https://github.com/statechannels/go-nitro/issues/883) containing proposals.

I think this means that it's possible to handle a proposal out of order, although I haven't seen a crash or any issue attributable to this. Most likely the proposals are mostly sorted anyways.

lalexgap:
Fixes #885.

This updates `handleMessage` to consume the proposals in a sorted order (by channelId, and turn num) instead of whatever order they happen to be delivered in. I didn't really see a difference in performance with this chage.

### DIFF
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -252,7 +252,7 @@ func (e *Engine) handleMessage(message protocols.Message) (EngineEvent, error) {
 
 	}
 
-	for _, entry := range message.LedgerProposals {
+	for _, entry := range message.SortedProposals() {
 		id := getProposalObjectiveId(entry.Proposal)
 		objective, err := e.store.GetObjectiveById(id)
 		if err != nil {


geoknee:
No difference in performance... but are there other benefits? Perhaps a reduction in test flicker?

NiloCK:
Re: #885, I think it's probably reasonable to sort by turn number. Sorting by channelID could lead to some busy channels with IDs like `A`, `B`, `C` preventing proposals from channel with IDs `X`, `Y`, `Z` from ever being processed.

The same problem could exist with a naive sort-by-turn-number: several busy "young" channels with low turn numbers could prevent proposals from longer running channels with high turn numbers.

It's not *immediately* obvious to me that a sort function which swaps based on turn number only when channel IDs are the same doesn't have any weird side effects or corner cases to worry about.

lalexgap:
> No difference in performance... but are there other benefits? Perhaps a reduction in test flicker?
> 
> What do we expect to change?

I've taken a little time to dig into this and I think it may not be strictly necessary. When we send a message with multiple proposals we're sending the contents of the [ProposalQueue](https://github.com/statechannels/go-nitro/blob/cd34325f023e0cffac68bf880f07ad8574cde242/protocols/virtualfund/virtualfund.go#L658). The proposal queue will always be ordered by turn num (as the consensus channel generates a turn number when something is added to the proposal queue.) This also means that all the proposals in a message will be for the same channel.

It may be still be a good defensive practice to sort the proposals as I don't think we technically specify messages be in any specific sorted order.

geoknee:
Thanks @lalexgap! I've expanded on your findings a bit here https://github.com/statechannels/go-nitro/issues/885#issuecomment-1375540505

NiloCK:
> Re: #885, I think it's probably reasonable to sort by turn number. Sorting by channelID could lead to some busy channels with IDs like `A`, `B`, `C` preventing proposals from channel with IDs `X`, `Y`, `Z` from ever being processed.
> 
> The same problem could exist with a naive sort-by-turn-number: several busy "young" channels with low turn numbers could prevent proposals from longer running channels with high turn numbers.
> 
> It's not _immediately_ obvious to me that a sort function which swaps based on turn number only when channel IDs are the same doesn't have any weird side effects or corner cases to worry about.
> 
> (Similar sorts of concerns here to the reasoning behind golang's randomized `select` branching).

After further reflection, I this is a non issue. The change in this PR sorts only a "single burst" of bundled proposals, rather than something like an incoming stream. Alphanumerically-challenged channels can't get stuck because of it.

Whoops!

PR #1018, opened 2022-12-12
Allow Adjudicator to accepts "bad" ERC20 tokens (such as USDT)

### LINKED ISSUE #1013: Make `deposit` work with Bad ERC20 tokens
Some ERC20s do not return anything from `transfer` and `transferFrom`, however we expect them to (`MultiAssetHolder.deposit`):

``` solidity
require(
    IERC20(asset).transferFrom(msg.sender, address(this), amountDeposited),
    'Could not deposit ERC20s'
);
```

So if `transferFrom` does not return, `deposit` reverts. There are a lot of medium and small tokens affected and surprisingly, even USDT and BNB are.

More description in the article: https://medium.com/coinmonks/missing-return-value-bug-at-least-130-tokens-affected-d67bf08521ca

geoknee:
~Git history to be tidied up~

Closes #1013 

Changes
---
- Adds an example bad token and test case to our suite. Fails with existing `MultiAssetHolder.sol`.
- Changes the behaviour of  `MultiAssetHolder.sol` to use ~a low-level call whose return value is not inspected.~ the OpenZeppelin SafeERC20 library 
- ~Adds sanity checks to ensure the balance of the adjudicator was increased in the token contract (NOTE I am not really sure this is necessary)~


nitro-protocol/contracts/MultiAssetHolder.sol
```diff
@@ -57,10 +57,18 @@ contract MultiAssetHolder is IMultiAssetHolder, StatusManager {
...
+                abi.encodeWithSignature(
+                    'transferFrom(address,address,uint256)',
+                    msg.sender,
+                    address(this),
+                    amountDeposited
+                )
             );
+            require(success, 'Could not deposit ERC20s');
+            uint256 aft = IERC20(asset).balanceOf(address(this));
+            require(aft == before + amountDeposited, 'Could not deposit ERC20s');
```
geoknee, line 71: "Bad" ERC20 tokens do not return anything on successful deposit. We could simply rely on them reverting if the depositor has insufficient tokens (instead of using `balanceOf` to do a check, before and after the deposit)?

nksazonov (approved):

geoknee:
https://ethereum.stackexchange.com/questions/137049/why-does-safeerc20-assume-that-the-tokens-transfer-and-transferfrom-return-valu

geoknee:
Looks like openzeppelin provide a library for handling bad tokens that we should lean on, here. We don't want to break the existing check on the return value for "proper" erc20s, which currently this PR unfortunately does.

nitro-protocol/contracts/MultiAssetHolder.sol
```diff
@@ -56,11 +59,7 @@ contract MultiAssetHolder is IMultiAssetHolder, StatusManager {
         if (asset == address(0)) {
             require(msg.value == amount, 'Incorrect msg.value for deposit');
         } else {
-            // require successful deposit before updating holdings (protect against reentrancy)
-            require(
-                IERC20(asset).transferFrom(msg.sender, address(this), amountDeposited),
-                'Could not deposit ERC20s'
-            );
+            IERC20(asset).safeTransferFrom(msg.sender, address(this), amountDeposited);
```
kerzhner, line 62: Do we know what happens if this call is made for USDT but the sender has less than `amountDeposited` tokens? Silent failure? Or a an error?

kerzhner (approved):

nitro-protocol/contracts/MultiAssetHolder.sol
geoknee: Good question. Here's the verified source for USDT on mainnet https://etherscan.io/address/0xdac17f958d2ee523a2206206994597c13d831ec7#code . 

On L185 there, we have a SafeMath style subtraction which will revert the transaction if the sender has insufficient tokens.

PR #1023, opened 2023-01-04
Sort proposals only by turn number when processing

NiloCK:
This PR is an alternative / refinement to #1015, addressing issue #885.

See linked for discussion.

---

#### Code quality

#### Project management

### DIFF
--- a/protocols/messages.go
+++ b/protocols/messages.go
@@ -74,6 +74,8 @@ func (m Message) ProposalsSortedForProcessing() []consensus_channel.SignedPropos
 	return proposals
 }
 
+// proposalsSortedForSummary sorts the proposals by channelId and then by turn number.
+func (m Message) proposalsSortedForSummary() []consensus_channel.SignedProposal {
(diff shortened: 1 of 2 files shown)


protocols/messages.go
```diff
@@ -50,8 +50,32 @@ type Message struct {
...
+			// if the channel ids are different, do not sort
+			return i < j
+		}
+	})
+
+	return proposals
+}
+
+// proposalsSortedForSummary sorts the proposals by channelId and then by turn number.
+func (m Message) proposalsSortedForSummary() []consensus_channel.SignedProposal {
```
geoknee, line 78: I'm conflicted about this. On the one hand, it helps us understand that sorting by channel id is only really used for logging purposes and is not used in the core business logic. On the other hand, it means that we log something different to the core business logic, which seems a bit dangerous?

NiloCK: A good point. Highly legible misleading information is probably not great.

That said, with updated comments on #1015, I think it's safe to close this alternative off altogether: the problem in 1015 that this PR addressed was a figment of my imagination.

Issue #889, opened 2022-09-22
Consider: add contract address to FixedPart


NiloCK:
During Standup 2022-09-21, we discussed tooling for redeploying `NitroAdjudicator` on the existing hardhat docker instance used by our testground tests. A considered option was **redeploying new adjudicator builds to the existing long-running testnet**. This led to the question of whether there are real-world potentials for chains with multiple nitroadjudicators running simultaneously. It's not completely obvious that persons would benefit from this, but it *is* obvious that there's no way to prevent it.

Existing `FixedPart` stuct contains `ChainID` in order to prevent replay attacks. This should probably be extended to also include `AdjudicatorAddress`, so that signatures can't be replayed against different networks living on the same chain.

Optionally, the adjudicator address and chain ID could be collapsed into a single deterministic value called something like `networkID := hash(chainID + adjudicatorAddress)`

```
// solidity
struct FixedPart {
    uint256 chainId;
    address adjudicator; // add this
    address[] participants;
    uint48 channelNonce;
    address appDefinition;
    uint48 challengeDuration;
}

// go
FixedPart struct {
 ChainId           *types.Uint256
 Adjudicator       types.Address // add this
 Participants      []types.Address
 ChannelNonce      *types.Uint256 // uint48 in solidity
 AppDefinition     types.Address
 ChallengeDuration *types.Uint256
}
```

geoknee:
My preference would be to just add it into the FixedPart at the top level. And to avoid `networkID`, as I think that already has a meaning on Ethereum.

PR #1025, opened 2023-01-06
Remove chainID from FixedPart

### LINKED ISSUE #1016: Remove `ChainID` from FixedPart
Following various discussions:

`ChainID` is a part of fixedPart, and the ChannelID, for poor reasons.

ChainID was included in order to
1. situate channels on their chains
2. prevent replay attacks, where states & signatures applied to otherwise identical channels from different chains get 

But
1. is redundant: the existence of a channel on some chain speaks for itself
2. is not necessary, and moreover provides false assurances. Good nonce management is sufficient to prevent any nitro user from ever reusing a channel ID, which is sufficient to prevent replay attacks.

NiloCK:
Closes #1016

This PR removes `chainID` as a member of a channel's `FixedPart`. From the linked issue:

> Following various discussions:
> 
> `ChainID` is a part of fixedPart, and the ChannelID, for poor reasons.
> 
> ChainID was included in order to
> 
>     1. situate channels on their chains
> 
>     2. prevent replay attacks, where states & signatures intended for otherwise identical channels from different chains are applied against a vulnerable channel.
> 
> 
> But
> 
>     1. is redundant: the existence of a channel on some chain speaks for itself
> 
>     2. is not necessary, and moreover provides false assurances. Good nonce management is sufficient to prevent any nitro user from ever reusing a channel ID, which is sufficient to prevent replay attacks.

Notably: not ALL references to a chainID have been removed from the codebase. A go-nitro client will still need endpoints for a channel's adjudicator (chainID + contract address). In the case of single-adjudicator networks (current roadmap target), a go-nitro client and its chainservice can be configured with a single value for each.

**How Has This Been Tested?**

Modifications on inputs & expected outputs to existing test suite.

Probably not, because the change has been discussed in some detail, but

- [x] Is it security related?

is a yes.

---

#### Code quality

#### Project management


channel/channel_test.go
```diff
@@ -204,7 +204,6 @@ func TestChannel(t *testing.T) {
 		}
 
 		v := state.State{ // TODO it would be terser to clone s and modify it -- but s.Clone() is broken https://github.com/statechannels/go-nitro/issues/96
```
geoknee, line 206: Out of scope but we could remove this comment, since #96 is closed anyway.

nitro-protocol/test/contracts/NitroAdjudicator/concludeAndTransferAllAssets.test.ts
```diff
@@ -44,8 +44,6 @@ const token = setupContract(
 ) as unknown as Token & Contract;
 
 const provider = getTestProvider();
-const chainId = process.env.CHAIN_NETWORK_ID;
```
geoknee, line 47: I think there is some file(s) specifying this global variable checked in? May want to remove an entry there, too.

nitro-protocol/contracts/ForceMove.sol
```diff
@@ -50,7 +50,6 @@ contract ForceMove is IForceMove, StatusManager {
         SignedVariablePart memory candidate,
         Signature memory challengerSig
     ) external virtual override {
-        _requireCorrectChainId(fixedPart.chainId);
```
geoknee, line 53: Did you also remove the definition of this function (I think it is unused now?

protocols/directfund/directfund_test.go
```diff
@@ -24,7 +24,6 @@ const TEST_CHAIN_ID = 1337
 var alice, bob testactors.Actor = testactors.Alice, testactors.Bob
 
 var testState = state.State{
-	ChainId:           big.NewInt(TEST_CHAIN_ID),
```
geoknee, line 27: You could also remove the `TEST_CHAIN_ID` constant on L22 (there may be a few others floating around, too?)

geoknee (approved):
I'm still on-board with this change. It does make me want to expedite work to fully qualify assets in the exit-format, though. 

Pre- this change, participants sign the asset address and chain id together, so the qualification is there already. After the change, one wouldn't necessarily know whether the state channel manages (say) ETH or FIL just by looking at the state. You'd have to check through all adjudicators (see #889) on all chains (in principle). 

Would you agree that in practice, this information would exist out of band and doesn't need to be made explicit in the state? We already have that situation with the adjudicator contract address, so this change is just leaning in to that pattern while also allowing cross-chain applications.

geoknee:
Just realised it would be nice to update the docs at the same time. In particular this bit https://docs.statechannels.org/protocol-tutorial/0010-states-channels/#chain-id

mod:
Is there benefits of removing it?

geoknee:
> Is there benefits of removing it?

There are some benefits, yes:

1. Cross chain state channel applications: one way to do an atomic swap across chains would be to have a single state channel, funded by two adjudicators (one on each chain). Then, the same state can be finalized on both chains and funds released according to a multi-chain-multi-asset outcome. This isn't possible as long as states are explicitly bound to a single chain id. There are likely other cross chain applications (besides swaps) which would be enabled by this change. 
2. The current use of chain id **obscures** the best practice around security. We already have a `channelNonce` which is designed to prevent replaying states from one channel to another channel (it makes the channel id unique as long as participants choose a fresh nonce). With the chain id, it would be safe to reuse a channel nonce on a different chain. With the change of this PR (removing the chain id), it becomes very clear that the nonce is the only mechanism required to prevent replays.

Interested to know your feedback!

nitro-protocol/contracts/ForceMove.sol
NiloCK: removed in 308c17ff.

NiloCK:
> Pre- this change, participants sign the asset address and chain id together, so the qualification is there already. After the change, one wouldn't necessarily know whether the state channel manages (say) ETH or FIL just by looking at the state. You'd have to check through all adjudicators (see #889) on all chains (in principle).

In principle, it could manage both - although I don't know how much utility there is in that. By analogy, a given Ethereum key manages funds for the same address on other chains / rollups - but presumably you know where you have funds.

> 
> Would you agree that in practice, this information would exist out of band and doesn't need to be made explicit in the state? We already have that situation with the adjudicator contract address, so this change is just leaning in to that pattern while also allowing cross-chain applications.

:+1:

protocols/directfund/directfund_test.go
NiloCK: This const (and others still floating around, like the ENV variables) is used in methods that bubble out to the client API.

Currently, there's no way for the client to specify a chainID (or adjudicator address) at run-time, but presumably that would need to exist in some form or another, and the entry points for chainID would still need to be there.

In effect: taking these out would be doing some work that would eventually need to be re-done.

geoknee:
> By analogy, a given Ethereum key manages funds for the same address on other chains / rollups

Correct, although notable that Ethereum transactions include a signature over the chain id to prevent a transaction broadcast on one chain being replayed on another. I suppose this is where the original motivation for our use of chain id came from.  However, I still think we have a good justification for not using it.

PR #1027, opened 2023-01-09
Add unit tests for message proposal sorting

geoknee:
This will make it easier to see at-a-glance the differences between #1015  and #1023.

### DIFF
--- a/protocols/messages_test.go
+++ b/protocols/messages_test.go
@@ -114,26 +114,26 @@ func TestSortedProposals(t *testing.T) {
 	testCases := []TestCase{
 		{
 			Input: []consensus_channel.SignedProposal{
-				addProposal(types.Destination{'a'}, 1),
+				removeProposal(types.Destination{'a'}, 1),
 				addProposal(types.Destination{'a'}, 0),
 			},
 			Expectation: []consensus_channel.SignedProposal{
 				addProposal(types.Destination{'a'}, 0),
-				addProposal(types.Destination{'a'}, 1),
+				removeProposal(types.Destination{'a'}, 1),
 			},
 		},
 		{
 			Input: []consensus_channel.SignedProposal{
-				addProposal(types.Destination{'b'}, 1),
+				removeProposal(types.Destination{'b'}, 1),
 				addProposal(types.Destination{'b'}, 0),
-				addProposal(types.Destination{'a'}, 1),
+				removeProposal(types.Destination{'a'}, 1),
 				addProposal(types.Destination{'a'}, 0),
 			},
 			Expectation: []consensus_channel.SignedProposal{
 				addProposal(types.Destination{'a'}, 0),
-				addProposal(types.Destination{'a'}, 1),
+				removeProposal(types.Destination{'a'}, 1),
 				addProposal(types.Destination{'b'}, 0),
-				addProposal(types.Destination{'b'}, 1),
+				removeProposal(types.Destination{'b'}, 1),
 			},
 		},
 	}


client/engine/engine.go
```diff
@@ -252,22 +252,22 @@ func (e *Engine) handleMessage(message protocols.Message) (EngineEvent, error) {
 
 	}
 
-	for _, entry := range message.LedgerProposals {
+	for _, entry := range message.SortedProposals() {
```
geoknee, line 255:
```suggestion
        for _, entry := range message.LedgerProposals {
```
I didn't actually want to make this change, here.

protocols/messages_test.go
```diff
@@ -103,3 +105,47 @@ func TestMessage(t *testing.T) {
...
+func TestSortedProposals(t *testing.T) {
+	type TestCase struct {
+		Input       []consensus_channel.SignedProposal
+		Expectation []consensus_channel.SignedProposal
+	}
+	testCases := []TestCase{
+		{
+			Input: []consensus_channel.SignedProposal{
+				addProposal(types.Destination{'a'}, 1),
+				addProposal(types.Destination{'a'}, 0),
```
NiloCK, line 118: Obviously these are well defined in context of the unit under test (the sort algo), but it's maybe noteworthy that a ledger channel should never fund the same channelID more than once. I don't know whether it'd be worthwhile for a data structure at some point to prevent it.

protocols/messages_test.go
```diff
@@ -103,3 +105,47 @@ func TestMessage(t *testing.T) {
...
+			},
+			Expectation: []consensus_channel.SignedProposal{
+				addProposal(types.Destination{'a'}, 0),
+				addProposal(types.Destination{'a'}, 1),
+				addProposal(types.Destination{'b'}, 0),
+				addProposal(types.Destination{'b'}, 1),
+			},
+		},
+	}
```
NiloCK, line 139: Include a mixture of the proposal types?

```suggestion
	testCases := []TestCase{
		{
			Input: []consensus_channel.SignedProposal{
				removeProposal(types.Destination{'a'}, 1),
				addProposal(types.Destination{'a'}, 0),
			},
			Expectation: []consensus_channel.SignedProposal{
				addProposal(types.Destination{'a'}, 0),
				removeProposal(types.Destination{'a'}, 1),
			},
		},
		{
			Input: []consensus_channel.SignedProposal{
				removeProposal(types.Destination{'b'}, 1),
				addProposal(types.Destination{'b'}, 0),
				removeProposal(types.Destination{'a'}, 1),
				addProposal(types.Destination{'a'}, 0),
			},
			Expectation: []consensus_channel.SignedProposal{
				addProposal(types.Destination{'a'}, 0),
				removeProposal(types.Destination{'a'}, 1),
				addProposal(types.Destination{'b'}, 0),
				removeProposal(types.Destination{'b'}, 1),
			},
		},
	}
```

NiloCK (approved):
:+1:

protocols/messages_test.go
geoknee: It isn't very clear from reading the code, but these destinations do not represent target (virtual) channel Ids. They represent the id of the ledger channel. So they aren't falling foul of the rule you mention.

geoknee: Correction: perhaps that isn't what you meant. The `addProposal` helper uses a fixed guarantee targeting a fixed channel, so the proposals _are_ falling foul of that rule, independent of the arguments we pass in to `addProposal`.

Issue #1019, opened 2022-12-14
Remove workarounds from `eth_chainservice_FEVM_test.go`


geoknee:
Assuming the underlying issues are solved up stream, and a new version of wallaby (or other testnet) is deployed with those fixes, we should remove the following sections and have the test still pass without the workarounds:

https://github.com/statechannels/go-nitro/blob/eb199af1d4557631c4838829a2d9d03b2f629a1a/client/engine/chainservice/eth_chainservice_FEVM_test.go#L36-L40

https://github.com/statechannels/go-nitro/blob/eb199af1d4557631c4838829a2d9d03b2f629a1a/client/engine/chainservice/eth_chainservice_FEVM_test.go#L208-L228

This issue is effectively tracking the only remaining issues preventing us from being 100% compatible with FVM using unmodified code designed for Ethereum. The upstream issues are:

- https://github.com/filecoin-project/ref-fvm/issues/1182
- https://github.com/filecoin-project/ref-fvm/issues/1158

lalexgap:
FYI Wallaby is currently down for maintenance but should be up around the 9th: https://filecoinproject.slack.com/archives/C03KGBTJ0BY/p1673033966147019?thread_ts=1673033767.579779&cid=C03KGBTJ0BY

PR #1031, opened 2023-01-10
Remove chain id workaround in FEVM test

geoknee:
Towards #1019 .

See my comment on https://github.com/filecoin-project/ref-fvm/issues/1182 for proof that the chain id problem is fixed. 

Unfortunately the test does not pass 😢 .

It looks like the log topic from an emitted event does not match what we expect. A regression, perhaps?

I added a bunch of pointers to help run the test again after a network reset.

### DIFF
--- a/client/engine/chainservice/eth_chainservice_FEVM_test.go
+++ b/client/engine/chainservice/eth_chainservice_FEVM_test.go
@@ -31,13 +31,10 @@ const FEVM_PARSE_ERROR = "json: cannot unmarshal hex string \"0x\" into Go struc
 func TestEthChainServiceFEVM(t *testing.T) {
 	// Since this is hitting a contract on a test chain we only want to run it selectively
 	t.Skip()
-	// This is funded key on wallaby based on the test test ... fake mnemoic
-	pkString := "6b65fdf763faebfbcf9a43d5ab3dd2fb639a3d69c10df99eddc0a6eb30a99ba7"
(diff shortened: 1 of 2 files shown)


lalexgap:
@geoknee I've cleaned this up and slightly and I'm going to merge this in. While there's still the two outstanding issues:
- #1033 
- https://github.com/filecoin-project/ref-fvm/issues/1420

I think it makes sense to get the improved documentation and removed workaround in.

client/engine/chainservice/eth_chainservice_FEVM_test.go
```diff
@@ -31,14 +31,15 @@ const FEVM_PARSE_ERROR = "json: cannot unmarshal hex string \"0x\" into Go struc
 func TestEthChainServiceFEVM(t *testing.T) {
 	// Since this is hitting a contract on a test chain we only want to run it selectively
 	t.Skip()
-	// This is funded key on wallaby based on the test test ... fake mnemoic
+	// This is funded key on wallaby based on the
+	// test test test test test test test test test test test junk
+	// mnemoic
 	pkString := "6b65fdf763faebfbcf9a43d5ab3dd2fb639a3d69c10df99eddc0a6eb30a99ba7"
```
lalexgap, line 37: We should update this to the prefunded mnemonic we have.

lalexgap: I'll tackle this in a separate PR.

lalexgap (approved):

PR #1032, opened 2023-01-11
Add Signal/Wait functions to Objective Request

### LINKED ISSUE #1026: Objectives only get stored after crank which causes the query API to return no results
While working on #1020 I introduced a simple query API that queries the store directly. However it looks like we only store the objective after it's  been cranked. This means you can call something like `CreateLedgerChannel` which returns a channel id, but until that crank is finished the query API will return no results for that channel id.

For example something like this fails in a simple test:
```
	response := alpha.CreateLedgerChannel(*beta.Address, 0, outcome)
	ledgerInfo, err := alpha.GetLedgerChannel(response.ChannelId)
```

However by adding a minimal delay it passes:
```
        response := alpha.CreateLedgerChannel(*beta.Address, 0, outcome)
	time.Sleep(1 * time.Millisecond)
	ledgerInfo, err := alpha.GetLedgerChannel(response.ChannelId)
```

## Possible Solutions
- Live with it. It doesn't take that long for the objective to be stored. However I could see this behavior being confusing/annoying for developers
- The client API doesn't return a channel id until the channel is already in the store when creating/updating channels. For example `CreateLedgerChannel` would block until the `engine` signals that the ledger channel is in the store.
- The query API blocks somehow until the channel is actually in the store?

lalexgap:
Fixes #1026 

This PR adds Signal and Wait functions to `ObjectiveRequest` so the engine can alert the client API that the objective request has been processed and the objective has "started".

This means that any client API calls that create an objective will block until the objective has been created, cranked, and stored in the store. This ensures that querying the store after the API call returns will always contain the new objective.

The objective request interface has two new functions:
```
// ObjectiveRequest is a request to create a new objective.
type ObjectiveRequest interface {
	Id(types.Address, *big.Int) ObjectiveId
	WaitForObjectiveToStart()
	SignalObjectiveStarted()
}
```

`WaitForObjectiveToStart()` is called by the client API after submitting the objective request to the engine. It blocks until the engine calls `SignalObjectiveStarted()`.

I've also created constructor functions `NewObjectiveRequest` so we don't have to expose the chan we use for signalling.

### DIFF
--- a/protocols/directdefund/directdefund.go
+++ b/protocols/directdefund/directdefund.go
@@ -350,9 +350,9 @@ func (r ObjectiveRequest) SignalObjectiveStarted() {
 	r.objectiveStarted <- struct{}{}
 }
 
-// ObjectiveStarted returns a channel used to signal when the objective is started
-func (r ObjectiveRequest) ObjectiveStarted() <-chan struct{} {
-	return r.objectiveStarted
+// WaitForObjectiveToStart blocks until the objective starts
+func (r ObjectiveRequest) WaitForObjectiveToStart() {
+	<-r.objectiveStarted
 }
 
 // Id returns the objective id for the request.
--- a/protocols/directdefund/directdefund.go
+++ b/protocols/directdefund/directdefund.go
@@ -333,7 +333,26 @@ func (o *Objective) clone() Objective {
 
 // ObjectiveRequest represents a request to create a new direct defund objective.
 type ObjectiveRequest struct {
-	ChannelId types.Destination
+	ChannelId        types.Destination
+	objectiveStarted chan struct{}
+}
+
+// NewObjectiveRequest creates a new ObjectiveRequest.
+func NewObjectiveRequest(channelId types.Destination) ObjectiveRequest {
+	return ObjectiveRequest{
+		ChannelId:        channelId,
+		objectiveStarted: make(chan struct{}),
+	}
+}
+
+// SignalObjectiveStarted is used by the engine to signal the objective has been started.
+func (r ObjectiveRequest) SignalObjectiveStarted() {
+	r.objectiveStarted <- struct{}{}
+}
+
+// ObjectiveStarted returns a channel used to signal when the objective is started
+func (r ObjectiveRequest) ObjectiveStarted() <-chan struct{} {
+	return r.objectiveStarted
 }
 
 // Id returns the objective id for the request.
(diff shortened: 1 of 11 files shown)


protocols/directdefund/directdefund.go
```diff
@@ -333,7 +331,26 @@ func (o *Objective) clone() Objective {
...
+func NewObjectiveRequest(channelId types.Destination) ObjectiveRequest {
+	return ObjectiveRequest{
+		ChannelId:        channelId,
+		objectiveStarted: make(chan struct{}),
+	}
+}
+
+// SignalObjectiveStarted is used by the engine to signal the objective has been started.
+func (r ObjectiveRequest) SignalObjectiveStarted() {
+	close(r.objectiveStarted)
```
lalexgap, line 348: Instead of just sending a `struct{}` to the `chan` we can close it to indicate the chan is [complete](https://gobyexample.com/closing-channels).

geoknee (approved):
Looks good!

My only question is: do we want to allow the API consumer to not wait if they don't want to?

Some options which are small tweaks to this PR: 

* We could have `CreateChannelSync` as a wrapper for `CreateChannel`. Then the consumer can choose which to call.
* Or we could return the `WaitForObjectiveToStart` function from the API. The consumer can choose to call it immediately, later, or not at all. 

If we think there is no way the API consumer is going to want to have this option, then what we have here looks fine. We can also return to it later if we decide that should be an option.

lalexgap:
> Looks good!
> 
> My only question is: do we want to allow the API consumer to not wait if they don't want to?
>
> Some options which are small tweaks to this PR:
> 
>     * We could have `CreateChannelSync` as a wrapper for `CreateChannel`. Then the consumer can choose which to call.

Using the term `Sync` might be slightly confusing, as we're not blocking for the whole of the `CreateChannel` process, just until the objective has started. I also think it might be a slight anti-pattern in golang to expose sync/async functions of an API, since in golang you can turn a sync call to an async call or vice versa using go-routines.

>     * Or we could return the `WaitForObjectiveToStart` function from the API. The consumer can choose to call it immediately, later, or not at all.
> 
My one concern with exposing `WaitForObjectiveToStart` is that it might not be clear to developers they need to call it. It might be surprising that something like this wouldn't work:
```
response := alpha.CreateLedgerChannel(*beta.Address, 0, outcome)
ledger := alpha.GetLedgerChannel(response.ChannelId)
```

I think we'd have to do a really good job of documenting the importance of the wait function if we expose it.

> If we think there is no way the API consumer is going to want to have this option, then what we have here looks fine. We can also return to it later if we decide that should be an option.

I think the nice thing about golang, is that even if we don't provide an async API function, developers can use go routines and channels to call the API function asynchronously anyways (although it would be slightly more awkward than providing an async API function). So a develop won't be limited by our choice here.

My vote is to keep the current behavior for now and just modify it in the future if we need to. 

One use case that I could imagine might benefit from not blocking at all is creating channels in bulk. Perhaps we should expose an API function for that in the future?

PR #1034, opened 2023-01-12
Query in batches to avoid epoch error

### LINKED ISSUE #1033: Investigate invalid epoch range error
Right now when we poll for events in the chain service the first query we make is the from the [genesis](https://github.com/statechannels/go-nitro/blob/cd34325f023e0cffac68bf880f07ad8574cde242/client/engine/chainservice/eth_chainservice.go#L292) block to the current block. When run against the [wallaby node](https://wallaby.node.glif.io/rpc/v0) we get an `invalid epoch range` error back.

We should figure out how epochs work and if there's a way we can avoid running into this (maybe querying be epoch). We could also consider raising an issue with the filecoin tea  after we investigate.

lalexgap:
Fixes #1033 

This PR updates the chain service to fetch logs from the JSON-RPC in batches instead of all at once.

The main motivation behind this was to avoid the `invalid epoch range` error we ran against the wallaby FEVM node. We ran into this because their JSON-RPC API limits the block range you can query logs for to [2880](https://github.com/Zondax/rosetta-filecoin/blob/b395b3e04401be26c6cdf6a419e14ce85e2f7331/tools/wallaby/files/config.toml#L243). However when our chainservice starts we try to query from the genesis block to the current block.

In general, I think some kind of batching like this is a good idea. Most public JSON-RPC nodes have some [limitations](https://docs.infura.io/infura/networks/ethereum/json-rpc-methods/eth_getlogs) on the amount  of logs you can query, so this change will make it easier for the chainservice to work with them.

The block range I chose of 2000 was just a nice round number that was less than 2880. We can always finetune this (or allow it to be configurable) in the future.

### DIFF
--- a/client/engine/chainservice/eth_chainservice.go
+++ b/client/engine/chainservice/eth_chainservice.go
@@ -42,6 +42,12 @@ type EthChainService struct {
 	logger                   *log.Logger
 }
 
+// MAX_QUERY_BLOCK_RANGE is the maximum range of blocks we query for events at once.
+// Most json-rpc nodes restrict the amount of blocks you can search.
+// For example Wallaby supports a maximum range of 2880
+// See https://github.com/Zondax/rosetta-filecoin/blob/b395b3e04401be26c6cdf6a419e14ce85e2f7331/tools/wallaby/files/config.toml#L243
+const MAX_QUERY_BLOCK_RANGE = 2000
+
 // Since we only fetch events if there's a new block number
 // it's safe to poll relatively frequently.
 const EVENT_POLL_INTERVAL = 500 * time.Millisecond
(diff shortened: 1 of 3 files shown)


client/engine/chainservice/eth_chainservice.go
```diff
@@ -281,20 +287,60 @@ func (ecs *EthChainService) subscribeForLogs(ctx context.Context) {
...
+	logs = append(logs, fetchedLogs...)
+	return logs, nil
 
+}
+
+// pollForLogs periodically polls the chain client to check if there new events.
+// It can function over a chain node that does not support notifications.
+func (ecs *EthChainService) pollForLogs(ctx context.Context) {
+	toBlock := ecs.getCurrentBlockNum()
+	fetchedLogs, err := ecs.fetchLogsFromChain(big.NewInt(0), toBlock)
```
lalexgap, line 343: Instead of starting at the genesis block we should use the block the contract was deployed. I'm just not sure where to find that information.

geoknee: I think that information is not readily available, unless we were to store it when we do the deploy https://github.com/ethers-io/ethers.js/discussions/2716#discussioncomment-2215706

client/engine/chainservice/simulated_backend_chainservice.go
```diff
@@ -57,10 +58,30 @@ type SimulatedBackendChainService struct {
...
+	ethChainService, err := newPollingSimulatedChainService(sim,
+		bindings.Adjudicator.Contract,
+		bindings.Adjudicator.Address,
+		bindings.ConsensusApp.Address,
+		bindings.VirtualPaymentApp.Address,
+		txSigner,
+		logDestination)
+	if err != nil {
+		return &SimulatedBackendChainService{}, err
+	}
```
geoknee, line 75: I could be missing something, but it doesn't seem terribly worth it to have this wrapper constructor? Could we combine the two?

geoknee (approved):

client/engine/chainservice/simulated_backend_chainservice.go
lalexgap: I've merged `newPollingSimulatedChainService` into `NewPollingSimulatedBackendChainService` and renamed `newPollingSimulatedBackendChainService` to keep it private (since we only use it on our tests).

PR #1036, opened 2023-01-12
Сhange `requireStateSupported` from pure to view

nksazonov:
# Reasoning

Some state channel apps might need to read a blockchain to decide whether a state is supported or not. The most obvious example: getting chainId.

This is not possible in the current implementation as `requireStateSupported` has `pure` state mutability modifier, which disallows reading blockchain state.

# Changes

This PR relaxes `requireStateSupported` `pure` state mutability to `view`, allowing state channel applications to read a blockchain state. Still, some may decide `pure` is enough for them, and thus implement `requireStateSupported` with `view` state mutability.

# Follow up

We may have relaxed the conditions even more, removing state mutability modifiers, thus allowing reading and writing to a blockchain. However, this may have unexpected effects on protocol security, therefore such change is postponed until it is needed.


nitro-protocol/contracts/interfaces/IForceMoveApp.sol
```diff
@@ -10,7 +10,7 @@ import './INitroTypes.sol';
 interface IForceMoveApp is INitroTypes {
     /**
      * @notice Encodes application-specific rules for a particular ForceMove-compliant state channel. Must revert when invalid support proof and a candidate are supplied.
-     * @dev Encodes application-specific rules for a particular ForceMove-compliant state channel. Must revert when invalid support proof and a candidate are supplied.
+     * @dev Depending on application, it might be useful to stricken state mutability of an implementation to 'pure' for security reasons.
```
geoknee, line 13:
```suggestion
     * @dev Depending on the application, it might be desirable to narrow the state mutability of an implementation to 'pure' to make security analysis easier. 
```

geoknee:
@nksazonov can you say more about the logic you have in mind? A couple of questions occur to me:
* Why would you want to place chain id information in the variable part of a state update? Could the logic permit the chain id to change at some point?
* We have an existing check on the `chainid` in the fixed part of the state. But we plan to remove it, since we believe it to be unecessary. Are you 100% sure you need to have any explicit checks on the chain id?

nksazonov:
> * Why would you want to place chain id information in the variable part of a state update? Could the logic permit the chain id to change at some point?

We want to check chain id information for some extra security.
We are deploying `YellowAdjudicator` to several chains, thus there is a slight chance that signed VariablePart from one chain would substitute the other one (given there are two virtual channels with at least one same participant on several chains).
This would lead to unexpected results.

> * We have an existing check on the `chainid` in the fixed part of the state. But we plan to remove it, since we believe it to be unecessary. Are you 100% sure you need to have any explicit checks on the chain id?

Yes, we are 100% sure we want this check. Moreover, as Louis said, we might want to read blockchain state in `requireStateSupported` for other purposes than just chain id.

geoknee:
In that case I will draw your attention to #1025.

We have a general off-chain method to prevent replays -- the `channelNonce`. If participants are careful to choose this uniquely every time they start a new channel, they can avoid replay attacks without relying on any check with the `chainId`. 

We are inclined to accept this PR, based on the use-case of reading some chain state such as a price feed. But in the meantime, I'd like to continue discussing the `chainId` question, since I think we need to check each other's understanding on that.

geoknee (approved):

PR #1040, opened 2023-01-12
Use prefunded accounts for fevm test

lalexgap:
We now have a set of prefunded accounts on wallaby, that should get funded every time the network is reset. The first 25 accounts dervived from the mnemonic `army forest resource shop tray cluster teach cause spice judge link oppose` get funded with 500 FIL. We've already updated our testground test to uses these accounts, but our FEVM chain service test was using a non-prefunded account.

This PR updates the chain test to use prefunded accounts instead.


client/engine/chainservice/eth_chainservice_FEVM_test.go
```diff
@@ -62,8 +76,9 @@ func TestEthChainServiceFEVM(t *testing.T) {
 
 	// This is the deployed contract on wallaby
 	// If wallaby gets reset this will need to be redeployed by running:
-	// WALLABY_DEPLOYER_PK="<FUNDED PK>" npx hardhat deploy --network wallaby --deploy-scripts ./hardhat-deploy-fvm --reset
-	// NOTE the FUNDED PK should correspond to  hardhat account 0 (see nitro-protocol/accounts.json), which you need to fund using the faucet.
```
geoknee, line 66: Was this comment correct? Does the key you wrote here correspond to the hardhat account?

geoknee (approved):

client/engine/chainservice/eth_chainservice_FEVM_test.go
lalexgap: The key I wrote corresponds to the 0th account derived from `WALLABY_MNEMONIC` which is different than the hardhat account. This key should be prefunded.

geoknee: Right. I think when I tried it I thought I needed it to match some an account from the hardhat config. Take a look at this line

https://github.com/statechannels/go-nitro/blob/d4e2828f4f26778e0c1b4e1009304051c5fbb16b/nitro-protocol/hardhat-deploy-fvm/00-deploy.ts#L16

I'm not sure if the `WALLABY_DEPLOYER_PK` variable actually ends up affecting the deploy to wallaby in the way we expect?

So I don't think I was right to say the PK should correspond to the hardhat account. I think what may be important is that the hardhat account is funded. What do you think?

lalexgap: I think the `WALLABY_DEPLOYER_PK` gets read by hardhat [here](https://github.com/statechannels/go-nitro/blob/edcd760cd12210418c40de3064633d4741146689/nitro-protocol/hardhat.config.ts#L72).  The `deployer` gets set by hardhat to the wallaby key we provide.

You can see this by printing out the addresses
In golang
```
a, _ := wallet.Derive(hdwallet.MustParseDerivationPath(fmt.Sprintf("%s/%d", WALLABY_HD_PATH, 0)), false)
// fmt.Printf("Prefunded account: %s\n", a.Address)
// Prefunded account: 0xE39dce95b1A924E2472E24C20C55eA3559a09251
```

In the deploy script
```
console.log('DEPLOYER ADDRESS', deployer);
// DEPLOYER ADDRESS 0xE39dce95b1A924E2472E24C20C55eA3559a09251
```

PR #1045, opened 2023-01-19
Switch from wallaby to hyperspace FEVM testnet

### LINKED ISSUE #1035: FEVM tests: switch from wallaby to hyperspace testnet
https://filecoinproject.slack.com/archives/C04JEJB82RY/p1673444221760339

geoknee:
Closes #1035 

Notes

~1. There is still something funky with the deployment account not matching the mnemonic written in the test, contrary to what you expect from the comments.~ This may have been my mistake. I didn't notice we are using the 0th account and the 1st account (expected them to match). 

2. There are two endpoints to choose from. I found a different error with each, both lead to a failing test. See commit messages for more info
3. The FVM team are looking for rapid feedback about how this testnet is performing. So be good to dig into this early doors and try and then provide updates on slack.

### DIFF
--- a/client/engine/chainservice/eth_chainservice.go
+++ b/client/engine/chainservice/eth_chainservice.go
@@ -231,7 +231,7 @@ func (ecs *EthChainService) dispatchChainEvents(logs []ethTypes.Log) {
 			ecs.out <- event
 
 		default:
-			ecs.fatalF("Unknown chain event")
+			ecs.fatalF("Unknown chain event %+v", l)
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


lalexgap:
After a few small changes (and redeploying) we seem to be running into this [issue](https://github.com/filecoin-project/ref-fvm/issues/1420) that we saw on wallaby.

geoknee:
Thanks @lalexgap ! Hyperspace actually got[ patched earlier today](https://filecoinproject.slack.com/archives/C04JEJB82RY/p1674192169129659), so I re-ran this just now. The test actually passes, against both end points too. So I will clean this up and merge in. 

The issue you linked has not been closed, but it looks to me like some changes around logs did make it in to the patch. So good chance the underlying problem has gone away.

PR #1047, opened 2023-01-20
Fix chain service polling in batches

lalexgap:
Fixes two errors with batched log polling in the chain service.

- We were using the original parameter passed into the function instead of `fromBlock` which gets updated.
- We were using the final `toBlock` instead of `nextBlock` in our query,.

TODO:
Figure out a way to test this.


geoknee (approved):

geoknee:
> TODO:
> Figure out a way to test this.

I've started on something over here https://github.com/statechannels/go-nitro/pull/1049

PR #1059, opened 2023-01-23
Introduce API to allow querying for channel info

lalexgap:
This adds two key functions to the client API to allow querying channel information:
1. `GetPaymentChannel(id types.Destination) (PaymentChannelInfo, error)`
2. `GetLedgerChannel(id types.Destination) (LedgerChannelInfo, error)`

The `PaymentChannelInfo` and `LedgerChannelInfo` proxies contain information about the status and funding of the channels.

To test these functions there is a new integration tests that run a ledger channel and virtual channel through their lifecycle and use the new API functions to check the status/outcome.

Most of this functionality has been added in a new `query` package, a subpackage of `client`

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -185,5 +185,5 @@ func (c *Client) GetPaymentChannel(id types.Destination) (PaymentChannelInfo, er
 // GetLedgerChannel returns the ledger channel with the given id.
 // If no ledger channel exists with the given id an error is returned.
 func (c *Client) GetLedgerChannel(id types.Destination) (LedgerChannelInfo, error) {
-	return GetLedgerChannelInfo(id, c.store)
+	return getLedgerChannelInfo(id, c.store)
 }
--- a/client/client.go
+++ b/client/client.go
@@ -173,7 +173,7 @@ func (c *Client) Pay(channelId types.Destination, amount *big.Int) {
 	c.engine.PaymentRequestsFromAPI <- engine.PaymentRequest{ChannelId: channelId, Amount: amount}
 }
 
-// GetPaymentChannel returns the ledger channel with the given id.
+// GetPaymentChannel returns the payment channel with the given id.
 // If no ledger channel exists with the given id an error is returned.
 func (c *Client) GetPaymentChannel(id types.Destination) (query.PaymentChannelInfo, error) {
 
--- a/client/client.go
+++ b/client/client.go
@@ -2,7 +2,6 @@
 package client // import "github.com/statechannels/go-nitro/client"
 
 import (
-	"errors"
 	"io"
 	"math/big"
 	"math/rand"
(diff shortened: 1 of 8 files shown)


client/client.go
```diff
@@ -28,6 +28,8 @@ type Client struct {
 	failedObjectives    chan protocols.ObjectiveId
 	receivedVouchers    chan payments.Voucher
 	chainId             *big.Int
+	store               store.Store
+	vm                  *payments.VoucherManager
```
lalexgap, line 32: It would be nice if the voucher wrote to the store instead of the client having to query it directly.

client/query/query.go
```diff
@@ -0,0 +1,159 @@
...
+		PaidSoFar:      paidSoFar,
+		RemainingFunds: remaining,
+	}
+}
+
+// getLatestSupported returns the latest supported state of the channel
+// or the prefund state if no supported state exists
+func getLatestSupported(channel *channel.Channel) state.State {
+	if channel.HasSupportedState() {
+		supported, _ := channel.LatestSupportedState()
```
kerzhner, line 54: Seems like the error is dropped here. Is that bad?

client/query/query.go
```diff
@@ -0,0 +1,159 @@
...
+		return supported
+	}
+	return channel.PreFundState()
+}
+
+// getLedgerBalanceFromState returns the balance of the ledger channel from the given state
+func getLedgerBalanceFromState(latest state.State) LedgerChannelBalance {
+
+	// TODO: We assume single asset outcomesch
```
kerzhner, line 63:
```suggestion
	// TODO: We assume single asset outcomes
```
```suggestion
	// TODO: We assume single asset outcomesch
```

client/client.go
```diff
@@ -167,3 +172,16 @@ func (c *Client) Pay(channelId types.Destination, amount *big.Int) {
 	// Send the event to the engine
 	c.engine.PaymentRequestsFromAPI <- engine.PaymentRequest{ChannelId: channelId, Amount: amount}
 }
+
+// GetPaymentChannel returns the ledger channel with the given id.
```
kerzhner, line 176:
```suggestion
// GetPaymentChannel returns the payment channel with the given id.
```
```suggestion
// GetPaymentChannel returns the ledger channel with the given id.
```

kerzhner (approved):
lgtm!

client/client.go
```diff
@@ -43,7 +46,9 @@ func New(messageService messageservice.MessageService, chainservice chainservice
 		panic(err)
 	}
 	c.chainId = chainId
-	c.engine = engine.New(messageService, chainservice, store, logDestination, policymaker, metricsApi)
+	c.store = store
```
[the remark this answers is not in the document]
lalexgap: The client now directly accesses the `store`  (to query it) so it needs to keep a reference to it.

client/query/query.go
lalexgap: I think I was just being a bit lazy since we'd never expect `LatestSupportedState` to  error if there is a supported state.

I've updated this in 4fee4032bd04201d3974adebc9c62ec66829a01c to bubble up the error.

PR #1062, opened 2023-01-24
Go-Nitro RPC service

calj:
This PR is an attempt to implement a RPC service from the go-nitro library.

## Why do we need a Go-Nitro RPC service

The go-nitro protocol provides the ability for developers to plug their own application and use the power of state channels.

Having go-nitro as a service gives the ability to developers to implement their own server-side logic in any language and connect to go-nitro using simple protocol like REST and JSON RPC.

## Design of the code

The code contains few interface to improve the extensibility of the code:

### Transport (network/protocol)
This interface provide the transportation layer logic handling binary connections.

It supports transport over:
- Go Channel
- NATS

Possible implementations are:
- REST
- Websocket
- Libp2p

### Serde (network/serde)
This interface serialize/de-serialize messages from/to the transport layer.

Current PR supports:
- JSON RPC format

Possible implementations could be:
- MSG Pack

### Merge existing *messageservice* with the network package

*messageservice* and *network* packages are very similar, we could merge them into one flexible package.

In our case we would like to use libp2p to connect different peers and NATS between go-nitro service and our application.

### Do we want go-nitro repository to provide a binary of service?

Providing a ready to use binary can be handy for users. They can run quickly go-nitro service and focus on their application development. 
Doing so imply to add all supported transports/encoding libraries as dependencies, or implement plugins.

## TODO

- [ ] Fix client tests, add asserts
- [ ] Update the main example / documentation
- [ ] Improve the service handler to register response handler on request


network/protocol/parser/parser.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+func ParseDirectFundResponse(data map[string]any) directfund.ObjectiveResponse {
+	r := directfund.ObjectiveResponse{
+		Id:        protocols.ObjectiveId(data["id"].(string)),
+		ChannelId: types.Destination(common.HexToHash(data["channel_id"].(string))),
+	}
+
+	return r
+}
```
geoknee, line 68: Should there be some error handling here, in case we try to parse an invalid response?

channel/state/outcome/allocation.go
```diff
@@ -18,10 +18,10 @@ const (
...
 // Allocation declares an Amount to be paid to a Destination.
 type Allocation struct {
-	Destination    types.Destination // Either an ethereum address or an application-specific identifier
-	Amount         *types.Uint256    // An amount of a particular asset
-	AllocationType AllocationType    // Directs calling code on how to interpret the allocation
-	Metadata       []byte            // Custom metadata (optional field, can be zero bytes). This can be used flexibly by different protocols.
+	Destination    types.Destination `json:"destination"`     // Either an ethereum address or an application-specific identifier
+	Amount         *types.Uint256    `json:"amount"`          // An amount of a particular asset
+	AllocationType AllocationType    `json:"allocation_type"` // Directs calling code on how to interpret the allocation
+	Metadata       []byte            `json:"metadata"`        // Custom metadata (optional field, can be zero bytes). This can be used flexibly by different protocols.
```
geoknee, line 24: Is it important that we encode the JSON with lowercase first letter?

kerzhner:
@calj, thanks for creating this PR. We are convinced that go-nitro needs an RPC api, so this work is very timely.

At the moment, go-nitro contains serde functionality for various object. One example is [here](https://github.com/statechannels/go-nitro/blob/17797fdf582e5456806ccbf1b7b73c8539e574d5/channel/consensus_channel/serde.go#L1). This PR also introduces [serde code](https://github.com/openware/go-nitro/blob/eeeb3ab10cd9f4707e301629c091e6b33a6dd23b/network/serde/jsonrpc.go#L37). Do you think it is reasonable to unify the serialization/deserialization approaches of existing code and the new code?

client/client.go
```diff
@@ -22,12 +22,12 @@ import (
 
 // Client provides the interface for the consuming application
 type Client struct {
-	engine              engine.Engine // The core business logic of the client
+	Engine              engine.Engine // The core business logic of the client
```
kerzhner, line 25: Is this intended to be a long term approach or just some quick code to get the prototype working?

My 2c: the client intentionally abstracts the engine from api consumers. An rpc api is a way for remote consumers of a nitro node to call client methods. So the nitro node rpc api code would do the following:
1. Receives some data string from the network.
2. Parses the data string into a jsonrpc request.
3. Invokes a client method based on the jsonrpc request.

channel/state/outcome/allocation.go
calj: It's not.
I used this case to be consistent with JSON RPC structures.

client/client.go
calj: yes this was a fast approach to get it to work.
You suggest to implement methods at the client level to forward the data from/to the engine?

network/protocol/parser/parser.go
calj: yes, and we should probably change a bit the usage and set the response/error handlers per request.

examples/rpc_nats/main.go
```diff
@@ -0,0 +1,267 @@
...
+			logger.Fatal().Msg("unexpected empty args for direct funding method")
+			return
+		}
+
+		for i := 0; i < len(m.Args); i++ {
+			res := m.Args[i].(map[string]any)
+			req := parser.ParseDirectFundRequest(res)
+
+			logger.Info().Msgf("Objective Request: %v", req)
+			clientB.Engine.ObjectiveRequestsFromAPI <- req
```
geoknee, line 151: It should be possible to reuse this method on `clientB`, and then we don't need to expose the `Engine` (`engine`):

https://github.com/statechannels/go-nitro/blob/0d46d6c5a5d97a2ad38e9229ae0725713301ea91/client/client.go#L133-L151

client/client.go
kerzhner: The way I have been thinking about RPC design is:
1. RPC code receives a message/request.
2. RPC code deserializes the message.
3. RPC code calls a client method based on the message.

geoknee:
In favour of #1068 (which is in this repo, not a fork -- allowing us to collaborate more easily)

PR #1064, opened 2023-01-24
refactor challenge_test.go and extend to target hyperspace FEVM testnet

### LINKED ISSUE #1051: Try running challenge_test.go against a FEVM chain (e.g. hyperspace)
This would be a natural thing to try as a part of the "bulletproofing" efforts for the FEVM.

geoknee:
This expands the surface of FEVM that we are exercising with tests. 

Closes #1051


client/engine/chainservice/adjudicator/challenge_fevm_test.go
```diff
@@ -0,0 +1,165 @@
...
+const WALLABY_HD_PATH = "m/44'/1'/0'/0"
+
+func TestChallengeFEVM(t *testing.T) {
+	t.Skip() // We only want to run this test manually for now.
+	hyperspaceChainId := big.NewInt(3141)
+
+	// Setup transacting EOA
+	wallet, err := hdwallet.NewFromMnemonic(WALLABY_MNEMONIC)
+	if err != nil {
+		panic(err)
```
kerzhner, line 36: Any particular reason panic instead of t.Fatal is used here?

client/engine/chainservice/adjudicator/challenge_fevm_test.go
```diff
@@ -0,0 +1,165 @@
...
+	client, err := ethclient.Dial("https://api.hyperspace.node.glif.io/rpc/v0")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// When submitting a transaction it's signed against a specific chain id
+	// To get the correct signature we need to use the correct chain id that wallaby is expecting
+	txSubmitter, err := bind.NewKeyedTransactorWithChainID(pk, hyperspaceChainId)
+	if err != nil {
+		log.Fatal(err)
```
kerzhner, line 62: Any particular reason log.Fatal instead of t.Fatal is used here?

client/engine/chainservice/adjudicator/challenge_fevm_test.go
```diff
@@ -0,0 +1,165 @@
...
+		// Make assertion
+		if !bytes.Equal(statusOnChain[:], expectedOnChainStatus) {
+			t.Fatalf("Adjudicator not updated as expected, got %v wanted %v", common.Bytes2Hex(statusOnChain[:]), common.Bytes2Hex(expectedOnChainStatus[:]))
+		}
+
+	}
+
+	for _, turnNum := range []uint64{0} {
+		t.Run("turnNum = "+fmt.Sprint(turnNum), func(t *testing.T) { TestChallengeWithTurnNum(t, turnNum) })
+	}
```
kerzhner, line 164: Is the goal to eventually test with more than turn number 0?

client/engine/chainservice/adjudicator/challenge_fevm_test.go
```diff
@@ -0,0 +1,165 @@
...
+	na, err := NewNitroAdjudicator(naAddress, client)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	balance := new(big.Int)
+	balance.SetString("10000000000000000000", 10) // 10 eth in wei
+
+	TestChallengeWithTurnNum := func(t *testing.T, turnNum uint64) {
+
```
kerzhner, line 85: Nit, seems like our code base mixes zero or one new line after a definition. We should probably stick with one standard.

kerzhner (approved):
Thanks for putting this together! It would be really nice to write a test like this once and factor out all chain specific logic (instead of the current copy/paste/modify approach). But perhaps that's not worth the effort at the moment.

client/engine/chainservice/adjudicator/challenge_fevm_test.go
geoknee: Agreed. Do we have a preference?

kerzhner: I prefer no newlines.

geoknee:
> It would be really nice to write a test like this once and factor out all chain specific logic (instead of the current copy/paste/modify approach).

I did a bit of work in this direction on this PR. I think longer term, we will want the various "back ends" to their own packages, so we can easily import a handle on a "prepared" chain (with deployed contracts) from multiple tests.

PR #1068, opened 2023-01-26
Go-Nitro RPC server

calj:
This PR attempts to implement an RPC service from the go-nitro library.

## Why do we need a Go-Nitro RPC service

The go-nitro protocol allows developers to plug their applications and use the power of state channels.

Having go-nitro as a service allows developers to implement their server-side logic in any language and connect to go-nitro using a simple protocol like REST and JSON RPC.

## Design of the code

The code contains a few interface to improve the extensibility of the code:

### Transport (network/protocol)
This interface provides the transportation layer logic handling binary connections.

It supports transport over:
- Go Channel
- NATS

Possible implementations are:
- REST
- Websocket
- Libp2p

### Serde (network/serde)
This interface serializes/de-serializes messages from/to the transport layer.

Current PR supports:
- JSON RPC format

Possible implementations could be:
- MSG Pack

### Merge existing *messageservice* with the network package

*messageservice* and *network* packages are very similar. We could merge them into one flexible package.

In our case, we would like to use libp2p to connect different peers and NATS between the go-nitro service and our application.

### Do we want the go-nitro repository to provide a binary of services?

Providing a ready-to-use binary can be handy for users. They can run quick go-nitro services and focus on their application development. 
Doing so implies adding all supported transports/encoding libraries as dependencies or implementing plugins.

## TODO

- [ ] Fix client tests, add asserts
- [ ] Improve the service handler to register the response handler on request
- [ ] Add a main to run the RPC server


kerzhner:
Thank you for contributing this PR @calj! Very well done, and we learned about some nice tools from your code!

We built off your PR via https://github.com/statechannels/go-nitro/pull/1073. We are hoping to merge that pull request this week after adding some more tests and cleanups. It would be great to hear your feedback on that PR!

PR #1070, opened 2023-01-26
Use a zerolog logger in engine and chain service

geoknee:
This is a super easy change, and I'm interested to see if it has an impact on our performance in testground. 

There are other justifications to make this switch, such as the ability to set log levels and to be able to more-easily parse the logs.

Example output in `/artifacts/test_direct_fund.log`

```json
{"l":"debug","txSigner:":"0x9e3F65","t":1675079708,"caller":"eth_chainservice.go:74","m":"Notifications are supported by the chain. Using notifications to listen for events."}
{"l":"debug","txSigner:":"0x9112B1","t":1675079708,"caller":"eth_chainservice.go:74","m":"Notifications are supported by the chain. Using notifications to listen for events."}
{"l":"debug","engine":"0xAAA662","t":1675079708,"caller":"engine.go:112","m":"Constructed Engine"}
{"l":"debug","engine":"0xBBB676","t":1675079708,"caller":"engine.go:112","m":"Constructed Engine"}
{"l":"debug","engine":"0xAAA662","t":1675079708,"caller":"engine.go:368","m":"handling new objective request for DirectFunding-0xb9eed3fb2e6ee2a8aafaa14165be8fdc4233a32730a26791f31ff3db369ad938"}
{"l":"debug","engine":"0xAAA662","t":1675079708,"caller":"engine.go:513","m":"Objective DirectFunding-0x69ba76e574c44ff1c4ed4fb1bbb16be49fac8bc7783663be82b5253a2740a374 is WaitingForCompletePrefund"}
{"l":"debug","engine":"0xAAA662","t":1675079708,"caller":"engine.go:692","m":"Sending message {To:0xBBB676f9cFF8D242e9eaC39D063848807d3D1D94 PayloadSummaries:[{ObjectiveId:DirectFunding-0x69ba76e574c44ff1c4ed4fb1bbb16be49fac8bc7783663be82b5253a2740a374 Type:SignedStatePayload PayloadDataSize:774}] ProposalSummaries:[] Payments:[] RejectedObjectives:[]}"}
[...]

PR #1073, opened 2023-01-27
Add Nitro RPC

kerzhner:
This PR builds off https://github.com/statechannels/go-nitro/pull/1068.

The PR introduces a start for the JSON-RPC API to go-nitro. The API can be called from any client over any network. In this PR, we are using a nats server to unit test the client to node interaction.


client_test/rpcnats_test.go
```diff
@@ -1,42 +1,31 @@
...
-func setupClientWithP2PMessageService(pk []byte, port int, chain *chainservice.MockChainService, logDestination io.Writer) (client.Client, *p2pms.P2PMessageService) {
-	messageservice := p2pms.NewMessageService("127.0.0.1", port, pk)
-	storeA := store.NewMemStore(pk)
-
-	return client.New(messageservice, chain, storeA, logDestination, &engine.PermissivePolicy{}, nil), messageservice
-}
-
 // User real blockchain simulated_backend_service
 func TestRunRpcNats(t *testing.T) {
 	logger := zerolog.New(zerolog.ConsoleWriter{
```
geoknee, line 31: Inspired by this use of `zerolog`, I made #1070.

client_test/rpcnats_test.go
```diff
@@ -108,19 +100,7 @@ func TestRunRpcNats(t *testing.T) {
...
-	//	CounterParty:      alice.Address(),
-	//	ChallengeDuration: 100,
-	//	Outcome:           testdata.Outcomes.Create(alice.Address(), bob.Address(), 100, 100, types.Address{}),
-	//	AppDefinition:     chainServiceA.GetConsensusAppAddress(),
-	//	// Appdata implicitly zero
-	//	Nonce: rand.Uint64(),
-	//}
-
+	dfObjResponseChan := make(chan directfund.ObjectiveResponse, 1)
 	ntsA.RegisterRequestHandler(network.DirectFundRequestMethod, func(m *netproto.Message) {
```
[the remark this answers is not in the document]
lalexgap: As we chatted about, I'll take a look at organizing some of this intro a struct for the testground test.

client/rpc/client.go
```diff
@@ -0,0 +1,164 @@
...
+	objReq := directfund.NewObjectiveRequest(
+		counterparty,
+		100,
+		outcome,
+		uint64(rand.Float64()), // TODO: Since numeric fields get converted to a float64 in transit we need to prevent overflow
+		common.Address{})
+
+	// Create a channel and store it in the responses map
+	// We will use this channel to wait for the response
+	resRec := make(chan interface{})
```
geoknee, line 65: Is there scope for DRYing out some of the code common to the various methods?

client/rpc/client.go
```diff
@@ -0,0 +1,164 @@
...
+	chainId   *big.Int
+
+	// responses is a collection of channels that are used to wait until a response is received from the RPC server
+	responses safesync.Map[chan interface{}]
+
+	idsToMethods safesync.Map[serde.RequestMethod]
+}
+
+// NewRpcClient creates a new RpcClient
+func NewRpcClient(rpcServerUrl string, myAddress types.Address, chainId *big.Int, logger zerolog.Logger) *RpcClient {
```
geoknee, line 38: This constructor implementation currently relies on NATS. Is it worth reflecting that in the name of the constructor? Or perhaps we can do that at a later stage when we add a client which works over a different (or over an abstract) transport. It looks like the underlying struct is built against an interface, so that's good.

client/rpc/client.go
```diff
@@ -0,0 +1,164 @@
...
+	})
+}
+
+// handleResponse handles a response from the rpc server for the given client
+// It is not a member of the RpcClient so it can take advantage of generics
+func handleResponse[T serde.ResponsePayload](rc *RpcClient, data []byte) {
+	rpcResponse := serde.JsonRpcResponse[T]{}
+	err := json.Unmarshal(data, &rpcResponse)
+	if err != nil {
+		panic("could not unmarshal direct defund objective response")
```
geoknee, line 137: I don't think we know it is a direct defund objective response at this point?

client/rpc/client.go
```diff
@@ -0,0 +1,164 @@
...
+	objRes := <-resRec
+	return objRes.(directfund.ObjectiveResponse)
+}
+
+func (rc *RpcClient) CloseLedger(id types.Destination) protocols.ObjectiveId {
+	objReq := directdefund.NewObjectiveRequest(id)
+
+	// Create a channel and store it in the responses map
+	// We will use this channel to wait for the response
+	resRec := make(chan interface{})
```
geoknee, line 87: Can we narrow the type from `interface {}` to something better?

client/rpc/client.go
```diff
@@ -0,0 +1,164 @@
...
+// registerHandlers registers error and response handles for the rpc client
+func (rc *RpcClient) registerHandlers() {
+
+	rc.nts.RegisterErrorHandler(func(id uint64, data []byte) {
+		panic(fmt.Sprintf("Objective failed: %v", data))
+	})
+
+	rc.nts.RegisterResponseHandler(func(id uint64, data []byte) {
+		rc.nts.Logger.Trace().Msgf("Rpc client received response: %+v", data)
+		method, reqFound := rc.idsToMethods.Load(fmt.Sprintf("%d", id))
```
geoknee, line 116: Would it be better to infer the method from the data itself, rather than keeping some state in the client to correlate requests ids to methods?

client/rpc/server.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/statechannels/go-nitro/network/serde"
+	natstrans "github.com/statechannels/go-nitro/network/transport/nats"
+	"github.com/statechannels/go-nitro/protocols/directdefund"
+	"github.com/statechannels/go-nitro/protocols/directfund"
+)
+
+// RpcServer handles nitro rpc requests and executes them on the nitro client
+type RpcServer struct {
+	nts     *network.NetworkService
+	ns      *server.Server
```
geoknee, line 21: Just flagging the tight coupling to `nats-server` here.

client/rpc/util.go
```diff
@@ -0,0 +1,24 @@
+package rpc
+
+import (
+	"fmt"
+
+	"github.com/statechannels/go-nitro/network/serde"
+)
+
+// getTopics returns a list of topics that the client/server should subscribe to.
+func getTopics() []string {
```
geoknee, line 10: How central is the idea of "topics" to RPC in general? Is this something peculiar to NATS (which might not make sense if we used a different transport?)

client_test/rpc_test.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+	// Connect nitro P2P message services
+	msgA.AddPeers(peers)
+	msgB.AddPeers(peers)
+
+	defer msgA.Close()
+	defer msgB.Close()
+
+	alice := testactors.Alice
+	bob := testactors.Bob
```
geoknee, line 54: I believe these lines are unnecessary (try cmd-click on `alice` on L43).

network/serde/jsonrpc.go
```diff
@@ -0,0 +1,105 @@
...
+
+type MessageType int8
+
+const (
+	TypeRequest  MessageType = 1
+	TypeResponse MessageType = 2
+	TypeError    MessageType = 3
+)
+
+// JsonRpcMessage is union of jsonrpc request, response, and error
```
geoknee, line 34: Nit: I think it is an _intersection_, not a union (at least that is the terminology from typescript...)

network/service.go
```diff
@@ -0,0 +1,134 @@
...
+		msg, messageType, err := serde.Deserialize(data)
+		if err != nil {
+			p.Logger.Error().Err(err).Msg("failed to deserialize message")
+			return
+		}
+
+		// NOTE: we do not hande messages in a separate goroutine
+		// to ensure that messages are handled in the order they are received
+		// and to avoid inconsistencies in the state of the peer
+		p.handleMessage(msg.Id, msg.Method, messageType, data)
```
geoknee, line 83: I haven't thought about this deeply but it seems important... if we try handling in a go routine, do we see a test flicker somewhere?

network/service.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+}
+
+func (p *NetworkService) SendMessage(method serde.RequestMethod, data []byte) {
+	// FIXME: we can use one topic per app, but they have to be different
```
geoknee, line 88: Can you expand on this? I'm not sure I understand the implication.

network/service_test.go
```diff
@@ -0,0 +1,73 @@
...
+	service := NetworkService{
+		Logger:         zerolog.Logger{},
+		Connection:     &connMock{},
+		handlerRequest: sync.Map{},
+	}
+
+	return &service
+}
+
+// maybe logic will get more complicated and gonna keep this as example
```
geoknee, line 35:
```suggestion

```

network/transport/nats/nats_transport.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+type natsTransport struct {
+	connection transport.Connection
+}
+
+var _ transport.Transport = (*natsTransport)(nil)
+
+func NewNatsTransport(nc *nats.Conn, pubTopicNames []string) *natsTransport {
+	// wouldn't it be better to get messaged directly into the channel?
```
geoknee, line 15: Can we resolve this now?

network/transport/nats/nats_transport.go
```diff
@@ -0,0 +1,30 @@
...
+	connection := NewNatsConnection(nc, pubTopicNames)
+	natsTransport := &natsTransport{
+		connection: connection,
+	}
+
+	return natsTransport
+}
+
+func (t *natsTransport) PollConnection() (transport.Connection, error) {
```
geoknee, line 24: Can you explain why the "Poll" part?

geoknee (approved):
👏  Nice job @kerzhner @lalexgap @calj. This is a big step toward getting go-nitro running in its own process. 

It's a big changeset but things look pretty well organised to me.

client/client.go
```diff
@@ -98,6 +98,11 @@ func (c *Client) ReceivedVouchers() <-chan payments.Voucher {
 	return c.receivedVouchers
 }
 
+// IncomingObjectiveRequests is a chan that accepts objective requests from an RPC server.
+func (c *Client) IncomingObjectiveRequests() chan<- protocols.ObjectiveRequest {
+	return c.engine.ObjectiveRequestsFromAPI
+}
```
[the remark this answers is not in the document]
kerzhner: This is a great point. Initially, I followed the first approach you described. I would be curious to hear @lalexgap's thoughts.

client/rpc/client.go
kerzhner: This was pointed out in a comment in a currently open followup PR https://github.com/statechannels/go-nitro/pull/1084/files#r1094726558. Probably best to tackle DRYing out refactor in [that PR](https://github.com/statechannels/go-nitro/pull/1084) as that PR builds off this code.

client/rpc/client.go
kerzhner: Good point. We should probably introduce a second transport to make sure our code is transport agnostic. This was done as a short term solution.

Filed https://github.com/statechannels/go-nitro/issues/1097.

client/rpc/client.go
kerzhner: Good catch! https://github.com/statechannels/go-nitro/pull/1073/commits/7250a3242ac1b97ad503349f523c8e7d044b2f7f

client/rpc/client.go
kerzhner: Agreed! This should be addressed when DRYing out this code.

client/rpc/client.go
kerzhner: That is a good alternative approach. Filed https://github.com/statechannels/go-nitro/issues/1098

client/rpc/server.go
kerzhner: Good catch. Adding a second transport should be helpful in flushing out transport specific code like this.

client/rpc/util.go
kerzhner: In RPC, there are method requests and responses. We should get away from the concept of topics in transport agnostic code.

client_test/rpc_test.go
kerzhner: Good catch. [0fad994](https://github.com/statechannels/go-nitro/pull/1073/commits/0fad99498dc3bc5c2d6dc5520f95346c3cb92b7b)

network/serde/jsonrpc.go
kerzhner: Do you think this comment still applies if we disregard Typescript terminology?

network/service.go
kerzhner: Probably best answered @calj

network/service.go
kerzhner: We have not tried this. I believe you are highlighting that we should think deeper about message ordering. And that we should not rely on messages arriving in order. These are great points. Filed https://github.com/statechannels/go-nitro/issues/1099.

network/service_test.go
kerzhner: [98a1b6c](https://github.com/statechannels/go-nitro/pull/1073/commits/98a1b6c27616aa2676f73312d6ec190b6b0efd3b)

network/transport/nats/nats_transport.go
kerzhner: I am actually not 100% sure what this means. Maybe @calj can elaborate.

network/transport/nats/nats_transport.go
kerzhner: Great question. Maybe @calj can elaborate on the thinking behind the naming.

network/serde/jsonrpc.go
geoknee: I'm not really sure 🤷 . Anyway i think it's fine to leave as is, or we could say `JsonRpcMessage has all of the properties of jsonrpc request, response, and error` or something?

PR #1077, opened 2023-01-30
WIP: RPC 

lalexgap:
This PR organizes the code into an rpc client and server and makes some changes to get things working.

- https://github.com/statechannels/go-nitro/pull/1077/commits/5dfdb9e825a72f995ca7624c3e41edffcd57f9d9: Fix serialization of request response to include `method`. This was preventing response handlers from being triggered correctly.
- https://github.com/statechannels/go-nitro/pull/1077/commits/c04e0c5671d40958efb370e3a41127d1a32450d2: The `Result` field is expected to be a single struct not a slice, but we were passing in the `m.Args` slice. This caused parsing errors later on. 
- https://github.com/statechannels/go-nitro/pull/1077/commits/e7a0c5e8eb588d2792fac8a1a2aebe4e4ca583cb: It looks like numeric values are serialized to a `float64` when going across the wire. This means that any nonce larger than `float64` will be truncated.  As a temporary workaround I've limited the random nonce to the max size of `float64`

Right now there is a simple RPC test that passes (but flickers).


network/serde/jsonrpc_test.go
```diff
@@ -24,11 +24,12 @@ func TestJsonRpcSerializeResponse(t *testing.T) {
 	rpc := JsonRpc{}
 	data, err := rpc.Serialize(&netproto.Message{
```
geoknee, line 25: Could we nest `network` code be under the `rpc` directory? 

Longer term, if there is common / reused code between the rpc networking and the p2p networking, we could have a top level `network` package.

client_test/rpc_test.go
```diff
@@ -0,0 +1,61 @@
...
+	alice := testactors.Alice
+	bob := testactors.Bob
+
+	rpcServerA := rpc.NewRpcServer(&clientA, logger)
+	rpcClientA := rpc.NewRpcClient(rpcServerA.Url(), alice.Address(), clientA.ChainId, logger)
+	defer rpcServerA.Close()
+	defer rpcClientA.Close()
+	testOutcome := testdata.Outcomes.Create(alice.Address(), bob.Address(), 100, 100, types.Address{})
+
+	res := rpcClientA.CreateLedger(bob.Address(), 100, testOutcome)
```
kerzhner, line 54: I could be mistaken, but I don't think this test waits for the completion of the objective. It seems that the test only waits for the objective id that is included in the json rpc response.

client_test/rpc_test.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/statechannels/go-nitro/client/engine/chainservice"
+	p2pms "github.com/statechannels/go-nitro/client/engine/messageservice/p2p-message-service"
+	"github.com/statechannels/go-nitro/client/rpc"
+	"github.com/statechannels/go-nitro/internal/testactors"
+	"github.com/statechannels/go-nitro/internal/testdata"
+	"github.com/statechannels/go-nitro/types"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestRpcClient(t *testing.T) {
```
kerzhner, line 17: Do you think this test should replace the [rpcnats_test](https://github.com/statechannels/go-nitro/blob/2da58eaed4180f36b1d28cfaf89595307471f627/client_test/rpcnats_test.go#L30)?

lalexgap: Yup I think so

client_test/rpc_test.go
lalexgap: That's correct as I originally was only interested in making sure the API call worked. However I wonder if this is the cause for some of the flickering so I've updated the test to wait for completion in d3e5ef8

client_test/rpc_test.go
kerzhner: Excellent!

PR #1078, opened 2023-01-31
Use updated exit-format

NiloCK:
PR pulls in changes from the exit-format. Breaking change is that `SingAssetExit` now has a meta data descriptor enum:

```solidity
struct SingleAssetExit {
    address asset;
    bytes metadata;
    Allocation[] allocations;
}
```
was replaced with
```solidity
struct SingleAssetExit {
    address asset;
    AssetMetadata assetMetadata;
    Allocation[] allocations;
}
struct AssetMetadata {
    AssetType assetType;
    bytes metadata;
}
```

This required an amount of follow-on work with abi encoding, updating existing outcome literals, etc. No new test cases have been added.


nitro-protocol/package-lock.json
```diff
@@ -1,16 +1,16 @@
 {
   "name": "@statechannels/nitro-protocol",
-  "version": "2.0.0-alpha.2",
+  "version": "2.0.0-alpha.3",
```
geoknee, line 3: Any clue why there were some many deletions from this file? Looks like maybe we hadn't run `npm install` recently or something...

nitro-protocol/gas-benchmarks/gasResults.json
```diff
@@ -1,56 +1,56 @@
...
-      "challengeL": 121684,
-      "challengeV": 171987,
-      "reclaimL": 60350,
-      "transferAllAssetsL": 112705,
-      "total": 466726
+      "challengeL": 124789,
+      "challengeV": 175507,
+      "reclaimL": 61438,
+      "transferAllAssetsL": 113227,
+      "total": 474961
```
geoknee, line 53: Slightly painful ~2% increase in gas here :-/

geoknee (approved):
Looks good. Seeing as this changeset changes many files, perhaps let's wait for a quiet moment to merge?

nitro-protocol/package-lock.json
NiloCK: Not really - I think we've been running `npm ci` as standard practice for a while, so maybe some dependencies up the tree have pruned out their own dependency lists in the meantime?

One expects the opposite to happen on average though.

nitro-protocol/gas-benchmarks/gasResults.json
NiloCK: I was a little surprised at how much the gas costs had increased. It seems that the size of our data structures themselves make up a substantial amount of the costs, even when empty.

I don't know or have a good intuition on whether the nested struct `AssetMetadata` is more, less, or exactly equivalent in gas consumption to a flatter version (which is what's done on the allocations, for what it's worth - a little bit of a wart for these to be different).

NiloCK:
> Looks good. Seeing as this changeset changes many files, perhaps let's wait for a quiet moment to merge?

Sure thing. I think it shouldn't intersect much with the ongoing RPC work :crossed_fingers:.

PR #1081, opened 2023-02-01
Add Close Ledger to RPC API

lalexgap:
Adds the `CloseLedger` function to the RPC client which triggers an RPC call to the RPC server to start `direct_defunding`.

I've introduced some generic request/response structs that I think this should make adding the remaining API calls easier. 

The test has been updated to call `CloseLedger` and wait for the objective to close.

### DIFF
--- a/client/rpc/client.go
+++ b/client/rpc/client.go
@@ -13,6 +13,8 @@ import (
 	"github.com/statechannels/go-nitro/network"
 	"github.com/statechannels/go-nitro/network/serde"
 	natstrans "github.com/statechannels/go-nitro/network/transport/nats"
+	"github.com/statechannels/go-nitro/protocols"
+	"github.com/statechannels/go-nitro/protocols/directdefund"
 	"github.com/statechannels/go-nitro/protocols/directfund"
 	"github.com/statechannels/go-nitro/types"
 
@@ -26,7 +28,9 @@ type RpcClient struct {
 	chainId   *big.Int
 
 	// responses is a collection of channels that are used to wait until a response is received from the RPC server
-	responses safesync.Map[chan directfund.ObjectiveResponse]
+	responses safesync.Map[chan interface{}]
+
+	idsToMethods safesync.Map[serde.RequestMethod]
 }
 
 // NewRpcClient creates a new RpcClient
@@ -34,13 +38,24 @@ func NewRpcClient(rpcServerUrl string, myAddress types.Address, chainId *big.Int
 
 	nc, err := nats.Connect(rpcServerUrl)
 	handleError(err)
-	trp := natstrans.NewNatsTransport(nc, []string{fmt.Sprintf("nitro.%s", network.DirectFundRequestMethod)})
+	trp := natstrans.NewNatsTransport(nc, []string{
+		fmt.Sprintf("nitro.%s",
+			serde.DirectFundRequestMethod),
+		fmt.Sprintf("nitro.%s",
+			serde.DirectDefundRequestMethod),
+		fmt.Sprintf("nitro.%s",
+			serde.VirtualFundRequestMethod),
+		fmt.Sprintf("nitro.%s",
+			serde.VirtualDefundRequestMethod),
+		fmt.Sprintf("nitro.%s",
+			serde.PayRequestMethod),
+	})
 
 	con, err := trp.PollConnection()
 	handleError(err)
 	nts := network.NewNetworkService(con)
 	nts.Logger = logger
-	c := &RpcClient{nts, myAddress, chainId, safesync.Map[chan directfund.ObjectiveResponse]{}}
+	c := &RpcClient{nts, myAddress, chainId, safesync.Map[chan interface{}]{}, safesync.Map[serde.RequestMethod]{}}
 	c.registerHandlers()
 	return c
 }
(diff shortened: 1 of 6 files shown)


network/serde/jsonrpc.go
```diff
@@ -4,7 +4,21 @@ import (
...
+	"github.com/statechannels/go-nitro/protocols"
+	"github.com/statechannels/go-nitro/protocols/directdefund"
 	"github.com/statechannels/go-nitro/protocols/directfund"
+	"github.com/statechannels/go-nitro/protocols/virtualdefund"
+	"github.com/statechannels/go-nitro/protocols/virtualfund"
+)
+
+type RequestMethod string
+
+const (
```
lalexgap, line 16: I moved this here to avoid a circular import. I haven't given a ton of thought to where this `const` should actually live.

client/rpc/client.go
```diff
@@ -77,21 +119,47 @@ func (rc *RpcClient) Close() {
...
 		panic(fmt.Sprintf("Objective failed: %v", data))
 	})
-	rs.nts.RegisterResponseHandler(func(data []byte) {
-		rs.nts.Logger.Trace().Msgf("Rpc client received response: %+v", data)
 
-		rpcResponse := serde.JsonRpcDirectFundResponse{}
-		err := json.Unmarshal(data, &rpcResponse)
-		if err != nil {
-			panic("could not unmarshal direct fund objective response")
+	rs.nts.RegisterResponseHandler(func(id uint64, data []byte) {
```
lalexgap, line 126: Including the id in the function signature makes things easier for the RPC client handling a response. The RPC client can use the id to look up the original request type and then de-serialize `data` to the an exact type based on the original method type.

network/serde/jsonrpc.go
```diff
@@ -17,60 +31,49 @@ const (
...
-
-type JsonRpcDirectFundResponse struct {
-	Jsonrpc           string                       `json:"jsonrpc"`
-	Id                uint64                       `json:"id"`
-	ObjectiveResponse directfund.ObjectiveResponse `json:"result"`
-	Error             interface{}                  `json:"error"`
-}
-
-type JsonRpc struct{}
+func NewJsonRpcRequest[T RequestPayload](requestId uint64, method RequestMethod, objectiveRequest T) *JsonRpcRequest[T] {
```
lalexgap, line 61: Ideally we'd just infer `method` from the type of `objectiveRequest`. However it seems in golang you can't switch on the type of a generic parameter 😞

network/serde/jsonrpc.go
```diff
@@ -17,60 +31,49 @@ const (
...
 }
-
-type JsonRpcRequestResponse struct {
+type JsonRpcRequest[T RequestPayload] struct {
+	Jsonrpc string `json:"jsonrpc"`
+	Id      uint64 `json:"id"`
+	Method  string `json:"method"`
+	Params  T      `json:"params"`
+}
+type ResponsePayload interface {
```
lalexgap, line 51: We're not uniform in our response type. `directdefund` and `virtualdefund` just return an `ObjectiveId` instead of an `ObjectiveResponse`.

client_test/rpc_test.go
```diff
@@ -64,4 +64,10 @@ func TestRpcClient(t *testing.T) {
 
 	waitTimeForCompletedObjectiveIds(t, &clientA, defaultTimeout, res.Id)
 	waitTimeForCompletedObjectiveIds(t, &clientB, defaultTimeout, res.Id)
+
+	closeId := rpcClientA.CloseLedger(res.ChannelId)
```
kerzhner, line 68: Very nice to have a more robust unit test!

kerzhner (approved):
Thanks for doing this work!

client/rpc/client.go
```diff
@@ -26,21 +28,34 @@ type RpcClient struct {
...
 }
 
 // NewRpcClient creates a new RpcClient
 func NewRpcClient(rpcServerUrl string, myAddress types.Address, chainId *big.Int, logger zerolog.Logger) *RpcClient {
 
 	nc, err := nats.Connect(rpcServerUrl)
 	handleError(err)
-	trp := natstrans.NewNatsTransport(nc, []string{fmt.Sprintf("nitro.%s", network.DirectFundRequestMethod)})
+	trp := natstrans.NewNatsTransport(nc, []string{
+		fmt.Sprintf("nitro.%s",
```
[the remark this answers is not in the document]
lalexgap: I've hidden this behind a helper function in https://github.com/statechannels/go-nitro/pull/1081/commits/528c3cb3229c8d67f9556ddf47bd954b288717f3

client/rpc/client.go
```diff
@@ -55,43 +70,94 @@ func (rc *RpcClient) CreateLedger(counterparty types.Address, ChallengeDuration
...
+				panic("could not unmarshal direct defund objective response")
+			}
+
+			if resRec, ok := rc.responses.Load(string(rpcResponse.Result)); ok {
+				rc.responses.Delete(fmt.Sprintf("%v", rpcResponse.Id))
+				resRec <- rpcResponse.Result
+
+				rc.idsToMethods.Delete(string(rpcResponse.Result))
+			}
 		}
```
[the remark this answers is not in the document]
lalexgap: Yeah, they're quite similar. I've merged them into a generic function in 9aabc05

PR #1082, opened 2023-02-01
fix up much Go code given new exit format

geoknee:


NiloCK:
Changes from this PR have been manually ported to #1078. Thanks @geoknee!

PR #1083, opened 2023-02-01
Fail earlier in testground action

geoknee:
We don't want to post a link to a dashboard etc if the run has failed. 

Because of [a0f2a27](https://github.com/statechannels/go-nitro/pull/1083/commits/a0f2a27a6edc7d8fbf6285f1f9dad6af6943d33f), this has effectively been manually tested already (the run passed on [e062c68](https://github.com/statechannels/go-nitro/pull/1083/commits/e062c68619fa6096929b69f0a32729d80c66895f), but the successful outcome variable was never set and no link was posted). 

Apologies for the formatting changes... I can tidy that up on request.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -21,7 +21,7 @@ import (
 	"github.com/statechannels/go-nitro/types"
 )
 
-// Client provides the interface for the consuming application
+// Client provides an API for the consuming application
 type Client struct {
 	engine              engine.Engine // The core business logic of the client
 	Address             *types.Address
--- a/client/client.go
+++ b/client/client.go
@@ -21,7 +21,7 @@ import (
 	"github.com/statechannels/go-nitro/types"
 )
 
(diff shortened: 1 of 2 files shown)


client/client.go
```diff
@@ -21,7 +21,7 @@ import (
 	"github.com/statechannels/go-nitro/types"
 )
 
-// Client provides the interface for the consuming application
```
lalexgap, line 24: Nit but should we include this in the PR?

lalexgap (approved):

client/client.go
geoknee: No, [6971fb3](https://github.com/statechannels/go-nitro/pull/1083/commits/6971fb3db947ab6f2e660bffdcdbc63db3b4a1a6)

geoknee:
@lalexgap should we have logic like this in `testground-nightly.yml`?

PR #1084, opened 2023-02-01
Implement rest of mutation API

lalexgap:
Implements the rest of the API and updates the integration test to test them.

### DIFF
--- a/client/rpc/client.go
+++ b/client/rpc/client.go
@@ -16,6 +16,7 @@ import (
 	"github.com/statechannels/go-nitro/protocols"
 	"github.com/statechannels/go-nitro/protocols/directdefund"
 	"github.com/statechannels/go-nitro/protocols/directfund"
+	"github.com/statechannels/go-nitro/protocols/virtualdefund"
 	"github.com/statechannels/go-nitro/protocols/virtualfund"
 	"github.com/statechannels/go-nitro/types"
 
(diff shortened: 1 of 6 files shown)


network/transport/nats/nats_connection.go
```diff
@@ -38,6 +38,7 @@ func (c *natsConnection) subscribeToTopics() {
 		} else {
 			c.mutex.Lock()
 			c.natsSubscriptions = append(c.natsSubscriptions, sub)
+			c.mutex.Unlock()
```
lalexgap, line 41: This was causing some maddening behavior where only the first two topics would get subscribed to, as the loop would block forever on the second call to`c.mutex.Lock`. Since this is happening in a go-routine it was not obvious at all.

kerzhner: Very unfortunate. I wonder if we can replace the use of a mutex with go channels.

kerzhner (approved):

client/rpc/client.go
```diff
@@ -100,6 +153,27 @@ func (rc *RpcClient) CloseLedger(id types.Destination) protocols.ObjectiveId {
 	return objRes.(protocols.ObjectiveId)
 }
 
+func (rc *RpcClient) Pay(id types.Destination, amount uint64) {
+
```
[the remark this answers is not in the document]
lalexgap: Nope, I think this is one of the things (along with line length) that I forget the linter won't fix for me.

client/rpc/client.go
```diff
@@ -50,6 +51,58 @@ func NewRpcClient(rpcServerUrl string, myAddress types.Address, chainId *big.Int
...
+	message := serde.NewJsonRpcRequest(requestId, serde.VirtualFundRequestMethod, objReq)
+	data, err := json.Marshal(message)
+	if err != nil {
+		panic("Could not marshal direct fund request")
+	}
+	rc.nts.SendMessage(serde.VirtualFundRequestMethod, data)
+
+	objRes := <-resRec
+	return objRes.(virtualfund.ObjectiveResponse)
```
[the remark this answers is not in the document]
lalexgap: Yeah I started investigating a generic function to handle this but didn't get something working. I'll take a second look at this.

lalexgap:
> Very unfortunate. I wonder if we can replace the use of a mutex with go channels.

I'm actually not sure if we even need any blocking/locking. The only place we read or modify `natsSubscriptions` is when we `subcribe` and `unsubscribe`, so I don't think we'd ever have to worry about concurrent access to the slice.

network/transport/nats/nats_connection.go
lalexgap: I'm actually not sure if we even need any blocking/locking. The only place we read or modify `natsSubscriptions` is when we subscribe and unsubscribe, so I don't think we'd ever have to worry about concurrent access to the slice.

lalexgap: I've removed the mutex in https://github.com/statechannels/go-nitro/pull/1084/commits/d459aba3a5f448a8bb3a97bb81be3c3053a2e193

client/rpc/client.go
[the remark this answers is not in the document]
lalexgap: I'm struggling to implement this nicely with generics. Since I think we're looking at simplifying the request handlers in #1091 we may want to try refactoring this code then.

network/serde/jsonrpc.go
```diff
@@ -43,8 +44,13 @@ type JsonRpcMessage struct {
 	Message string        `json:"message"`
 	Data    interface{}   `json:"data"`
 }
+
+type PaymentRequest struct {
```
lalexgap, line 48: This is used to pass the payment information over the RPC interface. Maybe this should be a [voucher](https://github.com/statechannels/go-nitro/blob/d459aba3a5f448a8bb3a97bb81be3c3053a2e193/payments/vouchers.go#L25)?

PR #1088, opened 2023-02-02
add social plugin to mkdocs

geoknee:
This will give us previews like this: 
Screenshot 2023-02-02 at 15 23 03

when we post links to the site on social media and slack.


.gitignore
```diff
@@ -23,4 +23,7 @@ artifacts
 
 nitro-protocol/hardhat-network-output.log
 
-site
\ No newline at end of file
+site
+
+# Cache from mkdocs (docs website)
+.cache
```
NiloCK, line 29:
```suggestion
.cache

```

NiloCK:
:+1: 

This produced 
image
for me in slack, which is also good.

NiloCK (approved):

PR #1100, opened 2023-02-03
Update hyperspace endpoint

geoknee:
This was communicated over slack https://filecoinproject.slack.com/archives/C04JEJB82RY/p1675340828839479

But we now see the error:
FATAL ERROR
expected cbor type 'byte string' in input


geoknee:
In favour of #1064.

PR #1107, opened 2023-02-07
Add an interest-yielding application channel for Intermediary-Provider

NiloCK:
(forked from #1076 and rebased to include updated outcome-format)

PR adds an application channel for hubs to financialize their participation in ledger channels. Application is intended as a working prototype. It provides a simple incentive structure for intermediaries: deposit third-party security collateral in exchange for a yield on that collateral.

An application like this could potentially serve as the entire incentive mechanism for network intermediaries in a [clients]-[providers] style network, or as one piece in a more modular approach. (EG, maybe hub-client channels simultaneously scrape per-connection fees from clients). This application is less suitable for mesh network topology where most participants are payer-payees, because it's not then obvious who should collect interest. (Aside: how does a network behave where interest yield in each channel is bidirectional?)

# Application Mechanism

For a given hub and service provider (`Irene` and `Bob`), the basic flow is:

- Bob wishes to enter a marketplace as a provider of some service
- Irene wishes to service connections in this marketplace with capital and networking
- Irene agrees to finance Bob's participation in the network with `x` tokens, at an interest rate `r`
- Prefund is signed with the interest rate encoded in appData as a fraction - two `uint128`s representing the number & denominator. The total of the appData struct is:

```solidity
struct InterestAppData {
    // a per-day simple interest rate (daily percentage yield), as a fraction.
    // ie, 1% per day is represented with dpyNum = 1 and dpyDen = 100.
    //
    // Lower numbers (eg, fraction in simplest terms) produce least risk of overflow.
    uint128 dpyNum;
    uint128 dpyDen;
    // the block number of the latest principal adjustment
    uint256 blocknumber;
    // the current principal. Decreases as the serviceProvider earns via the channel.
    Funds principal;
    // the total interest collected so far. Strictly increasing.
    // The value of the intermediary's allocation can never be less than this:
    // ie, when the intermediary's collectedInterest grows to be equal to their
    // allocation, the channel is effectively spent and can be concluded.
    Funds collectedInterest;
}
```

The channel is used as a regular consensus_ledger_channel to fund & defund virtual channels between Bob and various consumers, but the application state and outcome are subject to two new update rules.

1. when Bob earns money by closing out a virtual channel:
  - his allocation in the `outcome` is adjusted
  - the `principal` is decreased by the same amount (these are no longer Irene's staked funds)
  - the blocknumbe is updated
2. As each day *ticks*, the interest rate is applied to the principal. The result is:
  - moved from `Bob` to `Irene` in the outcome
  - added to `collectedInterest`
  - the blocknumber is updated

The interest adjustments can be calculated lazily. EG, if a given channel is idle for 3 days, then Irene and Bob can update the blocknumber, outcome, and collected interest along with whatever action is going on to reboot the channel.

As part of the appData, interest rates could be adjusted inside a running channel (responding to market shifts, I guess).

## Challenge Mechanism

If Bob ghosts, or refuses to sign updates that respect his commitments, Irene can challenge with:
- one state `s` with Bob's signature
- another state which applies the interest due during the time elapsed since `s.appData.blocknumber`

The app enforces a fair interest calculation & outcome adjustment.

## Dead channel risk to Irene:
- if Bob enters the marketplace with an initial lockup of X tokens, but *never makes a sale*, the Irene will have "lost out" of the interest that she's entitled to according to the prefund agreement. She can recover her own deposit, but it will not have been effective in generating yield because there are no Bob-earnings to "claw back".
- if Bob has any sort of 'slow start' (where his earnings are under his owed interest) then he may be tempted to abandon this channel and seek a deposit from a different intermediary.

## Interest calculation: context

I originally intended a higher resolution `bPY` interest rate (block percentage yield) with compounding interest (eg, earn interest also on your earned interest). The absence of floating point numbers in the EVM makes authentic calculation of compounding interest error prone (overflows) and expensive (repeated multiplications & divisions).

I settled on simple interest calculation (earned interest itself never generates future yield) with a `dPY` (daily percentage yield). I think this cadence is reasonable in context of hub-serviceProvider channels.

## Reviewing

[x] Is it security related?

Application specific security. Not protocol security exactly, but as a base-layer channel application it warrants some scrutiny.

## Future work

- These data structures and update mechanics are not yet implemented in golang. PR contains solidity and a typescript test suite.
- This application is not part of the documentation.

---

#### Code quality

#### Project management


nitro-protocol/contracts/LedgerFinancingApp.sol
```diff
@@ -0,0 +1,151 @@
+pragma solidity 0.8.17;
+pragma experimental ABIEncoderV2;
+
+import './interfaces/IForceMoveApp.sol';
+import './libraries/NitroUtils.sol';
+import {ExitFormat as Outcome} from '@statechannels/exit-format/contracts/ExitFormat.sol';
+
+// LedgerFinancingApp is a ForceMoveApp that allows a intermediary to earn interest
+// on a deposit. It fuctions as a ConsensusApp with the following additional rule:
```
geoknee, line 9:
```suggestion
// on a deposit. It functions as a ConsensusApp with the following additional rule:
```

nitro-protocol/contracts/LedgerFinancingApp.sol
```diff
@@ -0,0 +1,151 @@
...
+        // a per-day simple interest rate (daily percentage yield), as a fraction.
+        // ie, 1% per day is represented with dpyNum = 1 and dpyDen = 100.
+        //
+        // Lower numbers (eg, fraction in simplest terms) produce least risk of overflow.
+        uint128 dpyNum;
+        uint128 dpyDen;
+        // the block number of the latest principal adjustment
+        uint256 blocknumber;
+        // the current principal. Decreases as the serviceProvider earns via the channel.
+        Funds principal;
```
geoknee, line 28: I had to look it up, but this seems like a good term to use https://www.investopedia.com/terms/p/principal.asp#toc-what-is-principal

geoknee: Should the `principal` ever change? I think we discussed this on notion... ideally application-specific constants could be stored in the IForceMoveApp itself (and passed as constructor params).

nitro-protocol/contracts/LedgerFinancingApp.sol
```diff
@@ -0,0 +1,151 @@
...
+
+                    require(claimed <= earned, 'earned<claimed');
+                }
+            }
+        }
+    }
+
+    function daysSince(uint256 blocknumber) private view returns (uint32) {
+        return uint32((block.number - blocknumber) / 7200); // 7200 == 24*60*60/12
+    }
```
geoknee, line 150: Curious about the best way to convert from block number to time.  Is using 12 second block time is potentially a bit risky? Since The Merge, Ethereum mainnet will have a 12 second blocktime much more reliably... but PoW chains as well as say Filecoin Mainnet may have different periods and its not impossible for them to vary over time.

There is also `block.timestamp` available (which may have its own risks).

(Side note -- good thing we did #1036 recently!)

I wonder if it makes the most sense to define a yield per block, rather than per unit time? Then, participants may adjust this off-chain in light of a change in blocktime (from switching networks, say). 

Of course then interest rates will be "smaller" (bigger denominators) but I don't think this makes integer arithmetic more risky in the end. In fact, a downside to how you are doing it here is to divide the number of elapsed blocks by 7200 _before_ you multiply by the interest numerator and principal amount. I think you want to do all of your multiplications up front and only divide after that? (I see some evidence you had that in mind above anyway).

nitro-protocol/contracts/LedgerFinancingApp.sol
```diff
@@ -0,0 +1,151 @@
...
+        uint128 dpyDen;
+        // the block number of the latest principal adjustment
+        uint256 blocknumber;
+        // the current principal. Decreases as the serviceProvider earns via the channel.
+        Funds principal;
+        // the total interest collected so far. Strictly increasing.
+        // The value of the intermediary's allocation can never be less than this:
+        // ie, when the intermediary's collectedInterest grows to be equal to their
+        // allocation, the channel is effectively spent and can be concluded.
+        Funds collectedInterest;
```
geoknee, line 33: I don't believe this is actually used?

nitro-protocol/contracts/LedgerFinancingApp.sol
```diff
@@ -0,0 +1,151 @@
...
+                    // combine prior balance with outstandingInterest
+                    earned += initialOutcome[j]
+                        .allocations[uint256(AllocationIndicies.intermediary)]
+                        .amount;
+                    // and compare against the claimed outcome
+                    uint256 claimed = finalOutcome[j]
+                        .allocations[uint256(AllocationIndicies.intermediary)]
+                        .amount;
+
+                    require(claimed <= earned, 'earned<claimed');
```
geoknee, line 142: Be nice to factor this into a separate function `requireOutcomeIsEarnedForAsset`?

nitro-protocol/contracts/LedgerFinancingApp.sol
```diff
@@ -0,0 +1,151 @@
...
+            );
+            require(
+                proof[0].variablePart.turnNum + 1 == candidate.variablePart.turnNum,
+                'turn(candidate) != turn(proof)+1'
+            );
+
+            Funds memory outstandingInterest = getOutstandingInterest(
+                abi.decode(proof[0].variablePart.appData, (InterestAppData))
+            );
+            requireOutcomeIsEarned(
```
geoknee, line 73: I would prefer something like `requireOutcomeAdjusted`.

nitro-protocol/test/contracts/LedgerFinancingApp/requireStateSupported.test.ts
```diff
@@ -0,0 +1,291 @@
...
+    await ledgerFinancingApp.requireStateSupported(
+      fixedPart,
+      [signedByBoth],
+      challengeWithIntermediarySignature
+    );
+  });
+
+  it('reverts if the proof block number is in the future', async () => {
+    // create a challenge where the intermediary fakes the blocknumber
+    // (in order to get a higher interest rate).
```
geoknee, line 159: Wouldn't it be in the interest of the intermediary to fake a block number in the _past_?

nitro-protocol/test/contracts/LedgerFinancingApp/requireStateSupported.test.ts
```diff
@@ -0,0 +1,291 @@
...
+      ...baseState,
+      appData: appDataABIEncode({...baseAppData, blocknumber: baseAppData.blocknumber + 1000000}),
+    };
+    const signedByIntermediary: RecoveredVariablePart = {
+      variablePart: getVariablePart(futureState),
+      signedBy: BigNumber.from(0b01).toHexString(),
+    };
+    await expectRevert(() =>
+      ledgerFinancingApp.requireStateSupported(fixedPart, [signedByBoth], signedByIntermediary)
+    );
```
geoknee, line 170: Where does this trigger a revert? I think it would be good to attach a named error / revert string to this case and assert on that.

nitro-protocol/test/contracts/LedgerFinancingApp/requireStateSupported.test.ts
```diff
@@ -0,0 +1,291 @@
...
+  });
+
+  it('rejects excessive interest calulations', async () => {
+    // construct proof+candidate test case with unfair interest calculation, assert failure.
+    // test case:
+    //  - the appData's interest rate is 1% per day
+    //  - the initial outcome is 500:500
+    //  - the chain is advanced by 1 day
+    //  - challenge outcome is 506:495. Fraud!
```
geoknee, line 179:
```suggestion
    //  - challenge outcome is 506:494. Fraud!
```

nitro-protocol/contracts/LedgerFinancingApp.sol
```diff
@@ -0,0 +1,151 @@
...
+        Funds memory outstandingInterest
+    ) private pure {
+        for (uint256 i = 0; i < outstandingInterest.asset.length; i++) {
+            address asset = outstandingInterest.asset[i];
+            uint256 earned = outstandingInterest.amount[i];
+
+            for (uint256 j = 0; j < finalOutcome.length; j++) {
+                if (finalOutcome[j].asset == asset) {
+                    require(initialOutcome[j].asset == asset, 'Asset mismatch');
+                    require(
```
geoknee, line 124: There are no checks on the rest of the outcome. Would this allow Irene to (say) burn Bob's funds?

nitro-protocol/contracts/LedgerFinancingApp.sol
```diff
@@ -0,0 +1,151 @@
...
+    // The outstanding interest is calculated based on:
+    //  - the latest consensus principal
+    //  - the channel's interest rate
+    //  - the time elapsed since the last principal adjustment
+    function getOutstandingInterest(InterestAppData memory appData)
+        private
+        view
+        returns (Funds memory)
+    {
+        uint32 numDays = daysSince(appData.blocknumber);
```
geoknee, line 92: To check my understanding: if less than 7200 blocks have passed, **no** interest is claimable by Irene?

nitro-protocol/test/contracts/LedgerFinancingApp/requireStateSupported.test.ts
```diff
@@ -0,0 +1,291 @@
...
+ * increase blocknumber on provider by 7500.
+ * each day is ~7200= blocks (24*60*60/12)
+ */
+function advanceOneDay() {
+  // note: this is a hacky way to advance so many blocks, and results
+  // in a slower test.
+  // The 'hardhat_mine' method is better, but it causes a different error
+  // in the test.
+  for (let i = 0; i < 7500; i++) {
+    provider.send('evm_mine', []);
```
geoknee, line 289: This ought to be speed up your tests https://hardhat.org/hardhat-network/docs/reference#hardhat_mine 

Possibly also useful https://hardhat.org/hardhat-network/docs/reference#evm_setnextblocktimestamp

geoknee: Apologies, just read the code comments!

nitro-protocol/test/contracts/LedgerFinancingApp/requireStateSupported.test.ts
```diff
@@ -0,0 +1,291 @@
...
+          [signedByMerchant, signedByIntermediary], // two proof states - should fail
+          signedByIntermediary
+        ),
+      '|proof| > 1'
+    );
+  });
+});
+
+/**
+ * increase blocknumber on provider by 7500.
```
geoknee, line 280: Nit: 7500 blocks is 25 hours. So this method could have a more accurate name.

nitro-protocol/test/contracts/LedgerFinancingApp/requireStateSupported.test.ts
geoknee: What's the error you see with `hardhat_mine`?

nitro-protocol/contracts/LedgerFinancingApp.sol
geoknee: OK I think I see how it works. 

**ON chain** -- there is only ever a _single_ transition allowed (from a ✔️ ✅  state to a ✅ state)
**OFF chain** -- the parties can agree to change `principal, blocknumber and Outcome` (according to the rules set out in this contract) by unanimous consensus (supporting a ✔️ ✅  state). 

In practice, is there any point in doing this off chain mechanism other than as a part of defunding the ledger channel? I suppose if `dpyDen` changed, they would want to "reset the counting" in this way.

geoknee (changes requested):
This is great 🚀 

I left some questions, particularly around block number -> time calculation. 

I think with an ADR and a bit more work we could merge this soon. We would then want to revisit it / plan around integrating it into the off chain flow.

geoknee:
> ```solidity
>   // Lower numbers (eg, fraction in simplest terms) produce least risk of overflow.
> ```

I want to unpack this some more.

According to the docs, there is only one way to get an overflow with the `/` operator https://docs.soliditylang.org/en/v0.8.17/types.html#division. Are you worried about overflowing with the `*` operator? 

We could eliminate that by hardcoding the numerator to be `1`. 

In my understanding, the main "risk" is a rounding error. The `/` operator rounds down to the nearest integer.

nitro-protocol/contracts/LedgerFinancingApp.sol
NiloCK: That's right, yes. This *can* be altered by replacing DPY with BPY (block percentage yield). Using BPY was *not* feasible when I had intended to calculate compounding interest (each elapsed block would be another multiplication / division).

Still, moving from DPY from BPY involves a bit some care, since it probably *increases* risk of overflow problems.

nitro-protocol/test/contracts/LedgerFinancingApp/requireStateSupported.test.ts
NiloCK: Alternatively I could just move it to do exactly 7200 and check that things work. 7500 was picked arbitrarily to avoid any off-by-one time accounting errors.

nitro-protocol/test/contracts/LedgerFinancingApp/requireStateSupported.test.ts
NiloCK: Something to do with broken gas estimations - I found some documentation around that issue and root causes, but nothing that related it to `hardhat_mine`.

nitro-protocol/contracts/LedgerFinancingApp.sol
```diff
@@ -0,0 +1,151 @@
...
+            return;
+        } else if (proof.length == 1) {
+            // check that proof[0] -> candidate respects the stated interest rate.
+            // Requires:
+            //  - proof state is unanimous
+            //  - candidate state immediately follows proof state (by turnNum)
+            //  - the intermediary has not taken more funds than owed according
+            //    to the interest rate agreement of the channel
+            require(
+                NitroUtils.getClaimedSignersNum(proof[0].signedBy) == 2,
```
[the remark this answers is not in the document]
NiloCK: L51 was originally also a hardcoded 2, but the linter blocked merging because `fixedPart` was unused in `requireStateSupported`.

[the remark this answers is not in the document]
geoknee: OK I think you can get around that by removing the variable name from the function signature.

geoknee:
One more thing: can we discuss the name of the contract? How about `InterestBearingLedgerChannel` or something?

nitro-protocol/contracts/LedgerFinancingApp.sol
NiloCK: This is interesting. No, this is not used in `requireStateSupported`, and so it could, in practical terms, be dropped from the appData. But it's instrumental in the correct operation of the channel, since it does the accounting on the intermediary's profits. When the Intermediary's allocation in the channel falls down to be equal to this amount, she has no future interest in the channel, so it should be closed.

If not in the appData, this information would have to be stored in a separate structure somewhere else (appMetaData). We have no infrastructure for storing runtime metadata for a specific channel.

Should we take this as a hint that we might want that?

nitro-protocol/contracts/LedgerFinancingApp.sol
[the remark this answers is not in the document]
NiloCK: I had forgotten (or never knew) that unnamed function parameters were a thing in solidity. That's better here because it decreases the mental load for understanding the function.

4a847b77

nitro-protocol/contracts/LedgerFinancingApp.sol
```diff
@@ -0,0 +1,151 @@
...
+        } else {
+            revert('|proof| > 1'); // does it pay to be this terse with revert messages?
+        }
+    }
+
+    // The outstanding interest is calculated based on:
+    //  - the latest consensus principal
+    //  - the channel's interest rate
+    //  - the time elapsed since the last principal adjustment
+    function getOutstandingInterest(InterestAppData memory appData)
```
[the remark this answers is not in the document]
NiloCK: bd969958

NiloCK:
> > ```solidity
> >   // Lower numbers (eg, fraction in simplest terms) produce least risk of overflow.
> > ```
> 
> I want to unpack this some more.
> 
> According to the docs, there is only one way to get an overflow with the `/` operator https://docs.soliditylang.org/en/v0.8.17/types.html#division. Are you worried about overflowing with the `*` operator?
> 
> We could eliminate that by hardcoding the numerator to be `1`.
> 
> In my understanding, the main "risk" is a rounding error. The `/` operator rounds down to the nearest integer.

A couple of things.
- Yes, the overflow concern was with the multiplication. days*interestrateNumerator*amount.
- having separate numerator and denominator was an accidental holdover from a previous pass where a 1% interest rate would be stored as `101/100` rather than `1/100`. Both were required there, but a denominator-only approach is fine now. (For small denominators, there is a large jump in the effective interest rate between each possible denominator - this is not an issue because even at a daily cadence any 2 digit denominator is an insanely high yield - big numbers will be the norm for practical reasons.).
- With BPY, there is still `elapsedBlocks * amount` as an overflow risk. How big might `elapsedBocks` be? Hard to say.

nitro-protocol/test/contracts/LedgerFinancingApp/requireStateSupported.test.ts
NiloCK: 76f8f179

This test case made sense when `computeOutstandingInterest` was reading the intermediary supplied block number (which was replaced with reading the block number from the chain itself).

nitro-protocol/test/contracts/LedgerFinancingApp/requireStateSupported.test.ts
NiloCK: 76f8f179

Test case is removed.

nitro-protocol/test/contracts/LedgerFinancingApp/requireStateSupported.test.ts
NiloCK: This change accidentally snuck in with 76f8f179, and it worked.

nitro-protocol/contracts/LedgerFinancingApp.sol
geoknee: I'm a bit confused by this. Why does the intermediary's allocation matter?

Let's say the principal is 100, and after some time the hub is due 3 coins of interest. Meanwhile, the provider has earned 97 coins so that the hub has 3 coins left. It seems to me that the hub does still have an interest in the channel, since if nothing else happens (just time lapses), he will start to earn more interest which should be claimable (the service provider has plenty of cash to pay the interest).

nitro-protocol/contracts/LedgerFinancingApp.sol
NiloCK: The checks here are signatures from each participant (green: serviceProvider, white: intermediary), yes?

re: off chain principal adjustment

As see it, the principal should diminish based on SP earnings in the channel. For Alice, Bob, Irene, channels `AI`, `IB`, and `XY_X := the amount in XY allocated to X`, balances change like this:

| time | IB_I | AI_I |
|------|------|------|
|   0  |  10  |   0  |
|   1  |   9  |   1  |

Here, Irene deposited 10 coins to onboard Bob to the network. Bob expects to make a profit, so he's agreed to pay a yield on the deposit.

But after `t = 1`, Irene's commitment in the channel is no longer 10. Her claimable assets have sloshed around in the network. I *think* it no longer makes good sense for Bob to be paying interest on the 1 coin belonging to Irene in `AI`. So we adjust `appData.principal` of `IB` down to nine.

From Irene's perspective, she still has ten coins "locked-in" to the network. Nine of them are still generating yield, one isn't. But the one that isn't is now explicitly "her problem" rather than "Bob's problem".

nitro-protocol/contracts/LedgerFinancingApp.sol
NiloCK: 5166b4c8

nitro-protocol/contracts/LedgerFinancingApp.sol
NiloCK: Second thought was that it's cheaper / cleaner to just check that bob hasn't been deducted "too much". Requires specific ordering on the outcome items, which isn't necessarily easy to document - thoughts on that issue recorded [here](https://www.notion.so/statechannels/Speculative-0ae5d8a85c18425d8f1919b42d3d53db?p=91b49d5459ff4fe0914485547a356eb3&pm=s). 

19b2883c

nitro-protocol/contracts/LedgerFinancingApp.sol
NiloCK: e7734ec7

nitro-protocol/test/contracts/LedgerFinancingApp/requireStateSupported.test.ts
NiloCK: This will be a non-issue with the conversion to BPY.

nitro-protocol/contracts/LedgerFinancingApp.sol
NiloCK: Yield per block implemented: [353e56b](https://github.com/statechannels/go-nitro/pull/1107/commits/353e56b4a179eb38fa9c9e7281e14eb392285789)

.adr/0010-interest-bearing-channel.md
```diff
@@ -0,0 +1,130 @@
+# 0010 - Interest Bearing Channel Application
```
geoknee, line 1: Could I ask you to:

* increase the counter to **12** (we had another couple of ADRs merged while this has been in preparation)
* cross reference in this docs and in ADR 0010 that this one supercedes that one?

🙏

NiloCK:
Closing, as the follow-up #1162 has been merged.

PR #1109, opened 2023-02-07
Fix possible flicker in query test

lalexgap:
We're seeing flickers on [CI for the query ledger channel test.](https://github.com/statechannels/go-nitro/actions/runs/4115238114/jobs/7103683401)

This is due to us asserting that the status of a ledger channel is `Proposed` after calling `CreateLedgerChannel`. However it's possible that the create ledger channel objective completes before we call the query function, since there's nothing blocking progress on the ledger channel once `CreateLedgerChannel` is called. This means that it's possible that we get get back a status of `Proposed` or `Ready` when calling query ledger channel immediately after creating it. It seems like this is more prevalent on CI, most likely due to constrained resources?

This PR updates the initial check after calling `CreateLedgerChannel` to check that the status is `Ready` **or** `Proposed` since either could be valid.

### DIFF
--- a/client_test/query_test.go
+++ b/client_test/query_test.go
@@ -33,9 +33,16 @@ func TestQueryLedgerChannel(t *testing.T) {
 	res := aliceClient.CreateLedgerChannel(irene.Address(), 0, outcome)
 	ledgerId := res.ChannelId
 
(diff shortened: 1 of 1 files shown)


kerzhner (approved):
🎉

PR #1112, opened 2023-02-07
Logging cleanup

lalexgap:
Performs some cleanup of how we log stuff using `zerologger` so the log file is easily parsible by `JQ`.
image

# Changes
- The message summary implements the `MarshalZerologObject` which means the message summaries gets nicely serialized out to a JSON entry in the log.
- The engine constructs it's own logger, so instead of passing in a logger (and getting nested entries) we now just pass the log file name into the engine.
- We now truncate the log file at the start of the RPC test
- We now call `panic` with the error we run into the engine, so the error will be visible in the console when the test fails

### DIFF
--- a/protocols/messages.go
+++ b/protocols/messages.go
@@ -211,10 +211,10 @@ type PaymentSummaries []PaymentSummary
 
 func (m MessageSummary) MarshalZerologObject(e *zerolog.Event) {
 	e.Str("To", m.To).
-		Array("PayloadSummaries", ObjectivePayloadSummaries(m.PayloadSummaries)).
-		Array("ProposalSummaries", ProposalSummaries(m.ProposalSummaries)).
-		Array("Payments", PaymentSummaries(m.Payments)).
-		Array("RejectedObjectives", RejectedIds(m.RejectedObjectives))
+		Array("PayloadSummaries", m.PayloadSummaries).
+		Array("ProposalSummaries", (m.ProposalSummaries)).
+		Array("Payments", m.Payments).
+		Array("RejectedObjectives", m.RejectedObjectives)
 
 }
 func (o ObjectivePayloadSummaries) MarshalZerologArray(a *zerolog.Array) {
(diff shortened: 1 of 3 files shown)


protocols/messages.go
```diff
@@ -201,3 +202,50 @@ func (m Message) Summarize() MessageSummary {
...
+	for _, p := range o {
+		a.Object(p)
+	}
+}
+
+func (o PaymentSummaries) MarshalZerologArray(a *zerolog.Array) {
+	for _, p := range o {
+		a.Object(p)
+	}
+}
```
kerzhner, line 236: Nit: these all look the same. Any way to factor this out?

protocols/messages.go
```diff
@@ -201,3 +202,50 @@ func (m Message) Summarize() MessageSummary {
...
+type ObjectivePayloadSummaries []ObjectivePayloadSummary
+type ProposalSummaries []ProposalSummary
+type PaymentSummaries []PaymentSummary
+
+func (m MessageSummary) MarshalZerologObject(e *zerolog.Event) {
+	e.Str("To", m.To).
+		Array("PayloadSummaries", ObjectivePayloadSummaries(m.PayloadSummaries)).
+		Array("ProposalSummaries", ProposalSummaries(m.ProposalSummaries)).
+		Array("Payments", PaymentSummaries(m.Payments)).
+		Array("RejectedObjectives", RejectedIds(m.RejectedObjectives))
```
kerzhner, line 217: I am probably missing something, but are these type casts needed?

kerzhner (approved):
lgtm!

protocols/messages.go
lalexgap: You're right, they're not needed! I think they were added by autocomplete. Removed in 182bcd8.

protocols/messages.go
lalexgap: I've used some generics to clean this up in https://github.com/statechannels/go-nitro/pull/1112/commits/33948e8764038d33c30557e44b7ba4075ab66e9a

PR #1113, opened 2023-02-07
Concurrent test

### LINKED ISSUE #1105: RPC test deadlocks when ledger channels are created concurrently
Running TestRpcClient locally on https://github.com/statechannels/go-nitro/commit/73b2fe854b29670a459f105b7c9f5f4181878a10 results in:

```
--- FAIL: TestRpcClient (10.07s)
    helpers_test.go:69: Objective ids [DirectFunding-0xee2c00258f3283d405f19fcd9e9abf0eb6ce992f5ba0dca1387075d2212414cc] failed to complete on client 0x111A00868581f73AB42FEEF67D235Ca09ca1E8db within 10s
FAIL
exit status 1
```

lalexgap:
Fixes #1105 

Previously we would wait for Irene's ledger objectives to complete by two separate calls to `waitTimeForCompletedObjectiveIds`. However `waitTimeForCompletedObjectiveIds` will greedily read any completed objectives from the client, even if it's not the objective it's waiting for. This means that the first call to `waitTimeForCompletedObjectiveIds` could consume both completed objectives, causing the second call to block forever.

To resolve this we now make one call to `waitTimeForCompletedObjectiveIds` and pass in both objective ids we're interested in. This means we'll never "drop" a completed objective id.

### DIFF
--- a/client_test/rpc_test.go
+++ b/client_test/rpc_test.go
@@ -67,9 +67,8 @@ func TestRpcClient(t *testing.T) {
 	assert.Regexp(t, "DirectFunding.0x.*", res.Id)
 
 	waitTimeForCompletedObjectiveIds(t, &clientA, defaultTimeout, res.Id)
-	waitTimeForCompletedObjectiveIds(t, &clientI, defaultTimeout, res.Id)
 	waitTimeForCompletedObjectiveIds(t, &clientB, defaultTimeout, bobResponse.Id)
-	waitTimeForCompletedObjectiveIds(t, &clientI, defaultTimeout, bobResponse.Id)
+	waitTimeForCompletedObjectiveIds(t, &clientI, defaultTimeout, res.Id, bobResponse.Id)
 
 	vRes := rpcClientA.CreateVirtual(
 		[]types.Address{irene.Address()},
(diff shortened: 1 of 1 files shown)


lalexgap:
Locally I've been able to run `go test ./client_test  -count=1000 -shuffle=on --failfast --run TestRpcClient` successfully.

kerzhner:
> Previously we would wait for Irene's ledger objectives to complete by two separate calls to `waitTimeForCompletedObjectiveIds`. However `waitTimeForCompletedObjectiveIds` will greedily read any completed objectives from the client, even if it's not the objective it's waiting for. This means that the first call to `waitTimeForCompletedObjectiveIds` could consume both completed objectives, causing the second call to block forever.

Is this the API behavior that we want?

kerzhner (approved):

lalexgap:
> > However `waitTimeForCompletedObjectiveIds` will greedily read any completed objectives from the client, even if it's not the objective it's waiting for.
> 
> Is this the API behavior that we want?

Probably not.  I think there's two options we could look at to solve this:
1. Change `waitTimeForCompletedObjectiveIds` so it polls the client for the status. That way we don't have to worry about consuming from the `CompletedObjectives` in the wrong function.
2. Implement a struct like the the [completion monitor](https://github.com/statechannels/go-nitro-testground/blob/main/utils/monitor.go) that records objective completion and can be queried for the status of an objective.

Issue #1096, opened 2023-02-03
Reconsider client IncomingObjectiveRequests


kerzhner:
I'm interested in how this "generic" method (which the RPC server wraps) relates to the other methods in this file. 

 I feel like long term, it increases the risk that the "RPC API" and the "library API" diverge (which is probably not what we want).

Longer term options that I see (not a prerequisite to merging this PR):
1.  We remove this method and have the RPC API always call into one of the existing `client` methods.
2.  We could simply remove the entire `client` package (which would make #1092 unnecessary).  Then the RPC API connects directly to `engine.ObjectiveRequestsFromAPI`. There is no longer a library API.

2 would allow us a bit more freedom to rethink our overall API from scratch... but it would break quite a lot of tests.

_Originally posted by @geoknee in https://github.com/statechannels/go-nitro/pull/1073#discussion_r1095078947_

Issue #1094, opened 2023-02-02
Strategy to avoid reinitializing objectives 


kerzhner:
This card tracks https://github.com/statechannels/go-nitro/blob/79bae0c25cf4f522f265f112af50c99b6d57a355/client/rpc/server.go#L68

PR #1117, opened 2023-02-08
RPC server refactor

kerzhner:
Fixes https://github.com/statechannels/go-nitro/issues/1096, fixes https://github.com/statechannels/go-nitro/issues/1094.

This PR:
- Factors out copy/pasted code between server request handlers.
- Removes IncomingObjectiveRequest from the client library. Now, when a server receives a jsonrpc request, the server maps that request to a client library api.

### DIFF
--- a/rpc/server.go
+++ b/rpc/server.go
@@ -8,6 +8,7 @@ import (
 	"github.com/nats-io/nats.go"
 	"github.com/rs/zerolog"
 	nitro "github.com/statechannels/go-nitro/client"
+	"github.com/statechannels/go-nitro/protocols"
 	"github.com/statechannels/go-nitro/protocols/directdefund"
 	"github.com/statechannels/go-nitro/protocols/directfund"
 	"github.com/statechannels/go-nitro/protocols/virtualdefund"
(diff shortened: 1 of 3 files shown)


rpc/server.go
```diff
@@ -39,178 +40,84 @@ func NewRpcServer(nitroClient *nitro.Client, chainId *big.Int, logger zerolog.Lo
...
-		panic(err)
-	}
-}
 
-// handleError "handles" an error by panicking
-func handleError(err error) {
 	if err != nil {
-		panic(err)
+		return err
 	}
```
geoknee, line 115: Curious why you decided to remove `handleError` after all?

rpc/server.go
```diff
@@ -39,178 +40,84 @@ func NewRpcServer(nitroClient *nitro.Client, chainId *big.Int, logger zerolog.Lo
...
-
+func subcribeToRequest[T serde.RequestPayload, U serde.ResponsePayload](rs *RpcServer, method serde.RequestMethod, processPayload func(T) U) error {
+	return rs.connection.Subscribe(method, func(data []byte) []byte {
 		rs.logger.Trace().Msgf("Rpc server received request: %+v", data)
-
-		rpcRequest := serde.JsonRpcRequest[directfund.ObjectiveRequest]{}
+		rpcRequest := serde.JsonRpcRequest[T]{}
 		err := json.Unmarshal(data, &rpcRequest)
 		if err != nil {
 			panic("could not unmarshal direct fund objective request")
```
geoknee, line 68: I think we need a more generic message here, since we haven't drilled down to the type of objective request yet? Same goes for L76.

rpc/server.go
```diff
@@ -39,178 +40,84 @@ func NewRpcServer(nitroClient *nitro.Client, chainId *big.Int, logger zerolog.Lo
...
+	}
+
 	return rs
 }
 
-// registerHandlers registers the handlers for the rpc server
-func (rs *RpcServer) registerHandlers() {
-	err := rs.connection.Subscribe(serde.DirectFundRequestMethod, func(data []byte) []byte {
-
+func subcribeToRequest[T serde.RequestPayload, U serde.ResponsePayload](rs *RpcServer, method serde.RequestMethod, processPayload func(T) U) error {
```
geoknee, line 62: Super Nit: would it be nice to place this function below `registerHandlers`? Might be a personal thing but I like to read the higher abstractions first and then drill down into the details.

geoknee (approved):

rpc/server.go
```diff
@@ -39,178 +40,84 @@ func NewRpcServer(nitroClient *nitro.Client, chainId *big.Int, logger zerolog.Lo
...
-		msg := serde.NewJsonRpcResponse(rpcRequest.Id, objRes)
-		messageData, err := json.Marshal(msg)
-		if err != nil {
-			panic("Could not marshal direct fund response message")
-		}
+}
 
-		return messageData
+// registerHandlers registers the handlers for the rpc server
+func (rs *RpcServer) registerHandlers() error {
```
geoknee, line 84: This is now a nice, well-contained place to add new "routes" from the RPC API to the "internal" API 🚀

rpc/server.go
geoknee: [c69ad90](https://github.com/statechannels/go-nitro/pull/1117/commits/c69ad9013ec041768a0400d3b16953d670d7ad09)

rpc/server.go
geoknee: [8c18607](https://github.com/statechannels/go-nitro/pull/1117/commits/8c186077b892ee14ff8e08cc829fa843dcb7f229)

rpc/server.go
kerzhner: A couple reasons:
- A `handleError` function that panics is misleading.
- There is not a unifying error strategy for the rpc package. There are panics mixed with returned errors.

In general, the golang approach is to bubble up an error to the outermost layer. And I suggest that's what we should adopt for the rpc package.

rpc/server.go
kerzhner: Good catch!

PR #1122, opened 2023-02-10
Allow for manual testground workflow runs

kerzhner:
From https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow


geoknee (approved):
Why would we want to do this? Re-running failed (automatic) runs?

kerzhner:
This will allow us to run testground workflow for branches without a pull request.

PR #1123, opened 2023-02-10
Allow RPC Server to be configured with a port, and drive entire RPC test through RPC clients

geoknee:


client_test/rpc_test.go
```diff
@@ -55,24 +55,20 @@ func TestRpcClient(t *testing.T) {
 	defer msgB.Close()
 	defer msgI.Close()
 
-	rpcServerA := rpc.NewRpcServer(&clientA, chainId, createLogger(logDestination, "alice", "server"))
-	rpcClientA, err := rpc.NewRpcClient(rpcServerA.Url(), alice.Address(), chainId, createLogger(logDestination, "alice", "client"))
-	if err != nil {
-		t.Fatal(err)
-	}
-
-	defer rpcServerA.Close()
```
geoknee, line 64: Because of the greater encapsulation In introduce, the test no longer defers this function call. I think it is OK, since the server ought to be garbage collected anyway.

client_test/rpc_test.go
```diff
@@ -83,18 +79,56 @@ func TestRpcClient(t *testing.T) {
...
+func setupNitroNodeWithRPCClient(
+	pk []byte,
+	msgPort int,
+	rpcPort int,
+	chain *chainservice.MockChainService,
+	logDestination *os.File,
+) (*rpc.RpcClient, *p2pms.P2PMessageService, func()) {
+	chainId, err := chain.GetChainId()
+	if err != nil {
+		panic(err)
```
kerzhner, line 108: Not a big deal here, but, in general, I would prefer to bubble up errors to the outermost caller when possible.

client_test/rpc_test.go
```diff
@@ -83,18 +79,56 @@ func TestRpcClient(t *testing.T) {
...
+		logDestination,
+		&engine.PermissivePolicy{},
+		nil)
+	rpcServer := rpc.NewRpcServer(
+		&node,
+		chainId,
+		createLogger(logDestination, node.Address.Hex(), "server"),
+		rpcPort)
+	rpcClient, err := rpc.NewRpcClient(rpcServer.Url(), alice.Address(), chainId, createLogger(logDestination, node.Address.Hex(), "client"))
+	if err != nil {
```
kerzhner, line 125: Same comment about errors.

client_test/rpc_test.go
```diff
@@ -51,28 +51,24 @@ func TestRpcClient(t *testing.T) {
...
-	if err != nil {
-		t.Fatal(err)
-	}
+	defer cleanupFnA()
+	defer cleanupFnB()
+	defer cleanupFnC()
 
-	defer rpcServerA.Close()
 	defer rpcClientA.Close()
+	defer rpcClientB.Close()
```
kerzhner, line 59: Are these `Close` calls needed given the cleanup functions?

kerzhner (approved):
Looks great. This is a more interesting test now.

client_test/rpc_test.go
geoknee: Good catch. No, they are not.

client_test/rpc_test.go
geoknee: I'm with you there, but was planning on hiding behind the "it's only test code" defence :-P.

client_test/rpc_test.go
geoknee: [2f3f24c](https://github.com/statechannels/go-nitro/pull/1123/commits/2f3f24c82be900d5b06d8f570be51e07a82142fa)

client_test/rpc_test.go
kerzhner: It seems like the preferred pattern is to use `t.Fatalf` to fail a test. Bubbling up an error (even in test code) allows us to follow that pattern.

Issue #1119, opened 2023-02-09
Consider converting waitTimeForCompletedObjectiveIds to pub/sub


kerzhner:
At the moment, [waitTimeForCompletedObjectiveIds](https://github.com/statechannels/go-nitro/blob/d31553177acba50693d458da9a3ef2774af30c29/client_test/helpers_test.go#L31) uses the [CompletedObjectives](https://github.com/statechannels/go-nitro/blob/d31553177acba50693d458da9a3ef2774af30c29/client/client.go#L92) channel listen for completed objectives. If more than one consumer calls  `waitTimeForCompletedObjectiveIds` with the same objective id, then only one consumer will be unblocked when the objective completes. This is the result of using golang channel producer/consumer pattern. The rest of consumers will block forever.

We should consider modifying `waitTimeForCompletedObjectiveIds` into a publish/subscribe implementation. This way consumers can listen for the objective completed notification for same objective id. We could even modify `completedObjectives` channel into a  publish/subscribe implementation.

One note is to be aware of race conditions. Golang channel producer/consumer pattern very nicely avoids race conditions. With publisher/subscriber model, a subscriber will need to start listening for notifications before a publisher posts a notification.

kerzhner:
This issue is a duplicate of https://github.com/statechannels/go-nitro/issues/1114.

Issue #1114, opened 2023-02-07
 waitTimeForCompletedObjectiveIds can consume objectives it's not interested in 


lalexgap:
> > However `waitTimeForCompletedObjectiveIds` will greedily read any completed objectives from the client, even if it's not the objective it's waiting for.
> 
> Is this the API behavior that we want?

Probably not.  I think there's two options we could look at to solve this:
1. Change `waitTimeForCompletedObjectiveIds` so it polls the client for the status. That way we don't have to worry about consuming from the `CompletedObjectives` in the wrong function.
2. Implement a struct like the the [completion monitor](https://github.com/statechannels/go-nitro-testground/blob/main/utils/monitor.go) that records objective completion and can be queried for the status of an objective.

_Originally posted by @lalexgap in https://github.com/statechannels/go-nitro/issues/1113#issuecomment-1421524655_

kerzhner:
https://github.com/statechannels/go-nitro/issues/1119 is a duplicate of this issue.

geoknee:
It would be nice if the solution also worked around the race condition between the objective completing and the `waitTimeForCompletedObjectiveIds` call being made. In other words, `waitTimeForCompletedObjectiveIds` should return immediately (and not hang forever) if the objective has already completed.

geoknee:
Also, I think there is a good case for including the solution in the production code, not just a helper function.

PR #1125, opened 2023-02-13
Fix rpc test race condition

kerzhner:
We are seeing timeouts in github ci such as https://github.com/statechannels/go-nitro/actions/runs/4146194983/jobs/7171543355 As is, it's hard to tell which test is causing the timeout. The hope is that using the go test timeout flag will give more debug information.

Follow up: the timeouts are caused by an rpc test race condition. The race condition has been fixed in [6a1b0e0](https://github.com/statechannels/go-nitro/pull/1125/commits/6a1b0e09b948d56a4bfccd068afe23bdadbbe4bb)

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -31,8 +31,7 @@ jobs:
         run: go build -v ./...
 
       - name: Test
-        timeout-minutes: 1
-        run: go test ./... -count=2 -shuffle=on
+        run: go test ./... -count=2 -shuffle=on -timeout 1m
 
       - name: Archive logs
         if: always()
--- a/rpc/client.go
+++ b/rpc/client.go
@@ -111,6 +111,7 @@ func (rc *RpcClient) subscribeToNotifications() error {
 		}
 		rc.completedObjectives <- rpcRequest.Params
 	})
+	rc.logger.Trace().Msg("Subscribed to notifications")
 	return err
 }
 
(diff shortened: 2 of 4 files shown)


.github/workflows/go.yml
```diff
@@ -31,8 +31,7 @@ jobs:
         run: go build -v ./...
 
       - name: Test
-        timeout-minutes: 1
-        run: go test ./... -count=2 -shuffle=on
+        run: go test ./... -count=2 -shuffle=on -timeout 1m
```
geoknee, line 34: 👍

rpc/client.go
```diff
@@ -128,9 +130,29 @@ func waitForRequest[T serde.RequestPayload, U serde.ResponsePayload](rc *RpcClie
...
-		if expectedObjectiveId == receivedObjectiveId {
+		isObjectiveExpected := false
+		for _, expectedObjectiveId := range expectedObjectiveId {
+			if receivedObjectiveId == expectedObjectiveId {
+				isObjectiveExpected = true
+			}
+		}
+		if !isObjectiveExpected {
+			err := fmt.Errorf("received unexpected objective completion notification for objective %v", receivedObjectiveId)
+			panic(err)
```
geoknee, line 145: Is it good to panic here?

geoknee (approved):
Well done tracking this down. To me, it underlines the importance of #1114 #1119.

rpc/client.go
kerzhner: Good call. `WaitForObjectiveCompletion` now returns an error in [a4d5507](https://github.com/statechannels/go-nitro/pull/1125/commits/a4d550714ca6ceed03347a4dfdaac84d17c92bd0)

PR #1130, opened 2023-02-22
WIP debugging gas benchmark timeouts.

geoknee:
Enforce lockfile checked in: In my understanding, this is a better method than `npm ci` for ensuring that developer's local install matches that on CI.


geoknee:
Closing in favour of #1144 and #1136

PR #1131, opened 2023-02-22
SPIKE: use an event emitter pattern for `waitTimeForCompletedObjectiveIds`

geoknee:
Towards #1114 

It's pretty similar to the completion monitor solution mentioned in that issue. 

TODO
- [x] figure out how to make this work across the RPC interface
- [ ] roll out something similar for failed objectives, and possibly vouchers too
- [ ] decide whether to loop in the `store` (to make it robust to restarts). In other words, should the cache be stored in a more durable way


geoknee:
In favour of #1170

Issue #1097, opened 2023-02-03
Introduce an alternative transport to NATS


kerzhner:
With a single transport, transport specifics can creep into what needs to be transport agnostic code.

kerzhner:
Another option is to keep the topic parameter in the Connection Request/Response interface. Websockets can:
- Ignore the topic parameter when sending a message.
- On receiving a message, the message can be unmarshaled (using a general JsonRpcRequest struct) to parse out the topic.

PR #1132, opened 2023-02-22
Websocket RPC connection

kerzhner:
Fixes https://github.com/statechannels/go-nitro/issues/1097.

This PR introduces the websocket connection implementation that adheres to the same interface as the nats connection implementation.


rpc/transport/ws/client.go
```diff
@@ -31,12 +31,17 @@ func NewWebSocketConnectionAsClient(url string) (*clientWebSocketConnection, err
 	return wsc, nil
 }
 
-func (wsc *clientWebSocketConnection) Request(id uint64, data []byte) ([]byte, error) {
+func (wsc *clientWebSocketConnection) Request(data []byte) ([]byte, error) {
 	// Any request payload type will do here since we are only interested in the id
 	responseChan := make(chan []byte, 1)
-	wsc.responseHandlers[id] = responseChan
+	unmarshaledRequest := serde.JsonRpcMessage{}
```
geoknee, line 36: It's a shame that the transport layer is coupled to the serde layer like this. I hadn't anticipated this problem when suggesting we remove `id` from the inputs to this method. 

I suppose the NATS transport didn't need to use the `id` since it has request/response "built in". Here, we're leaning heavily on the JSON-RPC `id` field to associate a reponse with a request. 

I don't want to block getting this work merged, but it would be nice to return to this issue and clean it up -- for example, not having the different role for `id` depending on transport / having a clear separation of concerns between the different layers in the rpc stack. 

Overall I'm not entirely sure how we should be organising this.

geoknee (approved):

rpc/transport/ws/client.go
kerzhner: Agreed that there is probably a better way to associate requests with responses. Filed https://github.com/statechannels/go-nitro/issues/1133 to keep track of this work.

Issue #1135, opened 2023-02-23
Investigate why hardhat refuses connections when it's ready


lalexgap:
All of a sudden we noticed that the benchmark action in our github workflows would seem to run forever. An example can be found [here](https://github.com/statechannels/go-nitro/actions/runs/4245645880/jobs/7381432709)

It looks like the root cause of this is hardhat refusing the connection when wait-on makes an http requests. It appears hardhat is ready but is refusing the http requests.

For now to get around this we just assume the hardhat server refusing a connection means it's ready. We probably want to further investigate this and make our waiting mechanism a bit more robust.

lalexgap:
Closing as this is stale

PR #1136, opened 2023-02-23
Implement primitive wait-on alternative

lalexgap:
It looks like hardhat is refusing connections from `wait-on` even when it's ready.

To get around this I've implemented a primitive alternative to `wait-on` that takes the connection refusal as a sign that the hardhat server is ready.

I've created #1135 to revisit this and implement a more robust solution.

### DIFF
--- a/nitro-protocol/gas-benchmarks/jestSetup.ts
+++ b/nitro-protocol/gas-benchmarks/jestSetup.ts
@@ -21,7 +21,7 @@ declare global {
 const logFile = './hardhat-network-output.log';
 const hardHatNetworkEndpoint = 'http://localhost:9546'; // the port should be unique
 
-jest.setTimeout(15_000); // give hardhat network a chance to get going
+jest.setTimeout(30_000); // give hardhat network a chance to get going
 if (existsSync(logFile)) truncateSync(logFile);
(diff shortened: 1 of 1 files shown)


nitro-protocol/gas-benchmarks/jestSetup.ts
```diff
@@ -21,6 +21,28 @@ declare global {
...
+  let statusCode = 0;
+
+  while (statusCode !== 200) {
+    try {
+      statusCode = (await axios.get(url)).status;
+    } catch (e) {
+      // It seems the hardhat node will refuse connections even when it is ready
+      // We take the refusal to mean that the node is ready otherwise we could be waiting for a long time
+      if (axios.isAxiosError(e) && e.code === 'ECONNREFUSED') {
+        return;
```
kerzhner, line 38: That is a very confusing behavior...

kerzhner (approved):

nitro-protocol/gas-benchmarks/jestSetup.ts
lalexgap: Agreed! I think there's something more going on here but for now this appears to work 🤷

Issue #984, opened 2022-11-08
Implement a durable store


lalexgap:
Currently we are using a in-memory store to store all our data, meaning clients cannot survive being restarted.

We should investigate and implement some kind of storage solution that allows us to persist data in a more permanent manner. 

We probably want something relatively light-weight and low configuration for now. Maybe a local SQL db?

geoknee:
Another benefit will be debugging. If we can access client's stores after a failed test run, we can check the date in their stores.

PR #1138, opened 2023-03-05
introduce crash tolerance test

geoknee:
Towards #984


client_test/crash_test.go
```diff
@@ -0,0 +1,59 @@
...
+	clientA := client.New(messageserviceA, chainA, storeA, logDestination, &engine.PermissivePolicy{}, nil)
+
+	clientB, _ := setupClient(bob.PrivateKey, chainB, broker, logDestination, 0)
+	// End Client setup
+
+	// test successful condition for setup / teadown of unused ledger channel
+	{
+		channelId := directlyFundALedgerChannel(t, clientA, clientB, types.Address{})
+
+		anotherClientA := client.New(messageserviceA, chainA, storeA, logDestination, &engine.PermissivePolicy{}, nil)
```
geoknee, line 53: Here we spin up a new client with the same messaging service, chain service and store, and use that for the rest of the test. If the test fails, it's a sign that something important exists only in memory outside of the store.

The test currently fails.

geoknee: I believe the test failed because the old client continued to consume events from the chain service, starving the new client of that information. 

Stopping the old client before starting the new one makes the test pass [1733da1](https://github.com/statechannels/go-nitro/pull/1138/commits/1733da120a161475229fefdcbe97839111d0ad62) 🎉

client/engine/engine.go
```diff
@@ -122,6 +125,10 @@ func (e *Engine) ToApi() <-chan EngineEvent {
 	return e.toApi
 }
 
+func (e *Engine) Stop() {
```
lalexgap, line 128: What do you think about `Close` instead of `Stop`? I think `Close` may be more [idiomatic](https://pkg.go.dev/io#Closer) for golang.

client/client.go
```diff
@@ -185,3 +185,8 @@ func (c *Client) GetPaymentChannel(id types.Destination) (query.PaymentChannelIn
 func (c *Client) GetLedgerChannel(id types.Destination) (query.LedgerChannelInfo, error) {
 	return query.GetLedgerChannelInfo(id, c.store)
 }
+
+// Stop stops the client's engine run loop
```
lalexgap, line 189: Nit but I don't think we should refer to the engine run loop in our API docs.

client/client.go
```diff
@@ -185,3 +185,8 @@ func (c *Client) GetPaymentChannel(id types.Destination) (query.PaymentChannelIn
...
+func (c *Client) Stop() {
+	c.engine.Stop()
```
lalexgap, line 191: Do we need to stop/cleanup  the handleEngineEvents [goroutine](https://github.com/statechannels/go-nitro/blob/95f9fd1e702310901e38cceea911483f9305ca3a/client/client.go#L68)?

lalexgap (approved):
🌋

lalexgap:
FYI this has been rebased on #1142

geoknee:
Thanks @lalexgap ! I ended up rewriting the `Stop` logic extensively, so creating a new PR #1145 addressing all of your points.

PR #1140, opened 2023-03-07
add TestDirectDefundChallenge

geoknee:
Work in progress / test driven development. 

Under this approach, there is no new API method for challenging. Nor is it triggered automatically. Instead, it happens when:

- the user requests a channel be closed
- one or more counterparties does not respond before some timeout

Side note: I don't think we have a test (or even specify the behaviour) when trying to close a ledger channel with "open" virtual channels running from it.


lalexgap:
FYI this has been rebased on #1142

lalexgap:
@geoknee would it make sense to close this for now and reopen it once we start tackling this again? The merge conflicts are starting to pile up .

PR #1142, opened 2023-03-08
Resolve linting issue

lalexgap:
This gets the linting and static check working again. 

- I've updated to the latest version of `golibp2p`. Locally I was getting errors about an out of date `quic-go` package (that is imported by `golibp2p`) when running the linter.
- Updated to the latest version of the linter and action.
- Updated to the latest version of the static checker and action, as the static checker seemed to fail after upgrading the linter.
- Update the linter to to use ` --modules-download-mode readonly` which fails if any changes to the `go.mod` are needed/made. This also seems to speed things up?

I think the root cause of this is CI probably switching to golang `1.20.0` (since that satisfies `^1.19.0`). This didn't work with the version of `golibp2p` that we were using.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -18,7 +18,7 @@ jobs:
         with:
           # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
           version: v1.51.2
-          args: --timeout 3m0s --verbose
+          args: --timeout 3m0s --verbose --modules-download-mode readonly
 
       - name: Run staticcheck # see: staticcheck.io
         uses: dominikh/staticcheck-action@v1.3.0


.github/workflows/go.yml
```diff
@@ -14,16 +14,16 @@ jobs:
...
           go-version: "^1.19.0"
       - name: Run golangci-lint
-        uses: golangci/golangci-lint-action@v3.2.0
+        uses: golangci/golangci-lint-action@v3.4.0
         with:
           # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
-          version: v1.48.0
-          args: --timeout 3m0s --verbose
+          version: v1.51.2
+          args: --timeout 3m0s --verbose --modules-download-mode readonly
```
geoknee, line 21: I wonder how this new flag interacts / compares with the later steps in this action (running `go mod tidy` and `git diff --exit-code`)?

Perhaps we don't need to do both?

geoknee (approved):
👏

.github/workflows/go.yml
lalexgap: I think you're right that the ` --modules-download-mode readonly` flag should be able to replace `go mod tidy` and `git diff`

PR #1144, opened 2023-03-09
Some CI improvements

geoknee:
This PR cherry-picks from #1130 -- keeping some useful changes which may aid future debugging sessions.

### DIFF
--- a/.github/workflows/node.yml
+++ b/.github/workflows/node.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches: [main]
   pull_request:
-    paths: ["nitro-protocol/**"]
+    paths: ["nitro-protocol/**", ".github/workflows/node.yml"]
 
 jobs:
   build:
--- a/.github/workflows/node.yml
+++ b/.github/workflows/node.yml
@@ -16,7 +16,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
       - name: Install dependencies
-        run: npm --legacy-peer-deps
+        run: npm install --legacy-peer-deps
       - name: check git tree is clean
         # This will fail the job if any previous step (re)generated a file
         # that doesn't match what you checked in (or forgot to check in)
--- a/.github/workflows/node.yml
+++ b/.github/workflows/node.yml
@@ -16,11 +16,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
       - name: Install dependencies
-        run: npm install --legacy-peer-deps
-      - name: check git tree is clean
-        # This will fail the job if any previous step (re)generated a file
-        # that doesn't match what you checked in (or forgot to check in)
-        run: git diff --exit-code
+        run: npm ci --legacy-peer-deps
       - name: Compile contracts
         run: npx hardhat compile
       - name: Run eslint (including prettier)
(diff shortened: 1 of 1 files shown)


.github/workflows/node.yml
```diff
@@ -16,12 +16,31 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
       - name: Install dependencies
-        run: npm ci --legacy-peer-deps
+        run: npm install --legacy-peer-deps
```
NiloCK, line 19: Why the replacement of `ci` with `i`? From https://stackoverflow.com/questions/52499617/what-is-the-difference-between-npm-install-and-npm-ci:

> Use `npm ci` if you need a deterministic, repeatable build. For example during continuous integration, automated jobs, etc. and when installing dependencies for the first time, instead of `npm install`.

NiloCK (approved):
LGTM, but curious about my specific comment.

.github/workflows/node.yml
geoknee: Very reasonable question given the documentation around that command. 

While `npm ci` will give a deterministic, repeatable build -- it won't necessarily give the same as `npm install`, which is what developers tend to run locally. In particular, it won't make any changes to the lockfile. 

Perhaps we should keep `npm ci`, as it is clearly the recommended command. In theory it errors out when there is a mismatch between manifest and lockfile (e.g when a dependency was added but the lockfile was not regenerated). 

That's not as strong a guarantee as "you'll get exactly the same state as running `npm install` locally", which is what I was after.

I believe what motivated my change was that I ran `npm install` on `main` and saw a change to the lockfile (which I didn't expect). This happens right now on `main` (I think the changes are inconsequential, but still).

NiloCK: I think the downside of running `npm install` (on CI) is that new publications of dependencies, which our `package.json` accept (eg, a patch bump) will trigger changes to our lockfile.

This can add some noise to the dev cycle:
- do some local work
- (some dependency bumps its patch version number (`x.y.z` -> `x.y.z+1`)
- push to GH
- CI fails, because the `x.y.z+1` dependency is not in sync with the lockfile.
- now the local changes have to either include this dependency update (unrelated) or be rebased to a separate branch which incudes them (hassle)

geoknee: As discussed today:

I think in actual fact in the scenario you mention there will _not_ necessarily be a change to the lockfile under `npm install` when a dependency bumps its patch number. Changes only happen when the lockfile dependencies fail to satisfy the `package.json` dependencies, which usually doesn't happen when a patch is released (dependencies are commonly specified using `^version` meaning "_accept_ (not require) any new minor and patch versions over version").

Something like that does happen when you run `npm update`. 

In summary, `npm ci` and `npm install && git diff --exit-code` do very similar things but the latter provides a slightly stronger guarantee (possibly only differing in very edge cases). 

Still it seems that my local lockfile is still out of sync, even with this change. I suspect this is because I am running an (old) version of node and or npm. So we might want to lock down those versions, too.

geoknee: I ended up going back to `npm ci` 😆 [0e5337e](https://github.com/statechannels/go-nitro/pull/1144/commits/0e5337efa85d5d919775a2a0deaafd9a4395a966)

PR #1145, opened 2023-03-09
Refine `client.Close()` method by having engine, services implement `Closer()`

geoknee:
In response to comments on #1138.

The commit messages should provide a pretty good guide to these changes.

The overall philosophy here is to try and have the tree of go routines spawned by a client tidied up / garbage collected when the `.Close()` method is called. This should happen synchronously. I verified this by watching goroutines come and go when debugging the crash tolerance test (where `.Close()` is called). 

I followed the design where a **sender** closes a channel used to communicate with a go routine (where possible). In many cases this enough (when we use `range` or `<-`). In other cases (when we use `select`) we need an additional channel to signal the go routine should finish.

In some other situations we are using a `context` to communicate with a 3rd party API such as go-ethereum -- here the context is cancelled. We therefore expect those 3rd party go routines to be cleaned up eventually (but cannot wait on that happening). 

---

Open questions:

1. Does this PR need more documentation (godoc, ADR, etc?)
2. Is there an easy way to unit-test these `Close()` methods?

### DIFF
--- a/client/engine/chainservice/eth_chainservice.go
+++ b/client/engine/chainservice/eth_chainservice.go
@@ -43,7 +43,7 @@ type EthChainService struct {
 	txSigner                 *bind.TransactOpts
 	out                      chan Event
 	logger                   zerolog.Logger
-	cancelSubscription       context.CancelFunc
+	quit                     chan struct{}
 }
 
 // MAX_QUERY_BLOCK_RANGE is the maximum range of blocks we query for events at once.
@@ -74,14 +74,14 @@ func NewEthChainService(chain ethChain, na *NitroAdjudicator.NitroAdjudicator,
 	context, cancel := context.WithCancel(context.Background())
 
 	// Use a buffered channel so we don't have to worry about blocking on writing to the channel.
-	ecs := EthChainService{chain, na, naAddress, caAddress, vpaAddress, txSigner, make(chan Event, 10), logger, cancel}
+	ecs := EthChainService{chain, na, naAddress, caAddress, vpaAddress, txSigner, make(chan Event, 10), logger, make(chan struct{})}
 
 	if ecs.subscriptionsSupported() {
 		logger.Printf("Notifications are supported by the chain. Using notifications to listen for events.")
-		go ecs.subscribeForLogs(context)
+		go ecs.subscribeForLogs(context, cancel)
 	} else {
 		logger.Printf("Notifications are NOT supported by the chain. Using polling to listen for events.")
-		go ecs.pollForLogs(context)
+		go ecs.pollForLogs()
 	}
 	return &ecs, nil
 }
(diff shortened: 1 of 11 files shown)


client/engine/engine.go
```diff
@@ -125,8 +125,12 @@ func (e *Engine) ToApi() <-chan EngineEvent {
 	return e.toApi
 }
 
-func (e *Engine) Stop() {
-	e.stop <- struct{}{}
+func (e *Engine) Close() error {
+	close(e.stop)
+	close(e.toApi)
```
geoknee, line 130: This will stop the `handleEngineEvents` goroutine in the client.

client/engine/chainservice/eth_chainservice.go
```diff
@@ -71,14 +72,15 @@ func NewEthChainService(chain ethChain, na *NitroAdjudicator.NitroAdjudicator,
...
 
 	// Use a buffered channel so we don't have to worry about blocking on writing to the channel.
-	ecs := EthChainService{chain, na, naAddress, caAddress, vpaAddress, txSigner, make(chan Event, 10), logger}
+	ecs := EthChainService{chain, na, naAddress, caAddress, vpaAddress, txSigner, make(chan Event, 10), logger, make(chan struct{})}
 
 	if ecs.subscriptionsSupported() {
 		logger.Printf("Notifications are supported by the chain. Using notifications to listen for events.")
-		go ecs.subscribeForLogs(context.Background())
+		context, cancel := context.WithCancel(context.Background())
+		go ecs.subscribeForLogs(context, cancel)
```
lalexgap, line 80: Is this the right way to use `context.WithCancel`?   I thought the `cancel` function is generally called by the context constructor/provider, and it triggers  [Done](https://pkg.go.dev/context#Context.Done) on the context consumers.

I think the way this is implemented now the call to `cancel` by `ecs.subscribeForLogs` will do nothing as there is nothing that actually makes use of the context.

client/engine/chainservice/eth_chainservice.go
```diff
@@ -377,7 +381,7 @@ func (ecs *EthChainService) pollForLogs(ctx context.Context) {
 
 				ecs.dispatchChainEvents(fetchedLogs)
 			}
-		case <-ctx.Done():
+		case <-ecs.quit:
```
lalexgap, line 384: Is there any benefit/downside to uses contexts to trigger closing vs the `quit` channel?

lalexgap (approved):

client/engine/chainservice/eth_chainservice.go
geoknee: I think so: using contexts is essentially asynchronous whereas closing via a channel will block until the channel is actually closed.

client/engine/chainservice/eth_chainservice.go
geoknee: The context is threaded all the way through to the `ecs.chain.SubscribeFilterLogs(ctx, query, logs)` call. Although `ecs.subscribeForLogs` never reads `ctx.Done`, I think `chain.SubscribeFilterLogs` should?

This may be more confusing than necessary. It might be clearer / better to construct the `context` _inside_ of `ecs.subscribeForLogs`?

geoknee: 4ada56158196440d656ab3c2590c9060b2813912

Issue #822, opened 2022-08-15
Add a store for vouchers


lalexgap:
We want to be able to persist vouchers so our client can function after a crash/shutdown.

We should an interface for storing/fetching vouchers. We can then add this functionality to our existing memory store.

lalexgap:
I think this is still worth doing. It will get us ready for switching to a durable store.

We should add a `VoucherStore` interface that the mem store implements. The `VoucherManager` will then have to be update to use the memstore instead of it's in memory collection of vouchers.

PR #1149, opened 2023-03-09
Implement persisent storage

lalexgap:


client_test/crash_test.go
```diff
@@ -51,15 +57,18 @@ func TestCrashTolerance(t *testing.T) {
 		channelId := directlyFundALedgerChannel(t, clientA, clientB, types.Address{})
 
 		clientA.Stop()
+		storeA.Close()
```
lalexgap, line 60: Once https://github.com/statechannels/go-nitro/pull/1145 we should update `client.Close` to call `store.Close`

client/engine/store/persiststore.go
```diff
@@ -0,0 +1,465 @@
...
+	"github.com/statechannels/go-nitro/protocols/virtualfund"
+	"github.com/statechannels/go-nitro/types"
+	"github.com/tidwall/buntdb"
+)
+
+type PersistStore struct {
+	objectives         *buntdb.DB
+	channels           *buntdb.DB
+	consensusChannels  *buntdb.DB
+	channelToObjective *buntdb.DB
```
lalexgap, line 26: A `buntdb.DB` corresponds to a key/value store and gets stored in a single file.

client/engine/store/persiststore.go
```diff
@@ -0,0 +1,474 @@
...
+	"github.com/statechannels/go-nitro/protocols"
+	"github.com/statechannels/go-nitro/protocols/directdefund"
+	"github.com/statechannels/go-nitro/protocols/directfund"
+	"github.com/statechannels/go-nitro/protocols/virtualdefund"
+	"github.com/statechannels/go-nitro/protocols/virtualfund"
+	"github.com/statechannels/go-nitro/types"
+	"github.com/tidwall/buntdb"
+)
+
+type PersistStore struct {
```
lalexgap, line 22: Is there a better name for this? `FileStore`? `BuntDbStore`?

geoknee: > * [Sync always: 247 ms TTFP](http://34.168.92.245:3000/d/5OBBeW37k/time-to-first-payment?orgId=1&from=1678478595989&to=1678478699884)
> * [Sync every second: 86.5 ms  TTFP](http://34.168.92.245:3000/d/5OBBeW37k/time-to-first-payment?orgId=1&from=1678478669430&to=1678478825842&var-runId=cg5op0gnr2ghv5ug1m8g&var-jobCount=3&var-testDuration=1m0s&var-hubs=1&var-payees=5&var-jitter=1&var-latency=10&var-payers=2&var-payeepayers=0&var-nitroVersion=v0.0.0-20230310171721-486f70744942&var-storeSyncFrequency=1)
> * [Never Sync Policy: 84.1 ms TTFP](http://34.168.92.245:3000/d/5OBBeW37k/time-to-first-payment?orgId=1&from=1678479249909&to=1678479339937&var-runId=cg5o96gnr2ghv5ug1m60&var-jobCount=3&var-testDuration=1m0s&var-hubs=1&var-payees=5&var-jitter=1&var-latency=10&var-payers=2&var-payeepayers=0&var-nitroVersion=v0.0.0-20230310171721-486f70744942&var-storeSyncFrequency=1)

This is really interesting data to have. Interesting that they only offer these three levels of sync frequency. Going every second seems to have a negligible effect on TTFP (which is great!). I wonder what the TTFP vs sync period graph would look like 🤔 . It might be that sync'ing every 10ms is also having only a small effect. I guess it will be hard for us to know without hacking around with `buntdb`?

geoknee: > Is there a better name for this? `FileStore`? `BuntDbStore`?

I don't have a strong opinion. `BuntDbStore` seems reasonable as this implementation seems fairly tightly coupled to `buntdb`.

client/client.go
```diff
@@ -188,5 +188,9 @@ func (c *Client) GetLedgerChannel(id types.Destination) (query.LedgerChannelInfo
 
 // Close stops the client from responding to any input.
 func (c *Client) Close() error {
+	err := c.store.Close()
```
geoknee, line 191: Out of scope of this PR: I just want to flag the slight discrepancy between the relationship between `client`, `store` and `engine` here versus in the organisation of our packages. 

In the latter, the `store` belongs to the `engine`. Here, it belongs to the `client`.

client/engine/store/persiststore.go
lalexgap: Yup, each database can be configured seperately, using `db.SetConfig(config)`. This means we could have a different sync policy for each DB.

lalexgap:
This [PR](https://github.com/statechannels/go-nitro/pull/1152) is related to this PR, where I look at manually triggering the data to be written to disk.

client/client.go
lalexgap: That's a good point! I think the `client` gained access to the `store` when we introduced the query API. The `client` can directly query the `store` instead of having to go through the engine to look things up. We should probably be consistent and have either the `client` or `engine` "own" the `store` and be responsible for closing it.

geoknee: Yep. Probably the `store` should be a sibling of `engine`, not a child. But likely more pain than it is worth to fix.

geoknee:
> > * Should we consider switching all our tests to use the persistent store?
> 
> Since the new `persistStore` supports both in-memory and writing to file should we deprecate the `MemStore` in favour of this new store?

Maybe....
Some reasons _not_ to:
*  I think the `MemStore` is still slightly faster than the new store, even without syncing? Probably due to lack of a need to marshal things
* Is there some value in having multiple implementations of our `store` interface? 
* We have kept the `TestMessageService` around. The `MemStore` feels similar to that. 

Could we convert all integration tests over to the new store, but retain at least one integration test using the old store?

lalexgap:
> Could we convert all integration tests over to the new store, but retain at least one integration test using the old store?

I'd be happy with that! Or we could write some kind of test helper that runs an integration test against multiple stores (and maybe multiple message services)?

geoknee:
> > Could we convert all integration tests over to the new store, but retain at least one integration test using the old store?
> 
> I'd be happy with that! Or we could write some kind of test helper that runs an integration test against multiple stores (and maybe multiple message services)?

Yes I think that makes sense. I guess that would be like matrix testing?

client/engine/store/persiststore.go
```diff
@@ -0,0 +1,446 @@
...
+	}
+
+	// Objective ownership can only be transferred if the channel is not owned by another objective
+	var prevOwner protocols.ObjectiveId
+	var isOwned bool = false
+	err = ms.channelToObjective.View(func(tx *buntdb.Tx) error {
+		res, err := tx.Get(string(obj.OwnsChannel().String()))
+		if err != nil {
+			return nil
+		}
```
geoknee, line 159: On first glance this looks like a mistake. I think it would improve readability (and safety), to do 
```go
		if err == buntdb.ErrNotFound {
			return nil
		}
```

geoknee: Generally this section is much harder to parse than the corresponding section of the `MemStore`. Some of that difficulty may be unavoidable... but still I wonder if it could be tidied up some more?

client/engine/store/persiststore.go
```diff
@@ -0,0 +1,446 @@
...
+
+	key     string // the signing key of the store's engine
+	address string // the (Ethereum) address associated to the signing key
+}
+
+func (ms *PersistStore) checkError(err error) {
+	if err != nil {
+		panic(err)
+	}
+}
```
geoknee: Nit: maybe we should call this `panicOnError` or something?

Also, we want to avoid panicking in production code and generally we should prefer handling errors. Perhaps it could be justified here since this code runs at startup?

geoknee (approved):
🌋 

There seems to be a fair bit of repeated logic (~500 lines) here compared with the memstore. Not necessary for this PR, but I wanted to float the idea of having _another_ layer of abstraction to reduce code duplication. For example, there could be a "DB" interface of just `{CreateTable, Read,Write,Delete}`, and the store is then a sort of "driver" that sits above that and coordinates calls to that interface. 

As a case in point, take `SetObjective`. This method serializes the objective, writes it, then calls a method to infer the channels stored on that objective, writes those channels and finally updates the `channelToObjective` table to indicate that the objective is now the owner of its channel. This seems to be a bit of a mix of abstract application logic (which depends on state channel specific data structures and concepts, and is likely we will want to alter but not maintain multiple versions of) and lower level CRUD type logic (which should be very stable and a thin wrapper over something like `buntdb` or any other DB we want to drop in). Interested to know your thoughts.

client_test/crash_test.go
```diff
@@ -53,13 +59,14 @@ func TestCrashTolerance(t *testing.T) {
 		clientA.Close()
 		anotherMessageserviceA := messageservice.NewTestMessageService(alice.Address(), broker, 0)
 		anotherChainA, err := chainservice.NewSimulatedBackendChainService(sim, bindings, ethAccounts[0], logDestination)
+		anotherStoreA := store.NewPersistStore(alice.PrivateKey, dataFolder)
```
geoknee: Funky idea but I wonder about mixing the persist store and the mem store in a test like this?

geoknee:
This might close #984 ? But I think probably we need to do https://github.com/statechannels/go-nitro/issues/822 too...

client/engine/store/persiststore.go
```diff
@@ -0,0 +1,474 @@
...
+	}
+	return ch, nil
+}
+
+// GetConsensusChannel returns a ConsensusChannel between the calling client and
+// the supplied counterparty, if such channel exists
+func (ps *PersistStore) GetConsensusChannel(counterparty types.Address) (channel *consensus_channel.ConsensusChannel, ok bool) {
+
+	err := ps.consensusChannels.View(func(tx *buntdb.Tx) error {
+		return tx.Ascend("", func(key, chJSON string) bool {
```
NiloCK, line 329: These nested passed functions are a bit of a stretch for my conceptual capacities, but I don't have any concrete suggestions for simplifying.

NiloCK (approved):
This is great.

On potential addition (could be a new issue instead of tack-on to this PR) is an additional DB file which just contains something like `running` after startup and `shutdown` after a successful `Close` call. This is checked at startup for a clean shut down - if not, engine could get in touch with peers for latest states / vouchers on various channels.

client_test/crash_test.go
lalexgap: Wouldn't all the data in the `memStore` get lost, so the persist store would be starting fresh?

client/engine/store/persiststore.go
lalexgap: > Also, we want to avoid panicking in production code and generally we should prefer handling errors. Perhaps it could be justified here since this code runs at startup?

One thing to note is that we could use [recover](https://gobyexample.com/recover) to capture these panics and prevent them from escaping go-nitro ([It looks like some parts of the go standard library do that](https://github.com/golang/go/blob/07a25a9f6e46e83a6335d43b0174dfb250bca960/src/fmt/print.go#L587)). 

Also for debugging purposes `panics` are quite powerful, they immediately stop things at the site of the problem and require almost no additional error handling logic. Obviously this isn't ideal for production code but I'd prefer code that `panic`s on an error instead of ignoring it (which has been the cause of some painful bugs).

I think this could be fertile ground for an ADR or RFC? Some parts of the standard library do [panic](https://github.com/golang/go/blob/2d84fbbad8e939ccf82be1b9addf26f73f39f2dc/src/bufio/scan.go#L167) so it could be worthwhile considering the use of `panic`s and basic error handling in general.

lalexgap:
> This might close #984 ? But I think probably we need to do #822 too.
> 
> We may need an integration test which exposes the intermediate state we are in (potentially losing vouchers and losing money).

Good point! I'll tackle #822 in a separate PR as this one already has a lot of conversation on it.

lalexgap:
> > > Could we convert all integration tests over to the new store, but retain at least one integration test using the old store?
> > 
> > 
> > I'd be happy with that! Or we could write some kind of test helper that runs an integration test against multiple stores (and maybe multiple message services)?
> 
> Yes I think that makes sense. I guess that would be like matrix testing?

I've introduced a simple test that spins up clients using different store implementations in https://github.com/statechannels/go-nitro/pull/1149/commits/140a3349e96cc0f306cbbd8585f246049e51669c

PR #1155, opened 2023-03-15
Durable Storage ADR

lalexgap:
This adds an ADR specifying the details of why we selected `BuntDB` to implement our persistent store.

### DIFF
--- a/.adr/0011-persistent-storage.md
+++ b/.adr/0011-persistent-storage.md
@@ -1,0 +1,61 @@
+# 0011 Implementing a Persistent Store
+
+## Status
+
+Accepted
+
+## Context
+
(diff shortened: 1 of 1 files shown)


geoknee (approved):
Thanks @lalexgap!

The overall diff includes many source code files, too. So let's make sure to pinch those out before we merge this.

lalexgap:
> Thanks @lalexgap!
> 
> The overall diff includes many source code files, too. So let's make sure to pinch those out before we merge this.

Ah good catch, I forgot to update this PR after some rebases upstream.

geoknee:
Could we hold of merging for a little while, I just want to coordinate with some other ADR PRs we have open.

.adr/0011-persistent-storage.md
```diff
@@ -0,0 +1,64 @@
+# 0011 Implementing a Persistent Store
```
geoknee, line 1:
```suggestion
# 0011 Implementing a Durable Store
```

PR #1158, opened 2023-03-16
Introduce store compatibility integration test

lalexgap:
This PR adds an integration test spins up clients using different store implementations and runs through a basic test virtual funding and voucher payment test.

It should help serve as a sanity check that our store implementations work interchangeably.


lalexgap:
Closing in favour of #1193

PR #1159, opened 2023-03-16
Add request validation to jsonrpc server

kerzhner:
This PR adds sanity checks to jsonrpc requests. The checks are:
- Is the jsonrpc version 2.0?
- Is the method a string?
- Is the id an integer?

Note that [jsonrpc spec](https://www.jsonrpc.org/specification) allows for string ids. Our jsonrpcs interface is more restrictive for clarity of code. We can implement string ids if want to adhere fully to the jsonrpc spec. https://github.com/statechannels/go-nitro/issues/1160 tracks this work.

### DIFF
--- a/rpc/error.go
+++ b/rpc/error.go
@@ -3,6 +3,7 @@ package rpc
 import "github.com/statechannels/go-nitro/rpc/serde"
 
 var parseError = serde.JsonRpcError{Code: -32700, Message: "Parse error"}
+var invalidRequestError = serde.JsonRpcError{Code: -32600, Message: "Invalid Request"}
 var methodNotFoundError = serde.JsonRpcError{Code: -32601, Message: "Method not found"}
 var unexpectedRequestUnmarshalError = serde.JsonRpcError{Code: -32010, Message: "Unexpected unmarshal error"}
 var unexpectedRequestUnmarshalError2 = serde.JsonRpcError{Code: -32009, Message: "Unexpected unmarshal error"}
--- a/rpc/server.go
+++ b/rpc/server.go
@@ -126,7 +126,7 @@ func validateRequest(requestData []byte, logger *zerolog.Logger) validationResul
 	}
 
 	// jsonrpc spec says id can be a string, number.
-	// We only support numbers.
+	// We only support numbers: https://github.com/statechannels/go-nitro/issues/1160
 	// When golang unmarshals JSON into an interface value, float64 is used for numbers.
 	requestId := request["id"]
 	if !assertType[float64](request["id"]) {
--- a/rpc/server.go
+++ b/rpc/server.go
@@ -51,13 +51,12 @@ func (rs *RpcServer) registerHandlers() error {
 			return marshalResponse(parseError, rs.logger)
 		}
 
-		requestJson := serde.JsonRpcMessage{}
-		err := json.Unmarshal(requestData, &requestJson)
-		if err != nil {
-			return marshalResponse(unexpectedRequestUnmarshalError, rs.logger)
+		validationResult := validateRequest(requestData, rs.logger)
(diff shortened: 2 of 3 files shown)


rpc/server.go
```diff
@@ -116,6 +110,64 @@ func marshalResponse(response any, log *zerolog.Logger) []byte {
...
+	var request map[string]interface{}
+	vr := validationResult{}
+	err := json.Unmarshal(requestData, &request)
+	if err != nil {
+		vr.Error = marshalResponse(unexpectedRequestUnmarshalError, logger)
+		return vr
+	}
+
+	// jsonrpc spec says id can be a string, number.
+	// We only support numbers.
```
geoknee, line 129: Maybe dump a link to #1160 here.

rpc/server.go
```diff
@@ -116,6 +110,64 @@ func marshalResponse(response any, log *zerolog.Logger) []byte {
...
+	// When golang unmarshals JSON into an interface value, float64 is used for numbers.
+	requestId := request["id"]
+	if !assertType[float64](request["id"]) {
+		vr.Error = marshalResponse(invalidRequestError, logger)
+		return vr
+	}
+
+	// is the id an integer?
+	fRequestId := requestId.(float64)
```
geoknee, line 138: Is this not repeating the work done on L132 ?

rpc/server.go
```diff
@@ -116,6 +110,64 @@ func marshalResponse(response any, log *zerolog.Logger) []byte {
...
+	}
+	vr.Method = request["method"].(string)
+
+	return vr
+}
+
+func assertType[T string | float64](i interface{}) bool {
+	if i == nil {
+		return false
+	}
```
geoknee, line 166: I'm not sure I understand this special case. What happens if you remove it?

rpc/server.go
```diff
@@ -116,6 +110,64 @@ func marshalResponse(response any, log *zerolog.Logger) []byte {
 	return responseData
 }
 
+type validationResult struct {
+	Error  []byte
```
geoknee, line 114: Don't feel strongly, but curious about bundling the error and result together like this rather than returning a tuple `(result,error)` (which I am more used to seeing)?

rpc/error.go
```diff
@@ -3,6 +3,7 @@ package rpc
 import "github.com/statechannels/go-nitro/rpc/serde"
 
 var parseError = serde.JsonRpcError{Code: -32700, Message: "Parse error"}
+var invalidRequestError = serde.JsonRpcError{Code: -32600, Message: "Invalid Request"}
```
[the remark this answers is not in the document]
kerzhner: These errors are part of the spec: https://www.jsonrpc.org/specification

rpc/server.go
kerzhner: Done in d21bbfc5.

rpc/server.go
kerzhner: Good catch, I am overthinking this part of the code base. Removed in f18992f7.

rpc/server.go
kerzhner: I ended up removing `assertType` in f18992f7

rpc/server.go
kerzhner: We could certainly return a tuple with an error. The reason the "error" is part of the struct is that it is not a golang error but a slice of bytes to be sent over to the requester.

geoknee: That makes sense!

rpc/server.go
geoknee: Did you actually remove it? Looks like you just removed the use of it, but left it there and added a unit test?

kerzhner: Good catch! Fixed in f3b6b840.

PR #1161, opened 2023-03-17
move safesync package to interal dir

### LINKED ISSUE #887: The `safesync` package could probably live outside the `store`
Probably beyond the scope of this PR but the `safesync` package could probably live outside the `store` now that it's being used beyond the `store` package.

That might prevent some annoying import cycles in the future.

_Originally posted by @lalexgap in https://github.com/statechannels/go-nitro/pull/884#discussion_r976894064_

geoknee:
Closes #887


geoknee:
Testground failing because it imports `safesync` from the old location, which this PR removes.

Seeing as I chose the `internal` directory as the new home... we can't import from testground any more. 

Options:

1. Choose a different location than `internal` (this is fine, but perhaps a missed opportunity to reduce our API surface... do we want to "support" or expose `safesync`?
2. Use a 3rd party library to do what `safesync` does . Example: https://github.com/puzpuzpuz/xsync 
3. copypaste `safesync` into testground repo 🤮 
4. place testground code inside `internal` directory of this repo?

lalexgap:
> Testground failing because it imports `safesync` from the old location, which this PR removes.
> 
> Seeing as I chose the `internal` directory as the new home... we can't import from testground any more.
> 
> Options:
> 
>     1. Choose a different location than `internal` (this is fine, but perhaps a missed opportunity to reduce our API surface... do we want to "support" or expose `safesync`?
> 
>     2. Use a 3rd party library to do what `safesync` does . Example: https://github.com/puzpuzpuz/xsync
> 
>     3. copypaste `safesync` into testground repo 🤮
> 
>     4. place testground code inside `internal` directory of this repo?

`safesync` is a pretty small(~50 LoC) wrapper around a `sync.map`, is there any chance we could just remove `safesync` and use `sync.map`s?

geoknee:
We are going to go with option 3, but rip out safesync from testground.

geoknee:
See https://github.com/statechannels/go-nitro-testground/pull/153. If that get's merged, we should be able to rerun the testground action on this PR and see it pass.

NiloCK (approved):
:+1: 

If we find ourselves updating the copy-pasted utility more than once or twice, we could also consider pulling safeSync to its own package.

PR #1162, opened 2023-03-17
Add an interest-bearing Channel Application

NiloCK:
This is a replay of #1107 against current `main`.

Suggestion: read the ADR before the implementation / test.

- [x] Is it security related?
- [x] Is it a significant change to architecture or design?


nitro-protocol/contracts/InterestBearingApp.sol
```diff
@@ -0,0 +1,150 @@
...
+//   - the lender can unilaterally transition from state n to state n+1,
+//     forcing calculated interest into the lender's Outcome allocation
+contract InterestBearingApp is IForceMoveApp {
+    struct Funds {
+        address[] asset;
+        uint256[] amount;
+    }
+
+    struct InterestAppData {
+        // the per-block interest rate, expressed as the denominator of a fraction
```
lalexgap, line 19: This might be particular to my brain but it might help to specify the numerator is `1`?

nitro-protocol/contracts/InterestBearingApp.sol
```diff
@@ -0,0 +1,150 @@
...
+        Funds principal;
+        // the total interest collected so far. Strictly increasing.
+        // The value of the lender's allocation can never be less than this:
+        // ie, when the lender's collectedInterest grows to be equal to their
+        // allocation, the channel is effectively spent and can be concluded.
+        Funds collectedInterest;
+    }
+
+    enum AllocationIndicies {
+        borrower, // intends to earn service fees up to a limit of the financier's deposit
```
lalexgap, line 37:
```suggestion
        borrower, // intends to earn service fees up to a limit of the lender's deposit
```
Should we use the term `lender` since that's what we use for the enum below?

nitro-protocol/contracts/InterestBearingApp.sol
```diff
@@ -0,0 +1,150 @@
...
+
+    function requireStateSupported(
+        FixedPart calldata,
+        RecoveredVariablePart[] calldata proof,
+        RecoveredVariablePart calldata candidate
+    ) external view override {
+        if (proof.length == 0) {
+            // unanimous consensus check
+            require(
+                NitroUtils.getClaimedSignersNum(candidate.signedBy) == 2,
```
lalexgap, line 49: Do we enforce that there are only two participants in the channel?  Or need a stronger consensus check like [ConensusApp](https://github.com/statechannels/go-nitro/blob/7490468c0a5fed0cc64c6fa1c68c30760566315a/nitro-protocol/contracts/libraries/signature-logic/Consensus.sol#L19)?

nitro-protocol/contracts/InterestBearingApp.sol
```diff
@@ -0,0 +1,150 @@
...
+                NitroUtils.getClaimedSignersNum(proof[0].signedBy) == 2,
+                '!unanimous proof state'
+            );
+            require(
+                proof[0].variablePart.turnNum + 1 == candidate.variablePart.turnNum,
+                'turn(candidate) != turn(proof)+1'
+            );
+
+            Funds memory outstandingInterest = computeOutstandingInterest(
+                abi.decode(proof[0].variablePart.appData, (InterestAppData))
```
lalexgap, line 70: Is there anything stopping the `lender` from changing the `interestPerBlockDivisor` in the `candidate`, effectively letting them charge whatever interest rate they want? I think we'd want to enforce here that the `interestPerBlockDivisor` hasn't changed from `proof->candidate`?  The `interestPerBlockDivisor` can still be changed co-operatively by the borrower and lender.

nitro-protocol/contracts/InterestBearingApp.sol
NiloCK: :+1: 32047eb4

nitro-protocol/contracts/InterestBearingApp.sol
lalexgap: I guess my concern is that it would be pretty easy for a `lender` to include another participant they control to the channel. Since the App only needs 2 participants to support a channel, they could unilateral support any channel they want.

nitro-protocol/contracts/InterestBearingApp.sol
lalexgap: Ah good point, I missed that we're using the app data from the proof `proof[0].variablePart.appData`

lalexgap (approved):
Overall looks great! I would like to resolve this [comment](https://github.com/statechannels/go-nitro/pull/1162#discussion_r1143652938) but besides that I'm happy to approve.

nitro-protocol/contracts/InterestBearingApp.sol
NiloCK: The core protocol provides protection against that class of attack. Challenges against a particular cID are vetted so that the provided FixedPart has the correct participant list. Signatures attached to a challenge state do not "get counted" unless the recovered signer is in the participant list.

Convincing myself of all this did lead me to discover [a very closely related attack](https://github.com/statechannels/go-nitro/issues/1176), where the lender (or any participant 0) just stuffs a bunch of their OWN signature into the process.

PR #1164, opened 2023-03-17
Implement a store for vouchers

### LINKED ISSUE #822: Add a store for vouchers

lalexgap:
Fix #822.

This PR introduce a new interface `payments.VoucherStore`. The `VoucherManager` now uses the `VoucherStore` to store and fetch vouchers, instead of using a `safesync.Map`. I've defined `VoucherStore` in payments to avoid an import cycle if I were to declare it in the `store` package.

Both the `MemStore` and `PersistStore` have been updated to implement `payments.VoucherStore`. The `VoucherManager` now expects a `VoucherStore` when constructed and it uses that to fetch and store voucher related information.

TODO:
- [x] ~Expand crash test to test losing vouchers.~ Created https://github.com/statechannels/go-nitro/issues/1167

### DIFF
--- a/client/engine/store/persiststore.go
+++ b/client/engine/store/persiststore.go
@@ -10,6 +10,7 @@ import (
 	"github.com/statechannels/go-nitro/channel"
 	"github.com/statechannels/go-nitro/channel/consensus_channel"
 	"github.com/statechannels/go-nitro/crypto"
+	"github.com/statechannels/go-nitro/payments"
 	"github.com/statechannels/go-nitro/protocols"
 	"github.com/statechannels/go-nitro/protocols/directdefund"
 	"github.com/statechannels/go-nitro/protocols/directfund"
@@ -24,6 +25,7 @@ type PersistStore struct {
 	channels           *buntdb.DB
 	consensusChannels  *buntdb.DB
 	channelToObjective *buntdb.DB
+	vouchers           *buntdb.DB
 
 	key     string // the signing key of the store's engine
 	address string // the (Ethereum) address associated to the signing key
@@ -45,6 +47,7 @@ func NewPersistStore(key []byte, folder string, config buntdb.Config) Store {
 	ps.channels = ps.openDB("channels", config)
 	ps.consensusChannels = ps.openDB("consensus_channels", config)
 	ps.channelToObjective = ps.openDB("channel_to_objective", config)
+	ps.vouchers = ps.openDB("vouchers", config)
 
 	return &ps
 }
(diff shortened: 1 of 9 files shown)


payments/voucher_manager.go
```diff
@@ -108,31 +114,36 @@ func (vm *VoucherManager) Receive(voucher Voucher) (*big.Int, error) {
...
-	_, ok := vm.channels.Load(channelId.String())
+	_, ok := vm.store.GetVoucherInfo(channelId)
 	return ok
 
 }
 
 // Balance returns the balance of the channel
-func (vm *VoucherManager) Balance(channelId types.Destination) (Balance, error) {
-	data, ok := vm.channels.Load(channelId.String())
+func (vm *VoucherManager) Balance(channelId types.Destination) (initial, paid, remaining *big.Int, err error) {
```
lalexgap, line 141: I decided to remove the `Balance` struct and just embed `paid/remaining` in the `VoucherInfo`. As a result this function now returns named return values.  I think it cleans things up slightly but it might be a matter of preference.

payments/vouchers.go
```diff
@@ -27,10 +28,15 @@ type Voucher struct {
 	Signature state.Signature
 }
 
-// Balance stores the remaining and paid funds in a channel.
-type Balance struct {
-	Remaining *big.Int
-	Paid      *big.Int
+// VoucherInfo contains the largest voucher we've received on a channel.
+// As well as details about the balance and who the payee/payer is.
+type VoucherInfo struct {
```
lalexgap, line 33: This was based on the `paymentStatus` we were storing in maps in the voucher manager.

client/engine/store/persiststore.go
```diff
@@ -468,3 +471,36 @@ func (ps *PersistStore) checkError(err error) {
 		panic(err)
 	}
 }
+
+func (ps *PersistStore) SetVoucherInfo(channelId types.Destination, vs payments.VoucherInfo) error {
```
NiloCK, line 475: nit: why `vs` instead of `v` or just `voucher`? (I think the greatly-abbreviated convention is more generally for receiver variables than inputs).

NiloCK (approved):
Looks good to me! One substantive suggestion on `VoucherInfo` struct that you could take, leave, or copy to new issue.

client/engine/store/persiststore.go
lalexgap: Good catch. Originally I had named `VoucherInfo` `VoucherStatus` which is where the abbreviation `vs` comes from. I'll probably rename to `v` or `vouchInfo`

lalexgap: > I think the greatly-abbreviated convention is more generally for receiver variables than inputs

Technically I think the recommended convention is that short variable names should be preferred everywhere, including [parameters](https://go.dev/talks/2014/names.slide#7). Personally I'm in favor of longer, descriptive names 🤷

lalexgap: I've renamed to `v` in https://github.com/statechannels/go-nitro/pull/1164/commits/1f9c7cbcf8e36e8cd207122b3ba6cd654c0424c2

PR #1170, opened 2023-03-20
Use a "broadcast chan" and mutex to emit "objective completed" events

### LINKED ISSUE #1114: waitTimeForCompletedObjectiveIds can consume objectives it's not interested in

geoknee:
Please see detailed commit messages. 

Closes #1114.

Replaces #1131 .


client/engine/engine.go
```diff
@@ -33,6 +34,11 @@ func (uce *ErrUnhandledChainEvent) Error() string {
 	return fmt.Sprintf("chain event %#v could not be handled by objective %#v due to: %s", uce.event, uce.objective, uce.reason)
 }
 
+type CompletedObjectives struct {
+	l sync.Locker
+	c map[protocols.ObjectiveId]chan struct{}
```
lalexgap, line 39: does `c` mean `completed`?

client/engine/engine.go
```diff
@@ -33,6 +34,11 @@ func (uce *ErrUnhandledChainEvent) Error() string {
 	return fmt.Sprintf("chain event %#v could not be handled by objective %#v due to: %s", uce.event, uce.objective, uce.reason)
 }
 
+type CompletedObjectives struct {
+	l sync.Locker
```
lalexgap, line 38: Is there a reason we use a `mutex` over a `sync.Map`?

rpc/client.go
```diff
@@ -20,12 +21,17 @@ import (
 	"github.com/statechannels/go-nitro/channel/state/outcome"
 )
 
+type CompletedObjectives struct {
```
lalexgap, line 24: Should this be private? Since it has no functions and all it's members are private, it's not super useful outside of the `rpc` package.

client/engine/engine.go
```diff
@@ -721,3 +734,12 @@ func (e *Engine) recordMessageMetrics(message protocols.Message) {
 	e.metrics.RecordQueueLength(fmt.Sprintf("msg_payload_size,sender=%s,receiver=%s", e.store.GetAddress(), message.To), totalPayloadsSize)
 	e.metrics.RecordQueueLength(fmt.Sprintf("msg_size,sender=%s,receiver=%s", e.store.GetAddress(), message.To), len(raw))
 }
+
+func (e *Engine) CompletedObjectiveChan(id protocols.ObjectiveId) <-chan struct{} {
```
lalexgap, line 738: Instead of returning an empty struct should we return something with the objective id? While not strictly necessary it could be nice for debugging purposes (and maybe developer ergonomics?)

client/engine/engine.go
geoknee: Good call. There is no good reason, I think I just ended up here circuitously and forgot about `sync.Map`!

From the godoc of the `sync` package:

> The Map type is optimized for ... when multiple goroutines read, write, and overwrite entries for disjoint sets of keys. In [this case], use of a Map may significantly reduce lock contention compared to a Go map paired with a separate Mutex or RWMutex.

client/engine/engine.go
geoknee: I'm not sure that would help a lot... for developer ergonomics:

```go
<-CompletedObjectiveChan(id)
```
versus

```go
sameId := <-CompletedObjectiveChan(id)
```

Maybe for debugging but I don't see that yet. 

One option would be to return an enum for "rejected, failed, succeeded" for each id. 

The thing with this pattern is it does not work for repeated events, so we couldn't use it easily to do "objective updated" event emission multiple times. For that we would want to use a condition variable [`sync.Cond`.](https://pkg.go.dev/sync#Cond)

client_test/helpers_test.go
```diff
@@ -34,42 +36,39 @@ const PERSIST_STORE_FOLDER = "../data/client_test"
 // If the timeout lapses and the objectives have not all completed, the parent test will be failed.
 func waitTimeForCompletedObjectiveIds(t *testing.T, client *client.Client, timeout time.Duration, ids ...protocols.ObjectiveId) {
```
geoknee, line 37: As discussed in the commit messages, it would not be difficult to entirely deprecate this function. 

While I think it is a good function, it constitutes quite some complexity and "cleverness". If we want to keep it, it would be better to push _inside_ the production code. 

Complexity in test code implies complexity for the API consumer.

rpc/client.go
```diff
@@ -130,32 +128,10 @@ func waitForRequest[T serde.RequestPayload, U serde.ResponsePayload](rc *RpcClie
...
-				done = false
-			}
-		}
-		if done {
-			return nil
-		}
-	}
-	return nil
+// ObjectiveCompleteChan returns a chan that receives an empty struct when the objective with given id is completed
+func (rc *RpcClient) ObjectiveCompleteChan(id protocols.ObjectiveId) <-chan struct{} {
```
lalexgap, line 132: I wonder if we should rename this to avoid using `chan`, which is golang specific?

.adr/0012-event-emission.md
```diff
@@ -0,0 +1,83 @@
...
+
+It is possible to make a sort-of Javascript-style event system by having the `go-nitro` Client store a slice of `chans` for each objective id. Then, the consumer can "subscribe" by calling a method which appends a `chan` to that slice and returns it. Then, when the `client` learns from the engine that an objective has copmleted, it iterates through the slice and closes each `chan`. 
+
+This approach would not be idiomatic and has the following drawbacks:
+* if the subscriber is too late it can "miss" the event
+* trying to solve that issue with a cache leads to a race condition (a new subscription call reads the cache, infers the event has not fired, and then appends to the slice of `chans` _after_ the event fires and therefore still misses it)
+
+### Condition variable
+
+This is actually the most idiomatic approach if we want to emit events more than once. For example, if we wanted to have an "objective updated' event stream. See `sync.Cond` https://pkg.go.dev/sync#Cond .
```
lalexgap, line 43: Instead of closing the channel couldn't we just leave the channel open and send empty structs, allowing us to implement "objective updated"?

One of the [links](https://blogtitle.github.io/go-advanced-concurrency-patterns-part-3-channels/) from the official documentation for `sync.Cond` says:

> [Cond](https://golang.org/pkg/sync/#Cond) is the most controversial type in the sync package. I think it’s a dangerous primitive and it is too easy to use incorrectly. I never use it because I don’t trust myself to use it right and during code review I always suggest to use other primitives instead. Even Bryan Mills (who is in the Go team and works a lot on sync primitives) got to the point of [proposing to remove it](https://github.com/golang/go/issues/21165).


which makes me slightly nervous 😟

client_test/helpers_test.go
lalexgap: I'd be in favour of deprecating this, it would force us to use our own API.

.adr/0012-event-emission.md
geoknee: Interesting about `sync.Cond`. We could add this extra context to the ADR if you like. 

We can't simply send empty structs on the channel, because they would be read by only one consumer at a time. That doesn't give us a true "broadcast" pattern -- more like a "signal" pattern. 

We can also add this to the ADR to increase clarity.

rpc/client.go
geoknee: I think it is OK to be golang specific, here, since this is the golang RPC client. 

When we come to write a JS RPC client, it would likely have a different method set.

.adr/0012-event-emission.md
lalexgap: > We can't simply send empty structs on the channel, because they would be read by only one consumer at a time. That doesn't give us a true "broadcast" pattern -- more like a "signal" pattern.

Ah right!

client_test/helpers_test.go
geoknee: #1175

PR #1172, opened 2023-03-20
Rename PersisentStore to DurableStore

lalexgap:
This updates the store to following the naming convention that we're using in the ADR; **Durable** Store instead of Persist Store.

Since the testground tests uses the store constructor, renaming the constructor means we need to [update the testground test](https://github.com/statechannels/go-nitro-testground/pull/156), otherwise the testground GH action will fail. To avoid blocking this PR I've added a `NewPersistStore` function that is just a wrapper around `NewDurableStore`, allowing the old testground test to work (since there is still a `store.NewPersistStore` function to call). 

Once  https://github.com/statechannels/go-nitro-testground/pull/156 is merged we can remove `NewPersistStore`

### DIFF
--- a/client/engine/store/durablestore.go
+++ b/client/engine/store/durablestore.go
@@ -20,7 +20,7 @@ import (
 	"github.com/tidwall/buntdb"
 )
 
-type PersistStore struct {
+type DurableStore struct {
 	objectives         *buntdb.DB
 	channels           *buntdb.DB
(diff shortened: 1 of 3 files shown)


geoknee:
This is another data point supporting merging the `go-nitro-testground` repo with this one.

geoknee (approved):

PR #1174, opened 2023-03-21
Unit tests clean up goroutines

kerzhner:
The goal of this PR is for every unit test to not leave any orphaned goroutines. An orphaned goroutine is defined as a goroutine that is waiting on a signal that will never come. Note that this PR does **not** attempt to block a unit test until all goroutines started by the unit test exit. Some goroutines spun up by a unit test persist past the unit test but are expected to exit in a finite (and hopefully small) amount of time. Not ideal, but it is not clear there is an elegant way to block a unit test on all goroutine exits.

This PR follows these rules:
- A channel is only closed when necessary. It is necessary to close a channel when there is a goroutine that is waiting for the channel output. Only channel writers should close a channel.
- goroutines are only used when asynchronicity is necessary. Otherwise, functions are used.
- "if variable == condition then work else exit" pattern must be avoided when condition can be written by another goroutine. This pattern creates a race condition that can be solved by a mutex. But mutexes are not the preferred golang way.

### DIFF
--- a/client/engine/chainservice/mock_chainservice.go
+++ b/client/engine/chainservice/mock_chainservice.go
@@ -58,6 +58,8 @@ func (mc *MockChainService) GetChainId() (*big.Int, error) {
 }
 
 func (mc *MockChainService) Close() error {
-	close(mc.txListener)
-	return mc.chain.Close()
+	if mc.txListener != nil {
+		close(mc.txListener)
+	}
+	return nil
 }
--- a/client/engine/chainservice/mock_chainservice.go
+++ b/client/engine/chainservice/mock_chainservice.go
@@ -10,9 +10,8 @@ import (
 
 // MockChainService adheres to the ChainService interface. The constructor accepts a MockChain, which allows multiple clients to share the same, in-memory chain.
 type MockChainService struct {
-	chain      *MockChain
-	txListener chan protocols.ChainTransaction // this is used to broadcast transactions that have been received
-	eventFeed  <-chan Event
+	chain     *MockChain
+	eventFeed <-chan Event
 }
 
 // NewMockChainService returns a new MockChainService.
@@ -22,20 +21,8 @@ func NewMockChainService(chain *MockChain, address common.Address) *MockChainSer
 	return &mc
 }
 
-// NewMockChainWithTransactionListener returns a new MockChainService that will send transactions to the supplied chan.
-// This lets us easily rebroadcast transactions to other MockChainServices.
-func NewMockChainWithTransactionListener(chain *MockChain, address common.Address, txListener chan protocols.ChainTransaction) *MockChainService {
-	mc := NewMockChainService(chain, address)
-	mc.txListener = txListener
-	return mc
-}
-
 // SendTransaction responds to the given tx.
 func (mc *MockChainService) SendTransaction(tx protocols.ChainTransaction) error {
-	if mc.txListener != nil {
-		mc.txListener <- tx
-	}
-
 	return mc.chain.SubmitTransaction(tx)
 }
 
@@ -58,8 +45,5 @@ func (mc *MockChainService) GetChainId() (*big.Int, error) {
 }
 
 func (mc *MockChainService) Close() error {
-	if mc.txListener != nil {
-		close(mc.txListener)
-	}
 	return nil
 }
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -125,12 +125,11 @@ func (e *Engine) ToApi() <-chan EngineEvent {
 }
 
 func (e *Engine) Close() error {
-	close(e.stop)
-	close(e.toApi)
-	err := e.msg.Close()
-	if err != nil {
+	if err := e.msg.Close(); err != nil {
 		return err
 	}
+	close(e.stop)
+	close(e.toApi)
 	return e.chain.Close()
 }
 
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -128,7 +128,7 @@ func (e *Engine) Close() error {
 	if err := e.msg.Close(); err != nil {
 		return err
 	}
-	close(e.stop)
+	e.stop <- struct{}{}
 	close(e.toApi)
 	return e.chain.Close()
 }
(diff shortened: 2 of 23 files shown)


client/engine/engine.go
```diff
@@ -126,11 +126,12 @@ func (e *Engine) ToApi() <-chan EngineEvent {
 }
 
 func (e *Engine) Close() error {
+	if err := e.msg.Close(); err != nil {
```
lalexgap, line 129: Nit but it might be nice to document why we're closing things in a particular order.

client_test/helpers_test.go
```diff
@@ -26,7 +26,9 @@ import (
 )
 
 const TEST_CHAIN_ID = 1337
-const defaultTimeout = 10 * time.Second
+
+// TODO: change me back
```
lalexgap, line 30: Change back?

lalexgap (approved):

lalexgap:
> * A channel is only closed when necessary. It is necessary to close a channel when there is a goroutine that is waiting for the channel output. Only channel writers should close a channel.
> 
> * goroutines are only used when asynchronicity is necessary. Otherwise, functions are used.
> 
> * "if variable == condition then work else exit" pattern must be avoided when condition can be written by another goroutine. This pattern creates a race condition that can be solved by a mutex. But mutexes are not the preferred golang way.

Should we write these down somewhere?

geoknee:
> * A channel is only closed when necessary

I wonder about this. If we instead stuck to the principle of closing a channel _whenever no more data will be sent_, then this would be future proof against code edits that spawn additional goroutines consuming that channel. 

On the other hand, we could have a principle where, if you are using the `go` keyword, the onus is on you to ensure that it will be cleaned up appropriately (and therefore at this point you may need to go ahead and close an existing channel, say).

client/engine/engine.go
```diff
@@ -128,9 +128,11 @@ func (e *Engine) ToApi() <-chan EngineEvent {
...
-	e.chain.Close()
-	return nil
+	err := e.msg.Close()
+	if err != nil {
+		return err
+	}
+	return e.chain.Close()
 }
 
 // Run kicks of an infinite loop that waits for communications on the supplied channels, and handles them accordingly
```
geoknee, line 137: Nit: we could amend this godoc to describe the conditions under which the loop will exit.

client/engine/chainservice/mock_chainservice.go
```diff
@@ -58,6 +58,8 @@ func (mc *MockChainService) GetChainId() (*big.Int, error) {
 }
 
 func (mc *MockChainService) Close() error {
-	close(mc.txListener)
-	return mc.chain.Close()
+	if mc.txListener != nil {
+		close(mc.txListener)
+	}
```
geoknee, line 63: This jumps out a bit -- why is it necessary?

client/engine/messageservice/test-messageservice.go
```diff
@@ -115,7 +115,6 @@ func (tms TestMessageService) routeFromPeers() {
 // Close stops the TestMessagerService from sending or receiving messages.
 func (tms TestMessageService) Close() error {
 	close(tms.fromPeers)
-	close(tms.out)
```
geoknee, line 118: I'm not sure I agree that 

> There is no goroutine waiting for the output of the channel, so
closing the channel is not necessary.

The engine run loop is one such goroutine. It does alias the name of the channel, so that may explain the confusion. 

That goroutine is likely being cleaned up by another mechanism. But it's not unlikely that we refactor things such that there is a dedicated goroutine reading from this channel.

geoknee (approved):

client/engine/engine.go
kerzhner: I am actually not convinced that this order is the correct order. This was one of the early, experimental changes when I knew less about how an engine is shut down.

kerzhner:
> > * A channel is only closed when necessary
> 
> I wonder about this. If we instead stuck to the principle of closing a channel _whenever no more data will be sent_, then this would be future proof against code edits that spawn additional goroutines consuming that channel.
> 
> On the other hand, we could have a principle where, if you are using the `go` keyword, the onus is on you to ensure that it will be cleaned up appropriately (and therefore at this point you may need to go ahead and close an existing channel, say).

Both approaches make sense to me. If we choose to always close a channel, whenever a `select` statement is used to consume from a channel, the select statement must either be ready to handle a null value or use a 2 value form of receive. Currently, our code base does not do either.

client/engine/messageservice/test-messageservice.go
kerzhner: I am open to adding this back with the caveat of https://github.com/statechannels/go-nitro/pull/1174#issuecomment-1481420768. We should settle on a pattern and stick with it. In this PR I recommend one pattern (only close channels when necessary), but I am open to alternative patterns.

kerzhner:
> In this case I think we should follow @lalexgap's idea of sending an empty struct instead of closing the chan. That would make the `.Close()` call block until the aforementioned risk in the `select` statement has passed. The `stop` chan would also need to be unbuffered (it already is).

I think that's a good idea and an easy change: cf349771

client_test/helpers_test.go
kerzhner: Good catch: 44711242

client/engine/chainservice/mock_chainservice.go
kerzhner: Removed in e917c11c

client/engine/engine.go
kerzhner: Good idea. Done in a99488d4

kerzhner:
> Should we write these down somewhere?

That's a good idea once https://github.com/statechannels/go-nitro/pull/1174#issuecomment-1481420768 is settled.

geoknee:
> > Should we write these down somewhere?
> 
> That's a good idea once [#1174 (comment)](https://github.com/statechannels/go-nitro/pull/1174#issuecomment-1481420768) is settled.

Perhaps let's discuss this during standup today, and try to reach a consensus.

PR #1177, opened 2023-03-22
Only use TCP transport in the p2p message service

lalexgap:
Previously we were using `DefaultTransports` when constructing our lib p2p message service. The default transports includes a [couple of transports](https://github.com/libp2p/go-libp2p/blob/950151e861320426a62757121e95ad264ec39721/defaults.go#L44), however our [message service is only listening on TCP](https://github.com/statechannels/go-nitro/blob/4aa1c6887094c41dfab42de3470074f530ee6e08/client/engine/messageservice/p2p-message-service/service.go#L87), so we're only ever use the TCP transport.

The quic transport seems to be spinning up a go-routine, so by removing the QUIC transport we don't have another go-routine in our logs.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -188,9 +188,8 @@ func (c *Client) GetLedgerChannel(id types.Destination) (query.LedgerChannelInfo
 
 // Close stops the client from responding to any input.
 func (c *Client) Close() error {
-	err := c.store.Close()
-	if err != nil {
+	if err := c.engine.Close(); err != nil {
 		return err
 	}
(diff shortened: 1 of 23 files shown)


kerzhner (approved):

kerzhner:
There were a bunch of merge conflicts, so I recreated the work in [ccd5b0e](https://github.com/statechannels/go-nitro/pull/1174/commits/ccd5b0ead784dee8a80066ec7434bac5223089ea). Thanks for putting this PR together to guide me!

PR #1182, opened 2023-03-23
WIP: Fix testground stalling on CI

lalexgap:
Address the testground test occasionally stalling out by adding logic to make multiple attempts to resubscribe for events.

This should gracefully handle the timeout error we see when trying to use a sub


lalexgap:
Closing for now. Things seem stable after the upgrade 🤞

PR #1186, opened 2023-03-23
Seed go nitro random numbers using current time

kerzhner:
https://gobyexample.com/random-numbers: "The default number generator is deterministic, so it’ll produce the same sequence of numbers each time by default. To produce varying sequences, give it a seed that changes."

### DIFF
--- a/client/engine/chainservice/adjudicator/challenge_test.go
+++ b/client/engine/chainservice/adjudicator/challenge_test.go
@@ -104,7 +104,7 @@ func runChallengeWithTurnNum(t *testing.T, turnNum uint64, pc preparedChain) {
 	}
 
 	// Use random nonces so we can run this test multiple times against the same chain
-	nonce := uint64(rand.GetRandGenerator().Int63())
+	nonce := uint64(rand.Int63())
 	var s = state.State{
 		Participants: []types.Address{
 			Actors.Alice.Address,
--- a/client/engine/chainservice/adjudicator/challenge_test.go
+++ b/client/engine/chainservice/adjudicator/challenge_test.go
@@ -104,7 +104,7 @@ func runChallengeWithTurnNum(t *testing.T, turnNum uint64, pc preparedChain) {
 	}
 
 	// Use random nonces so we can run this test multiple times against the same chain
-	nonce := uint64(rand.Int63())
+	nonce := rand.Uint64()
 	var s = state.State{
 		Participants: []types.Address{
 			Actors.Alice.Address,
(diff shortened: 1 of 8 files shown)


rand/rand.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"math/rand"
+	"time"
+)
+
+var r *rand.Rand
+
+// GetRandGenerator seeds a random number generator based on current time
+func GetRandGenerator() *rand.Rand {
```
lalexgap, line 12: Instead of exposing a `GetRandGenerator` function, we could just expose a `Uint64()` function instead, since that's all we use (and probably will use in the foreseeable future)

rand/rand.go
```diff
@@ -0,0 +1,19 @@
...
+	if r != nil {
+		return r
+	}
+	source := rand.NewSource(time.Now().UnixNano())
```
lalexgap, line 16: Could be nice to add a small commen like:
```
// The default number generator is deterministic
// so we need to use an unique seed to avoid generating the same numbers
// See https://gobyexample.com/random-numbers
```

client/engine/chainservice/adjudicator/challenge_test.go
```diff
@@ -104,7 +104,7 @@ func runChallengeWithTurnNum(t *testing.T, turnNum uint64, pc preparedChain) {
 	}
 
 	// Use random nonces so we can run this test multiple times against the same chain
-	nonce := uint64(rand.NewSource(time.Now().Unix()).Int63())
+	nonce := uint64(rand.GetRandGenerator().Int63())
```
lalexgap, line 107: This is outside the scope of your changes but instead of using `Int63()` and casting to a `uint64` we can just call `Uint64()`

lalexgap (approved):
🌋 👍

rand/rand.go
lalexgap: Upon closer inspection it looks like we do use some other rand functions too, so maybe not worth the effort 🤷

kerzhner: Not a bad idea for now: 3309d66a

If we expand the variety of rand functions used, we should probably revert this approach.

rand/rand.go
kerzhner: 3309d66a

client/engine/chainservice/adjudicator/challenge_test.go
kerzhner: I was wondering if the reason we are using the current approach is to fit into some evm constraint. Sounds like maybe that's not the case...

kerzhner: 264d6f7c

Issue #1183, opened 2023-03-23
Consider On-chain versus off-chain security checks.


lalexgap:
The [discussion](https://github.com/statechannels/go-nitro/pull/1162#discussion_r1142739377) started around the  `InterestBearingApp` contract which checks a state is [supported by 2 participants](https://github.com/statechannels/go-nitro/blob/725a4bde1380c775c980b1295b1f70f599a7c313/nitro-protocol/contracts/InterestBearingApp.sol#L50), but never enforces that the channel has only two participants. Instead we rely on off-chain behavior of our client to never use the interest bearing app for a 3 participant channel (since it could be dangerous to do so).

This raised the broader question of when we should check things on-chain versus off-chain. Should we push as many checks off-chain to reduce gas usage? Or put more checks on-chain to rely less on the security guarantees of our go-nitro client.

One piece of the puzzle could be writing a  pure/readonly `validateFixedPart` function for our App contract.  This could be run on-chain when the contract is called, but could also be called off-chain (using a local evm) to peform validation checks in our go-nitro client.

PR #1192, opened 2023-03-28
[InterestBearingApp] require full consensus on supplied states

NiloCK:
Per discussions, the prior implementation put unnecessary vigilance requirements on wallet implementations. Here's the attack:

- Alice, Bob agree to open an interest-bearing channel
- Alice crafts the fixedpart / opening state, which places her and Bob in the appropriate places, but also sneaks an Alice alias into the participant list
- Bob inspects the supplied channel opening state, sees that he and Alice are where they should be, and consents to open / fund the channel.

Now, Alice holds keys for two channel participants, which Bob didn't notice. Under the prior implementation, Alice could challenge with a state that included signatures from her and her alias (because it passes the 2-signers check).

This is an opportunistic attack, because it relies on some degree of negligence in Bob's wallet, which should have checked that he and Alice were the only participants.

This PR's fix moves this responsibility from implementing wallets (which could be poorly written or have bugs introduced over time) to the more durable and easily audited solidity contract.

### DIFF
--- a/nitro-protocol/contracts/InterestBearingApp.sol
+++ b/nitro-protocol/contracts/InterestBearingApp.sol
@@ -3,6 +3,7 @@ pragma experimental ABIEncoderV2;
 
 import './interfaces/IForceMoveApp.sol';
 import './libraries/NitroUtils.sol';
+import './libraries/signature-logic/Consensus.sol';
 import {ExitFormat as Outcome} from '@statechannels/exit-format/contracts/ExitFormat.sol';
 
 // InterestBearingApp is a ForceMoveApp that allows a lender to earn interest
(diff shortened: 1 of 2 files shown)


lalexgap:
I think this PR resolves https://github.com/statechannels/go-nitro/issues/1183 ?

NiloCK:
It fixes the motivating specific instance of #1183 but not the broader question. I have a running [notion doc](https://www.notion.so/statechannels/Chapp-Policies-Metadata-91b49d5459ff4fe0914485547a356eb3?pvs=4#e46e845582314c088c943a7cbbef6d18) on the broader question as well.

lalexgap (approved):

PR #1193, opened 2023-03-28
Remove repetitive integration tests

lalexgap:
Implements a new integration test that accepts a configurable test case. The goal of this integration test is to replace the existing collection of repetitive test with a more comprehensive test.

I've introduced a new `TestCase` struct that specifies the various parameters for a run of an integration test:
```
// TestCase is a test case for the client integration test.
type TestCase struct {
	Description    string
	Chain          ChainType
	MessageService MessageService
	NumOfChannels  uint
	NumOfPayments  uint
	MessageDelay   time.Duration
	LogName        string
	NumOfHops      uint
	Participants   []TestParticipant
}

```

**NOTE:** This is based on https://github.com/statechannels/go-nitro/pull/1201  and will fail until that PR (and the accompany testground PR) is merged in.

I've created some issues for remaining work or issues to solve around this new integration test:
- https://github.com/statechannels/go-nitro/issues/1204
- https://github.com/statechannels/go-nitro/issues/1203
- https://github.com/statechannels/go-nitro/issues/1202

PR #1194, opened 2023-03-29
Stop generating ecdsa.PrivKey in P2P Message Service

lalexgap:
Previously we were generating a ECDSA private key for the p2p message service by hashing our private key and then using it a seed for a random source when calling `GenerateECDSAKeyPair`. This private key used to derive the Id of the message service. This was awkward and  made it difficult to precompute the IDs of message services,  since each call will use a different random number to generate the key.

To address this I've removed the key generation code from the P2P message service, it now just accepts an `*p2pcrypto.PrivKey` allowing us to generate the private key however we want **outside** of the message service. I've also added an `generateMessageKey` function that can be used to deterministically generate a message key for a given `pk`, that we can use in testing.  This is helpful if for some reason we want to know the ID of a message service in advance of the constructor (which would be very helpful in https://github.com/statechannels/go-nitro/pull/1193)

**Note**: Since this changes the message service constructor the testground test needs to be [updated](https://github.com/statechannels/go-nitro-testground/pull/157).  Once both PRs are approved I can merge them in together, until them the testground GH action will fail on this PR.

### DIFF
--- a/client_test/helpers_test.go
+++ b/client_test/helpers_test.go
@@ -1,6 +1,7 @@
 package client_test
 
 import (
+	"bytes"
 	"fmt"
 	"io"
 	"log"
@@ -12,6 +13,7 @@ import (
 	"time"
 
 	"github.com/google/go-cmp/cmp"
+	p2pcrypto "github.com/libp2p/go-libp2p/core/crypto"
 	"github.com/statechannels/go-nitro/channel/state/outcome"
 	"github.com/statechannels/go-nitro/client"
 	"github.com/statechannels/go-nitro/client/engine"
@@ -228,4 +230,23 @@ func closeSimulatedChain(t *testing.T, chain chainservice.SimulatedChain) {
 	if err := chain.Close(); err != nil {
 		t.Fatal(err)
 	}
+}
+
+// generateMessageKey generates a ECDSA private key deterministically using the given pk bytes.
+func generateMessageKey(pk []byte) p2pcrypto.PrivKey {
+
+	// We use he given
+	totalSize := 256
+	large := make([]byte, totalSize)
+	copy(large, pk)
+
+	messageKey, _, err := p2pcrypto.GenerateECDSAKeyPair(bytes.NewReader(large))
+	if err != nil {
+		panic(err)
+	}
+
+	if err != nil {
+		panic(err)
+	}
+	return messageKey
 }
--- a/client_test/helpers_test.go
+++ b/client_test/helpers_test.go
@@ -233,7 +233,7 @@ func closeSimulatedChain(t *testing.T, chain chainservice.SimulatedChain) {
 }
 
 // generateMessageKey generates a ECDSA private key using the given pk bytes, and is idempotent.
-func generateMessageKey(pk []byte) p2pcrypto.PrivKey {
+func generateMessageKey(t *testing.T, pk []byte) p2pcrypto.PrivKey {
 
 	// GenerateECDSAKeyPair expects a source to read random bytes from.
 	// We don't know exactly how many random bytes it will read, but we know it's under 256.
@@ -245,7 +245,7 @@ func generateMessageKey(pk []byte) p2pcrypto.PrivKey {
 
 	messageKey, _, err := p2pcrypto.GenerateECDSAKeyPair(bytes.NewReader(large))
 	if err != nil {
-		panic(err)
+		t.Fatal(err)
 	}
 
 	return messageKey
--- a/client_test/helpers_test.go
+++ b/client_test/helpers_test.go
@@ -232,10 +232,13 @@ func closeSimulatedChain(t *testing.T, chain chainservice.SimulatedChain) {
 	}
 }
 
-// generateMessageKey generates a ECDSA private key deterministically using the given pk bytes.
+// generateMessageKey generates a ECDSA private key using the given pk bytes, and is idempotent.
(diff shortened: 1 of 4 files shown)


lalexgap:
> Why does the messaging service need an ecdsa private key? From what I can tell, this is just used to generate an ID. It seems that we can generate an ID from a _public_ key or just some bytes, too. Should we take the opportunity to remove the use of ecdsa private keys from our codebase? Their screams "audit me" (unnecessarily, here, it would seem).

The underlying libp2p library requires providing an ECDSA private key, to generate an identity (and I assume encrypt all traffic) and if we don't specify one we get a [randomly generated one](https://github.com/libp2p/go-libp2p/blob/d5a280e6f6ca48eb99ba9140707c4fad54207c76/libp2p.go#L45). The only thing it accepts is[ a private key](https://github.com/libp2p/go-libp2p/blob/8c466ab7f8d0e1eb12ea8a23933e77d25190a81a/options.go#L234).

We could just let the p2p library generate a random ID, but I think that may have some pain points. That means every time you start up the message service you would be generating a brand new random ID. If go-nitro crashes and you restart, your message service is now listening on [a new address with a new ID](https://github.com/statechannels/go-nitro/blob/feb4d5c696e9c81abd6691804a2c1f75304ff809/client/engine/messageservice/p2p-message-service/service.go#L59). This means that all other participants now have out of date peer information and you would need to update them with your new peer info.

By allowing the `messageKey` to be passed in we can keep using the same ID/address for messages. This also lets us do some nice things in tests like generate a slice of peers before we create message services.

client_test/helpers_test.go
```diff
@@ -229,3 +231,22 @@ func closeSimulatedChain(t *testing.T, chain chainservice.SimulatedChain) {
 		t.Fatal(err)
 	}
 }
+
+// generateMessageKey generates a ECDSA private key deterministically using the given pk bytes.
+func generateMessageKey(pk []byte) p2pcrypto.PrivKey {
+
```
kerzhner, line 237:
```suggestion
```

client_test/helpers_test.go
```diff
@@ -229,3 +231,22 @@ func closeSimulatedChain(t *testing.T, chain chainservice.SimulatedChain) {
...
+	copy(large, pk)
+
+	messageKey, _, err := p2pcrypto.GenerateECDSAKeyPair(bytes.NewReader(large))
+	if err != nil {
+		panic(err)
+	}
+
+	if err != nil {
+		panic(err)
+	}
```
kerzhner, line 250:
```suggestion
```

client_test/helpers_test.go
```diff
@@ -229,3 +231,22 @@ func closeSimulatedChain(t *testing.T, chain chainservice.SimulatedChain) {
...
+func generateMessageKey(pk []byte) p2pcrypto.PrivKey {
+
+	// We use he given
+	totalSize := 256
+	large := make([]byte, totalSize)
+	copy(large, pk)
+
+	messageKey, _, err := p2pcrypto.GenerateECDSAKeyPair(bytes.NewReader(large))
+	if err != nil {
+		panic(err)
```
kerzhner, line 245: Should we return the error?

client_test/helpers_test.go
```diff
@@ -229,3 +231,22 @@ func closeSimulatedChain(t *testing.T, chain chainservice.SimulatedChain) {
 		t.Fatal(err)
 	}
 }
+
+// generateMessageKey generates a ECDSA private key deterministically using the given pk bytes.
+func generateMessageKey(pk []byte) p2pcrypto.PrivKey {
+
+	// We use he given
+	totalSize := 256
+	large := make([]byte, totalSize)
```
kerzhner, line 240: Curious, why is it necessary to create a 256 byte array to create a byte reader? Does `GenerateECDSAKeyPair` only accept 256 byte array readers?

kerzhner (approved):

client_test/helpers_test.go
lalexgap: Since this is purely a test helper I think it's OK to `panic`?

lalexgap: Alternatively it could accept a `t *testing.T` and call `t.Fatal`

client_test/helpers_test.go
lalexgap: Sorry, for some reason I thought something like this would be caught by the linter 🤦

client_test/helpers_test.go
kerzhner: Makes sense, thanks for the explanation.

client_test/helpers_test.go
kerzhner: I like the second alternative!

client_test/helpers_test.go
lalexgap: It looks like there's an open issue for [gofmt](https://github.com/golang/go/issues/6996) but that led me to this [tool](https://github.com/mvdan/gofumpt)

client_test/helpers_test.go
lalexgap: Updated with a more thorough comment in https://github.com/statechannels/go-nitro/pull/1194/commits/2d0e31b528c6a3a37c78c09e68ddb0503ea61f69

client_test/helpers_test.go
lalexgap: Updated in https://github.com/statechannels/go-nitro/pull/1194/commits/be5ea3376db8a084983debe20dccf5b9ebeeac69

PR #1198, opened 2023-03-29
Switch from gofmt to gofumpt

lalexgap:
Switch from `gofmt` to `gofumpt` which is a fork of `gofmt`:
>Enforce a stricter format than gofmt, while being backwards compatible. That is, gofumpt is happy with a subset of the formats that gofmt is happy with.
>
>The tool is a fork of gofmt as of Go 1.19, and requires Go 1.19 or later. It can be used as a drop-in replacement to format your Go code, and running gofmt after gofumpt should produce no changes. For example:

There added rules can be found [here](https://github.com/mvdan/gofumpt#added-rules). 

One of the benefits of switching to `gofumpt` is that it enforces consistent behavior around trailing empty lines; something that I mess up often 🙈


kerzhner:
Are there concerns using a fork of gofmt? [The last realease](https://github.com/mvdan/gofumpt/releases) is from 9/2022. [gofmt seems to get many more updates](https://pkg.go.dev/cmd/gofmt?tab=versions)

lalexgap:
> Are there concerns using a fork of `gofmt`? [The last realease](https://github.com/mvdan/gofumpt/releases) is from 9/2022. [`gofmt` seems to get many more updates](https://pkg.go.dev/cmd/gofmt?tab=versions)

I think since `gofmt` is part of the standard library it gets built for every version of golang (that's why the version page lists every go release), so it might not be an apple to apple comparison when comparing against releases of `gofumpt`.

That said, I take your point! We'd be switching from a standard tool built into the language to a third party tool maintained by some developer, which introduces some risk.

I personally think it's worth switching for two reasons:
-  It catches a whole bunch of spacing errors that take up bandwidth in PRs
- Switching back to `gofmt` is pretty easy (we'd just need to revert https://github.com/statechannels/go-nitro/pull/1198/commits/a531112cb649699e454ee355db6cd8cc5f57999d) so if we do run into issues we're not committed to `gofumpt`

That said I don't feel super strongly about it and I'd be happy to just run `gofumpt` locally if we want to stick with `gofmt`.

kerzhner (approved):
Let's give `gofumpt` a try

PR #1200, opened 2023-03-30
Nitro RPC Server

geoknee:
This PR allows go-nitro to run as an RPC server executable.

# Running Nitro RPC Server

```shell
go run .
```

There are also a couple few CLI arguments that let you select your ports and the `pk` to use:
```
❯ ./nitro-rpc-server --help
Usage of ./nitro-rpc-server:
  -hardhatUrl string
    	Specifies the url for the hardhat node. (default "ws://127.0.0.1:8545")
  -msgPort int
    	Specifies the tcp port for the  message service. (default 3005)
  -pk string
    	Specifies the private key for the client. Default is Alice's private key. (default "2d999770f7b5d49b694080f987b82bbc9fc9ac2b4dcc10b0f8aba7d700f69c6d")
  -rpcPort int
    	Specifies the tcp port for the rpc server. (default 4005)
  -useDurableStore
    	Specifies whether to use a durable store or an in-memory store.
  -useWS
    	Specifies whether to use websockets or NATS for the rpc server.
```
You can also build a binary. Building a binary will embed the git commit hash into the binary, allowing for the version RPC call to return the commit hash.
```
go build  -o nitro-rpc-server
```
# Interacting with Nitro RPC Server
And then in a new terminal

```shell
wscat -c 127.0.0.1:4005 
```

Then

```
> {"jsonrpc":"2.0","id":1,"method":"version","params":{}}
< {"jsonrpc":"2.0","id":1,"result":"26df816b4729512b5f1c3e3b9b8e2e6c37e4f1de","error":null}
```

---

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -4,6 +4,7 @@ package client // import "github.com/statechannels/go-nitro/client"
 import (
 	"io"
 	"math/big"
+	"runtime/debug"
 
 	"github.com/statechannels/go-nitro/channel/state/outcome"
 	"github.com/statechannels/go-nitro/client/engine"
@@ -99,6 +100,29 @@ func (c *Client) handleEngineEvents() {
 }
 
 // Begin API
+
+// Version returns the go-nitro version
+func (c *Client) Version() string {
+	info, _ := debug.ReadBuildInfo()
+
+	version := info.Main.Version
+	// Depending on how the binary was built we may get back no version info.
+	// If there's no version we'll default to "(devel)", which seems to be the default.
+	// See https://github.com/golang/go/issues/51831#issuecomment-1074188363 for more details.
+	if version == "" {
+		version = "(devel)"
+	}
+
+	// If the binary was built with the -buildvcs flag we can get the git commit hash and use that as the version.
+	for _, s := range info.Settings {
+		if s.Key == "vcs.revision" {
+			version = s.Value
+			break
+		}
+	}
+
+	return version
+}
 
 // CompletedObjectives returns a chan that receives a objective id whenever that objective is completed. Not suitable fo multiple subscribers.
 func (c *Client) CompletedObjectives() <-chan protocols.ObjectiveId {
--- a/main.go
+++ b/main.go
@@ -35,13 +35,14 @@ import (
 )
 
 func main() {
-	var pkString string
+	var pkString, hardhatUrl string
 	var msgPort, rpcPort int
 	var useWs, useDurableStore bool
 
 	flag.BoolVar(&useWs, "useWS", false, "Specifies whether to use websockets or NATS for the rpc server.")
 	flag.BoolVar(&useDurableStore, "useDurableStore", false, "Specifies whether to use a durable store or an in-memory store.")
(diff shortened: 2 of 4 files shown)


client/client.go
```diff
@@ -100,6 +101,14 @@ func (c *Client) handleEngineEvents() {
 
 // Begin API
 
+// Version returns the go-nitro version
+func (c *Client) Version() string {
+	info, _ := debug.ReadBuildInfo()
+	// TODO this doesn't seem to work, at least locally.
```
lalexgap, line 107: It looks like this will depend on how we run `main.go`.  From this [comment](https://github.com/golang/go/issues/51831) it turns out if we run specify the file (`go run main.go`) we don't get version info but if we don't (`go run .`) we do get version information.

Although even when we do the version is just `(devel)`:
```
> {"jsonrpc":"2.0","id":1,"method":"version","params":{}}
< {"jsonrpc":"2.0","id":1,"result":"(devel)","error":null}
```

rpc/server.go
```diff
@@ -59,6 +59,10 @@ func (rs *RpcServer) registerHandlers() (err error) {
 		}
 
 		switch serde.RequestMethod(validationResult.Method) {
+		case serde.VersionMethod:
+			return processRequest(rs, requestData, func(T serde.VersionRequest) string { // TODO I would like to be able to pass a function without parameters, here
```
lalexgap, line 63: I'm not sure if I can accomplish this without spiraling into a generics hole. 🙃

client/client.go
lalexgap: For now I've added some logic in https://github.com/statechannels/go-nitro/pull/1200/commits/26df816b4729512b5f1c3e3b9b8e2e6c37e4f1de so that we either get back `(devel)` or the git commit hash (if we built with `-buildvcs`)

main.go
```diff
@@ -0,0 +1,68 @@
...
+	flag.IntVar(&msgPort, "msgPort", 3005, "Specifies the tcp port for the  message service.")
+	flag.IntVar(&rpcPort, "rpcPort", 4005, "Specifies the tcp port for the rpc server.")
+	flag.Parse()
+	pk := common.Hex2Bytes(pkString)
+
+	logDestination := os.Stdout
+
+	storeA := store.NewMemStore(pk)
+	chain := chainservice.NewMockChain()
+	chainService := chainservice.NewMockChainService(chain, *storeA.GetAddress())
```
lalexgap, line 35: Since each RPC server is using it's own mockchain, they'll be unable to work together.

main.go
```diff
@@ -0,0 +1,68 @@
...
+	messageservice := p2pms.NewMessageService("127.0.0.1", msgPort, *storeA.GetAddress(), pk)
```
lalexgap, line 36: Currently we never [add any peers](https://github.com/statechannels/go-nitro/blob/c67ab800fc9ff45151162a388378cddb1802d54a/client/engine/messageservice/p2p-message-service/service.go#L52)so the message service won't be able to communicate with anyone

lalexgap: This [inspired me to make use of libp2p peer discovery](https://github.com/statechannels/go-nitro/pull/1207) so we don't have to worry about manually adding peers.

lalexgap:
Closing in favour of #1213, where I'm the author of the PR

PR #1201, opened 2023-03-31
Use a secp256k1 key in p2p message service

lalexgap:
It turns out we can use a `secp256k1` key with `libp2p` by calling `UnmarshalSecp256k1PrivateKey`. This means we can use the existing private keys as a message key.

**Note:** The testground action will fail until this [PR](https://github.com/statechannels/go-nitro-testground/pull/159) is merged.

### DIFF
--- a/client/engine/messageservice/p2p-message-service/service.go
+++ b/client/engine/messageservice/p2p-message-service/service.go
@@ -69,16 +69,16 @@ func (ms *P2PMessageService) AddPeers(peers []PeerInfo) {
 
 // NewMessageService returns a running P2PMessageService listening on the given ip, port and message key.
 func NewMessageService(ip string, port int, me types.Address, pk []byte) *P2PMessageService {
-	h := &P2PMessageService{
+	ms := &P2PMessageService{
 		toEngine: make(chan protocols.Message, BUFFER_SIZE),
 		peers:    &safesync.Map[peer.ID]{},
 		me:       me,
 	}
 
 	messageKey, err := p2pcrypto.UnmarshalSecp256k1PrivateKey(pk)
-	h.checkError(err)
+	ms.checkError(err)
 
-	h.key = messageKey
+	ms.key = messageKey
 
 	options := []libp2p.Option{
 		libp2p.Identity(messageKey),
@@ -91,9 +91,9 @@ func NewMessageService(ip string, port int, me types.Address, pk []byte) *P2PMes
 	if err != nil {
 		panic(err)
 	}
-	h.p2pHost = host
+	ms.p2pHost = host
 
-	h.p2pHost.SetStreamHandler(PROTOCOL_ID, func(stream network.Stream) {
+	ms.p2pHost.SetStreamHandler(PROTOCOL_ID, func(stream network.Stream) {
 		reader := bufio.NewReader(stream)
 		// Create a buffer stream for non blocking read and write.
 		raw, err := reader.ReadString(DELIMITER)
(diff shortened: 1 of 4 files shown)


client_test/rpc_test.go
```diff
@@ -120,7 +120,7 @@ func setupNitroNodeWithRPCClient(
 	messageservice := p2pms.NewMessageService("127.0.0.1",
 		msgPort,
 		crypto.GetAddressFromSecretKeyBytes(pk),
```
lalexgap, line 122: By allowing the SC address to be passed in separately from the `pk`,  it allows the option of using a different message key than the SC key.

geoknee (commented):
This definitely seems like an improvement from a neatness perspective. 

I am concerned about spreading `pk` around like this, though. It would be better from a security / audit-ability standpoint to prevent passing copies of this secret data around.

What do you think of the following approach:

Derive a message key as an "app key" following the spec here https://eips.ethereum.org/EIPS/eip-1775 . Pass that into the messaging service. The `originString` could be something simple like the existing `PROTOCOL_ID` from the p2p messaging service. 

We could even take this a step further and have the store expose a method for generating app keys: The state channel signing key could be derived from a main L1 key. We could even generate a new key per channel if we think there are advantages to doing that.

In general I'm trying to align us with the conclusions in this document https://www.notion.so/statechannels/RFC-8-Securing-Private-StateChannel-Key-c4f2945172d84db2b84220d64502e38c#1caa69cdb0814abf9c6a04e3d58554a8

lalexgap:
> Derive a message key as an "app key" following the spec here https://eips.ethereum.org/EIPS/eip-1775 . Pass that into the messaging service. The `originString` could be something simple like the existing `PROTOCOL_ID` from the p2p messaging service.

This current PR allows for this approach, the message service requires some `pk` but it doesn't require the same `pk` used for the wallet. So you could generate a completely random `pk` and pass that into the message service. In our tests it's kinda nice to use the same private key as it cuts down on some code, but in production there's no requirement for that.

Also right now I think we only call `NewMessageService` in test code when constructing `Client`s for tests. That means the "app key" generation code would only be called by test code constructing the clients. This probably hints that we should have a production way of constructing a client (that takes some kind of config?). That client constructor could generate an "app key" and construct the message service with it.

Perhaps I should rename `pk` so it's clear that it doesn't have to be the wallet pk?

geoknee:
I think a production constructor is very important and it would really help clarify all of this. 

The lack of the existence of such a constructor had me reading the test files as "the only way" we do things at the moment. I take your point that in the wild we wouldn't do it that way.

As discussed during standup -- the production constructor could read from a config file, plus read sensitive information (private key) from the command line.

client_test/helpers_test.go
```diff
@@ -225,21 +223,3 @@ func closeSimulatedChain(t *testing.T, chain chainservice.SimulatedChain) {
 		t.Fatal(err)
 	}
 }
-
-// generateMessageKey generates a ECDSA private key using the given pk bytes, and is idempotent.
-func generateMessageKey(t *testing.T, pk []byte) p2pcrypto.PrivKey {
```
lalexgap, line 230: This seemed to generate different keys for the same `pk`. I didn't read through `GenerateECDSAKeyPair` but I suspect some time-based information is being used when generating the  key.

client/engine/messageservice/p2p-message-service/service.go
```diff
@@ -68,7 +68,18 @@ func (ms *P2PMessageService) AddPeers(peers []PeerInfo) {
 }
 
 // NewMessageService returns a running P2PMessageService listening on the given ip, port and message key.
-func NewMessageService(ip string, port int, me types.Address, messageKey p2pcrypto.PrivKey) *P2PMessageService {
+func NewMessageService(ip string, port int, me types.Address, pk []byte) *P2PMessageService {
+	h := &P2PMessageService{
```
kerzhner, line 72: What is the significance of `h` as the variable name?

kerzhner (approved):

client/engine/messageservice/p2p-message-service/service.go
lalexgap: That's a good question! I just rearranged the existing code that was using `h` but I think we should switch it to something that makes more sense.

lalexgap: Updated in https://github.com/statechannels/go-nitro/pull/1201/commits/055f5a882ac41d6da455ec9ac9940660a9c8361b

PR #1206, opened 2023-04-04
Add query API to the RPC API

### LINKED ISSUE #1020: Implement an API to query the store over RPC
Currently the only way our client API communicates status information back to the caller is by sending  "events"(objectiveIds or vouchers) on one-way golang channels.  This means there's no way for an API caller to check the status of a channel(like after a reboot/crash). This also forces the caller of the API to be the one responsible for figuring out the current status by monitoring for events.

As part of [RFC 23](https://www.notion.so/statechannels/RFC-23-Overhaul-go-nitro-API-2c83cd05f59946ebb7354296de09f30e) I've put together a proposed query API and API proxies, that we've broadly discussed and agreed to implement.

There is still some ongoing discussion but here's roughly what it could look like:
```
type ChannelStatus string

const Proposed ChannelStatus = "Proposed"
const ReadyToFund ChannelStatus = "ReadyToFund"
const Closing ChannelStatus = "Closing"
const Complete ChannelStatus = "Complete"

type LedgerChannelInfo struct {
	ID      types.Destination
	Status  ChannelStatus
	Balance LedgerChannelBalance
}
type LedgerChannelBalance struct {
	AssetAddress  types.Address
	Hub           types.Address
	Client        types.Address
	HubBalance    *big.Int
	ClientBalance *big.Int
}


type PaymentChannelBalance struct {
	AssetAddress   types.Address
	Payee          types.Address
	Payer          types.Address
	PaidSoFar      *big.Int
	RemainingFunds *big.Int
}
type PaymentChannelInfo struct {
	ID      types.Destination
	Status  ChannelStatus
	Balance PaymentChannelBalance
}

type NitroQueryAPI interface {
	// GetPaymentChannelByReceiver returns the first payment found channel with the given receiver.
	// If no payment channel exists nil is returned.
	GetPaymentChannelByReceiver(receiver types.Address) PaymentChannelInfo
	// GetLedgerChannelByHub returns the first ledger channel found with the given hub.
	// If no ledger channel exists nil is returned.
	GetLedgerChannelByHub(Hub types.Address) LedgerChannelInfo
	// GetLedgerChannel returnns the ledger channel for the given id.
	// If no ledger channel exists nil is returned.
	GetLedgerChannel(id types.Destination) LedgerChannelInfo
	// GetPaymentChannel returnns the ledger channel for the given id.
	// If no ledger channel exists nil is returned.
	GetPaymentChannel(id types.Destination) PaymentChannelInfo
}
```

lalexgap:
Fixes #1020 

This PR adds the two query functions to the RPC API and updates the RPC integration test to call and check the query api.

### DIFF
--- a/client/query/query.go
+++ b/client/query/query.go
@@ -124,9 +124,9 @@ func GetPaymentChannelInfo(id types.Destination, store store.Store, vm *payments
 	if channelFound {
 		status := getStatusFromChannel(c)
 
-		// Since ADR 0009 allows for intermediaries to not sign the post fund setup
-		// our standard check that everyone has signed the postfund setup will not work
-		// Instead we check the status of the virtual fund objective specifically
+		// Due to ADR 0009 intermediaries can exit the virtual fund protocol before receiving all the postfund signatures
+		// This means intermediaries may not have a fully signed postfund state even though the channel is "ready"
+		// To determine the the correct status we check the status of the virtual fund objective
 		fund, isVirtualFund := isVirtualFundObjective(id, store)
 		if isVirtualFund && fund.Status == protocols.Completed {
 			status = Ready
--- a/client/query/query.go
+++ b/client/query/query.go
@@ -10,6 +10,7 @@ import (
 	"github.com/statechannels/go-nitro/payments"
 	"github.com/statechannels/go-nitro/protocols"
 	"github.com/statechannels/go-nitro/protocols/virtualdefund"
+	"github.com/statechannels/go-nitro/protocols/virtualfund"
 	"github.com/statechannels/go-nitro/types"
 )
 
@@ -74,19 +75,37 @@ func getLedgerBalanceFromState(latest state.State) LedgerChannelBalance {
 	}
 }
 
-// GetPaymentChannelInfo returns the PaymentChannelInfo for the given channel
-// It does this by querying the provided store and voucher manager
-func GetPaymentChannelInfo(id types.Destination, store store.Store, vm *payments.VoucherManager) (PaymentChannelInfo, error) {
+func isVirtualFundObjective(id types.Destination, store store.Store) (*virtualfund.Objective, bool) {
+	// This is slightly awkward but if the virtual defunding objective is complete it won't come back if we query by channel id
+	// We manually construct the objective id and query by that
+	virtualFundId := protocols.ObjectiveId(virtualfund.ObjectivePrefix + id.String())
+	o, err := store.GetObjectiveById(virtualFundId)
+	if err != nil {
+		return nil, false
+	}
+	return o.(*virtualfund.Objective), true
+}
+
+func isVirtualDefundObjective(id types.Destination, store store.Store) (*virtualdefund.Objective, bool) {
 	// This is slightly awkward but if the virtual defunding objective is complete it won't come back if we query by channel id
 	// We manually construct the objective id and query by that
 	virtualDefundId := protocols.ObjectiveId(virtualdefund.ObjectivePrefix + id.String())
-	fetchedDefund, err := store.GetObjectiveById(virtualDefundId)
-	isVirtualDefund := err == nil
+	o, err := store.GetObjectiveById(virtualDefundId)
+	if err != nil {
+		return nil, false
+	}
+
+	return o.(*virtualdefund.Objective), true
+}
+
+// GetPaymentChannelInfo returns the PaymentChannelInfo for the given channel
+// It does this by querying the provided store and voucher manager
+func GetPaymentChannelInfo(id types.Destination, store store.Store, vm *payments.VoucherManager) (PaymentChannelInfo, error) {
+	defund, isVirtualDefund := isVirtualDefundObjective(id, store)
 
 	// Since virtual defunding stores all state updates on the objective
 	// instead of the store we need to manually check the objective
 	if isVirtualDefund {
-		defund := fetchedDefund.(*virtualdefund.Objective)
 		status := Closing
 
 		if defund.Status == protocols.Completed {
@@ -99,12 +118,19 @@ func GetPaymentChannelInfo(id types.Destination, store store.Store, vm *payments
 			Balance: getPaymentChannelBalance(defund.VFixed.Participants, []outcome.SingleAssetExit{defund.FinalOutcome}),
 		}, nil
 	}
-
 	// Otherwise we can just check the store
 	c, channelFound := store.GetChannelById(id)
 
 	if channelFound {
 		status := getStatusFromChannel(c)
+
+		// Since ADR 0009 allows for intermediaries to not sign the post fund setup
+		// our standard check that everyone has signed the postfund setup will not work
+		// Instead we check the status of the virtual fund objective specifically
+		fund, isVirtualFund := isVirtualFundObjective(id, store)
+		if isVirtualFund && fund.Status == protocols.Completed {
+			status = Ready
+		}
 		latest, err := getLatestSupported(c)
 		if err != nil {
 			return PaymentChannelInfo{}, err
--- a/rpc/server.go
+++ b/rpc/server.go
@@ -2,7 +2,6 @@ package rpc
 
 import (
 	"encoding/json"
-	"fmt"
 	"math/big"
 
 	"github.com/rs/zerolog"
@@ -58,12 +57,6 @@ func (rs *RpcServer) registerHandlers() (err error) {
 		if validationResult.Error != nil {
 			return validationResult.Error
 		}
-
-		defer func() {
-			if r := recover(); r != nil {
-				err = fmt.Errorf("panic: %v", r)
-			}
-		}()
 
 		switch serde.RequestMethod(validationResult.Method) {
 		case serde.DirectFundRequestMethod:
--- a/rpc/server.go
+++ b/rpc/server.go
@@ -2,10 +2,12 @@ package rpc
 
 import (
 	"encoding/json"
+	"fmt"
 	"math/big"
 
 	"github.com/rs/zerolog"
 	nitro "github.com/statechannels/go-nitro/client"
+	"github.com/statechannels/go-nitro/client/query"
 	"github.com/statechannels/go-nitro/protocols"
 	"github.com/statechannels/go-nitro/protocols/directdefund"
 	"github.com/statechannels/go-nitro/protocols/directfund"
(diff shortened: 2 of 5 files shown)


client/query/query.go
```diff
@@ -99,12 +118,19 @@ func GetPaymentChannelInfo(id types.Destination, store store.Store, vm *payments
 			Balance: getPaymentChannelBalance(defund.VFixed.Participants, []outcome.SingleAssetExit{defund.FinalOutcome}),
 		}, nil
 	}
-
 	// Otherwise we can just check the store
 	c, channelFound := store.GetChannelById(id)
 
 	if channelFound {
 		status := getStatusFromChannel(c)
```
lalexgap, line 125: Previously we were checking if `c.PostFundComplete()` to determine if the channel status was "Ready" or "Proposed". However since the changes for ADR 0009 intermediaries exit the protocol early before receiving all the postfund signatures.

This means we can rely on `c.PostFundComplete()` to determine the status of a virtual channel.

rpc/server.go
```diff
@@ -79,12 +80,30 @@ func (rs *RpcServer) registerHandlers() error {
 				rs.client.Pay(obj.Channel, big.NewInt(int64(obj.Amount)))
 				return obj
 			})
+		case serde.GetPaymentChannelRequestMethod:
+			return processRequest(rs, requestData, func(r serde.GetPaymentChannelRequest) query.PaymentChannelInfo {
+				var pc query.PaymentChannelInfo
+				pc, err = rs.client.GetPaymentChannel(r.Id)
```
kerzhner, line 86:
```suggestion
				pc, err := rs.client.GetPaymentChannel(r.Id)
				// do something with the error
```

I find the syntax in the PR deceiving. `err` is assigned, but the `err` value is never checked since the assignment happens in a function that does not return an error. Our options are:
1. Modify [`RegisterRequestHandler`](https://github.com/statechannels/go-nitro/blob/1e95bec4a531ad81c2e86166220c01b61970d8bc/rpc/transport/transport.go#L31) to accept a function that can return an error. Something like:
``` go
RegisterRequestHandler(func([]byte) ([]byte, error)) error
```
2. Panic if `GetPaymentChannel` returns an error.

rpc/server.go
```diff
@@ -79,12 +80,30 @@ func (rs *RpcServer) registerHandlers() error {
...
+
+				return pc
+			})
+		case serde.GetLedgerChannelRequestMethod:
+			return processRequest(rs, requestData, func(r serde.GetLedgerChannelRequest) query.LedgerChannelInfo {
+				var l query.LedgerChannelInfo
+				l, err = rs.client.GetLedgerChannel(r.Id)
```
kerzhner, line 93: Same comment about error as above.

client/query/query.go
```diff
@@ -74,19 +75,37 @@ func getLedgerBalanceFromState(latest state.State) LedgerChannelBalance {
...
+	// We manually construct the objective id and query by that
+	virtualFundId := protocols.ObjectiveId(virtualfund.ObjectivePrefix + id.String())
+	o, err := store.GetObjectiveById(virtualFundId)
+	if err != nil {
+		return nil, false
+	}
+	return o.(*virtualfund.Objective), true
+}
+
+func isVirtualDefundObjective(id types.Destination, store store.Store) (*virtualdefund.Objective, bool) {
```
kerzhner, line 89:
```suggestion
func isVirtualDefundObjective(id types.Destination, store store.Store) *virtualdefund.Objective {
```
Do we need the `bool` return value? It seems like the two return values are coupled. If the result is nil, the second return value is set to false. If the result is not nil, the second return value is set to true.

client/query/query.go
```diff
@@ -74,19 +75,37 @@ func getLedgerBalanceFromState(latest state.State) LedgerChannelBalance {
 	}
 }
 
-// GetPaymentChannelInfo returns the PaymentChannelInfo for the given channel
-// It does this by querying the provided store and voucher manager
-func GetPaymentChannelInfo(id types.Destination, store store.Store, vm *payments.VoucherManager) (PaymentChannelInfo, error) {
+func isVirtualFundObjective(id types.Destination, store store.Store) (*virtualfund.Objective, bool) {
```
kerzhner, line 78: Same comment about return values.

rpc/server.go
lalexgap: > I find the syntax in the PR deceiving. err is assigned, but the err value is never checked since the assignment happens in a function that does not return an error.

Perhaps I'm mistaken but I think the `err` value is being checked. `err` is the return value defined [here](https://github.com/statechannels/go-nitro/blob/1e95bec4a531ad81c2e86166220c01b61970d8bc/rpc/server.go#L48). Inside the scope of `processRequest` we assign the `err` value to the result of `GetPaymentChannel`. The actual check happens just outside the scope of `processRequest`  where [we inspect the value of err](https://github.com/statechannels/go-nitro/blob/1e95bec4a531ad81c2e86166220c01b61970d8bc/rpc/server.go#L48).

I think in this case it's pretty confusing because in `processRequest` we both return something(`pc`) and assign a variable that lives outside the function scope (`err`). So we're probably better off doing something else.


> 1. Modify [`RegisterRequestHandler`](https://github.com/statechannels/go-nitro/blob/1e95bec4a531ad81c2e86166220c01b61970d8bc/rpc/transport/transport.go#L31) to accept a function that can return an error. Something like:
> 
This is probably the right thing to do, I was just being lazy and avoiding that.

>     2. Panic if `GetPaymentChannel` returns an error.

There's also a 3rd way where we `panic` but then `recover` the `panic` and return an error. [That was actually what I was doing previously to handle errors.](https://github.com/statechannels/go-nitro/pull/1206/commits/1e95bec4a531ad81c2e86166220c01b61970d8bc) It also makes me wonder if some of our functions should be using `recover` to guard against panics?

client/query/query.go
lalexgap: This is mostly syntactical sugar that mimicks how casting from an interface works in go: `v, ok = i.(T)`. I personally prefer having the extra bool value, even if it's not strictly necessary ,as I think it makes things slightly neater. However I don't feel super strongly about it!

rpc/server.go
kerzhner: My understanding of the function is:
1. Inside `registerHandlers`, there is a call to `RegisterRequestHandler` with a handler. That call can result in an error. That error is the return value of `registerHandlers`.
2. As of this PR, some of the handler callbacks can result in an error. Note that when an error like that is created, the `registerHandlers` has already returned. So the assignment has no effect.

The compiler tries points out to this issue if the syntax of error assignment is changed. Try:
``` go
return processRequest(rs, requestData, func(r serde.GetPaymentChannelRequest) query.PaymentChannelInfo {
  pc, err := rs.client.GetPaymentChannel(r.Id)
  return pc
})
```

The compiler should show: "err declared and not used". Another way to see this error is to use a different variable name like `err1`.

As far as panic/recover methodology: what are the advantages of this approach over the standard golang approach of returning and inspecting errors?

client/query/query.go
kerzhner: Gotcha. Maybe the method name is throwing me off. This method seems different than a cast. This method is more of of `getVirtualDefundObjectiveIfExists`. I don't feel strongly about the return type either.

rpc/server.go
lalexgap: >As of this PR, some of the handler callbacks can result in an error. Note that when an error like that is created, the registerHandlers has already returned. So the assignment has no effect.

This is the part I missed 🤦  I didn't pay attention that most of the code was a function declaration, not code being called immediately. 

> The compiler tries points out to this issue if the syntax of error assignment is changed. Try:

Just to be slightly nitpick/pedantic, but changing the `err` variable to a local declaration would show the `"err declared and not used"` error in a lot of scenarios where a valid assignment was being made. A function can set variables that are declared outside it's scope, and might not know how those variables are handled.

Here's a pretty trivial example.
```
func (rs *RpcServer) DoStuff() (err error) {
	func() {
        
		err = fmt.Errorf("BOOM")
                // This results in the error `err declared and not used`
                // err := fmt.Errorf("BOOM")
	}()

	return
}
```

lalexgap: > My understanding of the function is:
> 
>     1. Inside `registerHandlers`, there is a call to `RegisterRequestHandler` with a handler. That call can result in an error. That error is the return value of `registerHandlers`.
> 
>     2. As of this PR, some of the handler callbacks can result in an error. Note that when an error like that is created, the `registerHandlers` has already returned. So the assignment has no effect.

Yup, I goofed here and forgot completely that these were callback handlers running after `registerHandlers` had been called. Sorry!

> The compiler tries points out to this issue if the syntax of error assignment is changed. Try:
> 
> ```go
> return processRequest(rs, requestData, func(r serde.GetPaymentChannelRequest) query.PaymentChannelInfo {
>   pc, err := rs.client.GetPaymentChannel(r.Id)
>   return pc
> })
> ```
> 
> The compiler should show: "err declared and not used". Another way to see this error is to use a different variable name like `err1`.
> 

Just to be slightly nitpicky/pedantic but switching `err` to a local declaration would show the `err declared and not used` error even in scenarios where the `err` value does get checked/used. Here's a trivial example:

```
func (rs *RpcServer) DoStuff() (err error) {
	func() {
		err = fmt.Errorf("BOOM")
		// This complains about err being unused
		// err := fmt.Errorf("BOOM")
	}()

	return
}
```

kerzhner: For the code example above, if `err := ...` declares a new variable that is scoped to the anonymous function context. So the compiler error is useful. For example, the following test fails:

``` go
import (
	"fmt"
	"testing"

	"github.com/stretchr/testify/assert"
)

func TestError(t *testing.T) {
	assert.Error(t, ReturnsError())
}

func ReturnsError() (err error) {
	func() {
		err := fmt.Errorf("BOOM")
		fmt.Print(err)
	}()

	return
}
```

lalexgap: For now I've updated it to just panic in https://github.com/statechannels/go-nitro/pull/1206/commits/63097d0f82af984ea599cfdd418f4b688a0efad9

kerzhner: All that being said, the way `registerHandlers` is written (by me, I believe:)  ) lends itself to a confusing developer experience. So maybe we should refactor that function.

lalexgap: > For the code example above, if `err := ...` declares a new variable that is scoped to the anonymous function context. So the compiler error is useful. For example, the following test fails:

I think I was just trying to clarify this comment in particular:

> The compiler tries points out to this issue if the syntax of error assignment is changed. Try:
> 
> ```go
> return processRequest(rs, requestData, func(r serde.GetPaymentChannelRequest) query.PaymentChannelInfo {
>   pc, err := rs.client.GetPaymentChannel(r.Id)
>   return pc
> })
> ```
> 
> The compiler should show: "err declared and not used". Another way to see this error is to use a different variable name like `err1`.

I just wanted to clarify that switching `err` to a local declaration and seeing it unused doesn't mean the previous `err` var wasn't being checked properly (although it wasn't).  It just proves that the `err` var isn't being checked in the local context.

kerzhner (approved):

PR #1207, opened 2023-04-07
use mdns to discover peers in the p2p message service

lalexgap:
This updates the P2P message service to to handle peer discovery instead of relying on `AddPeers` to be manually called. This is done using [mdns](https://docs.libp2p.io/concepts/discovery-routing/mdns/) which allows for peers to discover each other on a **local network**.

While `mdns` alerts us to new peers, it doesn't provide the peers statechannel address. To address this we gossip our peer information  (including our SC address) to every new peer we find, and rely on them doing the same so we can map SC addresses to a peer.ID.

The message service exposes a `PeerInfoReceived` chan that receives a `PeerInfo` whenever a new peer(and all their info) is discovered. This can be useful to wait  until all your peers are available.


client/engine/messageservice/p2p-message-service/service.go
```diff
@@ -15,30 +16,44 @@ import (
...
 	"github.com/statechannels/go-nitro/types"
 )
 
+// PeerInfo contains information about a peer
+type PeerInfo struct {
+	Id      peer.ID
+	Address types.Address
+
+	// TODO: These are only here for backwards compatibility with the old testground test
+	Port      int
```
lalexgap, line 32: These fields can be removed once https://github.com/statechannels/go-nitro-testground/pull/161 is merged

client/engine/messageservice/p2p-message-service/service.go
```diff
@@ -164,6 +242,20 @@ func (s *P2PMessageService) Out() <-chan protocols.Message {
...
 	return s.p2pHost.Close()
 }
+
+// PeerInfoReceived returns a channel that receives a PeerInfo when a peer is discovered
+func (s *P2PMessageService) PeerInfoReceived() <-chan PeerInfo {
+	return s.newPeerInfo
+}
+
+// AddPeers is DEFUNCT and is only here for testground compatibility
+func (ms *P2PMessageService) AddPeers(peers []PeerInfo) {
```
lalexgap, line 257: This can be removed once https://github.com/statechannels/go-nitro-testground/pull/161 is merged

lalexgap:
Moving back to draft as not everyone is receiving the peer info

lalexgap:
I originally started looking into libp2p for peer discovery while working on the RPC server in  #1200.  Having to manually add peers when spinning up an RPC server was kind of clunky. By using libp2p peer discovery I could easily spin up as many RPC servers as I wanted; and they would automagically talk to eachother.

One important thing to note is that `mdns` works only on a local network, and therefore this message service is **not production ready**. 

There are also [other peer discovery options that work with libp2p](https://docs.libp2p.io/concepts/discovery-routing/overview/).  I briefly investigated using a `dht` but `mdns` seemed the most simple/straightforward, so I went with that. 

@kerzhner has also brought up the point about the security of being able to receive messages from anyone. Previously we were whitelisting specific peers (via calling `AddPeers`) and now anyone could send us a message. (Perhaps this is something we should rely on policies for?)

I still think this functionality is worth implementing in some fashion as it makes spinning up RPC servers much more pleasant. However we have a couple options on how/where we implement the logic:
- `mdns` peer discovery is optional in p2p message service (and can be enabled by some flag) The previous `AddPeers` logic remains.
- The `mdns` peer discovery logic is moved to  [executable](https://github.com/statechannels/go-nitro/blob/d31cebd6c31df30b86989d2ffaaeda240c5b6b4a/main.go#L30) that handles running the RPC server. Might result in some duplication.
- We could create a new separate message service that handles peer discovery and leave the P2PMessageService alone. We use the `PeerDiscoveryP2PMessageService` when creating our RPC server. Probably will result in some duplicate code.

Outside the scope of this PR I think there is also a broader question about how we expect peer discovery to be handled. Right now we pass in a `P2PMessageService` when constructing a client/engine so it's easy to to call `AddPeers` to setup the peers. However, in the future, we're hoping to have [a constructor that just takes in a config](https://github.com/statechannels/go-nitro/issues/120) and returns a client.  That means we either need to handle peer discovery or expose some kind of `AddPeer` API on the client.

@kerzhner any thoughts?

client/engine/messageservice/p2p-message-service/service.go
```diff
@@ -91,26 +88,107 @@ func NewMessageService(ip string, port int, me types.Address, pk []byte) *P2PMes
...
+		return
+	}
+	ms.checkError(err)
+	m, err := protocols.DeserializeMessage(raw)
+	ms.checkError(err)
+	ms.toEngine <- m
+	stream.Close()
+}
+
+func (ms *P2PMessageService) gossipStreamHandler(stream network.Stream) {
```
lalexgap, line 133: If we were ok enforcing that the `pk` for the  message service is your statechannel key, we could probably derive the state channel address from the `peer.ID`. This means we wouldn't manually have to gossip our state channel address to our peers.

client/engine/messageservice/p2p-message-service/service.go
```diff
@@ -164,6 +242,20 @@ func (s *P2PMessageService) Out() <-chan protocols.Message {
...
 // Close closes the P2PMessageService
 func (s *P2PMessageService) Close() error {
+	s.cancel()
+	s.mdns.Close()
 	s.p2pHost.RemoveStreamHandler(PROTOCOL_ID)
 	return s.p2pHost.Close()
 }
+
+// PeerInfoReceived returns a channel that receives a PeerInfo when a peer is discovered
+func (s *P2PMessageService) PeerInfoReceived() <-chan PeerInfo {
```
lalexgap, line 252: Since we don't know exactly when the peers get added to a message service, `PeerInfoReceived` is very useful to wait until everyone has all the peer info in tests.

client/engine/messageservice/p2p-message-service/service.go
lalexgap: We can rely on the `PeerStore` , which stores a multiaddress for each of our peers. We don't need to bother storing the port/ipaddress on `PeerInfo`

client/engine/messageservice/p2p-message-service/service.go
```diff
@@ -91,26 +83,100 @@ func NewMessageService(ip string, port int, me types.Address, pk []byte) *P2PMes
...
+		return
+	}
+	ms.checkError(err)
+	m, err := protocols.DeserializeMessage(raw)
+	ms.checkError(err)
+	ms.toEngine <- m
+	stream.Close()
+}
+
+func (ms *P2PMessageService) peerExchangeHandler(stream network.Stream) {
```
kerzhner, line 127: Why is this handler needed if we already have HandlePeerFound?

client/engine/messageservice/p2p-message-service/service.go
```diff
@@ -91,26 +83,100 @@ func NewMessageService(ip string, port int, me types.Address, pk []byte) *P2PMes
...
+
+	if !hasPeer {
+		fmt.Printf("Setting info received from peer: %v\n", peerInfo)
+		ms.peers.Store(peerInfo.Address.String(), *peerInfo)
+		ms.newPeerInfo <- *peerInfo
+	}
+
+	// It's possible that the a message service is using a new peer ID for the same address.
+	// We allow overriding the existing peer info in this case.
+	if isNewInfo := prev.Id != peerInfo.Id || prev.Address != peerInfo.Address; hasPeer && isNewInfo {
```
kerzhner, line 175: Not sure I fully understand this condition. `ms.peers.Load` lookup is done by address. So if the incoming peer information address does not match looked up peer address, I believe `hasPeer` will always be false?

kerzhner (approved):
This approach looks reasonable to me.

client/engine/messageservice/p2p-message-service/service.go
```diff
@@ -91,26 +83,100 @@ func NewMessageService(ip string, port int, me types.Address, pk []byte) *P2PMes
...
+		stream.Close()
+		return
+	}
+	ms.checkError(err)
+
+	var peerInfo *basicPeerInfo
+	err = json.Unmarshal([]byte(raw), &peerInfo)
+	ms.checkError(err)
+
+	prev, hasPeer := ms.peers.Load(peerInfo.Address.String())
```
kerzhner, line 165: There is a race condition since the read (Load) and the write (Store) are not protected by an encapsulating mutex. So there can be a read, read, write, write instead of the expected read, write, read, write.

In this particular case, I am not sure the race condition has much of an impact.

client/engine/messageservice/p2p-message-service/service.go
lalexgap: There's currently two parts to peer discovery process:
1. MDNS triggers `HandlePeerFound` and a message service is alerted to a new peer ID/multiaddress. We still don't know the SC address of the peer.
2. Whenever we find a a new peer in `HandlePeerFound` we message that peer (using a two-way network stream).  On that stream we send: our peer information (including or SC address)  and then read the peer information from the other message service.

`peerExchangeHandler` is the logic for handling the message received in part 2, and it registered by `ms.p2pHost.SetStreamHandler` in the constructor.

I did find the name slightly confusing so I ended up [embedding](https://github.com/statechannels/go-nitro/blob/23f608b4d04babb0eb5e16891c83327b15f2f499/client/engine/messageservice/p2p-message-service/service.go#L96) the logic into the stream handler

client/engine/messageservice/p2p-message-service/service.go
lalexgap: Good point! Updated to `LoadOrStore` in https://github.com/statechannels/go-nitro/pull/1207/commits/738c9ef14429f8ac423289a7c1f1ed7daaccb974

client/engine/messageservice/p2p-message-service/service.go
lalexgap: Removed in https://github.com/statechannels/go-nitro/pull/1207/commits/738c9ef14429f8ac423289a7c1f1ed7daaccb974

Issue #1208, opened 2023-04-07
The websocket RPC Server crashes if there are no listeners


lalexgap:
The RPC server may crash if it completes an objective and there are no RPC clients listening for JSON-RPC notifications with the following error:
```
{"l":"trace","client":"0xBBB676f9cFF8D242e9eaC39D063848807d3D1D94","rpc":"server","scope":"","t":1680892402,"m":"Sending notification: DirectFunding-0x4fd071df208b69c95d267401925445810455ddf811bb87637709de934fb49e03"}
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x100 pc=0x104fd3974]

goroutine 119 [running]:
nhooyr.io/websocket.(*Conn).writer(0x0, {0x1058ed3b0?, 0x14000196000?}, 0x0?)
	/Users/lagap/go/pkg/mod/nhooyr.io/websocket@v1.8.7/write.go:111 +0x24
nhooyr.io/websocket.(*Conn).write(0x0, {0x1058ed3b0, 0x14000196000}, 0x40?, {0x140005d2370, 0xa5, 0xb0})
	/Users/lagap/go/pkg/mod/nhooyr.io/websocket@v1.8.7/write.go:122 +0x48
nhooyr.io/websocket.(*Conn).Write(0x140005d2370?, {0x1058ed3b0?, 0x14000196000?}, 0x10474bed0?, {0x140005d2370?, 0x1058a7200?, 0x14000175300?})
	/Users/lagap/go/pkg/mod/nhooyr.io/websocket@v1.8.7/write.go:42 +0x28
github.com/statechannels/go-nitro/rpc/transport/ws.(*serverWebSocketTransport).Notify(0x105697a20?, {0x140005d2370?, 0x0?, 0x140000e6fb8?})
	/Users/lagap/code/go-nitro/rpc/transport/ws/server.go:63 +0x4c
github.com/statechannels/go-nitro/rpc.(*RpcServer).sendNotifications.func1()
	/Users/lagap/code/go-nitro/rpc/server.go:201 +0x160
created by github.com/statechannels/go-nitro/rpc.(*RpcServer).sendNotifications
	/Users/lagap/code/go-nitro/rpc/server.go:193 +0x60
exit status 2
```
I'm running into this error in the following scenario:
- Spin up RPC server for Alice
- Spin up RPC server for Bob
- Spin up a (typescript) RPC client for Alice
[...]

PR #1210, opened 2023-04-10
Websocket race condition fix

kerzhner:
Fixes https://github.com/statechannels/go-nitro/issues/1208.

This PR originated from running unit tests with [`-race`](https://go.dev/doc/articles/race_detector) flag. This PR includes a minor fix for a race condition in the `rand` package and a major change to the websocket rpc implementation.

The websocket rpc client (which should be renamed to http client) now makes requests to the rpc server using http post. The client listens to notifications over a websocket. Previously, the client used a single websocket connection for requests and notifications. While this solution was seemingly simple, the implementation had race conditions as well as complexity around managing of the websocket connection.

### DIFF
--- a/rand/rand.go
+++ b/rand/rand.go
@@ -8,16 +8,10 @@ import (
 	"time"
 )
 
-var r *rand.Rand
-
 // getRandGenerator seeds a random number generator based on current time
 func getRandGenerator() *rand.Rand {
-	if r != nil {
-		return r
-	}
 	source := rand.NewSource(time.Now().UnixNano())
-	r = rand.New(source)
-	return r
+	return rand.New(source)
 }
 
 func Uint64() uint64 {
--- a/rpc/transport/ws/server.go
+++ b/rpc/transport/ws/server.go
@@ -105,6 +105,8 @@ func (wsc *serverWebSocketTransport) subscribe(w http.ResponseWriter, r *http.Re
 	defer wsc.notificationListeners.Delete(key)
 
 	// A client closes a connection by sending a message over the websocket
+	// This code converts the socket `Read` call to a channel.
+	// Ideally, all signals would be managed in the select statement below. But there is no channel API for the websocket read.
 	closeChan := make(chan error)
 	go func() {
 		_, _, err := c.Read(r.Context())
--- a/rpc/transport/ws/server.go
+++ b/rpc/transport/ws/server.go
@@ -72,7 +72,7 @@ func (wsc *serverWebSocketTransport) Close() {
 }
 
 func (wsc *serverWebSocketTransport) Url() string {
-	return "ws://" + webscocketServerAddress + wsc.port
+	return webscocketServerAddress + wsc.port
 }
 
 func (wsc *serverWebSocketTransport) request(w http.ResponseWriter, r *http.Request) {
--- a/rpc/transport/ws/server.go
+++ b/rpc/transport/ws/server.go
@@ -34,7 +34,7 @@ func NewWebSocketTransportAsServer(port string) (*serverWebSocketTransport, erro
 	}
 
 	wsc.httpServer = &http.Server{
-		Handler:      wsc,
+		Handler:      &wsc.serveMux,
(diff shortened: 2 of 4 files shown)


rpc/transport/ws/server.go
```diff
@@ -71,18 +72,60 @@ func (wsc *serverWebSocketTransport) Close() {
...
+		select {
+		case err = <-closeChan:
+			done = true
+		case <-r.Context().Done():
+			err = r.Context().Err()
+			done = true
+		case notificationData := <-notificationChan:
+			err := c.Write(r.Context(), websocket.MessageText, notificationData)
+			if err != nil {
+				done = true
```
lalexgap, line 125: Outside the scope of this PR but we could consider using a `break` [statement with a label](https://go.dev/ref/spec#Break_statements).  I think the explicit behavior of `break` reduces some mental overhead of making sure nothing executes after `done=true` is set.

rpc/transport/ws/server.go
```diff
@@ -71,18 +72,60 @@ func (wsc *serverWebSocketTransport) Close() {
...
-	wsc.serverWebsocket = c
 	defer c.Close(websocket.StatusInternalError, "server initiated websocket close")
+	notificationChan := make(chan []byte)
+	key := strconv.Itoa(int(rand.Uint64()))
+	wsc.notificationListeners.Store(key, notificationChan)
+	defer wsc.notificationListeners.Delete(key)
+
+	// A client closes a connection by sending a message over the websocket
+	closeChan := make(chan error)
+	go func() {
```
lalexgap, line 109: Is this meant to block the server until we receive a close signal from the client? It looks like we expect an `err` from `c.Read`, is that because we expect some kind of closed websocket error (caused by the client closing their ws)? 

Should there be a way for the server to close without waiting on the client? Like a `select` statement that also checks for some `ctxt.Done()`

rpc/transport/ws/server.go
```diff
@@ -71,18 +72,60 @@ func (wsc *serverWebSocketTransport) Close() {
...
-	return "ws://" + webscocketServerAddress + wsc.port
+	return webscocketServerAddress + wsc.port
+}
+
+func (wsc *serverWebSocketTransport) request(w http.ResponseWriter, r *http.Request) {
+	if r.Method != "POST" {
+		http.Error(w, http.StatusText(http.StatusMethodNotAllowed), http.StatusMethodNotAllowed)
+		return
+	}
+	body := http.MaxBytesReader(w, r.Body, 8192)
```
lalexgap, line 83: Is there any reason we chose `8192` as the maximum amount of bytes we accept for a request? Maybe we could throw it in a constant?

rand/rand.go
```diff
@@ -8,16 +8,10 @@ import (
...
-
 // getRandGenerator seeds a random number generator based on current time
 func getRandGenerator() *rand.Rand {
-	if r != nil {
-		return r
-	}
 	source := rand.NewSource(time.Now().UnixNano())
-	r = rand.New(source)
-	return r
+	return rand.New(source)
```
lalexgap, line 14: What problem arose from using a shared `rand.Rand`? Did sharing `rand.Rand` mean that we were essentially using the same  "seed" and just getting the next number in the sequence.

rpc/transport/ws/server.go
kerzhner: Good idea! I wasn't aware of labels. af9e49a8

rpc/transport/ws/server.go
kerzhner: Good question. This code converts the socket `Read` call to a channel. Ideally, all signals would be managed in the select statement below. But there is no channel API for the websocket read. A godoc explanation has been added in 8273d82d

The loop below does monitor the `ctx.Done()` channel

rpc/transport/ws/server.go
kerzhner: This seemed like a reasonable choice. It has been made a constant in c4284943.

PR #1211, opened 2023-04-12
Synchronous ethereum log subscription

kerzhner:
Ethereum events are occasionally missed as flagged by a failing unit test: https://github.com/statechannels/go-nitro/actions/runs/4680109856/jobs/8291031994

One reason could be that Ethereum log subscription is an asynchronous operation in the chain service constructor. This change flips the subscription to be synchronous.

### DIFF
--- a/client/engine/chainservice/eth_chainservice.go
+++ b/client/engine/chainservice/eth_chainservice.go
@@ -71,7 +71,7 @@ func NewEthChainService(chain ethChain, na *NitroAdjudicator.NitroAdjudicator,
 
 	// Use a buffered channel so we don't have to worry about blocking on writing to the channel.
 	ecs := EthChainService{chain, na, naAddress, caAddress, vpaAddress, txSigner, make(chan Event, 10), logger, ctx, cancelCtx}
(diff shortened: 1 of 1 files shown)


client/engine/chainservice/eth_chainservice.go
```diff
@@ -219,33 +219,35 @@ func (ecs *EthChainService) subscribeForLogs() {
...
-	for {
-		select {
-		case <-ecs.ctx.Done():
-			sub.Unsubscribe()
-			return
-		case err := <-sub.Err():
-			if err != nil {
-				ecs.fatalF("received error from the subscription channel: %w", err)
-			}
+	go func() {
```
lalexgap, line 222: Nit but it might be nice to add a comment explaining that we're waiting for the synchronous subscribe before listening async

lalexgap (approved):

client/engine/chainservice/eth_chainservice.go
kerzhner: Added in 7a274372.

PR #1213, opened 2023-04-13
Nitro rpc server

lalexgap:
This PR allows go-nitro to run as an RPC server executable.

## Running the RPC Server

The RPC server can be started by running `go run .` in the `go-nitro` directory.

```shell
go run .
```
You can also build a binary and execute that. (This will embed the git commit info for the version API call).

```shell
go build  -o nitro-rpc-server
```
## RPC Server Options

There a a couple of CLI flags that control things like ports to listen on or the SC key to use. **Note:** All flags have a default value (seen below).
```
❯ ./nitro-rpc-server --help
Usage of ./nitro-rpc-server:
 ./nitro-rpc-server --help
Usage of ./nitro-rpc-server:
  -hardhatUrl string
    	Specifies the url for the hardhat node. (default "ws://127.0.0.1:8545")
  -msgPort int
    	Specifies the tcp port for the  message service. (default 3005)
  -pk string
    	Specifies the private key for the client. Default is Alice's private key. (default "2d999770f7b5d49b694080f987b82bbc9fc9ac2b4dcc10b0f8aba7d700f69c6d")
  -rpcPort int
    	Specifies the tcp port for the rpc server. (default 4005)
  -useDurableStore
    	Specifies whether to use a durable store or an in-memory store.
  -useNats
    	Specifies whether to use NATS or http/ws for the rpc server. (default true)
```

## Interacting with the RPC Server
Right now the easiest way of interacting with the RPC server is by using the [typescript RPC client](https://github.com/statechannels/nitro-rpc-client).

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -107,7 +107,7 @@ func (c *Client) Version() string {
 
 	version := info.Main.Version
 	// Depending on how the binary was built we may get back no version info.
-	// If there's no version we'll default to "(devel)", which seems to be the default.
+	// In this case we default to "(devel)".
 	// See https://github.com/golang/go/issues/51831#issuecomment-1074188363 for more details.
 	if version == "" {
 		version = "(devel)"
--- a/client/client.go
+++ b/client/client.go
@@ -4,6 +4,7 @@ package client // import "github.com/statechannels/go-nitro/client"
 import (
 	"io"
 	"math/big"
+	"runtime/debug"
 
 	"github.com/statechannels/go-nitro/channel/state/outcome"
 	"github.com/statechannels/go-nitro/client/engine"
@@ -99,6 +100,29 @@ func (c *Client) handleEngineEvents() {
 }
 
 // Begin API
+
+// Version returns the go-nitro version
+func (c *Client) Version() string {
+	info, _ := debug.ReadBuildInfo()
+
+	version := info.Main.Version
+	// Depending on how the binary was built we may get back no version info.
+	// If there's no version we'll default to "(devel)", which seems to be the default.
+	// See https://github.com/golang/go/issues/51831#issuecomment-1074188363 for more details.
+	if version == "" {
+		version = "(devel)"
+	}
+
+	// If the binary was built with the -buildvcs flag we can get the git commit hash and use that as the version.
+	for _, s := range info.Settings {
+		if s.Key == "vcs.revision" {
+			version = s.Value
+			break
+		}
+	}
+
+	return version
+}
 
 // CompletedObjectives returns a chan that receives a objective id whenever that objective is completed. Not suitable fo multiple subscribers.
 func (c *Client) CompletedObjectives() <-chan protocols.ObjectiveId {
--- a/main.go
+++ b/main.go
@@ -62,7 +62,7 @@ func main() {
 
 	chainService, err := newChainService(context.Background(), *ourStore.GetAddress(), hardhatUrl)
 	if err != nil {
-		panic(err)
+		panic(fmt.Errorf("connection to chain not established: %w", err))
 	}
 
 	messageservice := p2pms.NewMessageService("127.0.0.1", msgPort, *ourStore.GetAddress(), pk, logDestination)
@@ -99,7 +99,7 @@ func main() {
 		panic(err)
 	}
 
-	fmt.Println("Nitro as a Service listening on localhost:", rpcPort)
+	fmt.Println("Nitro as a Service listening on port", rpcPort)
 	sigs := make(chan os.Signal, 1)
 	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
 	sig := <-sigs
--- a/main.go
+++ b/main.go
@@ -6,7 +6,6 @@ import (
 	"encoding/hex"
 	"flag"
 	"fmt"
-	"log"
 	"math/big"
 	"os"
 	"os/signal"
@@ -45,9 +44,12 @@ func main() {
 	flag.StringVar(&hardhatUrl, "hardhatUrl", "ws://127.0.0.1:8545", "Specifies the url for the hardhat node.")
 	flag.IntVar(&msgPort, "msgPort", 3005, "Specifies the tcp port for the  message service.")
 	flag.IntVar(&rpcPort, "rpcPort", 4005, "Specifies the tcp port for the rpc server.")
+
 	flag.Parse()
+
 	pk := common.Hex2Bytes(pkString)
 	me := crypto.GetAddressFromSecretKeyBytes(pk)
+
 	logDestination := os.Stdout
 
 	var ourStore store.Store
@@ -57,7 +59,12 @@ func main() {
 	} else {
 		ourStore = store.NewMemStore(pk)
 	}
-	chainService := NewChainService(context.Background(), *ourStore.GetAddress(), hardhatUrl)
+
+	chainService, err := newChainService(context.Background(), *ourStore.GetAddress(), hardhatUrl)
+	if err != nil {
+		panic(err)
+	}
+
 	messageservice := p2pms.NewMessageService("127.0.0.1", msgPort, *ourStore.GetAddress(), pk)
 
 	node := client.New(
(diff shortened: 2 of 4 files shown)


client/client.go
```diff
@@ -100,6 +101,29 @@ func (c *Client) handleEngineEvents() {
 
 // Begin API
 
+// Version returns the go-nitro version
+func (c *Client) Version() string {
```
lalexgap, line 105: I wonder if we even need to bother with a `Version`  API anymore. If you're not using a built executable it returns `(devel)` which is not super helpful.  The `GetAddress` API call can be used instead of `Version` if we want some kind of easy sanity check. Similar to `Version` `GetAddress` also requires no parameters and it just returns a string.

main.go
```diff
@@ -0,0 +1,210 @@
...
+	"math/big"
+	"os"
+	"os/signal"
+	"syscall"
+
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/common"
+	ethcrypto "github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/ethclient"
+	hdwallet "github.com/miguelmota/go-ethereum-hdwallet"
```
NiloCK, line 18: The repo for this marked public archive as of about a year ago. A little surprising to have a security sensitive import like this from a personal github namespace.

main.go
```diff
@@ -0,0 +1,210 @@
...
+	Create2Deployer "github.com/statechannels/go-nitro/client/engine/chainservice/create2deployer"
+	p2pms "github.com/statechannels/go-nitro/client/engine/messageservice/p2p-message-service"
+	"github.com/statechannels/go-nitro/client/engine/store"
+	"github.com/statechannels/go-nitro/crypto"
+	"github.com/statechannels/go-nitro/rpc"
+	"github.com/statechannels/go-nitro/rpc/transport"
+	"github.com/statechannels/go-nitro/rpc/transport/nats"
+	"github.com/statechannels/go-nitro/rpc/transport/ws"
+	"github.com/statechannels/go-nitro/types"
+	"github.com/tidwall/buntdb"
```
NiloCK, line 33: Two nits:
- group bundtdb import w/ other external dependencies
- add blank line between external and internal dependencies

```suggestion
	"github.com/ethereum/go-ethereum/accounts/abi/bind"
	"github.com/ethereum/go-ethereum/common"
	ethcrypto "github.com/ethereum/go-ethereum/crypto"
	"github.com/ethereum/go-ethereum/ethclient"
	hdwallet "github.com/miguelmota/go-ethereum-hdwallet"
	"github.com/rs/zerolog"
	"github.com/tidwall/buntdb"
	
	"github.com/statechannels/go-nitro/client"
	"github.com/statechannels/go-nitro/client/engine"
	"github.com/statechannels/go-nitro/client/engine/chainservice"
	NitroAdjudicator "github.com/statechannels/go-nitro/client/engine/chainservice/adjudicator"
	Create2Deployer "github.com/statechannels/go-nitro/client/engine/chainservice/create2deployer"
	p2pms "github.com/statechannels/go-nitro/client/engine/messageservice/p2p-message-service"
	"github.com/statechannels/go-nitro/client/engine/store"
	"github.com/statechannels/go-nitro/crypto"
	"github.com/statechannels/go-nitro/rpc"
	"github.com/statechannels/go-nitro/rpc/transport"
	"github.com/statechannels/go-nitro/rpc/transport/nats"
	"github.com/statechannels/go-nitro/rpc/transport/ws"
	"github.com/statechannels/go-nitro/types"
```

main.go
```diff
@@ -0,0 +1,210 @@
...
+
+func main() {
+	var pkString, hardhatUrl string
+	var msgPort, rpcPort int
+	var useNats, useDurableStore bool
+
+	flag.BoolVar(&useNats, "useNats", true, "Specifies whether to use NATS or http/ws for the rpc server.")
+	flag.BoolVar(&useDurableStore, "useDurableStore", false, "Specifies whether to use a durable store or an in-memory store.")
+	flag.StringVar(&pkString, "pk", "2d999770f7b5d49b694080f987b82bbc9fc9ac2b4dcc10b0f8aba7d700f69c6d", "Specifies the private key for the client. Default is Alice's private key.")
+	flag.StringVar(&hardhatUrl, "hardhatUrl", "ws://127.0.0.1:8545", "Specifies the url for the hardhat node.")
```
NiloCK, line 44: Flags as the are sort of assume a testing environment. Maybe:

- hardhatUrl -> chainGateway 
- durableStore defaulting to `true` (annoying in the immediate, but definitely wants to be the default in future)

Not sure whether it matters, but I think flags usually prefer flat casing or hyphens. EG, `git log --oneline` or `git log --all-match`. Inconsistent exambples there, but both lower cased. (most seem to be hyphenated)

main.go
```diff
@@ -0,0 +1,210 @@
...
+	if useDurableStore {
+		dataFolder := fmt.Sprintf("./data/nitro-service/%s", me.String())
+		ourStore = store.NewDurableStore(pk, dataFolder, buntdb.Config{})
+	} else {
+		ourStore = store.NewMemStore(pk)
+	}
+
+	chainService, err := newChainService(context.Background(), *ourStore.GetAddress(), hardhatUrl)
+	if err != nil {
+		panic(err)
```
NiloCK, line 65:
```suggestion
		panic(fmt.Errorf("connection to chain not established: %w", err))
```

main.go
```diff
@@ -0,0 +1,210 @@
...
+		Str("client", ourStore.GetAddress().String()).
+		Str("rpc", "server").
+		Str("scope", "").
+		Logger()
+	_, err = rpc.NewRpcServer(&node, &logger, transport)
+	if err != nil {
+		panic(err)
+	}
+
+	fmt.Println("Nitro as a Service listening on localhost:", rpcPort)
```
NiloCK, line 102: less annoying than, eg, `localhost: 3452`

```suggestion
	fmt.Println("Nitro as a Service listening on port", rpcPort)
```

main.go
lalexgap: Happy to switch to this order, although I really wish the linter could handle this for me😞

NiloCK (approved):
This is generally great. It's very good to have a `main` function up and running.

Two general concerns:
1. a bunch of this code is directly geared toward go-nitro in context of testing environments, but mostly it isn't named or "quarantined" in any way that makes this obvious. EG: referencing hardhat's default funded accounts, deploying an adjudicator if not found, etc. 
2. The cli here receives a private key, which is used directly by the store & message service. But then a new key is derived for the chain service. This is incompatible with existing challenge workflow (at least - maybe other workflows) which requires that a challenge transaction is directly submitted by a channel participant. If the intention is toward ephemeral keys / keys per channel, then the chainservice probably needs to be asking the store for specific keys from each channel. (Also requires that challenging accounts are funded on chain for gas costs - bit of a can of worms).

I don't think either of these should necessarily block moving this in, but I think both points could stand to have some explicit planning around them.

main.go
lalexgap: Flags updated in https://github.com/statechannels/go-nitro/pull/1213/commits/0523e24f1a09ad1a1918c0e91ae72cf8d3b0c2e2

lalexgap:
> This is great. It's very good to have a `main` function up and running.
> 
> Two general concerns:
> 
>     1. a bunch of this code is directly geared toward go-nitro in context of testing environments, but mostly it isn't named or "quarantined" in any way that makes this obvious. EG: referencing hardhat's default funded accounts, deploying an adjudicator if not found, etc.

Fair point, we could:
- Add an option for a separate chain PK (instead of picking of a funded hardhat account)
- Add an option to allow auto deployment of the contract if its' not there.

>     2. The cli here receives a private key, which is used directly by the store & message service. But then a new key is derived for the chain service. This is incompatible with existing challenge workflow (at least - maybe other workflows) which requires that a challenge transaction is directly submitted by a channel participant. If the intention is toward ephemeral keys / keys per channel, then the chainservice probably needs to be asking the store for specific keys from each channel. (Also requires that challenging accounts are funded on chain for gas costs - bit of a can of worms).

Is this true? My understanding that was that anyone can submit the transaction, as long as the challenger has signed the "challenge" being submitted. Taking a quick look through the solidity code and I don't find anything that enforces the challenge transaction is submitted by the challenger (although I could be missing it!)

PR #1235, opened 2023-04-25
Add deploycontracts flag

lalexgap:
Adds a `deploycontracts` flag to the RPC server that determines whether the RPC deploys the Create2Deployer and Nitro Adjudicator. This allows the RPC server to work against any chain JSON-RPC node instead of relying on hardhat docker.

### DIFF
--- a/nitro-protocol/hardhat.config.ts
+++ b/nitro-protocol/hardhat.config.ts
@@ -61,7 +61,6 @@ const config: HardhatUserConfig & {watcher: any} = {
   networks: {
     hardhat: {
       chainId: 1337,
-      accounts: {count: 1000},
     },
     goerli: {
       url: infuraToken ? 'https://goerli.infura.io/v3/' + infuraToken : '',
--- a/nitro-protocol/hardhat.config.ts
+++ b/nitro-protocol/hardhat.config.ts
@@ -60,7 +60,8 @@ const config: HardhatUserConfig & {watcher: any} = {
   },
   networks: {
     hardhat: {
-      chainId: 31337,
+      chainId: 1337,
+      accounts: {count: 1000},
     },
     goerli: {
       url: infuraToken ? 'https://goerli.infura.io/v3/' + infuraToken : '',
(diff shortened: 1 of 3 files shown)


nitro-protocol/hardhat.config.ts
```diff
@@ -60,7 +60,7 @@ const config: HardhatUserConfig & {watcher: any} = {
   },
   networks: {
     hardhat: {
-      chainId: 31337,
+      chainId: 1337,
```
geoknee, line 63: Do we require / assume this `chainId` in go-nitro somewhere?

geoknee (approved):
LGTM. 

I'm wondering if we can push some of this code down into the chain service, to keep `main.go` nice and lean?

geoknee:
What happens if multiple nodes try to all deploy the adjudicator contract? The check for nonzero bytecode will fail if another node has submitted a tx to deploy the adjudicator, but that tx has not yet been mined.

I haven't yet found a definitive source, but it appears that a second deploy tx will **not revert** https://github.com/ethereum/go-ethereum/issues/17881 . I'm not sure if we want to change any of our code to exploit that...probably not.

nitro-protocol/hardhat.config.ts
lalexgap: Yup, most of our tests and the RPC server use a chainid of 1337. The RPC server should probably accept a chain id instead of using a hardcoded one.

lalexgap: Updated to accept a chain argument in https://github.com/statechannels/go-nitro/pull/1235/commits/fe77830642ee6968fc372e2022f2c643d0bfa564. If none is provided we use `1337`

lalexgap:
> LGTM.
> 
> I'm wondering if we can push some of this code down into the chain service, to keep `main.go` nice and lean?

Good idea, done in https://github.com/statechannels/go-nitro/pull/1235/commits/bb3962e9476c5289f860e8da987e430223922eff

PR #1238, opened 2023-04-25
WIP: Bring more harmony to `virtualfund`

geoknee:
By embedding a reference to the underlying virtual channel, rather than duplicating information directly on the objective. 

This means that things like finalization proofs can be read from the store by other tasks/objectives. I think this will be important when it comes to handling challenges.


geoknee:
In favour of #1241

Issue #1216, opened 2023-04-17
Ledger defund completes before virtual defund


kerzhner:
Virtual channel defunding includes finalizing the virtual channel and removing guarantees that fund the virtual channel from ledger channels. In the current go-nitro implementation, it is possible for the following scenario to occur:
1. Virtual defunding is initiated and all parties sign the final state.
2. Proposals for ledger channel guarantee removal are sent around.
3. When a participant has all proposals that the participant needs, the participant considers virtual defunding completed.
4. The participant can now initiate ledger direct defunding.
5. The counterparty in the ledger channel happily participates in ledger defunding even when its virtual defunding objective has not completed. It is possible to start (and even finish) ledger defunding before a virtual defund objective completes.

This brings up open questions:
1. Is there value in multiple objectives running for a single channel? If the answer is no, then it will be possible for objectives to be rejected due to other running objectives.
2. Do we need other guardrails for the types of objectives/states a go-nitro client accepts? As a note, there exists a [policymaker](https://github.com/statechannels/go-nitro/blob/e75d30833fead299085b0fcc39060baa310baa8c/client/engine/policy_maker.go#L15).
3. If multiple objectives for a channel can run simultaneously, should the objectives be aware of each other? More concretely, how does a ledger defund objective block until the virtual defund objective has completed?

Attached is a trimmed log file from [a failing CI unit test](https://github.com/statechannels/go-nitro/actions/runs/4694988561/attempts/1) that exposed this issues: [trimmed.log](https://github.com/statechannels/go-nitro/files/11256356/trimmed.log)

lalexgap:
> This brings up open questions:
> 
>  1. Is there value in multiple objectives running for a single channel? If the answer is no, then it will be possible for objectives to be rejected due to other running objectives.

I think our stores already implement the idea of an [objective owning a channel](https://github.com/statechannels/go-nitro/blob/1d2d74a89b9c45fad65b81f94a989fee56aec4a4/client/engine/store/memstore.go#L110), which means we should never have multiple active objectives for a channel.

> * The counterparty in the ledger channel happily participates in ledger defunding even when its virtual defunding objective has not completed. It is possible to start (and even finish) ledger defunding before a virtual defund objective completes.

I'm actually surprised the counterparty didn't `panic` due to the the `store` logic mentioned above.

lalexgap:
[...]

PR #1239, opened 2023-04-25
Add non fatal errors to the engine

kerzhner:
When initially developing `go-nitro`, we adopted the methodology of panicking the engine when an error is encountered. This is approach works well during the development phase as we avoid introducing any errors during normal nitro node operation.

In production, a nitro node should panic infrequently. The panic scenario is when the node reaches an irrecoverable state. This PR moves in that direction by introducing a class of errors for which the node logs an error but does not panic. We then use the new non-fatal errors to fix https://github.com/statechannels/go-nitro/issues/1216.

That issue involves a duplicate, delayed message causing the store to look up a deleted consensus channel. While many different approaches exist to solving this particular issue (such as not deleting consensus channels or merging consensus channels and vanilla channels), this PR choses to classify the failure to fetch an objective as a non fatal error.

### DIFF
--- a/channel/consensus_channel/consensus_channel.go
+++ b/channel/consensus_channel/consensus_channel.go
@@ -36,10 +36,10 @@ const (
 type ConsensusChannel struct {
 	// constants
 
-	MyIndex        ledgerIndex
-	fp             state.FixedPart
 	Id             types.Destination
+	MyIndex        ledgerIndex
 	OnChainFunding types.Funds
+	fp             state.FixedPart
 
 	// variables
 
@@ -894,7 +894,10 @@ func (c *ConsensusChannel) Clone() *ConsensusChannel {
 	for i, p := range c.proposalQueue {
 		clonedProposalQueue[i] = p.Clone()
 	}
-	d := ConsensusChannel{c.MyIndex, c.fp.Clone(), c.Id, c.OnChainFunding.Clone(), c.current.clone(), clonedProposalQueue}
+	d := ConsensusChannel{
+		MyIndex: c.MyIndex, fp: c.fp.Clone(),
+		Id: c.Id, OnChainFunding: c.OnChainFunding.Clone(), current: c.current.clone(), proposalQueue: clonedProposalQueue,
+	}
 	return &d
 }
 
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -34,6 +34,18 @@ func (uce *ErrUnhandledChainEvent) Error() string {
 	return fmt.Sprintf("chain event %#v could not be handled by objective %#v due to: %s", uce.event, uce.objective, uce.reason)
 }
 
+type ErrGetObjective struct {
+	wrappedError error
+	objectiveId  protocols.ObjectiveId
+}
+
+func (e *ErrGetObjective) Error() string {
+	return fmt.Sprintf("unexpected error getting/creating objective %s: %v", e.objectiveId, e.wrappedError)
+}
+
+// nonFatalErrors is a list of errors for which the engine should not panic
+var nonFatalErrors = []error{&ErrGetObjective{}}
+
 // Engine is the imperative part of the core business logic of a go-nitro Client
 type Engine struct {
 	// inbound go channels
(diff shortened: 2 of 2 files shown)


client/engine/engine.go
```diff
@@ -721,3 +727,22 @@ func (e *Engine) recordMessageMetrics(message protocols.Message) {
 	e.metrics.RecordQueueLength(fmt.Sprintf("msg_payload_size,sender=%s,receiver=%s", e.store.GetAddress(), message.To), totalPayloadsSize)
 	e.metrics.RecordQueueLength(fmt.Sprintf("msg_size,sender=%s,receiver=%s", e.store.GetAddress(), message.To), len(raw))
 }
+
+func (e *Engine) checkError(err error) {
+	if err != nil {
+
+		e.logger.Err(err).Msgf("%s, error in run loop", e.store.GetAddress())
+
+		for _, nonFatalError := range nonFatalErrors {
```
geoknee, line 736: Would it make sense to invert this logic, and instead of keeping a list of nonFatalErrors, keep a list of fatalErrors and check against that? I think in production we want the default to be non-fatal. So far we have (I think) many examples of non fatal errors, but not so many (none?) fatal error.

client/engine/engine.go
```diff
@@ -721,3 +727,22 @@ func (e *Engine) recordMessageMetrics(message protocols.Message) {
...
+			if errors.Is(err, nonFatalError) {
+				return
+			}
+		}
+
+		<-time.After(1000 * time.Millisecond) // We wait for a bit so the previous log line has time to complete
+		// TODO do not panic if in production.
```
geoknee, line 743: We could remove this TODO now, I think -- point being, we have pinched off errors that actually do require a panic in production.

client/engine/engine.go
```diff
@@ -721,3 +727,22 @@ func (e *Engine) recordMessageMetrics(message protocols.Message) {
...
+		panic(err)
+		// TODO report errors back to the consuming application
```
geoknee, line 745: I think we could move this TODO to L738. More important to report errors that are (currently) being swallowed.

channel/consensus_channel/consensus_channel.go
```diff
@@ -894,7 +894,10 @@ func (c *ConsensusChannel) Clone() *ConsensusChannel {
 	for i, p := range c.proposalQueue {
 		clonedProposalQueue[i] = p.Clone()
 	}
-	d := ConsensusChannel{c.MyIndex, c.fp.Clone(), c.Id, c.OnChainFunding.Clone(), c.current.clone(), clonedProposalQueue}
+	d := ConsensusChannel{
+		MyIndex: c.MyIndex, fp: c.fp.Clone(),
+		Id: c.Id, OnChainFunding: c.OnChainFunding.Clone(), current: c.current.clone(), proposalQueue: clonedProposalQueue,
+	}
```
geoknee, line 900: Nit: not opposed to this change, but seems very unrelated to the PR. Also, if you are forming literals with named arguments, the order of the fields doesn't matter... so I'm not really understanding the benefit?

client/engine/engine.go
kerzhner: I am open to either approach. My thinking was that for now we could still lean toward panicking to highlight unexpected errors

channel/consensus_channel/consensus_channel.go
kerzhner: The order of the fields of `consensus_channel` was changed to match that of channel. As a result, this struct construction started failing since the values were not named. Happy to move https://github.com/statechannels/go-nitro/pull/1239/commits/1ff9feb595192863d760c7d4b72a9adb6ee8a0fe to a separate PR if there is a strong preference.

client/engine/engine.go
geoknee: Sure, OK. Seems safer this way.

client/engine/engine.go
kerzhner: I chose to rephrase this todo and place it above the relevant line (the panic) ac9aaf36

geoknee: Maybe I'm confused, but I think we don't necessarily want to say "instead of panic" now. I believe we've got a model now where we want _some_ errors to result in a panic. Probably all errors should be sent to the API consumer / user.

geoknee (approved):

PR #1243, opened 2023-04-26
Specify node version for node related CI

lalexgap:
Specify a version of node (the LTS version of node) for the node-related CI workflows. This seems to resolve the [error](https://github.com/statechannels/go-nitro/actions/runs/4802488125/jobs/8565167719) we were seeing on main?


geoknee (approved):

geoknee:
Unfortunately it does not seem to have solved the problem https://github.com/statechannels/go-nitro/actions/runs/4818879629/jobs/8581399950 😞

Issue #1221, opened 2023-04-19
Don't panic


geoknee:
https://github.com/statechannels/go-nitro/blob/c7716670427e72948cf91a5dba75f20ecb5e19c3/client/engine/engine.go#L165-L170

geoknee:
<img width="294" alt="Screenshot 2023-04-24 at 15 20 11" src="https://user-images.githubusercontent.com/1833419/234024975-0d7d48d7-81a4-4be2-b7fb-e7caee95207b.png">

We have 48 calls to `panic()` to hunt down and remove.

geoknee:
Relevant to #1239

geoknee:
> One way to preserve that is to have an error channel / event stream that the consumer can listen to. In our tests we could setup a listener which fails the test as soon as an error occurs.

We could use an approach similar to this https://github.com/statechannels/go-nitro/pull/1223

geoknee:
https://github.com/statechannels/go-nitro/blob/0d72e0d6d8745ef860973ec7e12a5f80b345d11c/channel/state/state.go#L82-L96

Here's somewhere where we currently use a `panic` where we should arguably be returning an error. But doing so will involve putting error handling into dozens of call sites. 

Nevertheless I think we should do so: it is not infeasible to receive a malformated request or message which cannot be packed / abi-encoded. That could crash go-nitro, which would be bad.

PR #1249, opened 2023-05-02
Convert panics into errors

geoknee:
Towards #1221

### DIFF
--- a/protocols/messages.go
+++ b/protocols/messages.go
@@ -25,13 +25,13 @@ type PayloadType string
 
 // CreateObjectivePayload generates an objective message from the given objective id and payload.
 // CreateObjectivePayload handles serializing `p` into json.
-func CreateObjectivePayload(id ObjectiveId, payloadType PayloadType, p interface{}) ObjectivePayload {
+func CreateObjectivePayload(id ObjectiveId, payloadType PayloadType, p interface{}) (ObjectivePayload, error) {
 	b, err := json.Marshal(p)
 	if err != nil {
-		panic(err)
+		return ObjectivePayload{}, err
 	}
 
-	return ObjectivePayload{PayloadData: b, ObjectiveId: id, Type: payloadType}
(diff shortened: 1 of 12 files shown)


protocols/messages.go
```diff
@@ -24,13 +25,13 @@ type PayloadType string
 
 // CreateObjectivePayload generates an objective message from the given objective id and payload.
 // CreateObjectivePayload handles serializing `p` into json.
-func CreateObjectivePayload(id ObjectiveId, payloadType PayloadType, p interface{}) ObjectivePayload {
+func CreateObjectivePayload(id ObjectiveId, payloadType PayloadType, p interface{}) (ObjectivePayload, error) {
 	b, err := json.Marshal(p)
 	if err != nil {
-		panic(err)
+		return ObjectivePayload{}, err
```
NiloCK, line 31:
```suggestion
		return ObjectivePayload{}, fmt.Errorf("failed to marshal payload: %w", err)
```

Maybe? (fmt not already imported)

protocols/messages.go
```diff
@@ -183,8 +188,13 @@ func (m Message) Summarize() MessageSummary {
 
 	s.ProposalSummaries = make([]ProposalSummary, len(m.LedgerProposals))
 	for i, p := range m.LedgerProposals {
+		objId, err := GetProposalObjectiveId(p.Proposal)
+		objIdString := string(objId)
+		if err != nil {
+			objIdString = err.Error() // Use error message as objective id
```
NiloCK, line 194: Slightly cheeky but I think suitable for logging.

NiloCK (approved):
This LGTM. There seem to remain ~50 panic() calls in non-test code. Incrementally removing them is least painful way to do it.

protocols/messages.go
```diff
@@ -2,6 +2,7 @@ package protocols
 
 import (
 	"encoding/json"
+	"errors"
```
geoknee, line 5:
```suggestion
	"errors"
	"fmt"
```

PR #1250, opened 2023-05-04
Channel Update ADRS

lalexgap:
Adds two ADRS for the work done in #1223

### DIFF
--- a/.adr/0012-event-emission.md
+++ b/.adr/0012-event-emission.md
@@ -42,7 +42,7 @@ This approach would not be idiomatic and has the following drawbacks:
 
 A `sync.Cond` https://pkg.go.dev/sync#Cond is "a rendezvous point for goroutines waiting for or announcing the occurrence of an event.". It would allow events be emitted more than once by using the `Broadcast()` method. For example, if we wanted to have an "objective updated' event stream.
 
-There are many downsides to trying to build an event system with this type, however. Firstly, the event consumers would have to be goroutines. Secondly, no information can be passed with `Broadcast()`, so those goroutines would have to query information when they are woken up by `Broadcast()`. This introduces race conditions. Thirdly, usage of `sync.Cond` is veryy low and there are proposals to deprecate it https://github.com/golang/go/issues/21165.
+There are many downsides to trying to build an event system with this type, however. Firstly, the event consumers would have to be goroutines. Secondly, no information can be passed with `Broadcast()`, so those goroutines would have to query information when they are woken up by `Broadcast()`. This introduces race conditions. Thirdly, usage of `sync.Cond` is [very low](https://lukechampine.com/cond.html) and there are proposals to deprecate it https://github.com/golang/go/issues/21165.
 
 
 ## Decision
--- a/.adr/0012-event-emission.md
+++ b/.adr/0012-event-emission.md
@@ -40,14 +40,9 @@ This approach would not be idiomatic and has the following drawbacks:
 
 ### Condition variable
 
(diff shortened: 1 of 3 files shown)


.adr/0014-update-event-emission.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Alternatives considered
+
+### Using a Third Party Library
+
+We could use a third-party library, like [EventBus](https://github.com/asaskevich/EventBus), to handle the work of pubsub. However by writing our own eventhandler logic we can adjust it to our needs. One example is comparing the new event against the previously emitted event to prevent duplicate events being fired.
+
+### Sync.Cond
+
+While `sync.Cond` allows for broadcasting updates to multiple consumers, it requires the consumers and broadcaster to be goroutines. This means we'd have to spin up goroutines for each event listener which does not seem like a natural fit,.
```
geoknee, line 34: There's something that doesn't quite sit right with me here. 

In my head, listeners could quite naturally be go routines? When a listener subscribes, they get a chan which they can essentially use to trigger a callback by spawning a goroutine which first reads from the chan (and then does some action they want to be triggered by the event). If we were to use a `c := sync.Cond` , the listener can call `c.Wait()` to block until the event is fired and there's a similar pattern. 

Perhaps you have in mind that there are other patterns (maybe a `select` statement such as the one used in `checkNotifications`) which won't work so well with a `sync.Cond`? It seems to me that one can most likely work around that (you would still need goroutines, though).

geoknee: Thanks. Good point about not being able to communicate any data with `.Broadcast()`. 

I've taken the liberty of updating our discussion on `sync.Cond` across both ADRs [2f2d5c3](https://github.com/statechannels/go-nitro/pull/1250/commits/2f2d5c3e120ed84a38cb757a71b689580d400362).

geoknee:
Just a note that ADR numbering is broken (not the fault of this PR)

.adr/0012-event-emission.md
```diff
@@ -40,14 +40,9 @@ This approach would not be idiomatic and has the following drawbacks:
...
-
-Note in the godoc of this type: 
-
-> For many simple use cases, users will be better off using channels than a Cond (Broadcast corresponds to closing a channel, and Signal corresponds to sending on a channel).
-
-Note also that it has been proposed to [deprecate `sync.Cond`](https://github.com/golang/go/issues/21165).
+A `sync.Cond` https://pkg.go.dev/sync#Cond is "a rendezvous point for goroutines waiting for or announcing the occurrence of an event.". It would allow events be emitted more than once by using the `Broadcast()` method. For example, if we wanted to have an "objective updated' event stream.
 
+There are many downsides to trying to build an event system with this type, however. Firstly, the event consumers would have to be goroutines. Secondly, no information can be passed with `Broadcast()`, so those goroutines would have to query information when they are woken up by `Broadcast()`. This introduces race conditions. Thirdly, usage of `sync.Cond` is veryy low and there are proposals to deprecate it https://github.com/golang/go/issues/21165.
```
geoknee, line 45:
```suggestion
There are many downsides to trying to build an event system with this type, however. Firstly, the event consumers would have to be goroutines. Secondly, no information can be passed with `Broadcast()`, so those goroutines would have to query information when they are woken up by `Broadcast()`. This introduces race conditions. Thirdly, usage of `sync.Cond` is [very low](https://lukechampine.com/cond.html) and there are proposals to deprecate it https://github.com/golang/go/issues/21165.
```

PR #1256, opened 2023-05-04
lower default num of funded accounts to use

lalexgap:
[Anvil currently defaults to 10 funded accounts](https://book.getfoundry.sh/reference/anvil/), not 20. This updates the chain service utils use the first 10 derived PKs, not 20. This resolves an issue where a go-nitro RPC server could fail if it happened to grab an unfunded PK.


geoknee (approved):
Looks good but the godoc above needs some attention. 

We are still referencing:
- hardhat (does this make sense now?)
- modulating by 1000

also not sure "modulate" is correct usage?

lalexgap:
> Looks good but the godoc above needs some attention.
> 
> We are still referencing:
> 
>     * hardhat (does this make sense now?)
> 
>     * modulating by 1000
> 
> 
> also not sure "modulate" is correct usage?

Good point, updated in https://github.com/statechannels/go-nitro/pull/1256/commits/c13abb015f817993f18040953fa05c7b3e533e8d

PR #1261, opened 2023-05-08
add script to start rpc servers

lalexgap:
**Note:** This PR is based on #1259 

Adds a ts node script that handles spinning up anvil and 3 RPC servers. The goal is to simplify getting RPC servers up and running so it's easier to test things in `nitro-gui`. The script can be run by running `npx ts-node ./scripts/start-rpc-servers.ts` from the `go-nitro` directory.

The script assumes foundry is installed.  Foundry installation instructions can be found [here](https://book.getfoundry.sh/getting-started/installation)

### DIFF
--- a/scripts/start-rpc-servers.ts
+++ b/scripts/start-rpc-servers.ts
@@ -1,0 +1,67 @@
+#!/usr/bin/env ts-node
+import { ChildProcess, exec, spawn } from "child_process";
+
+enum Participant {
+  Alice = "alice",
+  Irene = "irene",
+  Bob = "bob",
+}
+enum Color {
+  black = "[30m",
+  red = "[31m",
+  green = "[32m",
+  yellow = "[33m",
+  blue = "[34m",
+  magenta = "[35m",
+  cyan = "[36m",
+  white = "[37m",
+  gray = "[90m",
+}
+
+// This is a simple script that:
+// - starts up a local anvil chain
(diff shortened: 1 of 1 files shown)


scripts/start-rpc-servers.ts
```diff
@@ -0,0 +1,67 @@
...
+}
+
+function setupRPCServer(participant: Participant, color: Color): ChildProcess {
+  let rpcServerProcess: ChildProcess;
+  switch (participant) {
+    case Participant.Alice:
+      rpcServerProcess = exec("go run . -autodeploy -usedurablestore");
+      break;
+    case Participant.Irene:
+      rpcServerProcess = exec(
```
geoknee, line 49: Will these processes all get tidied up nicely when the script exits?

geoknee (approved):
Nice!

Seems a little awkward to have a `.ts` script in `go-nitro` (notwithstanding our prior discussions about a TS/Go monorepo).

Did you consider doing this as a shell script or even in Go?

scripts/start-rpc-servers.ts
lalexgap: Yup, these child processes get automatically cleaned up when the node script complete. I've confirmed this by printing out the pid of a child process and checking it during and after running the RPC server.

lalexgap:
Closing in favour of the golang version #1266

PR #1266, opened 2023-05-09
add go version of `start-rpc-servers` script

lalexgap:
This takes the implementation of the ts-node script in #1261 and writes it in golang. It can be run from the `go-nitro` directory by running: `go run ./scripts/start-rpc-servers.go`.  This means we can avoid having to running `ts-node` in our golang specific repository.


lalexgap:
Moving back to draft as it looks like the running instances don't get cleaned up properly

kerzhner:
Looks awesome! Should we add a brief section to the readme about this command?

geoknee (approved):
🌋

geoknee:
> Looks awesome! Should we add a brief section to the readme about this command?

Yes and we should either link to if from here https://github.com/statechannels/nitro-gui#running-manual-integration-tests or just put the readme information in that repo.

PR #1268, opened 2023-05-09
Add more query functions to the API

lalexgap:
Adds two new query functions to the API and RPC API:
- `GetAllLedgerChannels` : Returns a `query.LedgerChannelInfo` for each ledger channel in the store. We check both the store's `consensuschannels`s as well as the store's `channels`.
- `GetPaymentChannelsByLedger` Returns all payment channels that are being actively funded by the ledger channel. We iterate through a ledger channel's `FundingTargets` and return a `query.PaymentChannelInfo` for each one.

The rpc test has been updated with some basic checks of these new query functions.

To support these new query functions a few functions have been added to the store interface. Allowing us to fetch channels in bulk.

### DIFF
--- a/client/engine/store/durablestore.go
+++ b/client/engine/store/durablestore.go
@@ -306,26 +306,31 @@ func (ds *DurableStore) GetChannelsByIds(ids []types.Destination) ([]*channel.Ch
 }
 
 // GetChannelsByAppDefinition returns any channels that include the given app definition
-func (ds *DurableStore) GetChannelsByAppDefinition(appDef types.Address) []*channel.Channel {
+func (ds *DurableStore) GetChannelsByAppDefinition(appDef types.Address) ([]*channel.Channel, error) {
 	toReturn := []*channel.Channel{}
+	var unmarshErr error
 	err := ds.channels.View(func(tx *buntdb.Tx) error {
-		err := tx.Ascend("", func(key, chJSON string) bool {
+		return tx.Ascend("", func(key, chJSON string) bool {
 			var ch channel.Channel
-			err := json.Unmarshal([]byte(chJSON), &ch)
-			if err != nil {
-				ds.checkError(err)
+			unmarshErr = json.Unmarshal([]byte(chJSON), &ch)
+			if unmarshErr != nil {
+				return false
 			}
 
 			if ch.AppDefinition == appDef {
 				toReturn = append(toReturn, &ch)
 			}
 
-			return true // channel not found: continue looking
+			return true
 		})
-		return err
 	})
-	ds.checkError(err)
-	return toReturn
+	if err != nil {
+		return []*channel.Channel{}, err
+	}
+	if unmarshErr != nil {
+		return []*channel.Channel{}, unmarshErr
+	}
+	return toReturn, nil
 }
 
 // GetChannelsByParticipant returns any channels that include the given participant
@@ -356,20 +361,25 @@ func (ds *DurableStore) GetChannelsByParticipant(participant types.Address) []*c
 
 func (ds *DurableStore) GetAllConsensusChannels() ([]*consensus_channel.ConsensusChannel, error) {
 	toReturn := []*consensus_channel.ConsensusChannel{}
+	var unmarshErr error
 	err := ds.consensusChannels.View(func(tx *buntdb.Tx) error {
 		return tx.Ascend("", func(key, chJSON string) bool {
 			var ch consensus_channel.ConsensusChannel
 
-			err := json.Unmarshal([]byte(chJSON), &ch)
-			if err != nil {
-				ds.checkError(err)
+			unmarshErr = json.Unmarshal([]byte(chJSON), &ch)
+			if unmarshErr != nil {
+				return false
 			}
 			toReturn = append(toReturn, &ch)
-			return true // channel not found: continue looking
+			return true
 		})
 	})
 	if err != nil {
 		return []*consensus_channel.ConsensusChannel{}, err
+	}
+
+	if unmarshErr != nil {
+		return []*consensus_channel.ConsensusChannel{}, unmarshErr
 	}
 	return toReturn, nil
 }
--- a/client/engine/store/durablestore.go
+++ b/client/engine/store/durablestore.go
@@ -266,6 +266,29 @@ func (ds *DurableStore) getChannelById(id types.Destination) (channel.Channel, e
 	return ch, nil
 }
 
+// GetChannelsByAppDefinition returns any channels that include the given participant
+func (ds *DurableStore) GetChannelsByAppDefinition(appDef types.Address) []*channel.Channel {
+	toReturn := []*channel.Channel{}
+	err := ds.channels.View(func(tx *buntdb.Tx) error {
+		err := tx.Ascend("", func(key, chJSON string) bool {
+			var ch channel.Channel
+			err := json.Unmarshal([]byte(chJSON), &ch)
+			if err != nil {
+				ds.checkError(err)
+			}
+
+			if ch.AppDefinition == appDef {
+				toReturn = append(toReturn, &ch)
+			}
+
+			return true // channel not found: continue looking
+		})
+		return err
+	})
+	ds.checkError(err)
+	return toReturn
+}
+
 // GetChannelsByParticipant returns any channels that include the given participant
 func (ds *DurableStore) GetChannelsByParticipant(participant types.Address) []*channel.Channel {
 	toReturn := []*channel.Channel{}
@@ -290,6 +313,26 @@ func (ds *DurableStore) GetChannelsByParticipant(participant types.Address) []*c
 	})
 	ds.checkError(err)
 	return toReturn
+}
+
+func (ds *DurableStore) GetAllConsensusChannels() ([]*consensus_channel.ConsensusChannel, error) {
+	toReturn := []*consensus_channel.ConsensusChannel{}
+	err := ds.consensusChannels.View(func(tx *buntdb.Tx) error {
+		return tx.Ascend("", func(key, chJSON string) bool {
+			var ch consensus_channel.ConsensusChannel
+
+			err := json.Unmarshal([]byte(chJSON), &ch)
+			if err != nil {
+				ds.checkError(err)
+			}
+			toReturn = append(toReturn, &ch)
+			return true // channel not found: continue looking
+		})
+	})
+	if err != nil {
+		return []*consensus_channel.ConsensusChannel{}, err
+	}
+	return toReturn, nil
 }
 
 // GetConsensusChannelById returns a ConsensusChannel with the given channel id
--- a/client/engine/store/durablestore.go
+++ b/client/engine/store/durablestore.go
@@ -305,6 +305,33 @@ func (ds *DurableStore) GetChannelsByIds(ids []types.Destination) ([]*channel.Ch
 	return toReturn, nil
 }
 
+// GetObjectiveByChannelIds returns a collection of objectives for the given channel ids
+func (ds *DurableStore) GetObjectiveByChannelIds(ids []types.Destination) (map[types.Destination]protocols.Objective, error) {
+	toReturn := map[types.Destination]protocols.Objective{}
+	var err error
+
+	txError := ds.channels.View(func(tx *buntdb.Tx) error {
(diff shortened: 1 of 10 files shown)


client/engine/store/durablestore.go
```diff
@@ -266,6 +266,29 @@ func (ds *DurableStore) getChannelById(id types.Destination) (channel.Channel, e
...
+			err := json.Unmarshal([]byte(chJSON), &ch)
+			if err != nil {
+				ds.checkError(err)
+			}
+
+			if ch.AppDefinition == appDef {
+				toReturn = append(toReturn, &ch)
+			}
+
+			return true // channel not found: continue looking
```
geoknee, line 323: This comment seems wrong: the channel may have been found. We do continue looking, of course.

client/engine/store/durablestore.go
```diff
@@ -292,6 +315,26 @@ func (ds *DurableStore) GetChannelsByParticipant(participant types.Address) []*c
...
+	err := ds.consensusChannels.View(func(tx *buntdb.Tx) error {
+		return tx.Ascend("", func(key, chJSON string) bool {
+			var ch consensus_channel.ConsensusChannel
+
+			err := json.Unmarshal([]byte(chJSON), &ch)
+			if err != nil {
+				ds.checkError(err)
+			}
+			toReturn = append(toReturn, &ch)
+			return true // channel not found: continue looking
```
geoknee, line 368: Remove "channel not found" in comment.

client/engine/store/memstore.go
```diff
@@ -179,6 +179,25 @@ func (ms *MemStore) getChannelById(id types.Destination) (channel.Channel, error
...
 }
 
+// GetChannelsByAppDefinition returns any channels that include the given participant
+func (ms *MemStore) GetChannelsByAppDefinition(appDef types.Address) []*channel.Channel {
+	toReturn := []*channel.Channel{}
+	ms.channels.Range(func(key string, chJSON []byte) bool {
+		var ch channel.Channel
+		err := json.Unmarshal(chJSON, &ch)
+		if err != nil {
+			return true // channel not found, continue looking
```
geoknee, line 220: There's different behaviour here in the memstore compared to the durablestore. In the latter, an unmarshaling error caused a panic. Here, it is essentially swallowed and we carry on.

client/engine/store/memstore.go
```diff
@@ -245,6 +264,26 @@ func (ms *MemStore) GetConsensusChannel(counterparty types.Address) (channel *co
...
+func (ms *MemStore) GetAllConsensusChannels() ([]*consensus_channel.ConsensusChannel, error) {
+	toReturn := []*consensus_channel.ConsensusChannel{}
+	var err error
+	ms.consensusChannels.Range(func(key string, chJSON []byte) bool {
+		var ch consensus_channel.ConsensusChannel
+
+		err = json.Unmarshal(chJSON, &ch)
+		if err != nil {
+			return false
+		}
```
geoknee, line 307: This seems like a better approach: stop iterating and then bubble the error up. Is there a reason we can't do that in all cases and in all our store implementations?

client/engine/store/memstore.go
```diff
@@ -245,6 +264,26 @@ func (ms *MemStore) GetConsensusChannel(counterparty types.Address) (channel *co
 	return
 }
 
+func (ms *MemStore) GetAllConsensusChannels() ([]*consensus_channel.ConsensusChannel, error) {
+	toReturn := []*consensus_channel.ConsensusChannel{}
+	var err error
```
geoknee, line 300: Nit: could call this `unmarshalErr` for more clarity?

client/engine/store/memstore.go
```diff
@@ -179,6 +179,37 @@ func (ms *MemStore) getChannelById(id types.Destination) (channel.Channel, error
...
+
+	ms.channels.Range(func(key string, chJSON []byte) bool {
+		var ch channel.Channel
+		err = json.Unmarshal(chJSON, &ch)
+		if err != nil {
+			return false
+		}
+
+		// If the channel is one of the ones we're looking for, add it to the list
+		if contains(ids, ch.Id) {
```
geoknee, line 196: Nit: I suppose there is an opportunity for optimization, here: no need to check the `ch.Id` against _all_ passed `ids` since any we have already found may be disregarded.

client/query/query.go
```diff
@@ -127,6 +127,60 @@ func GetPaymentChannelInfo(id types.Destination, store store.Store, vm *payments
...
+	toReturn := []LedgerChannelInfo{}
+
+	allConsensus, err := store.GetAllConsensusChannels()
+	if err != nil {
+		return []LedgerChannelInfo{}, err
+	}
+	for _, con := range allConsensus {
+		toReturn = append(toReturn, ConstructLedgerInfoFromConsensus(con))
+	}
+	allChannels := store.GetChannelsByAppDefinition(consensusAppDefinition)
```
geoknee, line 142: Seems a bit inconsistent that there is no error returned from this one?

client/query/query.go
```diff
@@ -127,6 +127,60 @@ func GetPaymentChannelInfo(id types.Destination, store store.Store, vm *payments
...
+	for _, c := range allChannels {
+		toReturn = append(toReturn, ConstructLedgerInfoFromChannel(c))
+	}
+	return toReturn, nil
+}
+
+func GetPaymentChannelsByLedger(ledgerId types.Destination, store store.Store, vm *payments.VoucherManager) ([]PaymentChannelInfo, error) {
+	// If a ledger channel is actively funding payment channels it must be in the form of a consensus channel
```
geoknee, line 151: Are we 100% sure about this? I think recently we encountered some raciness around defunding a ledger channel (the point when the consensus channel is destroyed and replaced by a regular channel).

client/query/query.go
```diff
@@ -127,6 +127,60 @@ func GetPaymentChannelInfo(id types.Destination, store store.Store, vm *payments
...
+	}
+	objectives, err := store.GetObjectiveByChannelIds(toQuery)
+	if err != nil {
+		return []PaymentChannelInfo{}, fmt.Errorf("could not query the store about ids %v: %w", toQuery, err)
+	}
+
+	toReturn := []PaymentChannelInfo{}
+	for _, p := range paymentChannels {
+		paid, remaining, err := GetVoucherBalance(p.Id, vm)
+		o := objectives[p.Id]
```
geoknee: Slightly out of scope: It would be really nice to not have to query objectives at all for this information. 

I think about objectives as storing:
* pointers to channels (which are the real source of truth and subject of any queries)
* extra metadata / convenience stuff which is more ephemeral / has a different nature.
* information about intentions, rather than facts

So anything important should in theory be stored on the channel itself.

rpc/server.go
```diff
@@ -118,6 +118,24 @@ func (rs *RpcServer) registerHandlers() (err error) {
 				}
 				return l
 			})
+		case serde.GetAllLedgerChannelsMethod:
+			return processRequest(rs, requestData, func(r serde.NoPayloadRequest) []query.LedgerChannelInfo {
+				ledgers, err := rs.client.GetAllLedgerChannels()
+				if err != nil {
+					// TODO: What's the best way to handle this error?
```
geoknee, line 125: Send back an error response to the RPC client?

client_test/rpc_test.go
```diff
@@ -235,6 +242,22 @@ func checkQueryInfo[T channelInfo](t *testing.T, expected T, fetched T) {
...
+	}
+	found := false
+	for _, fetched := range fetched {
+		if cmp.Equal(expected, fetched, cmp.AllowUnexported(big.Int{})) {
+			found = true
+			break
+		}
+	}
+	if !found {
+		panic(fmt.Errorf("did not find info %v in channel infos: %v", expected, fetched))
```
geoknee, line 270: is `t.Fatalf` better than `panic`?

client/query/query.go
```diff
@@ -144,16 +145,20 @@ func GetAllLedgerChannels(store store.Store, consensusAppDefinition types.Addres
...
 }
 
-func GetPaymentChannelsByLedger(ledgerId types.Destination, store store.Store, vm *payments.VoucherManager) ([]PaymentChannelInfo, error) {
+func GetPaymentChannelsByLedger(ledgerId types.Destination, s store.Store, vm *payments.VoucherManager) ([]PaymentChannelInfo, error) {
 	// If a ledger channel is actively funding payment channels it must be in the form of a consensus channel
-	con, err := store.GetConsensusChannelById(ledgerId)
+	con, err := s.GetConsensusChannelById(ledgerId)
+	// If the ledger channel is not a consensus channel we know that there are no payment channels funded by it
+	if errors.Is(err, store.ErrNoSuchChannel) {
+		return []PaymentChannelInfo{}, nil
```
geoknee, line 155: Not sure what the correct behaviour ought to be here. If I query the payment channels for some made up ledger id, should i get an empty list or an error?

client/query/query.go
```diff
@@ -127,6 +128,62 @@ func GetPaymentChannelInfo(id types.Destination, store store.Store, vm *payments
...
+
+	toReturn := []PaymentChannelInfo{}
+	for _, p := range paymentChannels {
+		paid, remaining, err := GetVoucherBalance(p.Id, vm)
+		if err != nil {
+			return []PaymentChannelInfo{}, err
+		}
+		// TODO: n+1 query problem
+		// We should query for the vfos in bulk, rather than one at a time
+		// Or we should be able to determine the status soley from the channel
```
geoknee, line 176: Strong preference to get status from the channel.

geoknee (approved):
Looks pretty good, I left some comments / queries. Be good to get a consistent approach to error handling sorted.

client/query/query.go
lalexgap: There's two possibilities here:
1. Querying for a made up ledger id.
2. Querying for a ledger channel that is no longer a consensus channel(Ie: it's being defunded)

I wanted to avoid throwing an error in 2.

client_test/rpc_test.go
lalexgap: > is t.Fatalf better than panic?

I think `t.Fatalf` is our preferred option, I think I switched to to `panic` so I could easily identify the error I'll switch it back

. One problem with `t.Fatalf` is that it doesn't give a full stack trace, so if it fails, it's often hard to figure out which call to your test function was the one causing the problem.

client/engine/store/memstore.go
lalexgap: Good point. Whenever we find a channel we could remove it from `ids`

client/engine/store/memstore.go
lalexgap: I think the behaviour here is incorrect, it should panic/error like the memstore.

client/engine/store/memstore.go
lalexgap: I think this makes sense to apply to other implementations. For the purposes of this PR I'll update the the functions added.

client/engine/store/memstore.go
```diff
@@ -179,6 +179,25 @@ func (ms *MemStore) getChannelById(id types.Destination) (channel.Channel, error
 	return ch, nil
 }
 
+// GetChannelsByAppDefinition returns any channels that include the given participant
```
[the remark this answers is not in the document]
lalexgap: Updated in fa91d1c96e7218de789062e00b67628be9bb4a0a

client/engine/store/durablestore.go
lalexgap: Resolved by https://github.com/statechannels/go-nitro/pull/1268/commits/79d587861afd733bc95cbdb634e6c635e7ee5521

client/engine/store/durablestore.go
```diff
@@ -266,6 +266,29 @@ func (ds *DurableStore) getChannelById(id types.Destination) (channel.Channel, e
 	return ch, nil
 }
 
+// GetChannelsByAppDefinition returns any channels that include the given participant
```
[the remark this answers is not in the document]
lalexgap: resolved by https://github.com/statechannels/go-nitro/pull/1268/commits/fa91d1c96e7218de789062e00b67628be9bb4a0a

rpc/server.go
lalexgap: Created #1275

PR #1269, opened 2023-05-10
Use `urfave/cli/v2` in `main.go`

### LINKED ISSUE #1205: Create a production constructor
I think a production constructor is very important and it would really help clarify all of this. 

The lack of the existence of such a constructor had me reading the test files as "the only way" we do things at the moment. I take your point that in the wild we wouldn't do it that way.

As discussed during standup -- the production constructor could read from a config file, plus read sensitive information (private key) from the command line.

_Originally posted by @geoknee in https://github.com/statechannels/go-nitro/issues/1201#issuecomment-1492315623_

geoknee:
⚠️ Based on #1267 

Closes #1205 (for now).

This changeset updates the `main.go` Nitro as a service entrypoint to use [a popular CLI library](https://cli.urfave.org/) (used by both `geth` and `lotus`).  It's similar to `yargs` npm package.

Advantages:
* help / `man` pages automatically generated
* ability to read from a config file (examples checked in, I updated the test orchestration script to read from 3x config files)
* rich featureset which we can use in the future
    - required flags
    - input validation
    
 This change is backwards compatible with the existing entrypoint.


scripts/start-rpc-servers.go
```diff
@@ -114,29 +114,21 @@ func waitForKillSignal() {
...
-		args = append(args, "-rpcport", "4005")
-		args = append(args, "-pk", "2d999770f7b5d49b694080f987b82bbc9fc9ac2b4dcc10b0f8aba7d700f69c6d")
-
+		args = append(args, "-config", "./scripts/test-configs/alice.toml")
 	case irene:
-		args = append(args, "-msgport", "3006")
-		args = append(args, "-rpcport", "4006")
-		args = append(args, "-pk", "febb3b74b0b52d0976f6571d555f4ac8b91c308dfa25c7b58d1e6a7c3f50c781")
-
+		args = append(args, "-config", "./scripts/test-configs/irene.toml")
```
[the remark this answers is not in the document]
geoknee: In production we would advise users to supply the key at the command line, not by storing it in a file. We could even make that a requirement. 

In testing, it is more convenient to use a file (in either case we have checked in the test pk to github already).

scripts/test-configs/alice.toml
```diff
@@ -0,0 +1,8 @@
+# Alice's test Config
+
+usedurablestore = true
```
lalexgap, line 3: If we removed this (and maybe let the script handle it) the test config files only contain actor specific info.

lalexgap (approved):

scripts/test-configs/alice.toml
geoknee: I considered that, but in the end I liked a more production-y setup which was more explicit and flexible. Definitely an option, though.

PR #1271, opened 2023-05-11
Remove channel info status dependency on objectives

lalexgap:
Due to ADR 0009 we were checking for the status of a `virtualfund` objective to determine is a channel was `Ready` or `Proposed`.  However this becomes awkward when you want to construct `PaymentChannelnfo` for a collection of `channel.Channel`s. You have to fetch an objective for each channel (or introduce a store function to fetch them in bulk). Instead this PR removes the objective check when constructing a `PaymentChannelInfo` and replaces it with a specific ADR 0009 check that only relies on the channel.

This makes it much simpler to call `ConstructPaymentInfo` as you no longer need to pass in a `virtualfund.Objective`.

### DIFF
--- a/channel/channel.go
+++ b/channel/channel.go
@@ -182,6 +182,11 @@ func (c Channel) PostFundComplete() bool {
 	return c.SignedStateForTurnNum[PostFundTurnNum].HasAllSignatures()
 }
 
+// PostFundSignedBy returns true if the participant with the given index has signed the post fund setup state, false otherwise.
+func (c Channel) PostFundSignedBy(participantIndex uint) bool {
+	return c.SignedStateForTurnNum[PostFundTurnNum].HasSignatureForParticipant(participantIndex)
+}
+
 // FinalSignedByMe returns true if the calling client has signed a final state, false otherwise.
 func (c Channel) FinalSignedByMe() bool {
 	for _, ss := range c.SignedStateForTurnNum {
--- a/channel/channel.go
+++ b/channel/channel.go
@@ -182,11 +182,6 @@ func (c Channel) PostFundComplete() bool {
 	return c.SignedStateForTurnNum[PostFundTurnNum].HasAllSignatures()
 }
 
-// PostFundSignedBy returns true if the participant with the given index has signed the post fund setup state, false otherwise.
-func (c Channel) PostFundSignedBy(participantIndex uint) bool {
-	return c.SignedStateForTurnNum[PostFundTurnNum].HasSignatureForParticipant(participantIndex)
-}
-
 // FinalSignedByMe returns true if the calling client has signed a final state, false otherwise.
 func (c Channel) FinalSignedByMe() bool {
 	for _, ss := range c.SignedStateForTurnNum {
--- a/client/query/query.go
+++ b/client/query/query.go
@@ -225,7 +225,7 @@ func ConstructLedgerInfoFromChannel(c *channel.Channel) LedgerChannelInfo {
 func ConstructPaymentInfo(c *channel.Channel, paid, remaining *big.Int) (PaymentChannelInfo, error) {
 	status := getStatusFromChannel(c)
 	// ADR 0009 allows for intermediaries to exit the protocol before receiving all signed post funds
-	// So for intermediaries we return ready once they have signed their post fund state
+	// So for intermediaries we return Open once they have signed their post fund state
 	amIntermediary := c.MyIndex != 0 && c.MyIndex != uint(len(c.Participants)-1)
 	if amIntermediary && c.PostFundSignedByMe() {
 		status = Open
--- a/client/query/query.go
+++ b/client/query/query.go
@@ -31,11 +31,9 @@ func getStatusFromChannel(c *channel.Channel) ChannelStatus {
 	return Ready
 }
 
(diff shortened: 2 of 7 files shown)


channel/channel.go
```diff
@@ -182,6 +182,11 @@ func (c Channel) PostFundComplete() bool {
 	return c.SignedStateForTurnNum[PostFundTurnNum].HasAllSignatures()
 }
 
+// PostFundSignedBy returns true if the participant with the given index has signed the post fund setup state, false otherwise.
+func (c Channel) PostFundSignedBy(participantIndex uint) bool {
+	return c.SignedStateForTurnNum[PostFundTurnNum].HasSignatureForParticipant(participantIndex)
+}
```
geoknee, line 188: I think in the end you didn't use this (dead code)?

geoknee (commented):
I think the thrust of this is super important, but I prepared an alternative solution over on #1272

client/query/query.go
```diff
@@ -24,12 +24,24 @@ func getStatusFromChannel(c *channel.Channel) ChannelStatus {
...
+
 	if !c.PostFundComplete() {
 		return Proposed
 	}
 	return Ready
 }
 
+func getVirtualChannelStatus(c *channel.Channel, virtualAppDef types.Address) ChannelStatus {
+	// ADR 0009: Allows for intermediaries to exit the protocol before receiving all signed post funds
+	isVirtual := c.AppDefinition == virtualAppDef
```
geoknee, line 36: Is it a bit awkward to call the function `getVirtualChannelStatus` and then test if the channel isVirtual inside the function body?

I also note that we have a `VirtualChannel` type, which we could use to differentiate / handle this as a special case. 

I think overall I am a bit wary of having a different status reported magically for intermediaries, when the underlying channel state is quite different. On my alternative #1272, I addressed this by introducing more fine-grained statuses to use.

geoknee: Having thought about it some more, I'm coming around to the idea that the query status can deviate from the "true channel status".

geoknee: I do think we want to tidy up the fn though, seems quite messy to have to thread `virtualAppDef` through everywhere. If we could use something like a type assertion / type check instead that might make it much cleaner.

geoknee:
> I think the thrust of this is super important, but I prepared an alternative solution over on #1272

I closed the alternative, so let's move this PR on towards merging it.

client/query/query.go
lalexgap: Actually I think we could just remove this check for now. Nowhere else in `GetPaymentChannelInfo` do we ensure that the channel is a virtual channel. So it's strange to do a specific check when getting a virtual channel status.

I do think we should have a check to make sure the channel is the right type, but that can be tackled outside of this PR.

lalexgap: I've updated this in https://github.com/statechannels/go-nitro/pull/1271/commits/32bf07346cb1d4e3621220cfde4067afefd6faf4 so we no longer check the `AppDef` or pass around the virtual app address.

I've also created #1280.

client/query/query.go
```diff
@@ -24,10 +24,11 @@ func getStatusFromChannel(c *channel.Channel) ChannelStatus {
 		}
 		return Closing
 	}
+
 	if !c.PostFundComplete() {
 		return Proposed
 	}
-	return Ready
+	return Open
```
lalexgap, line 31: I liked the term `Open` from #1272 so I've renamed `Ready` to `Open

client/query/query.go
```diff
@@ -224,14 +222,15 @@ func ConstructLedgerInfoFromChannel(c *channel.Channel) LedgerChannelInfo {
...
-func ConstructPaymentInfo(c *channel.Channel, vfo *virtualfund.Objective, paid, remaining *big.Int) (PaymentChannelInfo, error) {
+func ConstructPaymentInfo(c *channel.Channel, paid, remaining *big.Int) (PaymentChannelInfo, error) {
 	status := getStatusFromChannel(c)
-	if vfo != nil && vfo.Status == protocols.Completed {
-		// This means intermediaries may not have a fully signed postfund state even though the channel is "ready"
-		// To determine the the correct status we check the status of the virtual fund objective
-
-		status = Ready
+	// ADR 0009 allows for intermediaries to exit the protocol before receiving all signed post funds
+	// So for intermediaries we return ready once they have signed their post fund state
```
geoknee, line 228:
```suggestion
	// So for intermediaries we return Open once they have signed their post fund state
```

client/query/query.go
```diff
@@ -224,14 +222,15 @@ func ConstructLedgerInfoFromChannel(c *channel.Channel) LedgerChannelInfo {
...
+	amIntermediary := c.MyIndex != 0 && c.MyIndex != uint(len(c.Participants)-1)
+	if amIntermediary && c.PostFundSignedByMe() {
+		status = Open
 	}
```
geoknee, line 232: Just a general comment here: can we think of any way to try and keep this code in sync with what's going on in the protocol/objective?

Could we push the "ready to be executed" logic down into a method on the `Channel`/`VirtualChannel` class, and call into that from both i) the protocol and ii) the query package?

We don't need to block this PR on doing that, but if you like the idea we could make an issue or follow it up with a PR.

geoknee (approved):

PR #1272, opened 2023-05-11
Add an "enabled" channel status -> channel status need not depend on any objectives.

geoknee:
Alternative to #1271

### DIFF
--- a/channel/channel.go
+++ b/channel/channel.go
@@ -29,14 +29,14 @@ type Channel struct {
 	latestSupportedStateTurnNum uint64 // largest uint64 value reserved for "no supported state"
 }
 
-type ChannelStatus string
+type Status string
 
 const (
-	Proposed ChannelStatus = "Proposed"
-	Enabled  ChannelStatus = "Enabled"
-	Open     ChannelStatus = "Open"
-	Closing  ChannelStatus = "Closing"
-	Closed   ChannelStatus = "Closed"
+	Proposed Status = "Proposed"
+	Enabled  Status = "Enabled"
+	Open     Status = "Open"
+	Closing  Status = "Closing"
+	Closed   Status = "Closed"
 )
 
 // New constructs a new Channel from the supplied state.
@@ -123,7 +123,7 @@ func (c *Channel) UnmarshalJSON(data []byte) error {
 }
 
 // getStatusFromChannel returns the status of the channel
-func (c *Channel) Status() ChannelStatus {
+func (c *Channel) Status() Status {
 	if c.FinalSignedByMe() {
 		if c.FinalCompleted() {
 			return Closed
--- a/channel/channel.go
+++ b/channel/channel.go
@@ -29,6 +29,16 @@ type Channel struct {
 	latestSupportedStateTurnNum uint64 // largest uint64 value reserved for "no supported state"
 }
 
+type ChannelStatus string
+
+const (
+	Proposed ChannelStatus = "Proposed"
+	Enabled  ChannelStatus = "Enabled"
+	Open     ChannelStatus = "Open"
+	Closing  ChannelStatus = "Closing"
+	Closed   ChannelStatus = "Closed"
+)
+
 // New constructs a new Channel from the supplied state.
 func New(s state.State, myIndex uint) (*Channel, error) {
 	c := Channel{}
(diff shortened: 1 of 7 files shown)


client/query/query.go
```diff
@@ -20,14 +20,17 @@ import (
...
-		return Proposed
+	if c.PostFundSignedByMe() {
+		if c.PostFundComplete() {
+			return Open
+		}
+		return Enabled
 	}
-	return Ready
+	return Proposed
 }
```
geoknee, line 34: Here's the nub of the difference in approach.

geoknee:
There's an issue with `checkNotifications`.

We have statuses that differentiate between having _my_ signature vs having _all_ signatures on a (postfund,final) atate. Whether or not your engine actually cranks through one and then the other depends on message ordering (non deterministic). If I am the last to sign, I'll go from not having my signature, to having all without touching the intermediate state. 

So it would be acceptable to see _either_ `closing, closed` or _just_ `closed` (etc) in a test. 

`checkNotifications` requires the test author to specify _exactly_ what the notifications should be.

channel/channel.go
```diff
@@ -112,6 +122,23 @@ func (c *Channel) UnmarshalJSON(data []byte) error {
...
+func (c *Channel) Status() Status {
+	if c.FinalSignedByMe() {
+		if c.FinalCompleted() {
+			return Closed
+		}
+		return Closing
+	}
+	if c.PostFundSignedByMe() {
+		if c.PostFundComplete() {
+			return Open
```
lalexgap, line 135: In the case of an intermediary `c.PostFundComplete` could never be completed, so they would never get back a status of `Open`.

geoknee: I think it could be completed some times, but you are right that it would often remain incomplete.

I think it is OK -- but for sure it is leaking something which seems like an implementation detail across the API.

We have a mixture of information from the channel and from the objective. As long as we are focussed on the channel (and want to rip out any dependence on objective), I think we would probably have to be ok with that.

lalexgap:
I think my concern with this is that it seems like we're letting technical implementation details drive the statuses we present. It seems  like we're adding a specific status for `PostFundSignedByMe` because of ADR 0009, not because we think that status will be helpful or meaningful otherwise.

I think it also makes things slightly confusing. If I'm a hub I can have:
- a ledger channel with a status of `Enabled`  that is not ready to use yet (direct-fund requires `PostFundComplete`)
- a virtual channel with a status of `Enabled` that is being actively used. It will never be `Open`. (virtual-fund requires `PostFundSignedByMe` when intermediary)

I'm biased towards hiding the complexity and using proxies with simple statuses (as partially outlined in [RFC 23](https://www.notion.so/statechannels/RFC-23-Overhaul-go-nitro-API-2c83cd05f59946ebb7354296de09f30e?pvs=4#2c2feabbaf944775b15824bae396e300)), so I would prefer not to add a new status if we could.  In my view a `PaymentChannelInfo` or `LedgerChannelInfo` is a combination of channel, objective and voucher information. For performance reasons and simplicity it would be nice to avoid looking up objectives for a channel if we can, but I don't we should think of a `ChannelInfo` as separate from an objective.

geoknee:
> I think my concern with this is that it seems like we're letting protocol implementation details drive the statuses we present. It seems like we're adding a specific status for `PostFundSignedByMe` because of ADR 0009, not because we think that status will be helpful or meaningful otherwise.
> 
> I think it also makes things slightly confusing. If I'm a hub I can have:
> 
> * a ledger channel with a status of `Enabled`  that is not ready to use yet (direct-fund requires `PostFundComplete`)
> * a virtual channel with a status of `Enabled` that is being actively used. It will never be `Open`. (virtual-fund requires `PostFundSignedByMe` when intermediary)
> 
> I'm biased towards hiding the complexity and using proxies with simple statuses (as partially outlined in [RFC 23](https://www.notion.so/statechannels/RFC-23-Overhaul-go-nitro-API-2c83cd05f59946ebb7354296de09f30e?pvs=4#2c2feabbaf944775b15824bae396e300)), so I would prefer not to add a new status if we could. In my view a `PaymentChannelInfo` or `LedgerChannelInfo` is a combination of channel, objective and voucher information. For performance reasons and simplicity it would be nice to avoid looking up objectives for a channel if we can, but I don't we should think of a `ChannelInfo` as separate from an objective.

It's not so much that ADR 9 is driving things. It's that our protocols imply that the true off chain state in the store will be a certain way. The question is do we want to reflect all of that detail and subtlety in the query API. I think you are arguing that we do _not_ -- and I think that's a totally valid view, maybe even the best choice. 

I'll close this for now. I am actually wondering though if we can lean into the 'simple API' call even harder. I wonder if we could:
- rip out objectives entirely from the API (this is proposed in RFC 23)
- simplify the query channel status even more -- it could be `Ready` / `NotReady`. The status could merely reflect whether other API calls that "use" this channel are likely to succeed. Perhaps that is too simple... but I like the idea of moving away from anything remotely suggestive of implementation details. So maybe status names which pick out combinations of "can update" / "can close" / "can challenge" / etc.

An important feature of the API is that it helps consumers orchestrate their use of it.

PR #1278, opened 2023-05-12
Minor improvemenets to `pk` and `chainpk` test config

geoknee:


main.go
```diff
@@ -121,21 +117,29 @@ func main() {
 		Flags:  flags,
 		Before: altsrc.InitInputSourceWithContext(flags, altsrc.NewTomlSourceFromFlagFunc(CONFIG)),
 		Action: func(cCtx *cli.Context) error {
+			if pkString == "" {
```
lalexgap, line 120: I think if we set `Required: true` when constructing the flag then `urfave` will handle making sure that a value is provided. (See `naaddress`)

geoknee: https://github.com/urfave/cli/issues/849

PR #1279, opened 2023-05-12
Clean up chain service construction

### LINKED ISSUE #1270: It seems bad that the chain service requires a `NitroAdjudicator` to be passed in. It should probably construct that itself?
Out of scope, but it seems bad that the chain service requires a `NitroAdjudicator` to be passed in. It should probably construct that itself?

_Originally posted by @geoknee in https://github.com/statechannels/go-nitro/pull/1267#discussion_r1189496650_

geoknee:
Closes #1270

This removes `chainId` entirely from the "production constructor".

NOTE: needs some changes in testground repo to get CI passing

### DIFF
--- a/client/engine/chainservice/eth_chainservice.go
+++ b/client/engine/chainservice/eth_chainservice.go
@@ -15,6 +15,7 @@ import (
 	"github.com/rs/zerolog"
 	NitroAdjudicator "github.com/statechannels/go-nitro/client/engine/chainservice/adjudicator"
 	Token "github.com/statechannels/go-nitro/client/engine/chainservice/erc20"
+	chainutils "github.com/statechannels/go-nitro/client/engine/chainservice/utils"
 	"github.com/statechannels/go-nitro/internal/logging"
 	"github.com/statechannels/go-nitro/protocols"
 	"github.com/statechannels/go-nitro/types"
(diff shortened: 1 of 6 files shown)


client/engine/chainservice/eth_chainservice.go
```diff
@@ -59,6 +60,21 @@ const MAX_QUERY_BLOCK_RANGE = 2000
 // See https://github.com/ethereum/go-ethereum/blob/e14164d516600e9ac66f9060892e078f5c076229/eth/filters/filter_system.go#L43
 const RESUB_INTERVAL = 2*time.Minute + 30*time.Second
 
+// NewEthChainService2 is a convenient wrapper around NewEthChainService, which provides a simpler API
+func NewEthChainService2(chainUrl, chainPk string, naAddress, caAddress, vpaAddress common.Address, logDestination io.Writer) (*EthChainService, error) {
```
geoknee, line 64: Open to suggestions for a better name. 

Initially I had wanted to replace (not wrap) `NewEthChainService`... but this causes problems:
* the simulate back end chain service 
* the FEVM tests
want access to the inner constructor.

lalexgap: I think the simulated and FEVM tests are local to the `chainservice` package, so could we change the old constructor to a private one?

geoknee: Oh that would be nice. Looks like maybe testground uses it, too -- but I haven't dug in to see if that would be easy to fix.

geoknee: These are some changes that should get things working again with testground https://github.com/statechannels/go-nitro-testground/pull/166

geoknee: > I think the simulated and FEVM tests are local to the `chainservice` package, so could we change the old constructor to a private one?

[3a3f91e](https://github.com/statechannels/go-nitro/pull/1279/commits/3a3f91e6384fe7c0e3b4213373a856d0bf13ae1d)

lalexgap (approved):

PR #1287, opened 2023-05-16
Proliferate `VirtualChannel` type

### LINKED ISSUE #1280: Query functions should check channel is right "type"
Actually I think we could just remove this check for now. Nowhere else in `GetPaymentChannelInfo` do we ensure that the channel is a virtual channel. So it's strange to do a specific check when getting a virtual channel status.

I do think we should have a check to make sure the channel is the right type, but that can be tackled outside of this PR.

_Originally posted by @lalexgap in https://github.com/statechannels/go-nitro/pull/1271#discussion_r1192635414_

### LINKED ISSUE #1285: Consolidate status logic
Just a general comment here: can we think of any way to try and keep this code in sync with what's going on in the protocol/objective?

Could we push the "ready to be executed" logic down into a method on the `Channel`/`VirtualChannel` class, and call into that from both i) the protocol and ii) the query package?

We don't need to block this PR on doing that, but if you like the idea we could make an issue or follow it up with a PR.

_Originally posted by @geoknee in https://github.com/statechannels/go-nitro/pull/1271#discussion_r1192703608_

geoknee:
Closes #1280
Closes #1285 

Note: one awkward thing at present is that there is no discriminator for `Channel/VirtualChannel` in the store. This PR continues with the widespread practice of casting / wrapping / unwrapping to/from `VirtualChannel` based on the context (i.e. is a virtual protocol calling set or get). 

We could consider either:
* a distinct table in the store for virtual channels
* a discriminating enum that is serialized with the channel, and can be used by the store to know how to correctly hydrate data of the right type.

### DIFF
--- a/channel/channel.go
+++ b/channel/channel.go
@@ -321,3 +321,27 @@ func (c *Channel) SignAndAddState(s state.State, sk *[]byte) (state.SignedState,
 	}
 	return ss, nil
 }
+
+type Status string
+
+// TODO: Think through statuses
+const (
+	Proposed Status = "Proposed"
+	Open     Status = "Open"
+	Closing  Status = "Closing"
+	Complete Status = "Complete"
+)
+
+func (c *Channel) Status() Status {
+	if c.FinalSignedByMe() {
+		if c.FinalCompleted() {
+			return Complete
+		}
+		return Closing
+	}
+
+	if !c.PostFundComplete() {
+		return Proposed
+	}
+	return Open
+}
--- a/channel/virtual.go
+++ b/channel/virtual.go
@@ -39,3 +39,15 @@ func (v *VirtualChannel) Clone() *VirtualChannel {
 	w := VirtualChannel{*v.Channel.Clone()}
 	return &w
 }
+
+func (v *VirtualChannel) Status() Status {
+	s := v.Channel.Status()
+
+	// ADR 0009 allows for intermediaries to exit the protocol before receiving all signed post funds
+	// So for intermediaries we return Open once they have signed their post fund state
+	amIntermediary := v.MyIndex != 0 && v.MyIndex != uint(len(v.Participants)-1)
+	if amIntermediary && v.PostFundSignedByMe() {
+		s = Open
+	}
+	return s
+}
--- a/client/engine/engine.go
+++ b/client/engine/engine.go
@@ -372,7 +372,7 @@ func (e *Engine) handleMessage(message protocols.Message) (EngineEvent, error) {
 		if err != nil {
 			return EngineEvent{}, err
 		}
-		info, err := query.ConstructPaymentInfo(c, paid, remaining)
+		info, err := query.ConstructPaymentInfo(&channel.VirtualChannel{Channel: *c}, paid, remaining)
 		if err != nil {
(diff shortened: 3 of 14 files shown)


lalexgap:
> We could consider either:
> 
>     * a distinct table in the store for virtual channels

This feels kinda wrong to me. The underlying data we store for channels and virtual channels is the same, so it seems weird to store it separately. 

>     * a discriminating enum that is serialized with the channel, and can be used by the store to know how to correctly hydrate data of the right type.

Right now we currently store the `appDefinition`, so if the hydration logic knows the `VirtualAppDefinition` it can easily compare and determine if something is a virtual channel.

geoknee:
> > We could consider either:
> > ```
> > * a distinct table in the store for virtual channels
> > ```
> 
> This feels kinda wrong to me. The underlying data we store for channels and virtual channels is the same, so it seems weird to store it separately.
> 
> > ```
> > * a discriminating enum that is serialized with the channel, and can be used by the store to know how to correctly hydrate data of the right type.
> > ```
> 
> Right now we currently store the `appDefinition`, so if the hydration logic knows the `VirtualAppDefinition` it can easily compare and determine if something is a virtual channel.

That's a good idea. Seems a little brittle in the sense that we often pass in dummy addresses for virtual apps?

lalexgap:
> I take your point and I think I'm still undecided overall.
> 
> In the short term, I could drop the commits from this PR that move status calculation into `Channel`. I think it might be worth keeping the others though, since they harmonize use of `VirtualChannel` in the store and across the protocols. What do you think?
> 
> If we did that, we could slim down the contentious bit and carry on considering it.

I think that makes sense to me!

> > However I think we introduce a similar problem here by moving the logic to`Channel/VirtualChannel`; If a protocol changes we still have to remember to update how we determine a status in `Channel/VirtualChannel`.

This is annoying but probably not the end of the world. It seems unlikely that the protocols are going to change radically and if they do we just need to update the `query` package. We have an RPC test that would probably error out until the query logic is resolved.

> I'm wondering about having the protocol query the `Channel` to decide when it should finish. Then, there would be a single source of truth. Really where everything including ADR 9 stems from is the on-chain application definition. I think it is reasonable for a `Channel` to know about that, and let both i) protocols and ii) queries know all of the important information.
> 

Isn't this essentially what occurs now when cranking the objective? Based on the status of the chain and channels an objective decides if it's complete or waiting for something else. Or are you thinking the `Channel` would expose a status?


> If we are serious about not exposing anything objective-related to the API, we wouldn't want to hook objectives into queries I think.

I don't think we want to expose objective-related concepts via the API, but I think it's totally OK for queries to look at objectives. The `query` package is meant to combine objective and channel related information to construct simplified proxies.

geoknee:
> Isn't this essentially what occurs now when cranking the objective? Based on the status of the chain and channels an objective decides if it's complete or waiting for something else. Or are you thinking the `Channel` would expose a status?
> 
> Maybe the on-chain application should define a function for determining the status (that we'd run off-chain)?

Yes that is essentially what I am proposing. I think defining the status on chain is a great idea!

geoknee:
I'll close this for now in favour of #1299 . We can return to this discussion at a later point if we want to.

PR #1288, opened 2023-05-17
Clean up some low hanging panics

lalexgap:
This PR tries to clean up a bunch of `panic`s we have throughout the codebase. Some things to note:
-  In some places we now return a `errChan` from a goroutine, so errors cans be communicated back.
- The `store` interface has been modified to return an error on a few of its functions
- I ignored test files and test related helpers
- This doesn't tackle the `panic`s in the RPC package, and I've also ignored tests and test helpers.
- I've removed some functions that weren't being used.
- This will probably break testground, since it modifies store constructors.


channel/consensus_channel/consensus_channel.go
```diff
@@ -624,19 +620,19 @@ func (p SignedProposal) SortInfo() (types.Destination, uint64) {
...
+			return p.ToAdd.Target(), nil
 		}
 	case "RemoveProposal":
 		{
-			return p.ToRemove.Target
+			return p.ToRemove.Target, nil
 		}
 	default:
 		{
-			panic("invalid proposal type")
```
geoknee, line 639: This one is potentially debatable. If we hit this line, something has gone very badly wrong and it is definitely a programming error rather than something unexpected in the environment. The data in the store is in an invalid state. So one could argue that panicking is acceptable here.

protocols/virtualdefund/virtualdefund.go
```diff
@@ -199,22 +199,26 @@ func (o *Objective) initialOutcome() outcome.SingleAssetExit {
 	return o.V.PostFundState().Outcome[0]
 }
 
-func (o *Objective) generateFinalOutcome() outcome.SingleAssetExit {
+func (o *Objective) generateFinalOutcome() (outcome.SingleAssetExit, error) {
 	if o.MyRole != 0 {
-		panic("Only Alice should call generateFinalOutcome")
```
geoknee, line 204: Another example of something that "should never happen" unless there are bugs in the code.

client/engine/engine.go
```diff
@@ -745,25 +751,25 @@ func fromMsgErr(id protocols.ObjectiveId, err error) error {
...
 		{
 			const prefix = virtualdefund.ObjectivePrefix
 			channelId := p.ToRemove.Target.String()
-			return protocols.ObjectiveId(prefix + channelId)
+			return protocols.ObjectiveId(prefix + channelId), nil
 
 		}
 	default:
 		{
-			panic("invalid proposal type")
```
geoknee, line 766: Here's another example when we would really like the type system to help us out. It may be possible to refactor this so there are no errors and no panics.

geoknee: Here's one idea in that direction https://github.com/statechannels/go-nitro/commit/1ec641f2bace4620020e167576d64bbbdaa10eb7

geoknee: The lack of exhaustive pattern matching in Go is a real shame though https://news.ycombinator.com/item?id=9037557

lalexgap: > Here's one idea in that direction https://github.com/statechannels/go-nitro/commit/1ec641f2bace4620020e167576d64bbbdaa10eb7

Using a boolean would restrict us to two values, which works here, but I'm not sure we can do something similar if we have more than 2 types.  In fact instead of using a switch statement, we could just use if/else.

In general, I feel like this approach might be too clever for own good? I feel like we're trying to trick `golang` into having exhaustive pattern matching instead of working within the constraints of the language.

>This one is potentially debatable. If we hit this line, something has gone very badly wrong and it is definitely a programming error rather than something unexpected in the environment. The data in the store is in an invalid state. So one could argue that panicking is acceptable here.

I think you're right and we should just panic. When I first made these changes I wasn't really paying too much attention to the context of the `panic`. 

We don't ever expect a client to hit that panic and if they do something is seriously wrong and we should be panicking. I would advocate for continuing to use panics in exhaustive switch statements.

protocols/virtualdefund/virtualdefund.go
lalexgap: >Another example of something that "should never happen" unless there are bugs in the code.

I'm slightly on the fence about this once. You're right that it should never happen, but we only know it should never happen because the function is called in one place, guarded by an if statement.  If you're just reading `generateFinalOutcome` it's not clear why `o.MyRole != 0 ` will always be true.

geoknee: I guess when I say "it should never happen" I am not saying that it is impossible or even that I would bet money it wouldn't happen. What I mean more is, "if it does happen, that is fairly catastrophic and means we have a bug in our code". 

It's nice if we can _also_ become confident that it truly won't ever happen... and I take your point that in this case it does _not_ seem so safe to make that assumption.

geoknee:
Might be worth browsing open issues with label `unwanted-panic` to see if this PR can close any of them.

lalexgap:
> Might be worth browsing open issues with label `unwanted-panic` to see if this PR can close any of them.

Unfortunately I don't think it does. Two issues are related to RPC server not sending error responses and one is a general issue about our approach to panicking.

client/engine/store/durablestore.go
```diff
@@ -34,36 +34,57 @@ type DurableStore struct {
...
-	return &ps
+	return &ps, nil
 }
 
 // NewPersistStore provides backwards compatibility for testground tests built against the persist store
 // TODO: Remove this once https://github.com/statechannels/go-nitro-testground/pull/156 is merged
-func NewPersistStore(key []byte, folder string, config buntdb.Config) Store {
+func NewPersistStore(key []byte, folder string, config buntdb.Config) (Store, error) {
 	return NewDurableStore(key, folder, config)
 }
```
geoknee, line 76: We could clear this TODO now, I believe.

client/engine/store/durablestore.go
```diff
@@ -34,36 +34,57 @@ type DurableStore struct {
 
 // NewDurableStore creates a new DurableStore that uses the given folder to store its data
 // It will create the folder if it does not exist
-func NewDurableStore(key []byte, folder string, config buntdb.Config) Store {
+func NewDurableStore(key []byte, folder string, config buntdb.Config) (Store, error) {
```
geoknee, line 37: Since you've made the change already, I think this is fine. But otherwise probably a case of being ok to panic in a constructor?

 It's hard to think of the calling function handling the error and the program being able to continue. I suppose in theory it could fail over to a memstore if the durable store construction failed...

geoknee (approved):

client/engine/chainservice/eth_chainservice.go
```diff
@@ -90,7 +90,20 @@ func newEthChainService(chain ethChain, na *NitroAdjudicator.NitroAdjudicator,
 
 	// Use a buffered channel so we don't have to worry about blocking on writing to the channel.
 	ecs := EthChainService{chain, na, naAddress, caAddress, vpaAddress, txSigner, make(chan Event, 10), logger, ctx, cancelCtx}
-	ecs.subscribeForLogs()
+	errChan, err := ecs.subscribeForLogs()
+	// TODO: Return error from chain service instead of panicking
```
geoknee, line 94: What do you see as the longer term pattern for this?

Could the chain service (and other services) forward errors to the engine (perhaps on a new `chan`), which then funnels them (again on a new `chan`) to the consuming application / RPC server?

Under that pattern, the goroutine spawned on L95 wouldn't need to exist.

client/engine/engine.go
lalexgap: Yup, will do!

client/engine/store/durablestore.go
geoknee: #1312

PR #1293, opened 2023-05-17
Set Different App Definitions

lalexgap:
This PR updates the eth chain service to require two different addresses for the virtual payment app and consensus app. Since we rely on the appDef address in our `query` package to differentiate between virtual and ledger channels we want to enforce those addresses are different. 

This PR also updates `main.go` to pass into two different (dummy) addresses.

### DIFF
--- a/main.go
+++ b/main.go
@@ -33,10 +33,12 @@ func main() {
 		CHAIN_URL         = "chainurl"
 		CHAIN_PK          = "chainpk"
 		NA_ADDRESS        = "naaddress"
+		VPA_ADDRESS       = "vpaaddress"
+		CA_ADDRESS        = "caaddress"
 		MSG_PORT          = "msgport"
 		RPC_PORT          = "rpcport"
 	)
-	var pkString, chainUrl, naAddress, chainPk string
+	var pkString, chainUrl, naAddress, vpaAddress, caAddress, chainPk string
 	var msgPort, rpcPort int
 	var useNats, useDurableStore bool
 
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -130,7 +130,7 @@ func main() {
 			}
 
 			fmt.Println("Initializing chain service and connecting to " + chainUrl + "...")
-			chainService, err := chainservice.NewEthChainService(chainUrl, chainPk, common.HexToAddress(naAddress), common.Address{}, common.Address{}, os.Stdout)
```
geoknee, line 133: Why not let's go one step further and actually deploy and pass these addresses in?

I know we passed nil values in previously, but I'm uneasy about baking in these dummy values to our prod code (could be done in test code only).

geoknee (changes requested):

main.go
lalexgap: Updated in https://github.com/statechannels/go-nitro/pull/1293/commits/a936845689c590be73fdd619a2c5f307c690f54a

geoknee: Thanks! Will we need to update our rpc test orchestration script now?

lalexgap: That was done in https://github.com/statechannels/go-nitro/commit/a936845689c590be73fdd619a2c5f307c690f54a too! The `start-rpc-servers` script has been updated to pass in the required addresses

geoknee: Of course it was 🤦

geoknee (approved):

PR #1294, opened 2023-05-18
Make mdns optional for the message service

lalexgap:
This makes `mdns` optional for the p2p message service. In some complicated network environments `mdns` might not be that helpful. In the case of testground the `mdns` peer discovery was using the docker container ip address instead of the correct data network address.


geoknee (approved):
Long term I guess we can't be relying overly on mDNS since it only works on local networks.

PR #1298, opened 2023-05-18
Update big numbers to serde to hex string

### LINKED ISSUE #1297: Handle BigNumbers going over the wire
Currently the RPC serializes out BigNumbers as numbers, which will break if they get big.

lalexgap:
Fixes #1297 

This makes of go-ethereum's `hexutils.Big` that already implements serde to/from hex string (prefixed with `0x`).  For now I've only applied this to the query package, as this is the only JSON we serve up via the RPC interface.

However we could consider extending this across go-nitro to replace all our existing `big.Int`. We would need to update our code that uses big numbers to call `.ToInt()` but besides that I don't think it will be too painful.


geoknee:
> However we could consider extending this across go-nitro to replace all our existing `big.Int`

I think I prefer judiciously using `hexutils.Big` when we need to do serde, and keeping `big.Int` otherwise. 

Do we have confidence that the serde in the store is robust even when using very large values?

Looks like you have a failing PR check.

lalexgap:
> Do we have confidence that the serde in the store is robust even when using very large values?

I've added a simple test that checks both stores that it can store/fetch a channel with a large number.

> Looks like you have a failing PR check.

Resolved!

client/engine/store/store_test.go
```diff
@@ -208,3 +216,48 @@ func TestGetChannelsByParticipant(t *testing.T) {
...
+func TestBigNumberStorage(t *testing.T) {
+	pk := common.Hex2Bytes(`2af069c584758f9ec47c4224a8becc1983f28acfbe837bd7710b70f9fc6d5e44`)
+
+	dataFolder := fmt.Sprintf("%s/%d%d", STORE_TEST_DATA_FOLDER, rand.Uint64(), time.Now().UnixNano())
+	durableStore := store.NewPersistStore(pk, dataFolder, buntdb.Config{})
+	memStore := store.NewMemStore(pk)
+
+	for _, store := range []store.Store{durableStore, memStore} {
+		// Set the large amount to 100 * math.MaxInt64
+		// 9223372036854775807 * 100 = 922337203685477580700
```
geoknee, line 229: 💪

geoknee (approved):

PR #1299, opened 2023-05-19
Proliferate `VirtualChannel` type (but don't change the query package)

geoknee:
Supersedes #1287

### DIFF
--- a/client/query/query.go
+++ b/client/query/query.go
@@ -50,9 +50,9 @@ func getPaymentChannelBalance(participants []types.Address, outcome outcome.Exit
 	}
 }
 
-// getLatestSupported returns the latest supported state of the channel
+// getLatestSupportedOrPreFund returns the latest supported state of the channel
 // or the prefund state if no supported state exists
-func getLatestSupported(channel *channel.Channel) (state.State, error) {
+func getLatestSupportedOrPreFund(channel *channel.Channel) (state.State, error) {
 	if channel.HasSupportedState() {
 		return channel.LatestSupportedState()
 	}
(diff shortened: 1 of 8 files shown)


client/query/query.go
```diff
@@ -231,7 +231,7 @@ func ConstructPaymentInfo(c *channel.Channel, paid, remaining *big.Int) (Payment
 		status = Open
 	}
 
-	latest, err := getLatestSupported(c)
+	latest, err := getLatestSupportedOrPreFund(c)
```
geoknee, line 234: Are we sure we want this behaviour? If there is not yet a supported state, perhaps we should return a `nil` balance or something.

client/query/query.go
```diff
@@ -211,7 +211,7 @@ func ConstructLedgerInfoFromConsensus(con *consensus_channel.ConsensusChannel) L
 }
 
 func ConstructLedgerInfoFromChannel(c *channel.Channel) LedgerChannelInfo {
-	latest, err := getLatestSupported(c)
+	latest, err := getLatestSupportedOrPreFund(c)
```
geoknee, line 214: Same question here.

client/query/query.go
lalexgap: I think we should return what makes most sense to a user, which I think might be the existing balance? It would seem odd if you use the RPC client to create a ledger channel, specifying a certain balance, and then the first notification you get has a `nil` balance.

geoknee: We do have the `Proposed` status, so I guess that does make sense. 

I think in general we need to be careful about interpreting states which are not supported, though.

lalexgap (approved):
Nice, this seems much cleaner!

PR #1301, opened 2023-05-19
refactor: clean up variable names in rpc_test

NiloCK:
Minor PR for readibility. A few things had generic names on first instance (eg, `result`), then more specific names later.

PR adds participant initals and channel type (v or l) to some names, and a little whitespace.


NiloCK:
This work was included in #1315

Issue #1306, opened 2023-05-23
We sometimes see a a closing balance with notification of 100/0


lalexgap:
See https://github.com/statechannels/go-nitro/actions/runs/5034369990/jobs/9029151062

```
rpc_test.go:313: Received unexpected notification: {0xb5cf2bc6b41492c89ac8c38c714c18dd2b24903fce274092d51df7de183cb37f Closing {0x0000000000000000000000000000000000000000 0xBBB676f9cFF8D242e9eaC39D063848807d3D1D94 0xAAA6628Ec44A8a742987EF3A114dDFE2D4F7aDCE 0 100}}
```

Occasionally our RPC test flickers and fails with the error above.

It looks like sometimes we get a `Closing` notification with a balance of `100/0` instead of the expected value of `99/1`.  I think this could be legitimate, since it's possible that `Bob` happens to receive the close objective **before** the voucher? However I think we should dig into this and figure out what exactly is happening. 

Unfortunately I blew away the log files when rerunning the action.

lalexgap:
Here's another [example](https://github.com/statechannels/go-nitro/actions/runs/5061788149/jobs/9086644553). Attached are the log files.
[integration-test-logs.zip](https://github.com/statechannels/go-nitro/files/11548305/integration-test-logs.zip)

geoknee:
> Here's another [example](https://github.com/statechannels/go-nitro/actions/runs/5061788149/jobs/9086644553). [Here](https://github.com/statechannels/go-nitro/files/11548305/integration-test-logs.zip) are the log files.

Thanks @lalexgap. I piped the logs through `pino-pretty` and trawled through. Summary output below, with most lines trimmed and ordering preserved:

```log
[21:43:29.622] DEBUG <engine.go:683>: 0xBBB676  <  {"engine":"0xBBB676","message":"Constructing objective VirtualDefund-0xa1220ec1300a77856f72e63922ce7a269b0469c954dbd58c896c4b943518b0ed from message"}

[21:43:29.623] TRACE <engine.go:782>: 0xBBB676 0xBBB676 < 0xAAA662 {"engine":"0xBBB676","To":"0xBBB676","From":"0xAAA662","PayloadSummaries":[],"ProposalSummaries":[],"Payments":[{"ChannelId":"0xa1220ec1300a77856f72e63922ce7a269b0469c954dbd58c896c4b943518b0ed","Amount":1}],"RejectedObjectives":[],"message":"Received message"}

[...]

PR #1307, opened 2023-05-23
Workaround unexpected closing notification

lalexgap:
Workaround for #1306 

Sometimes when running the test on CI we see a flicker about an unexpected notification
```
{0xb5cf2bc6b41492c89ac8c38c714c18dd2b24903fce274092d51df7de183cb37f Closing {0x0000000000000000000000000000000000000000 0xBBB676f9cFF8D242e9eaC39D063848807d3D1D94 0xAAA6628Ec44A8a742987EF3A114dDFE2D4F7aDCE 0 100}}
```

This a closing notification for the virtual channel with a balance of `Alice:100,Bob:0`. However, based on the test, we expect a closing notification with a balance of `Alice:99,Bob:1` , since Alice calls `Pay` once. This PR updates the test to also accept the `100/0 notification as a workaround for the flicker.

 However we should probably investigate exactly what is happening when this happens. Does the voucher message just arrive after the close objective request? What happens to the voucher in that case?

### DIFF
--- a/client_test/rpc_test.go
+++ b/client_test/rpc_test.go
@@ -174,6 +174,9 @@ func executeRpcTest(t *testing.T, connectionType transport.TransportType) {
 	}
 	optionalVirtualNotifs := []query.PaymentChannelInfo{
 		expectedPaymentInfo(vRes.ChannelId, simpleOutcome(ta.Alice.Address(), ta.Bob.Address(), 99, 1), query.Closing),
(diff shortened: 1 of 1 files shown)


lalexgap:
We may want to leave this open until we capture the logs from another flicker

lalexgap:
> We may want to leave this open until we capture the logs from another flicker

Captures some logs [here](https://github.com/statechannels/go-nitro/issues/1306#issuecomment-1560101996).

PR #1310, opened 2023-05-25
downgrade libp2p to 0.26.2

lalexgap:
This downgrades our version of libp2p to `0.26.2` which is the same version `boost` and `lotus` uses. Otherwise adding `go-nitro` to `boost` causes libp2p to get upgraded and breaks `boost`/`lotus`:
```
make debug
rm -f boost boostd boostx
go build  -ldflags="-X=github.com/filecoin-project/boost/build.CurrentCommit=+git.4986e2f.dirty" -tags=debug -o boost ./cmd/boost
# github.com/filecoin-project/lotus/node/impl/net
../../go/pkg/mod/github.com/filecoin-project/lotus@v1.23.0/node/impl/net/net.go:148:22: autonat.PublicAddr undefined (type autonat.AutoNAT has no field or method PublicAddr)
```

I'm not sure if we want to merge this or just keep a separate branch for `boost` work. Or try to upgrade `boost`/`lotus` code to the latest libp2p 😨

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -151,7 +151,7 @@ require (
 	github.com/gorilla/websocket v1.5.0 // indirect
 	github.com/holiman/bloomfilter/v2 v2.0.3 // indirect
 	github.com/holiman/uint256 v1.2.0 // indirect
-	github.com/libp2p/go-libp2p v0.27.0
+	github.com/libp2p/go-libp2p v0.26.2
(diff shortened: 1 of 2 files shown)


lalexgap:
Actually it looks like we can just use a `replace` inside our fork of `boost`: https://github.com/statechannels/boost/pull/1/commits/44846fd34496952070a899ab49c2bbaea9f208bc. Closing this in favour of that.

Issue #830, opened 2022-08-19
Replace revert strings with custom errors


geoknee:
The reason they are a bit difficult to parse is because I am trying to keep them short to save gas. I need to refresh my memory about whether the gas costs are only significant for deployment, though.

Probably we should be adopting a more modern approach to errors https://docs.soliditylang.org/en/latest/contracts.html#errors-and-the-revert-statement .

_Originally posted by @geoknee in https://github.com/statechannels/go-nitro/pull/829#discussion_r949966871_

nksazonov:
This is a great idea!
Firstly we would need to migrate to a newer solidity version, specifically greater or equal than `0.8.4` - [the version `error` keyword was introduced in](https://github.com/ethereum/solidity/releases/tag/v0.8.4).
Thus, #556 is a must.

PR #1322, opened 2023-06-02
Refactor: replace `requireStateSupported(...)` with `stateIsSupported(...)`

NiloCK:
PR changes the shape of `IForceMove.requireStateSupported()` from

```
function requireStateSupported(
    FixedPart calldata fixedPart,
    RecoveredVariablePart[] calldata proof,
    RecoveredVariablePart calldata candidate
) external view;
```
to

```
function stateIsSupported(
    FixedPart calldata fixedPart,
    RecoveredVariablePart[] calldata proof,
    RecoveredVariablePart calldata candidate
) external view returns (bool, string memory);
```

and does follow-on work.

PR is prerequisite work for [Composition App](https://www.notion.so/statechannels/Speculative-0ae5d8a85c18425d8f1919b42d3d53db?p=9e825aa049eb4e8381c7e860ee7e2540&pm=s) - a channel application that allows for the creation of channels that mix and match application rules from existing deployed channel applications.

Notable: the existing channel applications in repository have been modified to adhere to the new interface method, but they still (mostly) enter the "failure" path by the pre-existing `revert` mechanism. This is OK - reverting remains a valid failure route. BUT channel app developers should prefer returning `(false, "some reason")` so that the app behaves as expected when consumed by the `CompositionApp`. Refactoring existing apps to do this is left as future work.

---

#### Code quality

  - uses existing tests 

#### Project management


nitro-protocol/contracts/interfaces/IForceMoveApp.sol
```diff
@@ -5,19 +5,19 @@ pragma experimental ABIEncoderV2;
...
      * @param proof Array of recovered variable parts which constitutes a support proof for the candidate. May be omitted when `candidate` constitutes a support proof itself.
      * @param candidate Recovered variable part the proof was supplied for. Also may constitute a support proof itself.
      */
-    function requireStateSupported(
+    function stateIsSupported(
         FixedPart calldata fixedPart,
         RecoveredVariablePart[] calldata proof,
         RecoveredVariablePart calldata candidate
-    ) external view;
+    ) external view returns (bool, string memory);
```
geoknee, line 22: This looks like Go code now :-D 

Is there a way to use solidity `errors` https://solidity-by-example.org/error/ instead of strings? I suspect it may not currently be possible, since errors are not first class ("just values" as in Go). But just wanted to flag that up as well as #830 .

.adr/0012-interest-bearing-channel.md
```diff
@@ -16,7 +16,7 @@ The Bitcoin Lightning Network (LN) is an in-production payment channel network (
 
 ## Similarities to Nitro-SCN
 
-In a PCN, payments between A and B are routed through intermediaries I_1, ..., I_n, where balances are pairwise adjusted along the route. Capital in intermediary channels is locked for the _duration_ of the payment - ie, until either the payment completes or times out.
+In a PCN, payments between A and B are routed through intermediaries I*1, ..., I_n, where balances are pairwise adjusted along the route. Capital in intermediary channels is locked for the \_duration* of the payment - ie, until either the payment completes or times out.
```
NiloCK, line 19:
```suggestion
In a PCN, payments between A and B are routed through intermediaries I_1, ..., I_n, where balances are pairwise adjusted along the route. Capital in intermediary channels is locked for the _duration_ of the payment - ie, until either the payment completes or times out.
```

nitro-protocol/test/tx-expect-wrappers.ts
```diff
@@ -0,0 +1,32 @@
...
+  expect(txResult.length).toBe(2);
+  expect(txResult[0]).toBe(true);
+  expect(txResult[1]).toBe('');
+}
+
+/**
+ * Wrapper for calls to `stateIsSupported` that are expected to fail.
+ * Checks that the reason for failure matches the supplied `reason` string.
+ */
+export async function expectFailure(fn: () => void, reason?: string) {
```
geoknee, line 26: Should we write a wrapper that will pass a test if the call results in _either_ a revert or a `false` return value? Or do you think it is OK for the tests to "know" which failure mode the implementation has plumped for?

nitro-protocol/test/tx-expect-wrappers.ts
```diff
@@ -0,0 +1,32 @@
+/**
+ * Wrapper for transactions that are expected to succeed with no return values.
+ */
+export async function expectSucceed(fn: () => void) {
```
geoknee, line 4: I wonder if `expectToSucceedWithNoReturnValues` would be better?

geoknee (approved):

nitro-protocol/gas-benchmarks/exportBenchmarkResults.ts
```diff
@@ -339,6 +339,7 @@ async function main() {
 
   writeFileSync(__dirname + '/gasResults.json', JSON.stringify(gasResults, null, 2));
   console.log('Benchmark results updated successfully!');
+  // todo: console.log a summary of the results for a commit message
```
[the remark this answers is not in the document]
NiloCK: https://github.com/statechannels/go-nitro/issues/1358

f394b948ece9a55f689a6291aff556f0d8b4df47

nitro-protocol/test/tx-expect-wrappers.ts
NiloCK: I think this is OK as is given current usage. Implementing the other wrapper shouldn't be much burden at time of need.

Did clarify intent here: 6dc6765f

Issue #1323, opened 2023-06-05
Consistent failures with rpc_test


NiloCK:
Since PR #1315, CI has been extremely brittle. #1315 introduced 0-hop and 2-hop virtual funding operations, mediated by the golang rpc_client, to the test suite (and CI). Previously, only the 1-hop virtual fund was under test.

recent example failures:

https://github.com/statechannels/go-nitro/actions/runs/5177367315/jobs/9327375026#step:7:578
```
...
rpc_test.go:423: bobVirtual received {0x981e13c023e2da0069ed544cc82c4ac7db43d7d1931eda38cca49bc127a9ef2c Complete {0x0000000000000000000000000000000000000000 0x6813Eb9362372EEF6200f3b1dbC3f819671cBA69 0x7E5F4552091A69125d5DfCb7b8C2659029395Bdf 0x1 0x63}}+
panic: send on closed channel

goroutine 2649 [running]:
github.com/statechannels/go-nitro/client/notifier.(*ledgerChannelListeners).Notify(0xc000258d20, {{0xe3, 0xb1, 0xaf, 0x5b, 0xd3, 0x40, 0x15, 0xf9, 0x1f, ...}, ...})
	/home/runner/work/go-nitro/go-nitro/client/notifier/listeners.go:89 +0x173
github.com/statechannels/go-nitro/client/notifier.(*ChannelNotifier).NotifyLedgerUpdated(0xc000a37770, {{0xe3, 0xb1, 0xaf, 0x5b, 0xd3, 0x40, 0x15, 0xf9, 0x1f, ...}, ...})
	/home/runner/work/go-nitro/go-nitro/client/notifier/channel-notifier.go:61 +0x1f8
github.com/statechannels/go-nitro/client.(*Client).handleEngineEvents(0xc0004c7860)
	/home/runner/work/go-nitro/go-nitro/client/client.go:107 +0x418
created by github.com/statechannels/go-nitro/client.New
	/home/runner/work/go-nitro/go-nitro/client/client.go:76 +0xbd3
FAIL	github.com/statechannels/go-nitro/client_test	0.935s
``` 
Here, sending on a closed channel generally indicates a go-nitro client attempting to communicate a protocol update message *after* it has externally been triggered to shut down (the closed channel in question is part of the event stream between go-nitro rpc-server and rpc-clients).

https://github.com/statechannels/go-nitro/actions/runs/5159120522/jobs/9293632835#step:7:751
```
...
[...]

PR #1325, opened 2023-06-06
Modify notifications check test to improve error reporting

NiloCK:
Toward #1323 

PR improves debug information from failed runs by failing more slowly. Instead of immediate fail on receipt of an unexpected notification, test now waits the test timeout until all required notifications have been seen.

After the timeout, it reports on all abnormalities (missing & required, present & unexpected).

### DIFF
--- a/client_test/rpc_test.go
+++ b/client_test/rpc_test.go
@@ -429,12 +429,14 @@ func checkNotifications[T channelInfo](t *testing.T, client string, required []T
 			t.Logf("%s received %v+", client, info)
 
 			// Check that the notification is a required or optional one.
-			_, found := acceptableNotifications[notifJSON]
-			if !found {
+			_, isExpected := acceptableNotifications[notifJSON]
+
+			if isExpected {
+				// To signal we received a notification we delete it from the map
+				delete(acceptableNotifications, notifJSON)
+			} else {
 				unexpectedNotifications[notifJSON] = true
 			}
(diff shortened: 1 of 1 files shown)


client_test/rpc_test.go
```diff
@@ -419,21 +425,26 @@ func checkNotifications[T channelInfo](t *testing.T, client string, required []T
...
 			// Check that the notification is a required or optional one.
-			_, found := acceptableNotifications[js]
+			_, found := acceptableNotifications[notifJSON]
 			if !found {
-				t.Fatalf("%s received unexpected notification: %v", client, info)
+				unexpectedNotifications[notifJSON] = true
 			}
 			// To signal we received a notification we delete it from the map
-			delete(acceptableNotifications, js)
+			delete(acceptableNotifications, notifJSON)
```
geoknee, line 437: This line will run even if `found=false`. What happens if you try and delete a key that doesn't exist?

Regardless, it might be nicer to do 

```go
if found {
    delete(acceptableNotifications, notifJSON)
} else {
    unexpectedNotifications[notifJSON] = true
}


```

geoknee (approved):

client_test/rpc_test.go
NiloCK: That delete of a non-key is a noop, but your suggestion is better anyhow.

7fdd7be1

PR #1328, opened 2023-06-07
Avoid starting MDNS before message service is ready

lalexgap:
Previously we would start up the `mdns` service before we had finished constructing the message service. This meant that if we're unlikely with timing `mdns` could call `HandlePeerFound` on the message service before the constructor had completed. We now start`mdns` at the end of constructor after everything has been initialized. 

I've also noticed that if we use the `-race` flag we get a race condition complaining about our use of our logger. I've created a new issue for that : #1329

### DIFF
--- a/client/engine/messageservice/p2p-message-service/service.go
+++ b/client/engine/messageservice/p2p-message-service/service.go
@@ -91,12 +91,7 @@ func NewMessageService(ip string, port int, me types.Address, pk []byte, useMdns
 	if err != nil {
 		panic(err)
 	}
-	if useMdnsPeerDiscovery {
-		mdns := mdns.NewMdnsService(host, "", ms)
-		err = mdns.Start()
-		ms.checkError(err)
-		ms.mdns = mdns
-	}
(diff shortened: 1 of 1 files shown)


client/engine/messageservice/p2p-message-service/service.go
```diff
@@ -106,6 +101,14 @@ func NewMessageService(ip string, port int, me types.Address, pk []byte, useMdns
 		stream.Close()
 	})
 
+	// Since the mdns service could trigger a call to  `HandlePeerFound` at any time once started
+	// We want to start mdns after the message service has been fully constructed
+	if useMdnsPeerDiscovery {
+		mdns := mdns.NewMdnsService(host, "", ms)
+		err = mdns.Start()
+		ms.checkError(err)
+		ms.mdns = mdns
```
geoknee, line 110: Super nit: why not set this property _before_ starting the service, to rule out any other race conditions?

geoknee (approved):

Issue #1330, opened 2023-06-07
RPC test flickers with send on closed channel error


lalexgap:
> Since PR #1315, CI has been extremely brittle. #1315 introduced 0-hop and 2-hop virtual funding operations, mediated by the golang rpc_client, to the test suite (and CI). Previously, only the 1-hop virtual fund was under test.
> 
> recent example failures:
> 
> https://github.com/statechannels/go-nitro/actions/runs/5177367315/jobs/9327375026#step:7:578
> 
> ```
> ...
> rpc_test.go:423: bobVirtual received {0x981e13c023e2da0069ed544cc82c4ac7db43d7d1931eda38cca49bc127a9ef2c Complete {0x0000000000000000000000000000000000000000 0x6813Eb9362372EEF6200f3b1dbC3f819671cBA69 0x7E5F4552091A69125d5DfCb7b8C2659029395Bdf 0x1 0x63}}+
> panic: send on closed channel
> 
> goroutine 2649 [running]:
> github.com/statechannels/go-nitro/client/notifier.(*ledgerChannelListeners).Notify(0xc000258d20, {{0xe3, 0xb1, 0xaf, 0x5b, 0xd3, 0x40, 0x15, 0xf9, 0x1f, ...}, ...})
> 	/home/runner/work/go-nitro/go-nitro/client/notifier/listeners.go:89 +0x173
> github.com/statechannels/go-nitro/client/notifier.(*ChannelNotifier).NotifyLedgerUpdated(0xc000a37770, {{0xe3, 0xb1, 0xaf, 0x5b, 0xd3, 0x40, 0x15, 0xf9, 0x1f, ...}, ...})
> 	/home/runner/work/go-nitro/go-nitro/client/notifier/channel-notifier.go:61 +0x1f8
> github.com/statechannels/go-nitro/client.(*Client).handleEngineEvents(0xc0004c7860)
> 	/home/runner/work/go-nitro/go-nitro/client/client.go:107 +0x418
> created by github.com/statechannels/go-nitro/client.New
> 	/home/runner/work/go-nitro/go-nitro/client/client.go:76 +0xbd3
> FAIL	github.com/statechannels/go-nitro/client_test	0.935s
> ```
> 
> Here, sending on a closed channel generally indicates a go-nitro client attempting to communicate a protocol update message _after_ it has externally been triggered to shut down (the closed channel in question is part of the event stream between go-nitro rpc-server and rpc-clients).
> 
[...]

PR #1331, opened 2023-06-07
Prevent race condition when closing client

lalexgap:
Fixes https://github.com/statechannels/go-nitro/issues/1330

This addresses a race condition that could occur when closing the nitro client. The nitro client starts a `handleEngineEvents` in a go-routine when it's constructed.  This functions listens on a chan from the engine for events and dispatches them to the `channelNotifier`.

The problem is that `handleEngineEvents` would to continue to run after we close the channel notifier in `Close`. This means `handleEngineEvents` would try to call `channelNotfier.Notify` after it's been closed, resulting in the error we see. Since the `toAPI` is [a buffered chan of size 100](https://github.com/statechannels/go-nitro/blob/4c820ee15dbf04648cb91e58617c75097cc1a26f/client/engine/engine.go#L142) it's quite possible that `handleEngineEvents` continues to execute for awhile after `Close` has been called.

To address this I've updated `handleEngineEvents` to accept a `context` so we can call `cancel` and stop the `handleEngineEvents` go-routine.

 While this let's us signal to the `handleEngineEvents` that it should stop, it's non-blocking, which means we can still end up closing the channel notifier before `handleEngineEvents` is done. To address this I've made use of `sync.Waitgroup` to block until the `handleEngineEvents` stops executing in `Close`. This way we can be sure that it's safe to close the channelNotifer.

I've run this on CI 8 times as a quick sanity check for the flicker. I did see one notification related failure but I think that is a [separate flicker](https://github.com/statechannels/go-nitro/issues/1323).

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -2,6 +2,7 @@
 package client // import "github.com/statechannels/go-nitro/client"
 
 import (
+	"context"
 	"io"
 	"math/big"
 	"runtime/debug"
@@ -40,6 +41,7 @@ type Client struct {
 	store                     store.Store
 	vm                        *payments.VoucherManager
 	logger                    zerolog.Logger
+	cancelEventHandler        context.CancelFunc
 }
 
 // New is the constructor for a Client. It accepts a messaging service, a chain service, and a store as injected dependencies.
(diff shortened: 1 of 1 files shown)


lalexgap:
Closing in favour of #1332

lalexgap:
> Closing in favour of #1332

Reopening as I think #1332 doesn't fix the problem.

client/client.go
```diff
@@ -68,56 +73,62 @@ func New(messageService messageservice.MessageService, chainservice chainservice
...
-			close(d)
-
-			// use a nonblocking send to the RPC Client in case no one is listening
-			select {
-			case c.completedObjectivesForRPC <- completed.Id():
-			default:
+func (c *Client) handleEngineEvents(ctx context.Context) {
+	for {
+		select {
+		case <-ctx.Done():
```
lalexgap, line 97: We actually don't have to use a `context` to signal `handleEngineEvents` to stop executing, we could just rely on the engine `ToAPI` chan getting closed when  we call `engine.Close` like we were doing previously.  

However I think this is less efficient as `close` has ["the effect of shutting down the channel after the last sent value is received"](https://pkg.go.dev/builtin#close), meaning this goroutine will only stop executing after it's gotten through the buffer of the `chan`.

By using a `context` and `cancel` the goroutine will stop as soon as it reads from `ctx.Done`

geoknee: I suppose the select statement could still end up "ignoring" the `ctx.Done()` path for some time as the buffer empties? It seems more likely that the goroutine is stopped sooner than via the existing behaviour, but I guess no certainty about that?

geoknee:
> The problem is that `handleEngineEvents` would to continue to run after we close the channel notifier in `Close`

How about if we changed the order of operations in `client.Close()`, so that the `engine` is closed _before_ the client notifier? I think it makes sense to stop things that send before the things that receive from them?

client/client.go
```diff
@@ -68,56 +73,62 @@ func New(messageService messageservice.MessageService, chainservice chainservice
...
 	c.channelNotifier = notifier.NewChannelNotifier(store, c.vm)
+
 	// Start the engine in a go routine
+
 	go c.engine.Run()
 
+	c.wg = &sync.WaitGroup{}
+	c.wg.Add(1)
+	ctx, cancel := context.WithCancel(context.Background())
+	c.cancelEventHandler = cancel
```
geoknee, line 84: When we resolve #1341, let's return to this section and add some comments to help the reader understand what is going on.

geoknee (approved):

geoknee (approved):

geoknee (approved):

PR #1332, opened 2023-06-07
Prevent race condition when closing client

lalexgap:
Fixes https://github.com/statechannels/go-nitro/issues/1330

This addresses a race condition that could occur when closing the nitro client. The nitro client starts a handleEngineEvents in a go-routine when it's constructed. This functions listens on a chan from the engine for events and dispatches them to the channelNotifier.

The problem is that handleEngineEvents would to continue to run after we close the channel notifier in Close. This means handleEngineEvents would try to call channelNotfier.Notify after it's been closed, resulting in the error we see. Since the toAPI is [a buffered chan of size 100](https://github.com/statechannels/go-nitro/blob/4c820ee15dbf04648cb91e58617c75097cc1a26f/client/engine/engine.go#L142) it's quite possible that handleEngineEvents continues to execute for awhile after Close has been called.

To address this I've moved the responsibility for closing the `channelNotifier` to the `handleEngineEvents` go routine,  so it only closes the channel notifier after it's done dispatching notifications.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -118,6 +118,7 @@ func (c *Client) handleEngineEvents() {
 	// If there are blocking consumers (for or select channel statements) on any channel for which the client is a producer,
 	// those channels need to be closed.
 	close(c.completedObjectivesForRPC)
+	c.channelNotifier.Close()
 }
 
 // Begin API
@@ -273,9 +274,6 @@ func (c *Client) GetLedgerChannel(id types.Destination) (query.LedgerChannelInfo
 
 // Close stops the client from responding to any input.
 func (c *Client) Close() error {
-	if err := c.channelNotifier.Close(); err != nil {
-		return err
-	}
 	if err := c.engine.Close(); err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


kerzhner (approved):

lalexgap:
It looks like we have pretty consistent flicker that results in this error:

```
 panic: nats: connection closed

goroutine 2521 [running]:
github.com/statechannels/go-nitro/rpc.(*RpcServer).sendNotifications.func1()
	/home/runner/work/go-nitro/go-nitro/rpc/server.go:233 +0x385
created by github.com/statechannels/go-nitro/rpc.(*RpcServer).sendNotifications
	/home/runner/work/go-nitro/go-nitro/rpc/server.go:223 +0x56
FAIL	github.com/statechannels/go-nitro/client_test	0.593s
```

Moving back to draft for now. I may revive #1331 if I can't get this sorted.

lalexgap:
One downside of this approach is that `client.Close` doesn't actually wait for `handleEngineEvents` to complete. I suspect that's what happening with the error above. The `handleEngineEvents` goroutine is still running and firing off notification from the `channelNotifier` to the rpc server to send off, but it's already closed.

PR #1336, opened 2023-06-08
Add jsonrpc api version to url

bitwiseguy:
Adding jsonrpc api version to the server's endpoint url

### Pros of this approach
Allows us to easily upgrade an entire api at the same time (instead of individually updating each jsonrpc method). This should make it easier to keep go-nitro services compatible with each other. If a client attempts to call a version of the api that is not supported by that go-nitro server, they will receive an error to let them know their request cannot be processed (instead of the request being processed by an old version of the api, potentially causing confusing/incorrect results).

### Cons of this approach
Potential added complexity if we want a server to support more than one version of the api simultaneously. We could choose to always support at least one previous version to give all nodes time to upgrade before they become incompatible with nodes that have already upgraded.

### Notes
We will need to update any clients (websocket or nats) after making this change since the server's rpc endpoint will be changed by this PR.


rpc/transport/ws/server.go
```diff
@@ -31,21 +33,23 @@ type serverWebSocketTransport struct {
...
+	tcpListener, err := net.Listen("tcp", websocketServerAddress+wsc.port)
 	if err != nil {
 		return nil, err
 	}
 
 	var serveMux http.ServeMux
 
-	serveMux.HandleFunc(path.Join("/", rpcPath), wsc.request)
-	serveMux.HandleFunc(path.Join("/", rpcPath, "subscribe"), wsc.subscribe)
+	serveMux.HandleFunc(apiVersionPath, wsc.request)
```
bitwiseguy, line 43: Instead of listening on the specific api version path (i.e. `/api/v1`), should we listen to the entire path `/api`, then we can lean on the logic in `wsc.request` to determine if this is a supported api version or not by parsing the url?

readme.md
```diff
@@ -50,17 +50,18 @@ Usage of ./nitro-rpc-server:
...
-  ```
-
-  but see https://github.com/statechannels/nitro-gui for an RPC client to do so programmatically. 
+
+```shell
+curl -X POST \
+  -H 'Content-Type: application/json' \
+  -d '{"jsonrpc":"2.0","id":1,"method":"get_address","params":{}}' \
+  http://localhost:4005/api/v1
+```
```
bitwiseguy, line 61: Not sure why but reformatting the curl command above allowed me to do a copy+paste into the terminal and run the command. The original format produced an error with the curl command

rpc/transport/nats/nats_transport.go
```diff
@@ -78,7 +78,7 @@ func (c *natsTransportClient) Request(data []byte) ([]byte, error) {
 	return msg.Data, err
 }
 
-func (c *natsTransportServer) RegisterRequestHandler(handler func([]byte) []byte) error {
+func (c *natsTransportServer) RegisterRequestHandler(apiVersion string, handler func([]byte) []byte) error {
```
bitwiseguy, line 81: Should the apiVersion be added to the `nitroRequestTopic`, similar to how the version is part of the url for websockets?
```golang
	nitroRequestTopic      = "nitro-request/api/v1"
```

bitwiseguy: Went with the topic string format shown above

rpc/transport/nats/server.go
```diff
@@ -91,24 +84,6 @@ func (c *natsTransportServer) RegisterRequestHandler(apiVersion string, handler
...
-	c.natsSubscriptions = append(c.natsSubscriptions, subscription)
-
-	return c.notificationChan, err
-}
-
-func (c *natsTransportClient) Close() {
-	c.natsTransport.Close()
-	close(c.notificationChan)
-}
-
```
geoknee, line 111: Thanks for tidying this up!

In an ideal world, we might split out this work into a separate PR -- especially since it is the kind of diff which looks more drastic than it actually is. 

I think it's fine to keep it here this time.

rpc/transport/ws/server.go
```diff
@@ -77,10 +81,24 @@ func (wsc *serverWebSocketTransport) Close() {
...
+		http.Error(w, "Invalid API version", http.StatusBadRequest)
+		return
+	}
+
+	apiVersion := pathSegments[2] // first segment is an empty string
+	handler, ok := wsc.requestHandlers[apiVersion]
+	if !ok {
+		http.Error(w, "Invalid API version", http.StatusBadRequest)
+		return
+	}
```
geoknee, line 100: I like this!

rpc/transport/ws/server.go
```diff
@@ -77,10 +81,24 @@ func (wsc *serverWebSocketTransport) Close() {
 }
 
 func (wsc *serverWebSocketTransport) Url() string {
-	return webscocketServerAddress + wsc.port
+	return websocketServerAddress + wsc.port + apiVersionPath
 }
 
 func (wsc *serverWebSocketTransport) request(w http.ResponseWriter, r *http.Request) {
+	// Pull api version from the url and determine if the version is supported
```
geoknee, line 88: Have we given any thought to how we plan on versioning the API? For example, does it make sense to use semantic versioning or not?

go.mod
```diff
@@ -129,11 +129,11 @@ require (
...
 	golang.org/x/exp v0.0.0-20230321023759-10a507213a29 // indirect
 	golang.org/x/mod v0.10.0 // indirect
-	golang.org/x/net v0.8.0 // indirect
-	golang.org/x/sync v0.1.0 // indirect
-	golang.org/x/text v0.8.0 // indirect
+	golang.org/x/net v0.10.0 // indirect
+	golang.org/x/sync v0.2.0 // indirect
+	golang.org/x/text v0.9.0 // indirect
 	golang.org/x/time v0.0.0-20220922220347-f3bd1da661af // indirect
-	golang.org/x/tools v0.7.0 // indirect
```
geoknee, line 136: I wasn't expecting any changes to `go.mod` or `go.sum`. Do you know why these are here?

Also the change to `.vscode/settings.json` seems unnecessary. Shall we snip that off this PR? We can always raise a new one if we feel strongly about that change.

geoknee (changes requested):
Thanks for this contribution @bitwiseguy !

Left a couple comments which I would like to see addressed. Otherwise, looks great!

As this is a breaking change, we can add a label to this PR for that. Also, it would be great practice to prepare the corresponding PR in http://github.com/statechannels/nitro-gui so that the GUI won't be broken for long (that repo will pull down the latest version of `go-nitro` on main branch). We can't guarantee the PRs will be merged atomically, but we can do it manually in quick succession if we get everything set up before hand.

go.mod
bitwiseguy: I will revert both of those changes. I ran a `go mod tidy` command that automatically updated the `go.mod` and `go.sum` files

rpc/transport/ws/server.go
bitwiseguy: To give us the highest likelihood for compatibility between go-nitro nodes and clients, I am proposing that we just use major api versions in the jsonrpc server url. If there is a breaking change, we can bump the api version (and maybe servers support at least one older version of the api?). If there is a non-breaking change (i.e. a minor revision bump) I don't think we would want to change the jsonrpc server endpoint because clients would have to explicitly change their requests. Also it seems like a single go-nitro might have to simultaneously support too many versions of the api (meaning lots of duplicated code) if we are differentiating between the minor revisions through the url. Maybe there is still a way to check the go-nitro node's overall version (through the commit hash or build tag), which could help differentiate between jsonrpc api minor versions without breaking compatibility.

That was my thinking, but happy to change if you feel like it should be built differently.

geoknee: Sounds like good thinking to me.

bitwiseguy:
@geoknee here is a corresponding PR to update the nitro-gui repo: https://github.com/statechannels/nitro-gui/pull/69

PR #1327, opened 2023-06-07
Add TestStartStop MWE of tcp connection refused error

geoknee:
I have a suspicion that we aren't cleaning up nitro nodes properly / possibly have a race condition around the message service shutdown.. 

I think related to #1323 

This new test seems to pass every time (on my machine) when I run it with `-count=1`. But if I run it with a higher count, I often get a failure. 

`go test ./client_test -run TestStartStop -v -count=10`

So I am wondering if there is something lingering on from the previous test which interferes with the next one. 

Error message:

```
{"level":"debug","engine":"0x111A00","time":1686154669417,"caller":"engine.go:151","message":"Constructed Engine"}
{"level":"trace","client":"0x111A00868581f73AB42FEEF67D235Ca09ca1E8db","rpc":"client","scope":"","time":1686154669417,"message":"Subscribed to notifications"}
{"level":"debug","engine":"0x3bc760","time":1686154669420,"caller":"engine.go:151","message":"Constructed Engine"}
{"level":"trace","client":"0x3bc7608cB3D5ea0572717C276386b8a12Fd8289B","rpc":"client","scope":"","time":1686154669420,"message":"Subscribed to notifications"}
{"level":"debug","engine":"0x9d31D3","time":1686154669421,"caller":"engine.go:151","message":"Constructed Engine"}
{"level":"trace","client":"0x9d31D3448D20B777C252EbBA830c1BbA8532BE43","rpc":"client","scope":"","time":1686154669422,"message":"Subscribed to notifications"}
{"level":"debug","message-service":"0x111A00","peerInfo":{"Id":"16Uiu2HAmDWx6b4aRap2nxiW3QwyyPSQCrYSU2wX84HWa3jWGENQW","Address":"0x3bc7608cb3d5ea0572717c276386b8a12fd8289b"},"time":1686154669424,"caller":"service.go:172","message":"New peer found"}
{"level":"debug","message-service":"0x3bc760","peerInfo":{"Id":"16Uiu2HAmHntR3SGeS7iF2tdeNBefSahXBhmTrqVozVLHydxzkaZn","Address":"0x111a00868581f73ab42feef67d235ca09ca1e8db"},"time":1686154669424,"caller":"service.go:172","message":"New peer found"}
[...]

PR #1337, opened 2023-06-08
handle any error in closing mdns

geoknee:
Let's do the right thing, seeing as mdns is in the dock right now #1327


lalexgap:
Ended up cherry-picking this into #1328

PR #1348, opened 2023-06-12
tidy up `natsTransport.unsubscribeFromTopic`

geoknee:
Currently, this function is recursive, requiring that some of its initial state is set by the calling function. If that state was set to a number greater than three, we could get an infinite loop and infinite recursion. The function also does not ever return an non-nil error. 

This change refactors the function to 
* make it non recursive
* have clear godoc explaining what the passed parameter means (it now sets the number of retries)
* return the most recent error after retrying the supplied number of times.

### DIFF
--- a/rpc/transport/nats/server.go
+++ b/rpc/transport/nats/server.go
@@ -35,7 +35,7 @@ func newNatsTransport(url string) (*natsTransport, error) {
 
 func (c *natsTransport) Close() {
 	for _, sub := range c.natsSubscriptions {
-		err := c.unsubscribeFromTopic(sub, 0)
+		err := c.unsubscribeFromTopic(sub, 3)
 		if err != nil {
 			log.Error().Err(err).Msgf("failed to unsubscribe from a topic: %s", sub.Subject)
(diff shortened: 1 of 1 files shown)


rpc/transport/nats/server.go
```diff
@@ -35,20 +35,23 @@ func newNatsTransport(url string) (*natsTransport, error) {
...
 
-func (c *natsTransport) unsubscribeFromTopic(sub *nats.Subscription, try int32) error {
+// unsubscribeFromTopic will attempt to unsubscribe the supplied subscription. On error, it will retry up to retries times.
+func (c *natsTransport) unsubscribeFromTopic(sub *nats.Subscription, retries uint32) error {
+	var try uint32
 	err := sub.Unsubscribe()
-	if err != nil && try < 3 {
-		return c.unsubscribeFromTopic(sub, try+1)
+	for err != nil && try < retries {
+		err = sub.Unsubscribe()
```
lalexgap, line 51: nit but should we log out any errors that occur?

PR #1351, opened 2023-06-12
Add ability to provide bearer auth token for chainservice

bitwiseguy:
In order for our chainservice to be able to send requests to a production Filecoin Lotus node, we need to be able to set a bearer auth token in the header of those requests. This PR allows us to pass a `--chainauthtoken` flag or set a `chainauthtoken` variable in the config file. This token will then get passed and set within the eth rpc client (wss or https) so that subsequent requests will include the auth header.

### Testing
To test that this was working I used two different Infura projects as the targeted `chainurl`. 

1. wss://mainnet.infura.io/ws/v3/<projectId_1> : no auth header required
2. wss://mainnet.infura.io/ws/v3/<projectId_2> : enabled the "JWT required" option, which says "Restrict API usage to requests that include a valid JWT. Select this to enable Bearer Auth."

To create and upload the JWT required for [2], I followed the Infura documentation [here](https://docs.infura.io/networks/ethereum/how-to/secure-a-project/use-jwts). 

If you want to test this code using my preconfigured Infura projects listed as [1] and [2] above, please reach out via Slack and I can provide those so that you don't have to run through those same prerequisite steps.

### DIFF
--- a/client/engine/chainservice/eth_chainservice.go
+++ b/client/engine/chainservice/eth_chainservice.go
@@ -61,11 +61,11 @@ const MAX_QUERY_BLOCK_RANGE = 2000
 const RESUB_INTERVAL = 2*time.Minute + 30*time.Second
 
 // NewEthChainService is a convenient wrapper around NewEthChainService, which provides a simpler API
-func NewEthChainService(chainUrl, chainPk string, naAddress, caAddress, vpaAddress common.Address, logDestination io.Writer) (*EthChainService, error) {
+func NewEthChainService(chainUrl, chainAuthToken, chainPk string, naAddress, caAddress, vpaAddress common.Address, logDestination io.Writer) (*EthChainService, error) {
 	if vpaAddress == caAddress {
 		return nil, fmt.Errorf("virtual payment app address and consensus app address cannot be the same: %s", vpaAddress.String())
 	}
-	ethClient, txSigner, err := chainutils.ConnectToChain(context.Background(), chainUrl, common.Hex2Bytes(chainPk))
+	ethClient, txSigner, err := chainutils.ConnectToChain(context.Background(), chainUrl, chainAuthToken, common.Hex2Bytes(chainPk))
 	if err != nil {
 		panic(err)
 	}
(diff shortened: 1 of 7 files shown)


lalexgap (approved):
Awesome, LGTM!

client/engine/chainservice/utils/utils.go
```diff
@@ -7,18 +7,33 @@ import (
...
 
 // ConnectToChain connects to the chain at the given url and returns a client and a transactor.
-func ConnectToChain(ctx context.Context, chainUrl string, chainPK []byte) (*ethclient.Client, *bind.TransactOpts, error) {
-	client, err := ethclient.Dial(chainUrl)
+func ConnectToChain(ctx context.Context, chainUrl, chainAuthToken string, chainPK []byte) (*ethclient.Client, *bind.TransactOpts, error) {
+	var rpcClient *rpc.Client
+	var err error
+
+	if chainAuthToken != "" {
+		fmt.Println("Adding bearer token authorization header to chain service")
```
bitwiseguy, line 19: Is there a service-wide logger I should use instead of `fmt.Print`?

client/engine/chainservice/eth_chainservice.go
```diff
@@ -61,11 +61,11 @@ const MAX_QUERY_BLOCK_RANGE = 2000
 const RESUB_INTERVAL = 2*time.Minute + 30*time.Second
 
 // NewEthChainService is a convenient wrapper around NewEthChainService, which provides a simpler API
```
geoknee, line 63: Out of scope, but
```suggestion
// NewEthChainService is a convenient wrapper around newEthChainService, which provides a simpler API
```
and also on L81.

client/engine/chainservice/utils/utils.go
geoknee: You could have this function accept a logger (say a suitable interface like `io.Writer`). Then, the calling context can pass in something suitable. 

For example, the chainservice has a logger https://github.com/statechannels/go-nitro/blob/3cb69815a77fe6daac09ac1d02bedb7c4d6b882e/client/engine/chainservice/eth_chainservice.go#L46. The `logDestination` is in scope when constructing the chain service, when `ConnectToChain` is called.

geoknee (approved):
Thanks @bitwiseguy. Left one suggestion. Up to you if you want to tackle that pre or post merge.

Issue #1326, opened 2023-06-07
Small spike: Investigate sending vouchers with the boost http request


geoknee:
Right now the nitro client is fully responsible for communicating a voucher to other participants. However it would be nice if we could just submit the payment for data right along with the request. 

We would just need to include the amount and the signature.

This would mean updating `go-nitro` to expose an API method to receive a voucher.

geoknee:
This is probably the neatest way to associate payments with retrieval requests.

Other solutions will require more book keeping to prevent payments being redeemed multiple times. 

I think the nicest API would be to have a `pushVoucher` or `pushMessage` API which returns a **payment amount**.

This would be a **non**-idempotent method, so that pushing a voucher a second (or third etc) time results in a payment amount of zero. 

We use replace-by-incentive vouchers. So:


```go
pushVoucher({c,10})
=> payment amount = 10
pushVoucher({c,20})
=> payment amount = 10
pushVoucher({c,20})
=> payment amount = 0
```

This makes it pretty easy for a retrieval provider to:
*  accept a voucher as http request params
*  push it into their nitro node
* inspect the returned amount to see if it covers the request.

If it doesn't cover the request, what happens? Obviously the request is not serviced. But the retrieval provider has now been paid some amount which won't count towards any future requests. That's something we either need to be comfortable with (can we fall back on the micropayments idea and not worry about it?), or solve somehow.

Clearly the `nitro.Pay` method needs to now return the voucher over the API on the client side of the interaction.

geoknee:
Via #1385

PR #1356, opened 2023-06-14
WIP: Receive Vouchers via RPC client

NiloCK:
Toward #1326.

PR introduces [CreatePayment](https://github.com/statechannels/go-nitro/blob/1c6d36a719fdf04a59f86bb65803d20f088cafef/rpc/client.go#L97) and [ReceivePayment](https://github.com/statechannels/go-nitro/blob/1c6d36a719fdf04a59f86bb65803d20f088cafef/rpc/client.go#L104) methods to the go-nitro rpc client, plumbing (go-nitro rpc-server & go nitro Client) to serve these requests, as well as a [test](https://github.com/statechannels/go-nitro/blob/1c6d36a719fdf04a59f86bb65803d20f088cafef/client_test/rpc_test.go#L61) for their correctness.

Some modest refactors along the way as well.


client/engine/engine.go
```diff
@@ -521,23 +521,23 @@ func (e *Engine) handlePaymentRequest(request PaymentRequest) (EngineEvent, erro
...
 		e.logMessage(message, Outgoing)
 		e.recordMessageMetrics(message)
 		e.msg.Send(message)
 	}
 }
 
-// executeSideEffects executes the SideEffects declared by cranking an Objective or handling a payment request.
-func (e *Engine) executeSideEffects(sideEffects protocols.SideEffects) error {
+// ExecuteSideEffects executes the SideEffects declared by cranking an Objective or handling a payment request.
+func (e *Engine) ExecuteSideEffects(sideEffects protocols.SideEffects) error {
```
NiloCK, line 540: This change is suspect. Before this change, all `sideEffect`s were manufactured locally by the running engine. They were "trusted" data. Executing them is presumed-safe (bugs notwithstanding), and `executeSideEffects` has not needed to be explicitly defensive in its implementation.

rpc/server.go
```diff
@@ -100,6 +102,66 @@ func (rs *RpcServer) registerHandlers() (err error) {
 				rs.client.Pay(obj.Channel, big.NewInt(int64(obj.Amount)))
 				return obj
 			})
+		case serde.ReceiveVoucherRequestMethod:
```
NiloCK, line 105: This breaks substantially from the established pattern in this file.

It performs validation of received voucher in order to return an immediate result to the calling client.

client/engine/engine.go
NiloCK: This was replaced with a more sensible PushMessage method, so incoming data is properly "scrutinized".

NiloCK:
Closing in favour of #1370, which was rebased on current main.

PR #1357, opened 2023-06-14
upgrade actions/setup-go

geoknee:


NiloCK (approved):
As far as I can tell, the [v3](https://github.com/actions/checkout/releases/tag/v3.0.0) is mostly about the default node runtime. Can't hurt in any case.

PR #1359, opened 2023-06-15
Return jsonrpc error objects from rpc server instead of panic

bitwiseguy:
Instead of panicking within the jsonrpc server, we should gracefully return error objects, as defined by the [jsonrpc spec](https://www.jsonrpc.org/specification). 

Addresses the pre-requisite to this boost issue: https://github.com/statechannels/boost/issues/8

### Pros to this approach
`JsonRpcError` now implements the standard `Error` interface, so the nitro core/client code can return either a JsonRpcError object or standard Error.

### Cons to this approach
To avoid circular dependencies, I had to move JsonRpcError code into the `types` package so that both the `rpc/` server and `client` packages could import and use those types.


rpc/server.go
```diff
@@ -82,72 +83,52 @@ func (rs *RpcServer) registerHandlers() (err error) {
...
-				return rs.client.CloseVirtualChannel(obj.ChannelId)
+			return processRequest(rs, requestData, func(obj virtualdefund.ObjectiveRequest) (protocols.ObjectiveId, error) {
+				return rs.client.CloseVirtualChannel(obj.ChannelId), nil
 			})
 		case serde.PayRequestMethod:
-			return processRequest(rs, requestData, func(obj serde.PaymentRequest) serde.PaymentRequest {
+			return processRequest(rs, requestData, func(obj serde.PaymentRequest) (serde.PaymentRequest, error) {
 				rs.client.Pay(obj.Channel, big.NewInt(int64(obj.Amount)))
-				return obj
+				return obj, nil
```
geoknee, line 112: I may be missing something, but it looks like a lot of these returned errors are set to `nil`? 

In some cases, that seems to make sense (returning the version for example). In other cases, it seems that we need to return an error (making a payment for example, in the case of insufficient funds). Is there more work to do to get those errors generated / wired up?

geoknee (approved):
Looks good to me, and I think relocation of types to common dependency is OK. 

Left a comment.

types/serde.go
```diff
@@ -16,3 +16,24 @@ func (d *Destination) UnmarshalText(text []byte) error {
...
+}
+
+var (
+	ParseError                       = JsonRpcError{Code: -32700, Message: "Parse error"}
+	InvalidRequestError              = JsonRpcError{Code: -32600, Message: "Invalid Request"}
+	MethodNotFoundError              = JsonRpcError{Code: -32601, Message: "Method not found"}
+	InvalidParamsError               = JsonRpcError{Code: -32602, Message: "Invalid params"}
+	InternalServerError              = JsonRpcError{Code: -32603, Message: "Internal error"}
+	UnexpectedRequestUnmarshalError  = JsonRpcError{Code: -32010, Message: "Unexpected unmarshal error"}
+	UnexpectedRequestUnmarshalError2 = JsonRpcError{Code: -32009, Message: "Unexpected unmarshal error"}
```
geoknee, line 38: I'm not sure I get why we want two errors like this?

bitwiseguy: I'm not sure what is the difference between the two `UnexpectedRequestUnmarshalError` error codes. Perhaps @kerzhner could provide some additional context and advise if we should keep both or reduce to one

rpc/server.go
bitwiseguy: Currently the underlying `client` methods do not all return an error object. For the ones that don't, I just assigned a `nil` error for now to satisfy the newly updated `processPayload` return types. I believe the underlying `client` methods should be updated to all return an error object. If you agree, and it won't create a breaking change for using go-nitro as a library, I can either include that update as part of this PR or a follow-up one

geoknee: Thanks, that makes sense! 

A follow-up PR would be great 🙏

types/serde.go
bitwiseguy: I spoke with @kerzhner and updated the `Message` for those two errors to distinguish between the two

PR #1364, opened 2023-06-20
Add request-specific validation of jsonrpc request params

bitwiseguy:
We already had some validation to ensure the jsonrpc request is properly formatted according to jsonrpc spec 2.0. This PR adds validation for the "params" field within the jsonrpc request, which is different for each jsonrpc method.

### DIFF
--- a/client_test/crash_test.go
+++ b/client_test/crash_test.go
@@ -84,7 +84,10 @@ func TestCrashTolerance(t *testing.T) {
 }
 
 func directlyDefundALedgerChannel(t *testing.T, alpha client.Client, beta client.Client, channelId types.Destination) {
-	id, _ := alpha.CloseLedgerChannel(channelId)
+	id, err := alpha.CloseLedgerChannel(channelId)
+	if err != nil {
+		t.Fatal(err)
+	}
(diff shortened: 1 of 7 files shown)


client_test/crash_test.go
```diff
@@ -84,7 +84,7 @@ func TestCrashTolerance(t *testing.T) {
 }
 
 func directlyDefundALedgerChannel(t *testing.T, alpha client.Client, beta client.Client, channelId types.Destination) {
-	id := alpha.CloseLedgerChannel(channelId)
+	id, _ := alpha.CloseLedgerChannel(channelId)
```
kerzhner, line 87: We could check the returned error and fail the test on error. This comment also applies to other changes in test files.

kerzhner (approved):
Looks great! Are you planning on adding unit or integration tests for the new errors?

client_test/crash_test.go
bitwiseguy: Good call. Fixed each of these tests to inspect the returned error

bitwiseguy:
> Looks great! Are you planning on adding unit or integration tests for the new errors?

Pushed a commit with some rpc request validation unit tests

Issue #1275, opened 2023-05-11
Don't panic: Implement error handling in RPC server and client


lalexgap:
We should handle sending back errors to RPC clients when something goes wrong. That means deciding on a collection of errors that we want to return via our JSON RPC API.

lalexgap:
This will resolve
- https://github.com/statechannels/go-nitro/issues/1257
- https://github.com/statechannels/go-nitro/issues/1303

PR #1368, opened 2023-06-21
Add input validation to query.GetVirtualChannel

bitwiseguy:
In order to reduce friction of go-nitro integration into boost, this PR adds some input validation to the rpcClient.GetVirtualChannel method. This means the validation occurs inside go-nitro code instead of the burden being placed on boost or another consumer

Resolves https://github.com/statechannels/boost/issues/8

### DIFF
--- a/rpc/client.go
+++ b/rpc/client.go
@@ -85,8 +85,8 @@ func (rc *RpcClient) GetPaymentChannel(chId types.Destination) query.PaymentChan
 	return waitForRequest[serde.GetPaymentChannelRequest, query.PaymentChannelInfo](rc, serde.GetPaymentChannelRequestMethod, req)
 }
 
-// CreateVirtual creates a new virtual channel
-func (rc *RpcClient) CreateVirtual(intermediaries []types.Address, counterparty types.Address, ChallengeDuration uint32, outcome outcome.Exit) virtualfund.ObjectiveResponse {
+// CreatePaymentChannel creates a new virtual channel
+func (rc *RpcClient) CreatePaymentChannel(intermediaries []types.Address, counterparty types.Address, ChallengeDuration uint32, outcome outcome.Exit) virtualfund.ObjectiveResponse {
 	objReq := virtualfund.NewObjectiveRequest(
 		intermediaries,
 		counterparty,
@@ -98,8 +98,8 @@ func (rc *RpcClient) CreateVirtual(intermediaries []types.Address, counterparty 
 	return waitForRequest[virtualfund.ObjectiveRequest, virtualfund.ObjectiveResponse](rc, serde.CreatePaymentChannelRequestMethod, objReq)
 }
 
-// CloseVirtual closes a virtual channel
-func (rc *RpcClient) CloseVirtual(id types.Destination) protocols.ObjectiveId {
+// ClosePaymentChannel closes a virtual channel
+func (rc *RpcClient) ClosePaymentChannel(id types.Destination) protocols.ObjectiveId {
 	objReq := virtualdefund.NewObjectiveRequest(
 		id)
 
@@ -123,7 +123,7 @@ func (rc *RpcClient) GetPaymentChannelsByLedger(ledgerId types.Destination) []qu
 }
 
 // CreateLedger creates a new ledger channel
-func (rc *RpcClient) CreateLedger(counterparty types.Address, ChallengeDuration uint32, outcome outcome.Exit) directfund.ObjectiveResponse {
+func (rc *RpcClient) CreateLedgerChannel(counterparty types.Address, ChallengeDuration uint32, outcome outcome.Exit) directfund.ObjectiveResponse {
 	objReq := directfund.NewObjectiveRequest(
 		counterparty,
 		100,
@@ -135,7 +135,7 @@ func (rc *RpcClient) CreateLedger(counterparty types.Address, ChallengeDuration 
 }
 
 // CloseLedger closes a ledger channel
-func (rc *RpcClient) CloseLedger(id types.Destination) protocols.ObjectiveId {
+func (rc *RpcClient) CloseLedgerChannel(id types.Destination) protocols.ObjectiveId {
 	objReq := directdefund.NewObjectiveRequest(id)
 
 	return waitForRequest[directdefund.ObjectiveRequest, protocols.ObjectiveId](rc, serde.CloseLedgerChannelRequestMethod, objReq)
--- a/rpc/client.go
+++ b/rpc/client.go
@@ -3,7 +3,9 @@ package rpc
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"net/http"
 	"os"
 	"sync"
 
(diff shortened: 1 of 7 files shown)


rpc/client.go
```diff
@@ -80,13 +82,16 @@ func NewHttpRpcClient(rpcServerUrl string) (*RpcClient, error) {
...
-func (rc *RpcClient) GetVirtualChannel(id types.Destination) query.PaymentChannelInfo {
-	req := serde.GetPaymentChannelRequest{Id: id}
+func (rc *RpcClient) GetVirtualChannel(chId types.Destination) (query.PaymentChannelInfo, int, error) {
+	if (chId == types.Destination{}) {
+		return query.PaymentChannelInfo{}, http.StatusBadRequest, errors.New("a valid channel id must be provided")
+	}
+	req := serde.GetPaymentChannelRequest{Id: chId}
 
-	return waitForRequest[serde.GetPaymentChannelRequest, query.PaymentChannelInfo](rc, serde.GetPaymentChannelRequestMethod, req)
+	return waitForRequest[serde.GetPaymentChannelRequest, query.PaymentChannelInfo](rc, serde.GetPaymentChannelRequestMethod, req), http.StatusOK, nil
```
geoknee, line 91: I'm not sure about returning an http status code. Although in practice we tend to use http as a transport for the RPC connection, it is in fact transport agnostic.

If the input validation were being performed by the RPC server (instead of the client), it could make more sense to return an http error (again depending on the transport). 

In fact, thinking about this some more, I wonder if it does in fact make more sense to the validation on the server side.

The main thrust behind having validation is to avoid the nitro node panicking https://github.com/statechannels/boost/pull/3#discussion_r1223402151 . 

The deeper in we put the validation, the more protection we get. Putting the validation in the go rpc client means we don't get protection when
* someone uses the TS rpc client (unless we duplicate the validation there too)
* someone uses go-nitro as a library, not as a service

So I think we should rethink our approach a little. Perhaps start with a test which triggers the go-nitro node to panic by sending in a query for a zero channel using the rpc client. Then, we would like to transform that panic to an error, bubble it up to the rpc sever, and send it back to the rpc client as an json rpc error response. Ultimately that will be received by the rpc client and it will still cause an error to be returned from `RpcClient.GetVirtualChannel`.

Keen to hear your thoughts! And apologies that the direction wasn't clear from the outset.

bitwiseguy: That makes sense to me and I agree that pushing validation deeper would be better. I noticed that in our jsonrpc server.go we have methods for `GetPaymentChannel` and `GetLedgerChannel`. Would `GetVirtualChannel` be the same as `GetPaymentChannel` or should I create an additional jsonrpc method?

bitwiseguy: There are a few other instances in the rpc client code where "Virtual" is used instead of "Payment". To improve the naming consistency, would you agree with renaming the following client methods?

`CreateVirtual` --> `CreatePaymentChannel`
`CloseVirtual` --> `ClosePaymentChannel`
`CreateLedger` --> `CreateLedgerChannel`
`CloseLedger` --> `CloseLedgerChannel`

bitwiseguy: I wonder if we might also want to align some of the json-rpc methods with the names of the core/node methods. The following method names are a little confusing:

json-rpc | node
-- | --
virtual_fund | CreateVirtualPaymentChannel
virtual_defund | CloseVirtualChannel
direct_fund | CreateLedgerChannel
direct_defund | CloseLedgerChannel

rpc/client.go
```diff
@@ -80,14 +79,14 @@ func NewHttpRpcClient(rpcServerUrl string) (*RpcClient, error) {
...
-	req := serde.GetPaymentChannelRequest{Id: id}
+func (rc *RpcClient) GetPaymentChannel(chId types.Destination) query.PaymentChannelInfo {
+	req := serde.GetPaymentChannelRequest{Id: chId}
 
 	return waitForRequest[serde.GetPaymentChannelRequest, query.PaymentChannelInfo](rc, serde.GetPaymentChannelRequestMethod, req)
 }
 
-// CreateLedger creates a new ledger channel
-func (rc *RpcClient) CreateVirtual(intermediaries []types.Address, counterparty types.Address, ChallengeDuration uint32, outcome outcome.Exit) virtualfund.ObjectiveResponse {
+// CreatePaymentChannel creates a new virtual channel
```
geoknee, line 88:
```suggestion
// CreatePaymentChannel creates a new virtual payment channel
```

rpc/client.go
```diff
@@ -96,15 +95,15 @@ func (rc *RpcClient) CreateVirtual(intermediaries []types.Address, counterparty
 		rand.Uint64(),
 		common.Address{})
 
-	return waitForRequest[virtualfund.ObjectiveRequest, virtualfund.ObjectiveResponse](rc, serde.VirtualFundRequestMethod, objReq)
+	return waitForRequest[virtualfund.ObjectiveRequest, virtualfund.ObjectiveResponse](rc, serde.CreatePaymentChannelRequestMethod, objReq)
 }
 
-// CloseVirtual closes a virtual channel
-func (rc *RpcClient) CloseVirtual(id types.Destination) protocols.ObjectiveId {
+// ClosePaymentChannel closes a virtual channel
```
geoknee, line 101:
```suggestion
// ClosePaymentChannel attempts to close the payment channel with supplied id
```

When we get on to working on our next grant, we should be able to drop the "attempts to" disclaimer. For now, closing a channel requires cooperative counterparties.

rpc/client.go
```diff
@@ -124,29 +123,29 @@ func (rc *RpcClient) GetPaymentChannelsByLedger(ledgerId types.Destination) []qu
...
+	return waitForRequest[directdefund.ObjectiveRequest, protocols.ObjectiveId](rc, serde.CloseLedgerChannelRequestMethod, objReq)
 }
 
 // Pay uses the specified channel to pay the specified amount
-func (rc *RpcClient) Pay(id types.Destination, amount uint64) {
+func (rc *RpcClient) Pay(id types.Destination, amount uint64) serde.PaymentRequest {
 	pReq := serde.PaymentRequest{Amount: amount, Channel: id}
 
-	waitForRequest[serde.PaymentRequest, serde.PaymentRequest](rc, serde.PayRequestMethod, pReq)
+	return waitForRequest[serde.PaymentRequest, serde.PaymentRequest](rc, serde.PayRequestMethod, pReq)
```
geoknee, line 148: Is this change intended / relevant?

rpc/serde/jsonrpc.go
```diff
@@ -13,17 +13,17 @@ import (
...
-	GetPaymentChannelRequestMethod   RequestMethod = "get_payment_channel"
-	GetLedgerChannelRequestMethod    RequestMethod = "get_ledger_channel"
-	GetPaymentChannelsByLedgerMethod RequestMethod = "get_payment_channels_by_ledger"
-	GetAllLedgerChannelsMethod       RequestMethod = "get_all_ledger_channels"
+	GetAddressMethod                  RequestMethod = "get_address"
+	VersionMethod                     RequestMethod = "version"
+	CreateLedgerChannelRequestMethod  RequestMethod = "create_ledger_channel"
+	CloseLedgerChannelRequestMethod   RequestMethod = "close_ledger_channel"
+	CreatePaymentChannelRequestMethod RequestMethod = "create_payment_channel"
+	ClosePaymentChannelRequestMethod  RequestMethod = "close_payment_channel"
```
geoknee, line 21: 👍

rpc/client.go
```diff
@@ -80,14 +79,14 @@ func NewHttpRpcClient(rpcServerUrl string) (*RpcClient, error) {
 	return c, nil
 }
 
-func (rc *RpcClient) GetVirtualChannel(id types.Destination) query.PaymentChannelInfo {
-	req := serde.GetPaymentChannelRequest{Id: id}
+func (rc *RpcClient) GetPaymentChannel(chId types.Destination) query.PaymentChannelInfo {
+	req := serde.GetPaymentChannelRequest{Id: chId}
 
 	return waitForRequest[serde.GetPaymentChannelRequest, query.PaymentChannelInfo](rc, serde.GetPaymentChannelRequestMethod, req)
```
geoknee, line 85: I'm not 100% sure but it looks like `waitForRequest` will panic in many error conditions, such that we will still crash the go-nitro server if we query for a channel with.a zero id. So still some more work to do to fix that! https://github.com/statechannels/go-nitro/issues/1275.

geoknee (approved):
Looks good!

geoknee:
@bitwiseguy I put the breaking change label on this, I think we will need to update TS rpc client in tandem with this change.

rpc/client.go
bitwiseguy: Reverted

rpc/client.go
bitwiseguy: I added [this line](https://github.com/statechannels/go-nitro/pull/1368/files#diff-e411c027b510060c0ba3026138d11ece9c213a5e5927164270fb1f9ed45a8249R180) to test that the server does not panic when passed a non-existent chId. I believe its working as expected.

bitwiseguy:
> @bitwiseguy I put the breaking change label on this, I think we will need to update TS rpc client in tandem with this change.

I will work on a PR for the TS rpc client and queue that one up before merging this one

rpc/client.go
geoknee: Awesome, I guess that settles it then 👍

bitwiseguy:
PR to make similar updates to TS client: https://github.com/statechannels/nitro-gui/pull/79

PR #1372, opened 2023-06-23
Add go-nitro docker image

kerzhner:
This PR adds docker to go-nitro. Using `make` commands, a docker image can be built and a container can be run.

At the moment, the docker container encapsulates three go-nitro nodes to simplify running demos. Down the line, a docker container should encapsulate a single nitro node.


rpc/transport/ws/server.go
```diff
@@ -36,7 +36,7 @@ type serverWebSocketTransport struct {
 func NewWebSocketTransportAsServer(port string) (*serverWebSocketTransport, error) {
 	wsc := &serverWebSocketTransport{port: port, notificationListeners: safesync.Map[chan []byte]{}}
 
-	tcpListener, err := net.Listen("tcp", websocketServerAddress+wsc.port)
+	tcpListener, err := net.Listen("tcp", ":"+wsc.port)
```
kerzhner, line 39: Listening on a hardcoded IP address of 127.0.0.1 is not working with Docker. [From `Listen` documentation](https://pkg.go.dev/net#Listen): "For TCP networks, if the host in the address parameter is empty or a literal unspecified IP address, Listen listens on all available unicast and anycast IP addresses of the local system."

geoknee (approved):

Issue #1362, opened 2023-06-16
Rename `LedgerChannelInfo` struct fields


NiloCK:
These calls to a go rpc client:

```
bobL := bobClient.GetLedgerChannel(bobLedger.ChannelId)
t.Log("Bob", bob.Address())
t.Log("BobLClient", bobL.Balance.Client)
t.Log("BobLHub", bobL.Balance.Hub)
```

Result in the following output:

```
rpc_test.go:182: Bob 0x6813Eb9362372EEF6200f3b1dbC3f819671cBA69
rpc_test.go:183: BobLClient 0x2B5AD5c4795c026514f8317c7a215E218DcCD6cF
rpc_test.go:184: BobLHub 0x6813Eb9362372EEF6200f3b1dbC3f819671cBA69
```

(Bob is not the hub in this scenario).

`Hub` `Client` seem to be referring here to `Follower` `Leader`, which are a consequence of which participant initiated the channel.

geoknee:
Do `Hub` and `Client` really make sense for multihop situations? I think we need to change this data structure:

https://github.com/statechannels/go-nitro/blob/3cb69815a77fe6daac09ac1d02bedb7c4d6b882e/client/query/types.go#L41-L48

PR #1377, opened 2023-06-26
Rename go-nitro/client --> go-nitro/node

bitwiseguy:
Renames the `go-nitro/client` package (and all references/comments) to `go-nitro/node`. No functionality changes within this PR.

### Questions
1. Should we also update adr .md files to replace `client` with `node`? Not sure if we want to preserve original language.
2. Should this field in the `LedgerChannelBalance` be `Client` or `Node`? https://github.com/statechannels/go-nitro/blob/main/client/query/types.go#L45


geoknee:
Thanks @bitwiseguy !
> 1. Should we also update adr .md files to replace `client` with `node`? Not sure if we want to preserve original language.

I think we _should_ replace those cases. We'll have the git history as a paper trail to recover the original documents if we want (they don't change much). 

> 2. Should this field in the `LedgerChannelBalance` be `Client` or `Node`? https://github.com/statechannels/go-nitro/blob/main/client/query/types.go#L45

I would just leave this for now. Those fields are badly named and we have an issue to track that https://github.com/statechannels/go-nitro/issues/1362 . Hopefully a fix is imminent which won't even use "client" terminology.

> 3. Where is the source diagram for [this architecture image](https://github.com/statechannels/go-nitro/blob/main/client/architecture.png) so that we can change `go-nitro client` to `go-nitro node`?

https://github.com/statechannels/go-nitro/blob/c607f08399fdb881d6357170b46da0a8ada16404/client/readme.md#L26

This diagram should be updated since it is quite old. I'm happy to do that but I may need pestering. I suggest we don't let updating the diagram block this PR (we could file a new issue to address it).

geoknee (approved):
Changes look good to me with a cursory glance. 

Can you explain your methodology a bit? Did you use e.g. VSCode features to help generate this changeset?

geoknee:
Checking out this branch, I have some gitignored data at `client/engine/data/store_test/6010186420187918401687437093330253000` etc. I imagine this is safe to delete / we may want to encourage devs to do that cleanup manually.

bitwiseguy:
> Changes look good to me with a cursory glance.
> 
> Can you explain your methodology a bit? Did you use e.g. VSCode features to help generate this changeset?

I used the following methodology:

1. Used the VSCode feature to "Rename symbol", which changed the package name and updated all instances where that package name is imported within this repo.
2. Used the VSCode "Find in folder" to find all instances of the word "client" within this repo. There were around 70 files that had that word. I manually went through each of those files and made the decision if the use of `client` should be replaced with `node` in each instance.

NiloCK:
>     2. Should this field in the `LedgerChannelBalance` be `Client` or `Node`? https://github.com/statechannels/go-nitro/blob/main/client/query/types.go#L45

Secondary comment on this: the usage of `client` here refers to one participant being the client of services provided by the hub. Still more overloading of the word client and fuel to the fire motivating this change.

PR #1384, opened 2023-06-27
Script to start servers and open channels

kerzhner:
This PR adds the `internal/nitro-network` package with the `InitializeNitroNetwork()` function. This package spins up 3 nitro rpc servers and creates 2 ledger channels and a virtual channel between those servers. This is used on startup of the go-nitro docker container.

With this work, demos of go-nitro no longer require running [a typescript create channels script](https://github.com/statechannels/nitro-gui/blob/main/packages/nitro-rpc-client/scripts/client-runner.ts). To demo go-nitro, the docker container is started and a web ui is pointed at the appropriate url.

Follow up work can be done to encapsulate a web ui inside the docker container.


.vscode/launch.json
```diff
@@ -4,6 +4,13 @@
   // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
   "version": "0.2.0",
   "configurations": [
+    {
```
kerzhner, line 7: This was added to help with debugging. It allows us to launch a go-nitro process in VSCode and set breakpoints.

internal/chain/chain.go
```diff
@@ -0,0 +1,83 @@
...
+func StartAnvil() (*exec.Cmd, error) {
+	chainCmd := exec.Command("anvil", "--chain-id", "1337")
+	chainCmd.Stdout = os.Stdout
+	chainCmd.Stderr = os.Stderr
+	err := chainCmd.Start()
+	if err == nil {
+		// Give the chain a second to start up
+		time.Sleep(1 * time.Second)
+	}
+	return chainCmd, err
```
geoknee, line 58: I'm a bit confused by this error handling. If there is no error, we wait for 1s. If there is an error, it will be returned, but with a non-nil `chainCmd.

internal/chain/chain.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+// DeployContracts deploys the  NitroAdjudicator contract.
```
geoknee, line 61:
```suggestion
// DeployContracts deploys the NitroAdjudicator, VirtualPaymentApp and ConsensusApp contracts.
```

internal/chain/chain.go
```diff
@@ -0,0 +1,83 @@
...
+func DeployContracts(ctx context.Context) (na common.Address, vpa common.Address, ca common.Address, err error) {
+	ethClient, txSubmitter, err := chainutils.ConnectToChain(context.Background(), "ws://127.0.0.1:8545", "", common.Hex2Bytes(FUNDED_TEST_PK))
```
geoknee, line 63:
```suggestion
	ethClient, txSubmitter, err := chainutils.ConnectToChain(context.Background(), ANVIL_CHAIN_URL, "", common.Hex2Bytes(FUNDED_TEST_PK))
```

node_test/rpc_test.go
```diff
@@ -13,9 +13,10 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"github.com/rs/zerolog"
 	"github.com/statechannels/go-nitro/channel/state/outcome"
-	"github.com/statechannels/go-nitro/internal/infra"
+	infraRpc "github.com/statechannels/go-nitro/internal/rpc"
```
geoknee, line 16: Nit: you have used the alias `interRpc` elsewhere.

internal/chain/chain.go
```diff
@@ -0,0 +1,83 @@
+package chain
```
geoknee, line 1: Something makes me a bit wary of having `internal/chain` and `internal/rpc` and `internal/utils` packages. I suppose it isn't obvious to me what they are for. Could we combine them all intro the `internal/nitro-network` package (could still keep the source files separate, but all in one folder)?

node_test/rpc_test.go
kerzhner: Good call. Updated in c1ac24e5.

internal/chain/chain.go
kerzhner: I initially started with the approach of one `internal/nitro-network` package. I then realized that I was pulling needed utilities out of other parts of our codebase and aggregating those in the `internal/nitro-network` package. So the package became a disorganized, miscellaneous directory of utilities.

My current philosophy is:
- If a utility is useful for consumers of `go-nitro`, the utility should be placed in a relevant package and exported. This is our usual pattern.
- If a utility is used only internally by a single package, the utility should be placed in that package and not exported. This is also our usual pattern.
- If a internal utility is used by multiple packages, it should be placed in a narrowly scoped internal package.

More than happy to discuss and reconsider this philosophy.

internal/chain/chain.go
geoknee: I see. I think it is valid to return a non-nil `chainCmd`, but I am not sure wha the convention is. I think almost everywhere else we return the default value alongside non-nil errors? I don't feel super strongly about it. 

More nitpicking -- it is also a bit unusual for us to do a check for nil rather than not nil.

```go
	if err != nil {
                 return nil, err
        }
        time.Sleep(1 * time.Second) // give time for chain to start up
	return chainCmd, nil
```

internal/chain/chain.go
geoknee: This sounds good to me 👍

internal/chain/chain.go
kerzhner: These are good points. Refactored in 499a4cec.

PR #1385, opened 2023-06-28
Allow for manual voucher exchange

lalexgap:
Based on this [comment](https://github.com/statechannels/go-nitro/pull/1370#discussion_r1244528266) I took a stab at avoiding the engine to create and receive vouchers. Instead the go-nitro node directly calls into the voucher manger to add or create vouchers.

I've created #1389 as tech debt to consider this approach vs #1370

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -196,6 +196,18 @@ func (c *Node) ReceivedVouchers() <-chan payments.Voucher {
 	return c.receivedVouchers
 }
 
+// CreateVoucher creates a voucher for the given channelId and amount and returns it.
+// It is the responsibility of the caller to send the voucher to the payee.
+func (c *Node) CreateVoucher(channelId types.Destination, amount *big.Int) (payments.Voucher, error) {
+	return c.vm.Pay(channelId, amount, *c.store.GetChannelSecretKey())
+}
+
+// ReceiveVoucher receives a voucher and returns the amount that was paid.
+// It can be used to add a voucher that was sent outside of the go-nitro system.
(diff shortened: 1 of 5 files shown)


rpc/serde/jsonrpc.go
```diff
@@ -99,7 +109,9 @@ type ResponsePayload interface {
 		query.LedgerChannelInfo |
 		VersionResponse |
 		GetAllLedgersResponse |
-		GetPaymentChannelsByLedgerResponse
+		GetPaymentChannelsByLedgerResponse |
+		payments.Voucher |
+		uint64
```
lalexgap, line 114: I used a `uint64` for convenience but this probably should be a `big.Int`

node/node.go
```diff
@@ -196,6 +196,18 @@ func (c *Node) ReceivedVouchers() <-chan payments.Voucher {
 	return c.receivedVouchers
 }
 
+// CreateVoucher creates a voucher for the given channelId and amount and returns it.
+// It is the responsibility of the caller to send the voucher to the payee.
```
NiloCK, line 200: Subtlety around increment vs running total here. Maybe:

```suggestion
// CreateVoucher creates and returns a voucher for the given channelId which increments the redeemable balance by amount.
// It is the responsibility of the caller to send the voucher to the payee.
```

rpc/serde/jsonrpc.go
lalexgap: https://github.com/statechannels/go-nitro/pull/1385/commits/5e799eb3694346f4ef47416d3b5bc2488fcd729f

NiloCK (approved):
:+1:

PR #1390, opened 2023-06-28
Return the amount added by a voucher

lalexgap:
Previously when receiving a voucher, our voucher manager/API just returns the total payment made on the channel (aka the largest voucher we've received).  However that doesn't let us know the change in balance due to the voucher, meaning we don't know how much the voucher was "worth".  If we want to know that we have to keep track of the channel balance before  and after`ReceiveVoucher` is called and compare the difference.

This PR updates the voucher manager and API's `ReceiveVoucher` to  return the difference between the incoming voucher and the largest voucher we have (alongside the total value). This helps simplify payment checks as you can easily check if a voucher resulted in a payment of `x`.

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -204,7 +204,7 @@ func (c *Node) CreateVoucher(channelId types.Destination, amount *big.Int) (paym
 
 // ReceiveVoucher receives a voucher and returns the amount that was paid.
 // It can be used to add a voucher that was sent outside of the go-nitro system.
-func (c *Node) ReceiveVoucher(v payments.Voucher) (*big.Int, error) {
+func (c *Node) ReceiveVoucher(v payments.Voucher) (total *big.Int, fromVoucher *big.Int, err error) {
 	return c.vm.Receive(v)
 }
 
--- a/node/node.go
+++ b/node/node.go
@@ -204,7 +204,7 @@ func (c *Node) CreateVoucher(channelId types.Destination, amount *big.Int) (paym
 
 // ReceiveVoucher receives a voucher and returns the amount that was paid.
 // It can be used to add a voucher that was sent outside of the go-nitro system.
-func (c *Node) ReceiveVoucher(v payments.Voucher) (total *big.Int, fromVoucher *big.Int, err error) {
+func (c *Node) ReceiveVoucher(v payments.Voucher) (total *big.Int, delta *big.Int, err error) {
 	return c.vm.Receive(v)
 }
 
--- a/payments/voucher_manager.go
+++ b/payments/voucher_manager.go
@@ -80,7 +80,7 @@ func (vm *VoucherManager) Pay(channelId types.Destination, amount *big.Int, pk [
 }
 
 // Receive validates the incoming voucher, and returns the total amount received so far as well as the amount received from the voucher
-func (vm *VoucherManager) Receive(voucher Voucher) (total *big.Int, fromVoucher *big.Int, err error) {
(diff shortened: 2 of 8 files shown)


node/node.go
```diff
@@ -204,7 +204,7 @@ func (c *Node) CreateVoucher(channelId types.Destination, amount *big.Int) (paym
 
 // ReceiveVoucher receives a voucher and returns the amount that was paid.
 // It can be used to add a voucher that was sent outside of the go-nitro system.
-func (c *Node) ReceiveVoucher(v payments.Voucher) (*big.Int, error) {
+func (c *Node) ReceiveVoucher(v payments.Voucher) (total *big.Int, fromVoucher *big.Int, err error) {
```
lalexgap, line 207: not sure if `total` and `fromVoucher` are the best terms.

rpc/serde/jsonrpc.go
```diff
@@ -92,6 +94,11 @@ type (
 	GetPaymentChannelsByLedgerResponse = []query.PaymentChannelInfo
 )
 
+type ReceiveVoucherResponse struct {
```
lalexgap, line 97: We could just return the fromVoucher value and avoid having to create a new response struct and modifying the schema in nitro-gui.  If someone does want to know the total value they can just call`GetPaymentChannel`

node/node.go
geoknee: I think they are OK. How about `total` and `delta`? In any case we should explain the meaning in the godoc.

payments/voucher_manager.go
```diff
@@ -79,43 +79,45 @@ func (vm *VoucherManager) Pay(channelId types.Destination, amount *big.Int, pk [
...
 	if err != nil {
-		return &big.Int{}, fmt.Errorf("channel not registered: %w", err)
+		return &big.Int{}, &big.Int{}, fmt.Errorf("channel not registered: %w", err)
 	}
 
 	// We only care about vouchers when we are the recipient of the payment
 	if vInfo.ChannelPayee != vm.me {
-		return &big.Int{}, nil
+		return &big.Int{}, &big.Int{}, nil
 	}
```
geoknee, line 92: Out of scope, but this seems at odd with th ebehaviour on L86. If the recipient is incorrect, or if the channel is not registered I think we should return an error.

payments/voucher_manager.go
```diff
@@ -79,43 +79,45 @@ func (vm *VoucherManager) Pay(channelId types.Destination, amount *big.Int, pk [
...
-	received := &big.Int{}
-	received.Set(voucher.Amount)
-	if types.Gt(received, vInfo.StartingBalance) {
-		return &big.Int{}, fmt.Errorf("channel has insufficient funds")
+
+	if types.Gt(voucher.Amount, vInfo.StartingBalance) {
+		return &big.Int{}, &big.Int{}, fmt.Errorf("channel has insufficient funds")
```
geoknee, line 95: Genuine Q: is there some value in returning the _actual_ total each time, even in the case of an error? It might make error handling less necessary for consumers...

rpc/serde/jsonrpc.go
geoknee: On the fence with this one -- I'm leaning towards making the breaking change, though. We shouldn't let the tail wag the dog.

geoknee (approved):

payments/voucher_manager.go
lalexgap: I don't think we can always return the `total` since we could fail to fetch the voucher info on L84, and never be able to figure out the current total. To keep things consistent I think it makes sense to not return the total here.

PR #1392, opened 2023-06-29
Add Address method to golang RPC client

kerzhner:
This work is a part of the refactor of golang create channels script.

### DIFF
--- a/node_test/rpc_test.go
+++ b/node_test/rpc_test.go
@@ -112,6 +112,13 @@ func executeNRpcTest(t *testing.T, connectionType transport.TransportType, n int
 	}
 	t.Logf("%d Clients created", n)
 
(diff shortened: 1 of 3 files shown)


NiloCK (approved):
:+1: 

Is there a case for the rpcClient holding onto the result locally and just recycling the cached result after a first run? Unlike lots of other query methods, I wouldn't expect the returned data to change over time.

That said, it'd only matter if we call this often and also care about how snappy it is.

kerzhner:
> Is there a case for the rpcClient holding onto the result locally and just recycling the cached result after a first run? Unlike lots of other query methods, I wouldn't expect the returned data to change over time.
> 
> That said, it'd only matter if we call this often and also care about how snappy it is.

Good idea. Implemented in ddfaa166.

PR #1396, opened 2023-06-30
Require correct `expectedHeld`, Do not give refunds

NiloCK:
A stricter version of `deposit`. Requires that `expectedHeld` and `held` are equal.

PR is work toward #1410.

Previously: `deposit` had guardrails that prevented users from "overfunding" their commitments to channels. If a user deposited into a channel whose holdings exceeded their expectation, the adjudicator would make adjustments to the incoming deposit and refund the difference to the caller.

Now: `deposit` simply rejects any deposits where existing holdings do not match expectation. Users still cannot over-deposit.


geoknee:
I'm on board with this. Refunds are an unnecessary feature which I don't think we envisaged anyone actually using.

This seems simpler, safer, and preserves the reorg-protection property.

NiloCK:
> I'm on board with this. Refunds are an unnecessary feature which I don't think we envisaged anyone actually using.
> 
> This seems simpler, safer, and preserves the reorg-protection property.

Will continue this branch then and use it as a base for the batch-deposit (which play nicest with this version anyhow).

nitro-protocol/contracts/MultiAssetHolder.sol
```diff
@@ -41,37 +41,23 @@ contract MultiAssetHolder is IMultiAssetHolder, StatusManager {
...
-        emit Deposited(channelId, asset, amountDeposited, nowHeld);
-
-        if (asset == address(0)) {
-            // refund whatever wasn't deposited.
-            uint256 refund = amount - amountDeposited;
-            (bool success, ) = msg.sender.call{value: refund}(''); //solhint-disable-line avoid-low-level-calls
-            require(success, 'Could not refund excess funds');
-        }
+        holdings[asset][channelId] += amount;
+        emit Deposited(channelId, asset, amount, holdings[asset][channelId]);
```
geoknee, line 60: Nit: I think we may be able to save some gas by reducing the number of storage reads? Currently there are 3, and I think in principle it should be do-able in 1. Might be worth an investigation (no need to hold up this PR though).

geoknee (approved):

nitro-protocol/contracts/MultiAssetHolder.sol
NiloCK: :+1: Tracked in #1410

PR #1397, opened 2023-06-30
fix: add `asset` docstring to `Deposited` event

NiloCK:
PR includes:
- Very small fix to a doc string (`asset` was not documented)
- a hint for readers of a piece of subtle bit manipulating code


nitro-protocol/test/contracts/libraries/Consensus/Consensus.test.ts
```diff
@@ -86,7 +86,9 @@ describe('requireConsensus', () => {
       if (reason) {
         await expectRevert(() => Consensus.requireConsensus(fixedPart, proof, candidate));
       } else {
-        await expectSucceedWithNoReturnValues(() => Consensus.requireConsensus(fixedPart, proof, candidate));
+        await expectSucceedWithNoReturnValues(() =>
```
NiloCK, line 89: These typescript changes were prompted by the CI failure https://github.com/statechannels/go-nitro/actions/runs/5423048165/jobs/9860465189

A little confusing for this to fail now - eslint is a devDependency pinned at 7.17.0., which was configured ~15 months ago. Running the same version on my own machine produces the same flagged issues

nitro-protocol/contracts/interfaces/IMultiAssetHolder.sol
```diff
@@ -71,6 +71,7 @@ interface IMultiAssetHolder {
     /**
      * @dev Indicates that `amountDeposited` has been deposited into `destination`.
      * @param destination The channel being deposited into.
+     * @param asset The asset being deposited. Zero address indicates ETH.
```
geoknee, line 74:
```suggestion
     * @param asset The asset being deposited. Zero address indicates the native token (e.g. ETH).
```

Just to cover all EVM chains.

geoknee (approved):

PR #1393, opened 2023-06-29
Refactor channel create script to use RPC clients

kerzhner:
- [ ] Client websocket close hangs

PR #1404, opened 2023-07-04
Fix double close of websocket connection

kerzhner:
Do not close the websocket connection from the server when the client initiated the websocket close.

When developing https://github.com/statechannels/go-nitro/pull/1393, rpc clients hang on close. Breakpoints show that [websocket Close](https://github.com/statechannels/go-nitro/blob/cabce74d3a1e85a28bf0f2f89c79f32e8dd44cf2/rpc/transport/ws/client.go#L69) is being called. But the [webscoket Read](https://github.com/statechannels/go-nitro/blob/cabce74d3a1e85a28bf0f2f89c79f32e8dd44cf2/rpc/transport/ws/client.go#L80) never resolves.

### DIFF
--- a/rpc/transport/ws/server.go
+++ b/rpc/transport/ws/server.go
@@ -154,7 +154,6 @@ func (wsc *serverWebSocketTransport) subscribe(w http.ResponseWriter, r *http.Re
 	if err != nil {
 		panic(err)
 	}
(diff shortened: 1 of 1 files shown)


geoknee (approved):

kerzhner:
Closing in favor of https://github.com/statechannels/go-nitro/pull/1407.

PR #1405, opened 2023-07-06
Format server error objects to match jsonrpc spec

bitwiseguy:
Previously the jsonrpc server was not returning json responses that adhered to the jsonrpc spec. This PR fixes that, which is a first step to providing our jsonrpc clients (go and TS) a reliable way to detect errors when they make requests.

### DIFF
--- a/rpc/server.go
+++ b/rpc/server.go
@@ -78,12 +78,15 @@ func (rs *RpcServer) registerHandlers() (err error) {
 		rs.logger.Trace().Msgf("Rpc server received request: %+v", string(requestData))
 
 		if !json.Valid(requestData) {
-			return marshalResponse(types.ParseError, rs.logger)
+			rs.logger.Error().Msg("request is not valid json")
+			errRes := types.JsonRpcResponse{Jsonrpc: "2.0", ErrorObj: types.ParseError}
+			return marshalResponse(errRes, rs.logger)
 		}
 
-		jsonrpcReq := validateJsonrpcRequest(requestData, rs.logger)
-		if jsonrpcReq.Error != nil {
-			return jsonrpcReq.Error
+		jsonrpcReq, errRes := validateJsonrpcRequest(requestData, rs.logger)
+		if errRes != nil {
+			rs.logger.Error().Msg("could not validate jsonrpc request")
+			return errRes
 		}
 
 		switch serde.RequestMethod(jsonrpcReq.Method) {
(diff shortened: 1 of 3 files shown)


types/serde.go
```diff
@@ -17,11 +17,24 @@ func (d *Destination) UnmarshalText(text []byte) error {
...
 }
 
+type JsonRpcRequest struct {
+	Jsonrpc string      `json:"jsonrpc"`
+	Id      uint64      `json:"id"`
+	Method  string      `json:"code"`
+	Params  interface{} `json:"params"`
+}
+
+type JsonRpcResponse struct {
```
bitwiseguy, line 27: There is also a `JsonRpcResponse` type in `rpc/serde/jsonrpc.go`. However, that one is closely tied to the specific jsonrpc method through the use of generics so I have not been able to get that one to work within the server for some of the errors (e.g. `MethodNotFoundError`). It does feel redundant to have this object defined in two places though so not sure what is the best option

geoknee: Reading https://www.jsonrpc.org/specification#response_object, it seems to me that we would like to have two types:
1. `JsonRpcSuccessResponse` (which has a result but no error)
2. `JsonRpcErrorResponse` (which has an error but no result).

types/serde.go
```diff
@@ -17,11 +17,24 @@ func (d *Destination) UnmarshalText(text []byte) error {
 	return nil
 }
 
+type JsonRpcRequest struct {
```
geoknee, line 20: Slightly out of scope for this PR: I'm not super clear why we want to split the types across this file and `rps/serde/jsonrpc.go`. 

Do you think we could tidy this up (even in a follow-up PR?).

rpc/server.go
```diff
@@ -210,55 +212,40 @@ func marshalResponse(response any, log *zerolog.Logger) []byte {
...
 	}
 
-	_, ok = request["method"].(string)
+	sMethod, ok := request["method"].(string)
 	if !ok {
-		requestError := types.InvalidRequestError
-		requestError.Id = vr.Id
-		vr.Error = marshalResponse(requestError, logger)
-		return vr
+		errRes := types.JsonRpcResponse{Jsonrpc: "2.0", Id: vr.Id, ErrorObj: types.InvalidRequestError}
```
geoknee, line 243: Nit: We have a pattern elsewhere of using a helper function to construct these objects, rather than doing it literally. Doing that here would reduce the need to have `"Jsonrpc:"2.0"` everywhere, and make future upgrades easier to do? Don't feel strongly though

geoknee (approved):
Thanks @bitwiseguy. It's good for us to stay on spec!

I think there's probably a little bit further we could go in terms of organising types -- see comments.

types/serde.go
geoknee: Let's discuss at standup. It might be nice encapsulation for the node to not know anything about json rpc?

PR #1407, opened 2023-07-06
Switch websockets from github.com/nhooyr to github.com/gorilla/websocket

kerzhner:
The primary reason for the change is that the websocket rpc client sometimes hangs on close. It is unclear whether this is due to an incorrect usage of the websocket library or an issue with the library itself.

Gorilla websockets is a popular library that is ergonomic to use. Unfortunately, the github repository is archived. However, it doesn't seem like these types of libraries see many updates anyways. Gorilla websockets was last updated in 7/2022 (https://github.com/gorilla/websocket/commits/master) and nhooyr websockets was last updated on 12/2022.

### DIFF
--- a/rpc/transport/ws/server.go
+++ b/rpc/transport/ws/server.go
@@ -12,9 +12,10 @@ import (
 	"sync"
 	"time"
 
+	"github.com/gorilla/websocket"
+
 	"github.com/statechannels/go-nitro/internal/safesync"
 	"github.com/statechannels/go-nitro/rand"
-	"nhooyr.io/websocket"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


rpc/transport/ws/server.go
```diff
@@ -141,54 +142,43 @@ func (wsc *serverWebSocketTransport) request(w http.ResponseWriter, r *http.Requ
...
-	// A client closes a connection by sending a message over the websocket
-	// This code converts the socket `Read` call to a channel.
-	// Ideally, all signals would be managed in the select statement below. But there is no channel API for the websocket read.
 	closeChan := make(chan error)
-	wsc.wg.Add(1)
-	go wsc.listenForClose(r.Context(), c, closeChan)
+
+	closeHandler := c.CloseHandler()
+	c.SetCloseHandler(func(code int, text string) error {
+		closeChan <- nil
```
lalexgap, line 165: Does it matter that we're signalling on the `closeChan` before `closeHandler` completes?

lalexgap (approved):
🦍

rpc/transport/ws/server.go
kerzhner: I can't think of any issues with that. Are you thinking about anything specific?

lalexgap: I think that means we'll signal the `closeChan` (and return from `subscribe`) before the `upgrader` has finished executing `closeHandler()`? I'm not quite sure what the `c.CloseHandler` does so it might be ok, but in general I think we should only signal something is done after all execution is complete. This [ADR](https://github.com/statechannels/go-nitro/blob/7b8a110a9a0a302a254a2412510bcdc05fd9c482/.adr/0016-closing-go-routines.md#L83) might be relevant.

PR #1408, opened 2023-07-06
Refactor channel create script to use RPC clients

kerzhner:
This PR refactors the create channel script so that the script does not leverage go-nitro internal. The script is used in the entrypoint for go-nitro docker container.

The PR also cleans up functions defined in internal package:
- functions that are only used by a single package are moved to that package.
- copy pasted logging is refactored into the internal package.


docker/entrypoint.sh
```diff
@@ -1,5 +1,14 @@
 #!/usr/bin/env bash
 
 set -e
+set -m # enable job control
 
-go run ./internal/nitro-network/main.go
+go run ./scripts/start-rpc-servers.go &
```
lalexgap, line 6: Is there any way we could support running this container against an external chain, by accepting parameters somehow? In our boost integration we do this with [env vars](https://github.com/statechannels/boost/blob/2d532f0271f90888ff1ff6b997cad47de27b434f/docker/devnet/booster-http/entrypoint.sh#L14) but there may be a more "dockery" way of doing that.

That would let us use this dockerized nitro with a local filecoin devnet.

scripts/start-rpc-servers.go
lalexgap, line 1: Maybe we should move this into a `start-rpc-servers` folder and rename it to `main.go` to keep consistent with create-channels?

scripts/start-rpc-servers.go
```diff
@@ -186,3 +185,37 @@ func newColorWriter(c color, w io.Writer) colorWriter {
 		color:  c,
 	}
 }
+
+// StopCommands stops the given executing commands
```
lalexgap, line 189:
```suggestion
// stopCommands stops the given executing commands
```

scripts/start-rpc-servers.go
```diff
@@ -186,3 +185,37 @@ func newColorWriter(c color, w io.Writer) colorWriter {
...
+			panic(err)
+		}
+		err = cmd.Process.Kill()
+		if err != nil {
+			panic(err)
+		}
+	}
+}
+
+// GenerateTempStoreFolder generates a temporary folder for storing store data and a cleanup function to clean up the folder
```
lalexgap, line 204:
```suggestion
// generateTempStoreFolder generates a temporary folder for storing store data and a cleanup function to clean up the folder
```

lalexgap (approved):
👍

internal/logging/logging.go
```diff
@@ -27,3 +30,20 @@ func configureZeroLogger() {
 		return file + ":" + strconv.Itoa(line)
 	}
 }
+
```
lalexgap, line 33:
```suggestion
NewLogWriter returns a writer for the given logDir and logFile
If the log file already exists it will be removed and a fresh file will be created
```

scripts/create-channels/main.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	return nil
+}
+
+// main creates channels between 3 participants: alice, irene, and bob
+// A ledger channel is opened between alice and irene
+// A ledger channel is opened between irene and bob
+// A virtual channel is opened between alice and bob
+func main() {
```
lalexgap, line 124: Probably beyond the scope of this PR but we may want to consider accepts parameters via CLI flags in this script ([similar to our TS create-channel script](https://github.com/statechannels/go-nitro/blob/8c3578087e9c5fcb9110beb48dbce814d9df9f7c/packages/nitro-rpc-client/scripts/client-runner.ts#L82)) . For example, it could be nice to run this script without creating ledger channels if we wanted to add some more virtual channels for whatever reason.

docker/entrypoint.sh
kerzhner: That's a good point. Created https://github.com/statechannels/go-nitro/issues/1467.

scripts/start-rpc-servers.go
kerzhner: Good point. Created https://github.com/statechannels/go-nitro/issues/1468.

scripts/create-channels/main.go
kerzhner: Good point. Created https://github.com/statechannels/go-nitro/issues/1469.

PR #1350, opened 2023-06-12
Towards batch deposits

NiloCK:
Toward #1333.

PR adds test scaffolding (gas results) and a draft implementation of batch deposits.

From `gasResults.json`:

```json
  "directlyFundAChannelWithETHFirst": {
    "satp": 47238
  },
  "directlyFundAChannelWithETHSecond": {
    "satp": 30150
  },
  "batchFundChannelsWithETHFirst": {
    "satp": {
      "1": 57011,
      "5": 192285,
      "10": 361401,
      "20": 699620
    }
  },
  "batchFundChannelsWithETHSecond": {
    "satp": {
      "1": 34305,
      "5": 78755,
      "10": 134341,
      "20": 245500
    }
  },
```

Glance math:
- batch funding 10 channels with first deposits comes in at about 75% of cost of existing deposit route.
- batch funding 10 channels with second deposits (or topups) comes in at about 40% of existing deposit route.

---

#### Code quality

- [x] I have written clear commit messages
- [ ] I have performed a self-review of my own code
- [x] This change does not have an unduly wide scope
- [x] I have separated logic changes from refactor changes (formatting, renames, etc.)
- [ ] I have commented my code wherever necessary (can be 0)
  - [ ] todo: doccomments for new deposit_batch function 
- [ ] I have added tests that prove my fix is effective or that my feature works, if necessary
  - Partial. Existing gas benchmark tests demonstrate basic functionality. To what extent does a function like this need a specific test suite aimed at it? It's a pretty dumb wrapper to an existing, [well tested](https://github.com/statechannels/go-nitro/blob/main/nitro-protocol/test/contracts/MultiAssetHolder/deposit.test.ts) function.
- [ ] ~I have added comprehensive [godoc](https://go.dev/blog/godoc) comments~

#### Project management

- [x] I have assigned myself to this PR
- [ ] I have linked the appropriate github issue
- [ ] I have assigned this PR to the appropriate GitHub project
- [ ] I have assigned this PR to the appropriate GitHub Milestone

PR #1409, opened 2023-07-07
Remove `amountDeposited` from `Deposited` event

NiloCK:
Toward #1410, related to #1350. This PR is a further gas optimizations along the `deposit` route.

On each deposit, the nitro adjudicator emits a `Deposited` event. This lets channel participants (or nosy outsiders) know that the on-chain state of things has changed. After hearing of a deposit, it may now be safe to either:
- make your own deposit
- sign a post-fund and begin off-chain transactions

In order to determine either of these, the _current balance_ of the channel is sufficient. The size of the current deposit is not relevant to the safety of the protocol. This is essentially an "unused variable" not caught by any linter because event logs cross language / runtime barriers.

Removal drops deposit costs by ~1%.

- [x] Is it security related?

It's feasible that there are security implications to this, but I don't think there are.

Submitted as draft because follow-up work on deposit tests (shape of emitted data) not done.

---

#### Code quality

#### Project management


NiloCK:
> I'm on board with this.
> 
> I think in principle consumers could catch the event, and then query the transaction history to figure out the amount deposited anyway?

Yes - anyone listening for these events has enough information to reproduce the entire history.

geoknee (approved):

PR #1413, opened 2023-07-07
Log to a log file instead of stdout for rpc tests

kerzhner:
When tests are run via github actions, stdout is polluted to the point where it is difficult to diagnose what test fails. This PR shifts the log lines to an artifact log file.

### DIFF
--- a/node_test/rpc_test.go
+++ b/node_test/rpc_test.go
@@ -478,7 +478,7 @@ func checkNotifications[T channelInfo](t *testing.T, logger zerolog.Logger, clie
 			// Log both to the test log file and to stdout
 			failMsg := fmt.Sprintf("%s timed out waiting for notification(s): \n%v", client, incompleteRequired(acceptableNotifications))
 			logger.Error().Msgf(failMsg)
-			t.Fatalf("%s timed out waiting for notification(s): \n%v", client, incompleteRequired(acceptableNotifications))
+			t.Fatalf(failMsg)
 		}
 	}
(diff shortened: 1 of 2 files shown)


node_test/rpc_test.go
```diff
@@ -78,10 +86,11 @@ func executeNRpcTest(t *testing.T, connectionType transport.TransportType, n int
 	// Setup
 	//////////////////////
 
-	t.Logf("Starting test with %d clients", n)
 	logFile := fmt.Sprintf("test_%d_rpc_clients_over_%s.log", n, connectionType)
 	logDestination := newLogWriter(logFile)
 	defer logDestination.Close()
+	logger := testLogger(logDestination)
```
NiloCK, line 92: I don't have super strong feelings about this, but here the engines and the test output write to the same log file, which reproduces the maybe-too-noisy outputs currently being ignored by readers of failed test runs.

Existing `t.Logf` calls should not be contributors to any logging performance issues, since they print to stdout only if the test fails.

kerzhner: That's a good point. As an fyi, we run unit tests [with -v flag in CI](https://github.com/statechannels/go-nitro/blob/e7d8a0585f363be1a1fa2730b3186ab81bd75ba1/.github/workflows/go.yml#L34). So if we leave `logf` in the codebase, we should turn off the verbose flag.

I also don't feel strongly about this change. The benefit of the proposed change is that all logs are in a single file. The added log lines will not grow the log file significantly.

PR #1419, opened 2023-07-10
Add support to start-rpc-servers to run against a local filecoin devnet

lalexgap:
**NOTE:** This supersedes https://github.com/statechannels/go-nitro/pull/1378

This updates the `start-rpc-servers-script` to accept some chain related flags, letting us target something that isn't anvil running locally.

The script has converted to a `urFav` CLI app and now takes the following flags:
- Whether to start a local anvil instance or not.
- The chain url to use. The default is the anvil url
- The chain auth token to use. The default is an empty string
- The chain pk of the deployer to use. The default is the `FUNDED_TEST_PK`

Since we're using sane defaults running `start-rpc-servers` with no arguments results in the previous behaviour, we spin up a local instance of anvil and use that.

To run against filecoin it would look something like this:
```
go run ./scripts/start-rpc-servers.go -a=false -cu=ws://127.0.0.1:1234/rpc/v1 -ct=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJBbGxvdyI6WyJyZWFkIiwid3JpdGUiLCJzaWduIiwiYWRtaW4iXX0.iKUOdDak2Gwhyh_neCukKS-nsJjy41XdUTTM7vZRKTU
```

I've also removed the logic setting gas limits in `ConnectToChain`. This was causing us to use legacy transactions which is not supported by filecoin. Without these gas limits things to deploy properly to both anvil and a local filecoin devnet so I don't think they were needed.


node/engine/chainservice/eth_chainservice.go
```diff
@@ -60,7 +60,8 @@ const MAX_QUERY_BLOCK_RANGE = 2000
 // We do this to avoid https://github.com/ethereum/go-ethereum/issues/23845
 // We use 2.5 minutes as the default filter timeout is 5 minutes.
 // See https://github.com/ethereum/go-ethereum/blob/e14164d516600e9ac66f9060892e078f5c076229/eth/filters/filter_system.go#L43
-const RESUB_INTERVAL = 2*time.Minute + 30*time.Second
+// This has been reduced to 15 seconds to support local devnets with much shorter timeouts.
+const RESUB_INTERVAL = 15 * time.Second
```
lalexgap, line 64: The fact that we keep running into problems with long running subscriptions makes me wonder if we should consider some kind of polling option, where we periodically poll the chain for new events.

I believe the filecoin devnet has an io timeout of 30 seconds, so we reduce to 15 seconds here to stay well under that.

geoknee (approved):

PR #1422, opened 2023-07-11
Allow store data folder to be specified as a CLI arg

lalexgap:
This updates go-nitro to accept a `durablestorefolder` that it will use for the durable store. Our scripting has been updated to take advantage of that by using a random temp folder for the store and ensuring it gets removed when the script is stopped.

### DIFF
--- a/internal/rpc/rpc.go
+++ b/internal/rpc/rpc.go
@@ -3,6 +3,7 @@ package rpc
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/rs/zerolog"
@@ -21,7 +22,7 @@ import (
 )
 
 func InitChainServiceAndRunRpcServer(pkString string, chainOpts chain.ChainOpts,
-	useDurableStore bool, useNats bool, msgPort int, rpcPort int,
+	useDurableStore bool, durableStoreFolder string, useNats bool, msgPort int, rpcPort int,
 ) (*rpc.RpcServer, *node.Node, *p2pms.P2PMessageService, error) {
 	if pkString == "" {
 		panic("pk must be set")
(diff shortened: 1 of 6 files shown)


scripts/start-rpc-servers.go
```diff
@@ -102,6 +102,13 @@ func main() {
 				running = append(running, anvilCmd)
 			}
 
+			dataFolder := "./data/nitro-service"
```
geoknee, line 105: Where is the corresponding line where we specify that data should be stored in this folder? Is it baked in to `go-nitro`?

lalexgap: It's part of the internal [RunRPCServer](https://github.com/statechannels/go-nitro/blob/8d7a7c11e6b3e46a2861d0415511995cdfa6bdee/internal/rpc/rpc.go#L64)

lalexgap: Probably. We should update our `go-nitro` executable to take in a flag to specify the store location (with some sane default). That way the start-rpc-servers script can decide on the data folder and make sure it's clean beforehand.

lalexgap: It slightly expands the scope of the PR but I've done this in https://github.com/statechannels/go-nitro/pull/1422/commits/96264f2e4aa2e958765b7c3614cccbff28fc443d go-nitro now accepts a `durablestorefolder` argument and we generate a random temp folder when we run our start rpc server script.

main.go
```diff
@@ -112,6 +113,13 @@ func main() {
 			Category:    "Connectivity:",
 			Destination: &rpcPort,
 		}),
+		altsrc.NewStringFlag(&cli.StringFlag{
+			Name:        DURABLE_STORE_FOLDER,
+			Usage:       "Specifies the folder for the durable store data storage.",
+			Category:    "Storage:",
+			Destination: &durableStoreFolder,
+			Value:       "./data/nitro-store",
```
lalexgap, line 121: No idea what we want here as a default

geoknee (approved):

scripts/start-rpc-servers.go
geoknee: Thanks! Perhaps update the PR title before merging 👍

internal/rpc/rpc.go
```diff
@@ -60,8 +61,9 @@ func RunRpcServer(pk []byte, chainService chainservice.ChainService,
 		Logger()
 
 	if useDurableStore {
-		logger.Info().Msg("Initialising durable store...")
-		dataFolder := fmt.Sprintf("./data/nitro-service/%s", me.String())
+		dataFolder := filepath.Join(durableStoreFolder, me.String())
```
lalexgap, line 64: Since we always use our address as a subfolder in `durableStoreFolder` we don't have to worry about making sure each go-nitro instance is given a unique `durableStoreFolder`. Multiple go-nitro instances can use the same `durableStoreFolder`, as long as they're each using a different SC address.

Issue #1411, opened 2023-07-07
Support arbitrary DirectlyFunded channels 


geoknee:
At this time, `go-nitro` supports:
* directly funded ledger channels and 
* virtually funded payment channels.

The broader nitro-protocol supports arbitrary 3rd party applications https://docs.statechannels.org/protocol-tutorial/0020-execution-rules/#application-rules. 

Proposed Changes:
1. API changes
    1.1. Add a `CreateDirectChannel` API method to the `node`. It should be very similar to https://github.com/statechannels/go-nitro/blob/217e0b49f03a72b4e37884b1bf6b62b12632c08d/node/node.go#L230, but allow for the application definition and application data to be passed in (and not queried from the store or implicitly set zero).
    1.2. Add an `UpdateChannel` method. It plays a similar role to `Pay`, but it should construct a new state from the supplied app data and then sign, store and send it.
    1.3. Add a `CloseChannel` method. It plays the same role as `CloseLedgerChannel`, but it shouldn't need to worry about handling `ConsensusChannels`.
2. (Optional, or can be done last or separately) Repeat step 1a,b for the rpc clients (Go and TS) and rpc server.
3. Rename existing `directfund` protocol to `directfundledgerchannel`. Write new protocol `directfund` which doesn't make any assumptions about the channel type. Or otherwise remove those assumptions from `directfund`. We do **not** want to create a `ConsensusChannel` at any point during this channel's lifetime.
4.  Allow for our wire protocol to send messages which are not bound to an objective. They just contain signed states and cause receivers to update their channels with the signed states.
5.  Look for opportunities to DRY out what we ended up with (e.g. have `CreateLedgerChannel` call into `CreateDirectChannel`).
6.  Integration test to show this all works. 

## Out of scope
Arbitrary Virtually Funded Channels

geoknee:
For 3, if we take the path of modifying `directfund` to not assume a consensusapp, we should ensure that. 
* `node.CreateLedgerChannel` still results in a `ConsensusChannel` in the store
* `node.CreateDirectlyFundedChannel({appDef:consensusApp})` does **not** result in a `ConsensusChannel` in the store.

PR #1425, opened 2023-07-11
feat: add support of directly funded channels

MaxMoskalenko:
Towards #1411 
Implement support of directly funded channels 

Yellow Network needs an node API to create, update and close direct channels (more specs here https://github.com/statechannels/go-nitro/issues/1411)

The summary may omit some of these details if this PR fixes a single ticket that includes these details.

- [ ] Is it security related?
- [X] Is it a significant process change?
- [ ] Is it a significant change to architecture or design?

---

#### Code quality

#### Project management

### DIFF
--- a/node/engine/engine.go
+++ b/node/engine/engine.go
@@ -9,9 +9,12 @@ import (
 	"math/big"
 	"sync"
 
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/rs/zerolog"
 	"github.com/statechannels/go-nitro/channel"
 	"github.com/statechannels/go-nitro/channel/consensus_channel"
+	"github.com/statechannels/go-nitro/channel/state"
+	"github.com/statechannels/go-nitro/directs"
 	"github.com/statechannels/go-nitro/internal/logging"
 	"github.com/statechannels/go-nitro/node/engine/chainservice"
 	"github.com/statechannels/go-nitro/node/engine/messageservice"
@@ -58,8 +61,9 @@ type Engine struct {
 	// inbound go channels
 
 	// From API
-	ObjectiveRequestsFromAPI chan protocols.ObjectiveRequest
-	PaymentRequestsFromAPI   chan PaymentRequest
+	ObjectiveRequestsFromAPI     chan protocols.ObjectiveRequest
+	PaymentRequestsFromAPI       chan PaymentRequest
+	UpdateChannelRequestsFromAPI chan UpdateChannelRequest
 
 	fromChain  <-chan chainservice.Event
 	fromMsg    <-chan protocols.Message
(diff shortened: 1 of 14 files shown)


protocols/directfund/directfund.go
```diff
@@ -188,6 +188,12 @@ func (dfo *Objective) GetStatus() protocols.ObjectiveStatus {
 	return dfo.Status
 }
 
+// IsLedgerChannel returns true if the objective is direct funding a ledger channel.
```
kerzhner, line 191: I think we'll need deeper changes to the direct fund protocol. At the moment, the protocol is geared toward funding a ledger channel. A ledger channel is a specific type of channel used to fund other (virtual) channels. We should probably refactor the direct fund protocol to remove the assumption that the protocol funds a ledger channel.

A ledger channel is also known as a consensus channel in the golang codebase.

mod: yes, but I can't pinpoint what makes a ledger channel different than a direct channel.
there is no flag at the state level.

MaxMoskalenko: I removed this changes for now. But we definitely need to refactor ledger channels, so they will be a subtype of a direct funded channels

geoknee: > yes, but I can't pinpoint what makes a ledger channel different than a direct channel. there is no flag at the state level.

That is true, and we "get away with it" because we only have two types of channel currently (namely, Ledger Channels and Virtual Payment Channels).

 Currently the flow from API to store is:

`CreateLedgerChannel`[ API method ](https://github.com/statechannels/go-nitro/blob/2b0bc7c5998abe021970704f448d030844a05928/node/node.go#L244) -> [`directfund.Objective`](https://github.com/statechannels/go-nitro/blob/2b0bc7c5998abe021970704f448d030844a05928/protocols/directfund/directfund.go#L41) -> [`Channel`](https://github.com/statechannels/go-nitro/blob/8940b55c99f0aeb84bec9b486e5b53fa31bab03c/node/engine/store/store.go#L32). 

When the directfund objective completes, the channel is converted into a ConsensusChannel. When the ledger channel is closed, it is converted back into a regular `Channel`.

We may need to attach a new `type` property to a `Channel`. It can be an enum `Application, Ledger`.
When we call `CreateLedgerChannel` we get  `type=Ledger`, when we call `CreateDirectlyFundedApplicationChannel` we get `Application`.

node/engine/engine.go
```diff
@@ -547,6 +595,64 @@ func (e *Engine) handlePaymentRequest(request PaymentRequest) (EngineEvent, erro
...
+		}
+
+		newState := latestSignedState.Clone().State()
+		newState.AppData = request.AppData
+		newState.TurnNum = newState.TurnNum + 1
+
+		s = &newState
+	}
+
+	ss, err := c.SignAndAddState(*s, e.store.GetChannelSecretKey())
```
geoknee, line 624: Here we are countersigning a state if the appdata matched what came over the API.

We need to be careful: the state may have a different [**Outcome**](https://docs.statechannels.org/protocol-tutorial/0030-outcomes/). I suggest we need to add the outcome to the relevant data structures and check against that. 

A more elegant solution would be to use the [**state hash**](https://docs.statechannels.org/protocol-tutorial/0010-states-channels/#state-commitments) to check if it is "the same state".

PR #1430, opened 2023-07-11
Update DeployContracts to wait for transactions to be mined

lalexgap:
When running against a local filecoin devnet I would occasionally see a `panic: no contract code at given address` if I started using `go-nitro` right after the deployment script succeed.

As @geoknee pointed out this is because we're not waiting for the transaction to be mined/confirmed(I'm not sure which is the correct term these days?) This PR updates the `DeployContract` function to call [WaitMined](https://github.com/ethereum/go-ethereum/blob/afe344bcf31bfb477a6e1ad5b862a70fc5c1a22b/accounts/abi/bind/util.go#L32) to wait for the deployment transaction to be mined.  I've also done a slight refactor with generics to DRY things out a bit.

Locally I'm no longer able to reproduce `panic: no contract code at given address` against a local filecoin devnet with this change. I've also verified this works when run against `anvil`

### DIFF
--- a/internal/chain/chain.go
+++ b/internal/chain/chain.go
@@ -7,9 +7,11 @@ import (
 	"os/exec"
 	"time"
 
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	b "github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
 	ethTypes "github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/ethclient"
 	"github.com/statechannels/go-nitro/node/engine/chainservice"
 	NitroAdjudicator "github.com/statechannels/go-nitro/node/engine/chainservice/adjudicator"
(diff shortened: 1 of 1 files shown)


geoknee (approved):
Is there anywhere else in the codebase where we want to make these changes?

https://github.com/statechannels/go-nitro/blob/1898ce618aa06fd7a721621f73d7a870675dee90/node/engine/chainservice/adjudicator/challenge_test.go#L202

lalexgap:
> Is there anywhere else in the codebase where we want to make these changes?
> 
> https://github.com/statechannels/go-nitro/blob/1898ce618aa06fd7a721621f73d7a870675dee90/node/engine/chainservice/adjudicator/challenge_test.go#L202

I think that code may be ok, since with the simulated backend we're able to call [sim.Commit](https://github.com/statechannels/go-nitro/blob/1898ce618aa06fd7a721621f73d7a870675dee90/node/engine/chainservice/adjudicator/challenge_test.go#L229) which forces the contracts to be deployed.

PR #1431, opened 2023-07-11
Retry nats request once

kerzhner:
We sometimes see "Test panicked: received nill data for request ... with error nats: no responders available for request" in [CI tests](https://github.com/statechannels/go-nitro/actions/runs/5522572367/jobs/10072347583).

It seems that an RPC client is making a nats request before the server has subscribed for the nats topic. This PR adds a client retry after 500 milliseconds to give the server time to register.

### DIFF
--- a/rpc/transport/nats/client.go
+++ b/rpc/transport/nats/client.go
@@ -23,11 +23,29 @@ func NewNatsTransportAsClient(url string) (*natsTransportClient, error) {
 }
 
 func (c *natsTransportClient) Request(data []byte) ([]byte, error) {
-	msg, err := c.nc.Request(nitroRequestTopic+apiVersionPath, data, 10*time.Second)
-	if msg == nil {
-		return nil, fmt.Errorf("received nill data for request %v with error %w", string(data), err)
+	requestFn := func(data []byte) (*nats.Msg, error) {
+		return c.nc.Request(nitroRequestTopic+apiVersionPath, data, 10*time.Second)
 	}
-	return msg.Data, err
+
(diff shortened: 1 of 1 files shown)


rpc/transport/nats/client.go
```diff
@@ -23,11 +23,29 @@ func NewNatsTransportAsClient(url string) (*natsTransportClient, error) {
...
+	var msg *nats.Msg
+	for i := 0; i < numTries; i++ {
+		msg, err = requestFn(data)
+		if msg != nil && err == nil {
+			return msg.Data, nil
+		}
+
+		// Skip sleep after the last try
+		lastTry := i == numTries-1
+		if lastTry {
```
lalexgap, line 41: nit but this could be in-lined in the if statement 
```suggestion
		if lastTry := i == numTries-1; lastTry {
```

rpc/transport/nats/client.go
```diff
@@ -23,11 +23,29 @@ func NewNatsTransportAsClient(url string) (*natsTransportClient, error) {
...
+			break
+		}
+
+		time.Sleep(500 * time.Millisecond)
```
lalexgap, line 45: Could we put this in a constant somewhere?

lalexgap (approved):

rpc/transport/nats/client.go
kerzhner: Since we only use this value once, to me, it is more readable for this to be inline. Happy to reconsider if you feel strongly about creating a constant.

PR #1433, opened 2023-07-11
Add rpc test cleanup log lines

kerzhner:
This should help diagnose deadlocks like https://github.com/statechannels/go-nitro/actions/runs/5525366833/jobs/10078937740


NiloCK (approved):
Good addition. From memory, the PKs here are just sequential, so they make for good identifiers on the outputs.

PR #1435, opened 2023-07-12
Add buffer to notification channel

kerzhner:
We are currently seeing deadlocks where a goroutine is blocked on sending to the notification channel.

```
goroutine 13030 [chan send]:
github.com/statechannels/go-nitro/rpc/transport/ws.(*clientWebSocketTransport).readMessages(0xc0035d8900)
	/home/runner/work/go-nitro/go-nitro/rpc/transport/ws/client.go:88 +0xa5
created by github.com/statechannels/go-nitro/rpc/transport/ws.NewWebSocketTransportAsClient
	/home/runner/work/go-nitro/go-nitro/rpc/transport/ws/client.go:41 +0x1b7
```

https://github.com/statechannels/go-nitro/actions/runs/5526236952/jobs/10080745506

### DIFF
--- a/rpc/transport/ws/client.go
+++ b/rpc/transport/ws/client.go
@@ -22,7 +22,7 @@ type clientWebSocketTransport struct {
 // NewWebSocketTransportAsClient creates a websocket connection that can be used to send requests and listen for notifications
 func NewWebSocketTransportAsClient(url string) (*clientWebSocketTransport, error) {
 	wsc := &clientWebSocketTransport{}
-	wsc.notificationChan = make(chan []byte)
+	wsc.notificationChan = make(chan []byte, 10)
 	wsc.url = url
 
 	subscribeUrl, err := urlUtil.JoinPath("ws://", url, "subscribe")


NiloCK (approved):
I think this change is OK and may allow certain previously locking runs of go-nitro to gracefully continue. But: 

Is there a real case for thread synchronization here?

Do we actually want the rpc client's transport layer to not forward notifications to the RPC client's main thread if the client has 10 unprocessed notifications? (Or one, as was the case previously).

I think no. We need a concurrency safe data structure here, but not the concurrency orchestration properties of a gochan. It's just message posting - not flow control. Something like a `safesync.Map` feels less likely to get into any resource locking trouble.

lalexgap:
> I think this change is OK and may allow certain previously locking runs of go-nitro to gracefully continue. But:
> 
> Is there a real case for thread synchronization here?
> 
> Do we actually want the rpc client's transport layer to not forward notifications to the RPC client's main thread if the client has 10 unprocessed notifications? (Or one, as was the case previously).
> 
> I think no. We need a concurrency safe data structure here, but not the concurrency orchestration properties of a gochan. It's just message posting - not flow control. Something like a `safesync.Map` feels less likely to get into any resource locking trouble.

**Note:** I'm going to use the term **chan** to refer to a go channel. I'll use the term producer go-routine for the go-routine that writes to a chan, in this case [`readMessages`](https://github.com/statechannels/go-nitro/blob/733f735589099f0b9cc0190c7ff3c2190528189b/rpc/transport/ws/client.go#L88). I'll use the term consumer go-routine for the go-routine that reads from a chan, in this case [`subscribeToNotifications`](https://github.com/statechannels/go-nitro/blob/733f735589099f0b9cc0190c7ff3c2190528189b/rpc/client.go#L77)

I think we're using a buffered chan here (and in other places in our codebase) to serve two purposes:
1. a mechanism to communicate between go-routines. By pushing data into `notificationChan` we activate the consuming go-routine to wake up and send out those notifications.
2.  As a queue to queue data that should be processed by the consuming go-routine. By writing to the buffered `notificationChan` we queue up work for the consuming go-routine to send out.

I think chans solve 1) pretty well since they include behavior for waking go-routines when there is new data on the chan. If we wanted to avoid chans and use `safeSync.Map` the consuming go-routine would have to periodically poll the `safeSync.Map` to see if there's new data. 

Maybe chans aren't the best fit for 2), as we generally use small buffers? Perhaps we should write (or find) a queue component. We can use non-buffered (or slightly buffered) chans to communicate between go-routines. The consuming go-routines can use the queue component to queue up work, ensuring that they read from the chan quickly, avoiding blocking the producing go-routine.

kerzhner:
One thing to note is that a transport (websocket transport in this case) is set up before the rpc client. So a transport can start receiving messages before a client is ready to process those messages. If that happens with a notification channel of buffer size 1, the client will not be waiting via a [websocket read](https://github.com/statechannels/go-nitro/blob/4d1ae760a41123cc291e2e913576894ec95248bb/rpc/transport/ws/client.go#L81) when data arrives from the server. It's hard to say if that can be a source of issues without digging into the websocket implementation.

PR #1436, opened 2023-07-12
Move nitro-gui code into go-nitro to create monorepo

bitwiseguy:
## PR summary
This PR moves code from the `nitro-gui` repo into this repo. 

* `./packages` contains the source code from `nitro-gui` repo
  * `site` package was renamed to `packages/nitro-gui`
* `./packages` contains documentation files (.png, README.md) from `nitro-gui` repo
  * README.md text updated to fit new repo and file structure
* Config files related to yarn are now at the root of this `go-nitro` repo
* Commits related to `nitro-gui/packages` now have new commit hashes and are interspersed with previous `go-nitro` commits

Related issue: https://github.com/statechannels/go-nitro/issues/1442

## Steps taken to preserve git history of `nitro-gui/packages` source code
 ```
### Filter so we only have the directory we want to move
cd nitro-gui
git filter-branch --prune-empty --subdirectory-filter packages

### Create new standalone "packages" repo that only contains history for that dir
cd ..
mkdir packages
cd packages
git init
git pull ../nitro-gui

### Pull packages into go-nitro and merge git history with go-nitro
cd ../go-nitro
git remote add packages ../packages
git fetch packages
git merge packages/master --allow-unrelated-histories

### Individual package dirs now exist at go-nitro root
### Need to create new packages/ parent directory and move packages there
mkdir packages
git mv boost-demo packages
git mv nitro-rpc-client packages
git mv site packages/nitro-gui
```
## Steps taken to move config/docs files from `nitro-gui` to `go-nitro`
I originally tried to preserve the git history of all individual files that were moved from `nitro-gui` repo root into the `go-nitro` root using the following commands:
```
cd nitro-gui
git log --pretty=email --patch-with-stat --reverse -- ./.editorconfig ./.eslintrc.js ./.gitattributes ./.gitignore ./.yarnrc.yml ./architecture.png ./package.json ./README.md ./yarn.lock > ~/patches

cd ../go-nitro
git am < ~/patches
```
This caused some errors initially. After working around them (by skipping some of the patches) it appeared the commits were added as a new collection and were not linked to other commits related to the `packages` source code. I decided to just copy and paste these files into the `go-nitro` repo (some are now at repo root and some are in `packages/` directory) to hopefully reduce any confusion and ensure we got the latest version of all of those files. This means the git history for these files is NOT preserved.

## Future work
1. Ensure `nitro-gui` files added to `go-nitro/.github/workflows` are hooked up and working correctly
2. Consider moving go-nitro/nitro-protocol into go-nitro/submodules directory. Can use `git mv` to preserve git history
    * Move from npm to yarn package manager to align with other submodules
    * Fix any references to go-nitro/nitro-protocol package from other areas of go-nitro code/docs
3. Serve nitro-gui from inside go-nitro node (instead of running it externally)

### CODE REVIEW

File: .github/workflows/yarn-build.yml [hunk 1]
```diff
@@ -0,0 +1,22 @@
+name: Yarn build and lint
+
+on:
+  pull_request:
+    paths: ['**/*.ts', '**/*.tsx', '**/*.js', '**/*.jsx', '**/*.json', '**/*.md']
```
lalexgap, line 5: We might want to consider scoping this to the `packages` folder so the yarn workflow doesn't run for changes outside of the `packages` folder. I think right now a change to any `readme.md` would trigger this workflow to run.

package.json
```diff
@@ -0,0 +1,26 @@
...
+  "version": "0.0.1",
+  "private": true,
+  "description": "Typescript packages for go-nitro statechannel network",
+  "homepage": "https://statechannels.org",
+  "bugs": {
+    "url": "https://github.com/statechannels/go-nitro/issues"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/statechannels/go-nitro.git"
```
lalexgap, line 12: Beyond the scope of this PR but I'm wondering if we should consider renaming the `go-nitro` repository to something language agnostic?

lalexgap (approved):
Overall this looks good to me! 

As a sanity check I used the monorepo to run the boost demo which uses the `nitro-rpc-client` and `boost-demo` typescript packages as well as the start-rpc-servers golang script, and it all worked 👍

.github/workflows/yarn-build.yml
bitwiseguy: Good call. I will push a commit to add the `packages/` prefix to each of those paths

geoknee:
Thanks @bitwiseguy 🥷 

I

> * Consider moving go-nitro/nitro-protocol into go-nitro/submodules directory. Can use `git mv` to preserve git history
>   
>   * Move from npm to yarn package manager to align with other submodules
>   * Fix any references to go-nitro/nitro-protocol package from other areas of go-nitro code/docs
> * Serve nitro-gui from inside go-nitro node (instead of running it externally)


In favour of all of this, but maybe let's wait until @NiloCK has wrapped up some of his work in `nitro-protocol` to avoid conflicts?

PR #1438, opened 2023-07-13
streamline Typescript RPC Client Test GHA

geoknee:
This script pulls from github twice, which it no longer needs to do (and shouldn't, to avoid problems). 

Also, it won't currently run unless the ts client itself changes, so we won't catch breaking changes made on the server side. I fixed that.

### DIFF
--- a/.github/workflows/ts-rpc-test.yml
+++ b/.github/workflows/ts-rpc-test.yml
@@ -2,7 +2,6 @@ name: TypeScript RPC Client Test
 
 on:
   pull_request:
-    paths: ["packages/nitro-rpc-client", ".github/workflows/ts-rpc-test.yml"]
   workflow_dispatch:
 jobs:
   run-rpc-test:
--- a/.github/workflows/ts-rpc-test.yml
+++ b/.github/workflows/ts-rpc-test.yml
@@ -20,9 +20,6 @@ jobs:
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@v1
 
-      - name: Tidy
-        run: go mod tidy
-
       - name: Run go-nitro RPC servers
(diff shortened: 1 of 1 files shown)


.github/workflows/ts-rpc-test.yml
```diff
@@ -2,37 +2,25 @@ name: TypeScript RPC Client Test
 
 on:
   pull_request:
-    paths: ['packages/nitro-rpc-client', '.github/workflows/ts-rpc-test.yml']
```
geoknee, line 5: This might be a little controversial, since this action takes 4 or 5 minutes to run. It will also force us to fix the TS client before merging any breaking changes to the go-nitro.

.github/workflows/ts-rpc-test.yml
```diff
@@ -2,37 +2,25 @@ name: TypeScript RPC Client Test
 
 on:
   pull_request:
```
NiloCK, line 4:
```suggestion
  pull_request:
    paths: ['packages/nitro-rpc-client', 'rpc', '.github/workflows/ts-rpc-test.yml']
```

Is this reasonable? We probably shouldn't be counting on rpc client tests to catch emergent errors inside of go nitro, but this set of paths probably catches any breaking change to the rpc server.

NiloCK (approved):
:+1:

.github/workflows/ts-rpc-test.yml
geoknee: I had considered that... but I concluded that we could still break the contract between go-nitro and the TS client without touching `rpc`. So an integration test would break but we wouldn't detect it immediately because it would not run. I'm not 100&% sure though.

PR #1449, opened 2023-07-13
Embed GUI into go-nitro binary 

### LINKED ISSUE #1442: Consider embedding the UI into go-nitro
We want to have the UI served from go-nitro itself. 

We can use the `boost` repo as an example, because they do this with their UI.

geoknee:
Closes #1442

Couple of things I would like to refine:

~1. Note removal of "CONNECT" button and text field. This will make local development slightly more painful, since we won't have hot reloading working well.~

~Open to suggestions on how to make this switchable / preserve that developer experience.~
We can recover the experience from before (mostly) by setting an environment variable. This will work fine for development against a single mock go nitro node.

~2. Choice of port number. For now, I used something hacky (`rpcPort - 100`). We could have either: 
    i.   `80` or `8080`. (tricky when running multiple processes on the same host)
    ii. `rpcPort` (need to coordinate with the actual rpc server which also binds to this port)~

I updated the PR to use the existing http server and port.

3. nitro-gui is now a hard dependency of go-nitro. This means we need to build nitro-gui before go-nitro. I haven't updated our CI script yet, since I wanted to get feedback on the idea of only optionally embedding the UI to allow for quicker CI checks / development. 

Screenshot 2023-07-13 at 12 02 28


geoknee:
Standup notes: be nice to have the "connect to remote node" button available optionally, to aid development and allow for more usage patterns.

geoknee:
> Standup notes: be nice to have the "connect to remote node" button available optionally, to aid development and allow for more usage patterns.

[7dad705](https://github.com/statechannels/go-nitro/pull/1449/commits/7dad705f7d28fee3cea099288818f5ece0610435)

geoknee:
> We would like a `--headless` flag or env var to enable building go-nitro without the yarn dependency for integration tests etc.

I decided to punt this to future work because it seems nontrivial [e721c00](https://github.com/statechannels/go-nitro/pull/1449/commits/e721c000ef86ee0924c0633ceac2148527c74483) . 

The compiler will complain if the embedded directory is empty. So the only solution I can think of is to check in a file to that directory, so that _something_ is always embedded. We can then do a runtime check on what that thing looks like, and if it doesn't look like the full site we can panic unless a headless flag is set. That's my best idea for now.

.github/workflows/go.yml
```diff
@@ -15,7 +15,7 @@ jobs:
           go-version: "1.20.0"
 
       - name: Install and build GUI dependency with yarn
-        run: yarn && yarn
+        run: yarn && yarn build
```
geoknee, line 18: Ideally I would rewrite history to `fixup` this commit (merging it with the previous one).

internal/nitro-network/main.go
```diff
@@ -79,7 +80,8 @@ func InitializeNitroNetwork() error {
 		dataFolder, cleanup := utils.GenerateTempStoreFolder()
 		defer cleanup()
 
-		server, node, msgService, err := interRpc.InitChainServiceAndRunRpcServer(nodeOpts.Pk, chainOpts, nodeOpts.UseDurableStore, dataFolder, false, nodeOpts.MsgPort, nodeOpts.RpcPort)
+		var emptyFS fs.FS // Serve a dummy static site (we do not currently support embedding the GUI here)
+		server, node, msgService, err := interRpc.InitChainServiceAndRunRpcServer(nodeOpts.Pk, chainOpts, nodeOpts.UseDurableStore, dataFolder, false, nodeOpts.MsgPort, nodeOpts.RpcPort, emptyFS)
```
geoknee, line 84: This isn't ideal but difficult to work around. We can't embed anything further up the directory tree than this file.

Options:
*place `packages` underneath `internal`
* our build script needs to copy the static site somewhere where it can be embedded.

main.go
```diff
@@ -13,6 +15,9 @@ import (
 	"github.com/urfave/cli/v2/altsrc"
 )
 
+//go:embed packages/nitro-gui/dist/*
```
lalexgap, line 18: What does this comment do? Could we add a comment above explaining?

main.go
```diff
@@ -136,7 +141,12 @@ func main() {
 				CaAddress:      common.HexToAddress(caAddress),
 			}
 
-			rpcServer, _, _, err := rpc.InitChainServiceAndRunRpcServer(pkString, chainOpts, useDurableStore, durableStoreFolder, useNats, msgPort, rpcPort)
+			ss, err := fs.Sub(fs.FS(staticSite), "packages/nitro-gui/dist")
```
lalexgap, line 144: Can this be controlled by a flag, so serving the nitro UI is optional?

rpc/transport/ws/server.go
```diff
@@ -46,6 +47,10 @@ func NewWebSocketTransportAsServer(port string) (*serverWebSocketTransport, erro
 
 	serveMux.HandleFunc(apiVersionPath, wsc.request)
 	serveMux.HandleFunc(path.Join(apiVersionPath, "subscribe"), wsc.subscribe)
+
+	// Serve static site
+	serveMux.Handle("/", http.FileServer(http.FS(staticSite)))
```
lalexgap, line 52: What happens if `staticSite` is empty/nil? Should we check so we don't bother trying to serve a static site if it doesn't exist?

main.go
lalexgap, line 144: Is it important that we only do this in `main`? Or could we call `fs.Sub` right in the websocket server, avoiding passing `ss` around.

rpc/transport/ws/server.go
```diff
@@ -34,7 +35,7 @@ type serverWebSocketTransport struct {
 }
 
 // NewWebSocketTransportAsServer starts an http server that accepts websocket connections
-func NewWebSocketTransportAsServer(port string) (*serverWebSocketTransport, error) {
+func NewWebSocketTransportAsServer(port string, staticSite fs.FS) (*serverWebSocketTransport, error) {
```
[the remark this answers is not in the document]
geoknee: Right now we have `http/ws` and `nats` transports. I'm not sure how we could serve a website (i.e. html, css, js) over nats? I think we would probably need to switch back to http at some point in the link.

Side remark: I have my doubts about the utility of maintaining the `nats` transport option going forward. We should check with @mod, though.

rpc/transport/ws/server.go
geoknee: Good shout: [d130638](https://github.com/statechannels/go-nitro/pull/1449/commits/d130638575713581a4f844289f2111d277a86a21)

main.go
geoknee: The `main` entrypoint is where the files are embedded. The `.go` file with the `embed` directive (L19( cannot search up the directory tree, so without rearranging the repo we can't `embed` in (say) the `rpc` package or websocket server. That's the thing that necessitates our passing _something_ down to the rpc server from this file. 

Given that we have to embed or "import" the static site in `main.go`, we should do whatever wrangling we need to prepare it to be passed to the rpc server in `main.go`, too. That's what this line is doing, essentially trimming down the unnecessary path prefix.



I've tidied this up a bit [f2c7ed8](https://github.com/statechannels/go-nitro/pull/1449/commits/f2c7ed87b903d84a112241b13258e5860a415f33)

geoknee: Yes [f2c7ed8](https://github.com/statechannels/go-nitro/pull/1449/commits/f2c7ed87b903d84a112241b13258e5860a415f33)

I don't think we can make embedding the UI optional, but we can make serving it optional. This is because the compiler will complain if we try to embed a directory with no matching files in it. 

If we really care about that, we can come up with a solution:
* Have a `build_fake_UI` script which just writes an empty or trivial html file into the `packages/nitro-gui/dist` directory. We can coordinate builds with a makefile and have a `make go-nitro/headless` command which calls that first.
* Or have a new folder which we check the trivial `index.html` into, and have the production gui build overwrite that folder with the full UI.

geoknee: Digging around in the `boost` repo, I believe they are just requiring their react frontend to be built before boost can be built (and living with it 🤷 ).

geoknee: Here's a sketch at something we could do https://github.com/statechannels/go-nitro/pull/1452

geoknee:
From standup conversation: 

We would prefer to have the GUI served on its own port, independent of the RPC and its transport. 
This allows us to still connect over NATS (in principle) and still have the GUI available.

main.go
```diff
@@ -13,6 +15,9 @@ import (
 	"github.com/urfave/cli/v2/altsrc"
 )
 
+//go:embed packages/nitro-gui/dist/*
+var staticSiteRaw embed.FS
```
lalexgap, line 19: I wonder if we can use build tags to optionally enable this. It looks like there is support for conditionally including files based on the build tag: https://www.digitalocean.com/community/tutorials/customizing-go-binaries-with-build-tags

geoknee: Nice find. Let me play around with that.

geoknee: I think it'll work. POC here https://github.com/statechannels/go-nitro/pull/1452/commits/0548dfd8646f6b5ca8c447c3e38d0ca1dcc510ad#diff-2cc94b7e0195048022d70d3f2deac587c20786b5d1effb31e3947a76c233ee2d

geoknee:
In favour of #1457

geoknee:
I think we still want some of the changes from this branch, we can potentially cherry pick them onto the other PR.

PR #1451, opened 2023-07-13
Fix: initialize rpc websocket logger

kerzhner:
The logger for websocket rpc transport is never initialized. This PR fixes this issues as well as removes the unused `NewHttpRpcClient` function.

### DIFF
--- a/rpc/client.go
+++ b/rpc/client.go
@@ -61,24 +61,12 @@ func NewRpcClient(logger zerolog.Logger, trans transport.Requester) (*RpcClient,
 
 // NewHttpRpcClient creates a new RpcClient using an http transport
 func NewHttpRpcClient(rpcServerUrl string) (*RpcClient, error) {
-	transport, err := ws.NewWebSocketTransportAsClient(rpcServerUrl)
+	logger := zerolog.New(os.Stdout)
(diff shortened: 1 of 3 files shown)


rpc/client.go
```diff
@@ -59,28 +57,6 @@ func NewRpcClient(logger zerolog.Logger, trans transport.Requester) (*RpcClient,
 	return c, nil
 }
 
-// NewHttpRpcClient creates a new RpcClient using an http transport
-func NewHttpRpcClient(rpcServerUrl string) (*RpcClient, error) {
```
lalexgap, line 63: We currently use this constructor in our[ boost integration](https://github.com/statechannels/boost/blob/f7d1b8ec9c98e3e4b72ed5c4ff032c279fcf9438/cmd/booster-http/server.go#L90). It is nice to have one constructor that does everything instead of manually constructing the transport

lalexgap (approved):
🌋

rpc/client.go
kerzhner: Good call!

kerzhner: `NewHttpRpcClient` method is refactored in https://github.com/statechannels/go-nitro/pull/1451/commits/957558706a840b891fc8457458d4a2b4789e2fd1

PR #1457, opened 2023-07-14
Embed the Nitro UI in go-nitro

lalexgap:
This allows the RPC-server to embed `nitro-gui` and serve the UI on the same port as the RPC server(if using the http/ws transport.

```
{"level":"info","time":1689370840077,"message":"Initializing websocket RPC transport..."}
Nitro as a Service listening on port 4008
Using transport http server to host UI at http://localhost:4008/
```

# Build Tags
There are two new files that expose a `HostNitroUI` function in the `main `package. We use a build tag to ensure that only one of those files is included and built. We do this by using the build directives
-  `//go:build embed_ui`  which includes the file if the build tag `embed_ui` is defined
- `//go:build !embed_ui` which includes the file if the build tag `embed_ui` is **not** defined

Since these build tag checks are evaluated at compile time the nitro ui data is only embedded in the executable if the build tag is set. This means we can build a slimmer executable.

To set the build tag it can be passed in with the `-tags` argument to either go build or go run
```bash
go run -tags embed-ui  .
go  build -tags embed-ui ./...
```

If the `embed-ui` build tag is not included we won't embed the UI data and we won't try to serve the UI over http. This means the default behavior is the same as before, there is no hard dependency on the nitro UI.

**Note** vs code will ignore(no autocomplete, error highlighting) `embed-ui.go` due to the `//go:build embed_ui` build flag. If you want to edit the file it can be helpful to remove `//go:build embed_ui` temporarily to enable autocomplete, etc..

# Hosting the UI on the RPC port

I've changed the RPC transport to use the `DefaultServeMux`. By using the `DefaultServeMux` we can easily register additional handlers by calling `http.Handler` from other sections of code.

We make use of this in our new `hostNitroUI` function, by calling `http.Handle` to register a handler against the `DefaultServeMux` and relying on the transport server to handle running the http server. We only try to start a http server to serve the UI if there is nothing running on port.

One downside of switching the rpc transport to use the `DefaultServeMux` is that it doesn't work in with our RPC tests. Our rpc tests has multiple RPC servers being spun up in the same go process, meaning the `DefaultServeMux` is shared between multiple RPC servers. That means we'll have multiple calls to the `DefaultServeMux` to register a handler for the same path. 

~To address this I've added a `isTest` flag that results in using a new `HttpServerMux` for the ws transport to avoid this causing the tests to fail.~ To address this I've updated `rpc_test` to reset the `DefaultServeMux` before trying to create the RPC server.

# Start RPC Servers script
The start rpc support script has been updated to optionally pass `-tags embed_ui` into the `go run` commands to run the nitro executable.  It accepts a new `hostui` flag to control that behavior:
```
go run  ./scripts/start-rpc-servers.go -hostui=true
```

# Nats support

If we're using nats we can't use the same RPC port to listen for http traffic, since nats is listening on that address.  Instead I've added a workaround of starting the UI server on port `rpcPort+500`.

# Questions
- ~Right now the build tag controls 1) whether the static file UI data is included in the build/run, 2) Whether we attempt to serve the UI or not. Should 2) be controlled by a CLI flag?~ See https://github.com/statechannels/go-nitro/pull/1457#issuecomment-1636409456


geoknee:
> This means we can build a slimmer executable

This is true but I think not the main benefit. The GUI is actually pretty lightweight (500 KB compared with 50 MB binary). It's more that building the UI from scratch is comparatively slow (I'm estimating 30s vs 5s).

I think it isn't actually _that_ slow, probably due to using yarn 3. But I do like us having the option to leave the UI out with the build tags route.

geoknee:
> * Should 2) be controlled by a CLI flag?

I think not. No point in trying to serve it if it isn't embedded, and not much point in not serving it if it is embedded 🤷

lalexgap:
> > This means we can build a slimmer executable
> 
> This is true but I think not the main benefit. The GUI is actually pretty lightweight (500 KB compared with 50 MB binary). It's more that building the UI from scratch is comparatively slow (I'm estimating 30s vs 5s).
> 
> The difference in build times is not huge, probably due to using yarn 3. But I do like us having the option to leave the UI out with the build tags route.

Yup fair point! I think one big benefit is that we can easily build/run `go-nitro` without having to rely on `nitro-gui` being built or even there.  For example  in this PR running `go build ./...` works even if the `nitro-gui` folder doesn't exist. The only time we expect the `nitro-gui` folder to be there is when we're explicitly setting a build tag to include it.

geoknee (approved):
Nice job!

I have a couple of reservations:

1. Using essentially a global variable `http.DefaultServeMux` to enable the main entrypoint to optionally share the http server with the deeper rpc server package. This requires some slightly hacky workarounds in our tests, as well as special casing the nats transport and needing to check if the server is already listening. Those things don't feel super scalable or easy to maintain.

I don't have a better suggestion just now -- although I am wondering whether doing some "props drilling" kind of thing might be nicer. As in, pass the static site down to the rpc server for it to serve.

2. It's nice to preserve existing behaviour by defaulting to not serving the UI. On the other hand, it might be good to default to serving it so that it gets used more. Of course we want to balance this against the small amount of developer friction it would introduce (developer needs to build the UI).  I wonder if we want to include a makefile entry for building go nitro which has the UI ON by default?  Might be a good compromise.

geoknee:
> Nice job!
> 
> I have a couple of reservations:
> 
> 1. Using essentially a global variable `http.DefaultServeMux` to enable the main entrypoint to optionally share the http server with the deeper rpc server package. This requires some slightly hacky workarounds in our tests, as well as special casing the nats transport and needing to check if the server is already listening. Those things don't feel super scalable or easy to maintain.
> 
> I don't have a better suggestion just now -- although I am wondering whether doing some "props drilling" kind of thing might be nicer. As in, pass the static site down to the rpc server for it to serve.
> 
> 2. It's nice to preserve existing behaviour by defaulting to not serving the UI. On the other hand, it might be good to default to serving it so that it gets used more. Of course we want to balance this against the small amount of developer friction it would introduce (developer needs to build the UI).  I wonder if we want to include a makefile entry for building go nitro which has the UI ON by default?  Might be a good compromise.

I've addressed 1 through some follow-one work merged into this PR.

PR #1459, opened 2023-07-17
Tweaks following "embed UI" changes

geoknee:
* Clean up frontend
* Enhance documentation

PR #1461, opened 2023-07-17
Use a distinct `guiPort`

geoknee:
Following today's standup discussion, this was felt to be a reasonable compromise. One more port, but quite a lot of code saved and no need to use a global variable. 

Stacked on #1459

### DIFF
--- a/main.go
+++ b/main.go
@@ -27,10 +27,11 @@ func main() {
 		CA_ADDRESS           = "caaddress"
 		MSG_PORT             = "msgport"
 		RPC_PORT             = "rpcport"
+		GUI_PORT             = "guiport"
 		DURABLE_STORE_FOLDER = "durablestorefolder"
 	)
 	var pkString, chainUrl, chainAuthToken, naAddress, vpaAddress, caAddress, chainPk, durableStoreFolder string
-	var msgPort, rpcPort int
+	var msgPort, rpcPort, guiPort int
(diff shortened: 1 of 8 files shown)


main.go
```diff
@@ -113,6 +114,13 @@ func main() {
 			Category:    "Connectivity:",
 			Destination: &rpcPort,
 		}),
+		altsrc.NewIntFlag(&cli.IntFlag{
+			Name:        GUI_PORT,
+			Usage:       "Specifies the tcp port for the Nitro Connect GUI.",
+			Value:       5005,
+			Category:    "Connectivity:",
```
NiloCK, line 121: Minor side issue:

Do these categories ever get used?
- yes: extract into constants
- no: maybe scrub, replace with empty string or `"categoriesnotused"` const

main.go
```diff
@@ -141,14 +149,7 @@ func main() {
 				return err
 			}
 
-			uiPort := uint(rpcPort)
-			// If we're using nats we can't re-use the RPC port for the UI, so we choose another
```
NiloCK, line 145: definitely good to get rid of port juggling logic as a function of other config options.

NiloCK (approved):
:+1:

main.go
geoknee: They do get used in the "help" function I believe, to organise the output. Good call on constants.

PR #1462, opened 2023-07-17
Kademlia-dht peer discovery

bitwiseguy:
## Motivation
Add a peer discovery mechanism suitable for production setting.

## Justification
`go-libp2p` is trusted by [many notable projects](https://github.com/libp2p/go-libp2p#notable-users), and specifically for peer discovery by IPFS, Lotus (filecoin node), Polkadot, and others. Ethereum 2.0 uses a custom discovery protocol named `discv5` that is based on the Kademlia DHT design, but is tailored for sharded subnets. `go-libp2p` is modular and provides multiple options for configuring different parts of a p2p network. The following peer discovery options are provided by the library:
1. **mDNS**: only useful in local networks. Nodes broadcast topics they’re interested in
2.  **rendezvous**: introduces some centralization. `rendezvous` nodes act as hubs to provide peerIds of other nodes
3.  **kademlia-dht**: uses a distributed hash table to find nodes and data on the network by using a routing table that organizes peers based on how similar their keys are (referred to as their "distance" from each other). In order to enter the network, a new node MUST be given some "bootstrap" node IDs so that it can start discovering other peers. Periodic calls are made to `FindPeers` to update a node's view of the dht.

How data is stored in the dht:
  * Keys: the hashes of peer IDs. The peer ID is a unique identifier assigned to each peer/node in the network, which is typically the hash of its public key.
  * Values: represent the peer records or peer info, which typically includes the peer's ID and the list of multiaddresses (IP address, port + other routing info) at which the peer can be reached

Difference between `Peerstore` and `RoutingTable` (dht)
  * `Peerstore`: stores metadata for all peers. When node needs to send a message to a peer, looks up peer's address in Peerstore to establish a connection
  * `RoutingTable` (dht): manages connections to other peers and stores "distance" info

## Summary of changes
1. Added two new cli/config options: `bootpeers` and `usemdns`
2. `P2PMessageService` can now be run in kademlia-dht mode (mDNS option still available for now) 
3. Added kademlia-dht integration test

## ToDo
- [x] Simplify data stored in `ms.peers` map
- [x] Automate test suite to run against network in dht mode
- [ ] Identify root of `WARN` logs printed by libp2p
- [ ] Tune bucket size of kademlia-dht
- [ ] Tune frequency of `FindPeers` calls
- [ ] Research meaning/ramifications of `libp2p.New` options
- [ ] Test on non-local network
- [ ] Decide if mDNS should remain as an option
- [ ] Add DDoS protection (rate limiting or peer reputation score?)
- [ ] Handle peer disconnected?
- [ ] Set limit on number of connected peers? (Lotus uses [these settings](https://github.com/filecoin-project/lotus/blob/2ddb52e1039963840b8e20b52634e33642575059/node/builder.go#L230))


internal/utils/utils.go
```diff
@@ -0,0 +1,54 @@
...
+	for sNum, s := range services {
+		for i := 0; i < len(services)-1; i++ {
+			peerInfo := <-s.PeerInfoReceived()
+			fmt.Printf("Service num: %d, peer num: %d, peerInfo: %v\n", sNum, i, peerInfo)
+		}
+	}
+}
+
+// StopCommands stops the given executing commands
+func StopCommands(cmds ...*exec.Cmd) {
```
lalexgap, line 23: I think there may be a rebase/merge issue here? I think a bunch of these util functions were refactored into [start-rpc-servers](https://github.com/statechannels/go-nitro/blob/a90cdb294aaa1902f93db15ed612d393253cee03/cmd/start-rpc-servers/main.go#L205)

internal/nitro-network/main.go
lalexgap, line 1: Is this a rebase/merge issue? This looks like a previous implementation of starting up a nitro network that may have gotten picked up by accident.

node/engine/messageservice/p2p-message-service/service.go
```diff
@@ -245,25 +313,52 @@ func (s *P2PMessageService) PeerInfoReceived() <-chan basicPeerInfo {
...
-		}
-		multi, _ := multiaddr.NewMultiaddr(fmt.Sprintf("/ip4/%s/tcp/%d/p2p/%s", p.IpAddress, p.Port, p.Id))
-		ms.p2pHost.Peerstore().AddAddr(p.Id, multi, peerstore.PermanentAddrTTL)
+func (ms *P2PMessageService) discoverPeers(ctx context.Context, topic string) {
+	routingDiscovery := routing.NewRoutingDiscovery(ms.dht)
+	_, err := routingDiscovery.Advertise(ctx, topic) // fires every 3 hours by default
+	ms.checkError(err)
 
-		ms.peers.Store(p.Address.String(), basicPeerInfo{p.Id, p.Address})
+	ticker := time.NewTicker(time.Second * 10)
```
lalexgap, line 337: Might be nice to throw this is a constant and add a comment about it

main.go
```diff
@@ -62,6 +65,13 @@ func main() {
 			Value:       false,
 			Destination: &useDurableStore,
 		}),
+		altsrc.NewBoolFlag(&cli.BoolFlag{
```
lalexgap, line 68: Completely beyond the scope of this PR (and not really that important) but it might be nice to support some kind of `EnumFlag`.  That would allow for a more descriptive flag, IE:`-peerdiscovery=dht` vs `-peerdiscovery=mdns`

It looks like `urFav` doesn't have an enum flag  out of the box but it [looks pretty easy to implement your own](https://github.com/urfave/cli/issues/602#issuecomment-416000118)

node/engine/messageservice/p2p-message-service/service.go
```diff
@@ -83,71 +93,123 @@ func NewMessageService(ip string, port int, me types.Address, pk []byte, useMdns
 	options := []libp2p.Option{
 		libp2p.Identity(messageKey),
 		libp2p.ListenAddrStrings(fmt.Sprintf("/ip4/%s/tcp/%d", ip, port)),
-		libp2p.Transport(tcp.NewTCPTransport),
-		libp2p.NoSecurity,
-		libp2p.DefaultMuxers,
+		// libp2p.Transport(tcp.NewTCPTransport),
```
lalexgap, line 96: Should these still be commented out?

node/engine/messageservice/p2p-message-service/service.go
```diff
@@ -83,71 +93,123 @@ func NewMessageService(ip string, port int, me types.Address, pk []byte, useMdns
...
+	ms.dht = kademliaDHT
 
-		ms.mdns = mdns
+	// Print out my own peerInfo
+	peerInfo := peer.AddrInfo{
+		ID:    ms.p2pHost.ID(),
+		Addrs: ms.p2pHost.Addrs(),
+	}
+	addrs, err := peer.AddrInfoToP2pAddrs(&peerInfo)
+	ms.checkError(err)
```
lalexgap, line 142: Completely outside the scope of this PR but we should probably consider some kind of proper error handling for the p2p message service. Right now we just `panic`. I'll create an issue for this.

internal/utils/utils.go
bitwiseguy: I'm not really sure what happened. I tried to pull in the latest changes using these commands:
```
git checkout main
git pull origin main
git checkout kademlia-dht
git merge main
```
Then I addressed all of the merge conflicts and ran these commands.
```
git add -A
git commit
```
Any idea what might have gone wrong in that sequence?

bitwiseguy: Maybe it was caused because the entire files were deleted on `main`. I removed the files `internal/utils` and `internal/nitro-network` again on this branch so it should be good now

node/engine/messageservice/p2p-message-service/service.go
bitwiseguy: Yes that's a good call. All messaging errors should not cause `panic`

internal/utils/utils.go
lalexgap: I generally `rebase` instead of `merge`, IE:`git fetch origin && git rebase origin/main` but what you did makes sense to me.

>Then I addressed all of the merge conflicts and ran these commands.

I suspect this is where things might have gone wrong. When resolving merge conflicts it's easy to include a file that you've modified that's been removed in `main`.

node/engine/messageservice/p2p-message-service/service.go
bitwiseguy: I forgot I had done that. I just added them back in

internal/testactors/actors.go
```diff
@@ -35,22 +36,25 @@ var Alice Actor = Actor{
...
 
-// Brian has the address 0xB2B22ec3889d11f2ddb1A1Db11e80D20EF367c01
-var Brian Actor = Actor{
-	common.Hex2Bytes("0aca28ba64679f63d71e671ab4dbb32aaa212d4789988e6ca47da47601c18fe2"),
+// Ivan has the address 0xA8d2D06aCE9c7FFc24Ee785C2695678aeCDfd7A0
+// peerId: 16Uiu2HAm1hgN2MkrhGen8JPrBBYyXACbZtfqJmraN53XiHYCeoFi
+var Ivan Actor = Actor{
+	common.Hex2Bytes("1ea91a2724b40fb8fed6a3648d49e9431996c09744fb841b718377fb0700f3e7"),
 	2,
 	"brian",
```
lalexgap, line 52:
```suggestion
	"brian",
	"ivan",
```

main.go
```diff
@@ -153,7 +170,11 @@ func main() {
 				CaAddress:      common.HexToAddress(caAddress),
 			}
 
-			rpcServer, _, _, err := rpc.InitChainServiceAndRunRpcServer(pkString, chainOpts, useDurableStore, durableStoreFolder, useNats, msgPort, rpcPort)
+			var peerSlice []string
+			if bootPeers != "" {
```
lalexgap, line 174: Nitpick but I think `string.Splits` [returns an empty slice if given an empty string](https://pkg.go.dev/strings#Split), so I don't think we need the if statement

node/engine/messageservice/p2p-message-service/service.go
```diff
@@ -93,61 +104,113 @@ func NewMessageService(ip string, port int, me types.Address, pk []byte, useMdns
...
-		err = mdns.Start()
-		ms.checkError(err)
+	// Add bootstrap peers
+	err = ms.dht.Bootstrap(ctx)
+	ms.checkError(err)
+	ms.addBootPeers(bootPeers)
+
+	expectedPeers := len(bootPeers)
+	if expectedPeers > 0 {
+		ticker := time.NewTicker(time.Second)
```
lalexgap, line 154: Could throw this is a const

lalexgap (approved):
🚀

main.go
bitwiseguy: I tried the following...
```
peerSlice := strings.Split(bootPeers, ",")
fmt.Printf("num boot peers: %d\n", len(peerSlice))
fmt.Printf("boot peers: %v\n", bootPeers)
```
... but I get the following output when I don't have any bootPeers.
```
num boot peers: 1
boot peers: 
```
I will leave for now and we can clean it up later if theres a better way

PR #1463, opened 2023-07-18
tweak layout of user-flows section

geoknee:


NiloCK (approved):
more indexing, less scroll requirements is generally good

But I don't see why this page doesn't currently include a ToC on the right hand side. See the difference between 

https://docs.statechannels.org/protocol-tutorial/0010-states-channels/

and

https://docs.statechannels.org/user-flows/0010-user-flows/

The states & channels doc contains clickable anchor links to different sections, which would probably also work well enough here.

geoknee:
> more indexing, less scroll requirements is generally good
> 
> But I don't see why this page doesn't currently include a ToC on the right hand side. See the difference between
> 
> https://docs.statechannels.org/protocol-tutorial/0010-states-channels/
> 
> and
> 
> https://docs.statechannels.org/user-flows/0010-user-flows/
> 
> The states & channels doc contains clickable anchor links to different sections, which would probably also work well enough here.

I think the framework is just quite opinionated about the heading levels. It might be that I should have stuck with one file and bumped the heading level down to `##`. Unfortunately I can't get the site to build on my machine so I am testing via netlify...

geoknee:
I think I prefer this actually https://github.com/statechannels/go-nitro/pull/1466

PR #1465, opened 2023-07-18
upgrade `mockchain.holdings` to a `safesync.Map`

### LINKED ISSUE #1398: Concurrent Map read and map write in MockChain
Recent CI run against a comments-only change set: https://github.com/statechannels/go-nitro/actions/runs/5423048162/jobs/9860465199

[https://github.com/statechannels/go-nitro/actions/runs/5423048162/jobs/9860465199#step:7:2159](Line 2159): fatal error: concurrent map read and map write

The output is the less-useful style with a dump of all running goroutines. I do not know whether the immediately following goroutine (Mockchain.SubmitTransaction) is the source of the error. (After inspection: seems likely. "Running" goroutine, line number, error type make sense together.)

geoknee:
Closes #1398

### DIFF
--- a/node/engine/chainservice/mock_chain.go
+++ b/node/engine/chainservice/mock_chain.go
@@ -14,7 +14,7 @@ import (
 type MockChain struct {
 	blockNum uint64
 	// holdings tracks funds for each channel.
-	holdings map[types.Destination]types.Funds
+	holdings safesync.Map[types.Funds]
 	// out maps addresses to an Event channel. Given that MockChainServices only subscribe
 	// (and never unsubscribe) to events, this can be converted to a list.
 	out safesync.Map[chan Event]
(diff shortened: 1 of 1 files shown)


node/engine/chainservice/mock_chain.go
```diff
@@ -33,21 +33,24 @@ func NewMockChain() *MockChain {
 // unlike an ethereum blockchain, MockChain accepts go-nitro protocols.ChainTransaction
 func (mc *MockChain) SubmitTransaction(tx protocols.ChainTransaction) error {
 	mc.blockNum++
+	channelIdString := tx.ChannelId().String()
+	h, _ := mc.holdings.Load(channelIdString)
```
NiloCK, line 37: Swallowed `ok`s or `error`s raise my eyebrows and slow my reading of code. A quick disclaimer may be a good idea when the swallowing isn't immediately & obviously correct.

Maybe

```suggestion
	h, _ := mc.holdings.Load(channelIdString) // ignore `ok` because the returned zero-value is what we want
```

In this case it is correct because:
- safesync.Map returns a zero-valued `T` (types.Funds in this case) when its lookup fails
- the zero-value of `types.Funds` is well behaved with its `Add` function

NiloCK (approved):
:+1:

node/engine/chainservice/mock_chain.go
geoknee: Great point.

Issue #1468, opened 2023-07-18
Unified directory structure for scripts


kerzhner:
Maybe we should move this into a `start-rpc-servers` folder and rename it to `main.go` to keep consistent with create-channels?

_Originally posted by @lalexgap in https://github.com/statechannels/go-nitro/pull/1408#discussion_r1264230327_

PR #1470, opened 2023-07-18
Move start-rpc-servers.go into own directory

kerzhner:
Fixes https://github.com/statechannels/go-nitro/issues/1468.

Go has the following restrictions:
- Only one package is allowed per directory
- An executable package must be named `main` and contain a main() function.

`scripts` directory is now a collection of subdirectories, each of which is a script package.


geoknee (approved):
Wondering if we should use `cmd` instead of `scripts` as the dir name? I see this in go-ethereum and boost etc.

kerzhner:
> Wondering if we should use `cmd` instead of `scripts` as the dir name? I see this in go-ethereum and boost etc.

That's a good idea. Done in 5d1c169b.

PR #1473, opened 2023-07-19
Implement an interface for the RPC client

lalexgap:
I was originally prompted to look into this when using `mockGen` to generate a nitro rpc client mock for a test in the boost repo ([based on how they've mocked out the boost API](https://github.com/statechannels/boost/blob/3cbc1ef7f21d928c879686e859ea1be78a0da928/cmd/booster-http/server.go#L33)). However the tool expects there to be an interface to use when generating the mock.

Ignoring mocking I think exposing an interface and hiding the implementation away might be nice. A RPC client is how we expect most people to interact with a nitro node so having a well defined interface for that probably makes sense.

In this PR:
- I've created an interface for the RPC client called `RpcClientApi`
- I've made the RPC client implementation struct private by renaming it to `rpcClient`

### DIFF
--- a/rpc/client.go
+++ b/rpc/client.go
@@ -26,8 +26,62 @@ import (
 	"github.com/statechannels/go-nitro/types"
 )
 
-// RpcClient is a client for making nitro rpc calls
-type RpcClient struct {
+// RpcClientApi provides various functions to make RPC API calls to a nitro RPC server
+type RpcClientApi interface {
+	// Address returns the address of the nitro node
+	Address() (common.Address, error)
+
+	// CreateVoucher creates a voucher for the given channelId and amount and returns it.
+	// It is the responsibility of the caller to send the voucher to the payee.
+	CreateVoucher(chId types.Destination, amount uint64) (payments.Voucher, error)
+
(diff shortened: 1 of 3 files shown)


rpc/client.go
```diff
@@ -26,8 +26,62 @@ import (
 	"github.com/statechannels/go-nitro/types"
 )
 
-// RpcClient is a client for making nitro rpc calls
-type RpcClient struct {
+// RpcClientApi provides various functions to make RPC API calls to a nitro RPC server
+type RpcClientApi interface {
```
lalexgap, line 30: I originally named the interface `RpcClient` but I was worried `RpcClient` vs `rpcClient` could be confused.

rpc/client.go
```diff
@@ -26,8 +26,62 @@ import (
 	"github.com/statechannels/go-nitro/types"
 )
 
-// RpcClient is a client for making nitro rpc calls
-type RpcClient struct {
+// RpcClientApi provides various functions to make RPC API calls to a nitro RPC server
```
lalexgap, line 29: This could live in a separate types or interfaces file if we wanted.

geoknee (approved):

PR #1483, opened 2023-07-20
Implement a basic reverse payment proxy

lalexgap:
This PR adds a simple reverse payment proxy. The proxy is a HTTP server that checks incoming requests for a voucher, and if the voucher is sufficient, makes a request to some destination server and serves back the response. If the voucher is insufficient or malformed than the `ReversePaymentProxy` serves an error message and a 402 status code.

We're relying on a [`http.ReverseProxy`](https://pkg.go.dev/net/http/httputil#ReverseProxy) to do most of the heavy lifting. It handles the work of rerouting the request to the destination server and serving the response. All we need to do is add a simple payment check before delegating responsibility to the `http.ReverseProxy`. This also means that our `ReversePaymentProxy` handles forwarding requests for **any** url. For example a request to the `ReversePaymentProxy` at `http://localhost:5511/some/path?SomeParams&ChannelId=0x123..` will cause a request to the destination server at `http://localhost:6622/some/path?SomeParams` (assuming your voucher was sufficient).

Right now the payment model is still pretty simple. We expect some hard-coded amount for every request.

I've added a basic test that spins up a destination server and makes sure that the `ReversePaymentProxy` both enforces payments and properly serves content when the payment is provided. I've also been able to manually test it by running a simple HTTP server, the `start-reverse-payment-proxy` script, and the nitro-rpc-client to create vouchers.

### DIFF
--- a/cmd/start-reverse-payment-proxy/main.go
+++ b/cmd/start-reverse-payment-proxy/main.go
@@ -14,7 +14,7 @@ import (
 
 const (
 	NITRO_ENDPOINT  = "nitroendpoint"
-	PORT            = "port"
+	PROXY_ADDRESS   = "proxyaddress"
 	DESTINATION_URL = "destinationurl"
 )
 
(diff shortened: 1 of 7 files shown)


cmd/start-reverse-payment-proxy/main.go
```diff
@@ -0,0 +1,70 @@
...
+				Usage: "Specifies the port to run the proxy on",
+				Value: 5511,
+			},
+			&cli.StringFlag{
+				Name:  NITRO_ENDPOINT,
+				Usage: "Specifies the endpoint of the Nitro RPC server",
+				Value: "localhost:4007/api/v1",
+			},
+			&cli.StringFlag{
+				Name:  DESTINATION_URL,
```
lalexgap, line 37: It could be nice to use a consistent format when passing in the different adresses

node_test/reverseproxy_test.go
```diff
@@ -0,0 +1,271 @@
...
+
+	logDestination := logging.NewLogWriter("../artifacts", logFile)
+
+	aliceClient, ireneClient, bobClient, cleanup := setupNitroClients(t, logDestination)
+	defer cleanup()
+
+	paymentChannel := createChannelData(t, aliceClient, ireneClient, bobClient)
+
+	// Startup a simple http server that will be used as the destination server
+	// It serves a simple text response and on two endpoints `resourceWithParams`` and `resource``
```
kerzhner, line 47:
```suggestion
	// It serves a simple text response and on two endpoints `resourceWithParams` and `resource`
```

node_test/reverseproxy_test.go
```diff
@@ -0,0 +1,271 @@
...
+	return aliceClient, ireneClient, bobClient, func() {
+		aliceCleanup()
+		ireneCleanup()
+		bobCleanup()
+		chain.Close()
+	}
+}
+
+// createChannelData creates ledgers channels and a payment channel between Alice and Bob
+func createChannelData(t *testing.T, aliceClient, ireneClient, bobClient *rpc.RpcClient) (paymentChannelId types.Destination) {
```
kerzhner, line 188: Is it worth it to factor out a utility given that we already have https://github.com/statechannels/go-nitro/blob/4f0c010645695f898e410f617c7cdf4d13792be3/cmd/create-channels/main.go#L37?

kerzhner (approved):

reverseproxy/proxy.go
```diff
@@ -0,0 +1,171 @@
...
+		if err := p.server.ListenAndServe(); err != http.ErrServerClosed {
+			p.logger.Err(err).Msg("ListenAndServe()")
+		}
+	}()
+
+	return nil
+}
+
+// Stop stops the proxy server and closes everything.
+func (p *ReversePaymentProxy) Stop() error {
```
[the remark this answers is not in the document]
lalexgap: https://github.com/statechannels/go-nitro/pull/1483/commits/ace8e79f09b201ab34b4de00c312b3ee01ea45fb

reverseproxy/proxy.go
```diff
@@ -0,0 +1,171 @@
...
+
+	// Forward the request to the destination server
+	p.reverseProxy.ServeHTTP(w, r)
+	p.logger.Debug().Msgf("Destination request URL %s", r.URL.String())
+}
+
+// webError is a helper function to return an http error.
+func (p *ReversePaymentProxy) webError(w http.ResponseWriter, err error, code int) {
+	// TODO: This is a hack to allow CORS requests to the gateway for the boost integration demo.
+	w.Header().Set("Access-Control-Allow-Origin", "*")
```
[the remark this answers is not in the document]
lalexgap: Good catch! I think this got copied over from the boost integration by accident. Removed in https://github.com/statechannels/go-nitro/pull/1483/commits/6e6231b8a7aec8f689b80bb67cbd1b609f09f4ee

node_test/reverseproxy_test.go
lalexgap: I'm going to leave this as is for now, since  `createLedgerChannel` would have to be refactored a bit to be able to serve as a utility that can be used from the `create-channels` script and in tests.  We also may want to consider other utility functions; for example we also have a [node_test helper to create ledger channels](https://github.com/statechannels/go-nitro/blob/6e6231b8a7aec8f689b80bb67cbd1b609f09f4ee/node_test/helpers_test.go#L180). I'll create an issue for that.

PR #1488, opened 2023-07-25
Use libp2p notifications

bitwiseguy:
This PR makes the following changes:
* Update libp2p configuration to work for both mDNS and dht mode
* Use libp2p connection notifications to trigger state channel signing address exchange instead of `discoverPeers` method (improves reliability/timing of peer info exchange during node initialization)
* Modify integration tests to store logs in separate files to prevent them from overwriting each other

### DIFF
--- a/node/engine/messageservice/p2p-message-service/notifier.go
+++ b/node/engine/messageservice/p2p-message-service/notifier.go
@@ -1,0 +1,22 @@
+package p2pms
+
+import (
+	"github.com/libp2p/go-libp2p/core/network"
+	"github.com/multiformats/go-multiaddr"
+)
+
+type NetworkNotifiee struct {
+	ms *P2PMessageService
+}
+
+func (nn *NetworkNotifiee) Connected(n network.Network, conn network.Conn) {
+	nn.ms.logger.Debug().Msgf("notification: connected to peer %s", conn.RemotePeer().Pretty())
+	go nn.ms.sendPeerInfo(conn.RemotePeer(), false)
+}
+
+func (nn NetworkNotifiee) Disconnected(n network.Network, conn network.Conn) {
+	nn.ms.logger.Debug().Msgf("notification: disconnected from peer: %s", conn.RemotePeer().Pretty())
+}
+
+func (nn NetworkNotifiee) Listen(network.Network, multiaddr.Multiaddr)      {}
+func (nn NetworkNotifiee) ListenClose(network.Network, multiaddr.Multiaddr) {}
--- a/node/engine/messageservice/p2p-message-service/notifier.go
+++ b/node/engine/messageservice/p2p-message-service/notifier.go
@@ -5,6 +5,8 @@ import (
 	"github.com/multiformats/go-multiaddr"
 )
 
+// Triggered by p2pHost.Network events. When new peers are connected, the
+// "Connected" notification handler will trigger nitro state channel address exchange
 type NetworkNotifiee struct {
 	ms *P2PMessageService
 }
--- a/node/engine/messageservice/p2p-message-service/service.go
+++ b/node/engine/messageservice/p2p-message-service/service.go
@@ -93,7 +93,7 @@ func NewMessageService(ip string, port int, me types.Address, pk []byte, useMdns
 	ms.key = messageKey
 	options := []libp2p.Option{
 		libp2p.Identity(messageKey),
-		libp2p.ListenAddrStrings(fmt.Sprintf("/ip4/%s/tcp/%d", "0.0.0.0", port)),
+		libp2p.ListenAddrStrings(fmt.Sprintf("/ip4/%s/tcp/%d", ip, port)),
 		libp2p.Transport(tcp.NewTCPTransport),
 		libp2p.NATPortMap(),
 		libp2p.DefaultMuxers,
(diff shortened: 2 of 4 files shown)


node/engine/messageservice/p2p-message-service/service.go
```diff
@@ -41,8 +40,9 @@ type peerExchangeMessage struct {
 }
 
 const (
-	PROTOCOL_ID                  protocol.ID = "/go-nitro/msg/1.0.0"
-	PEER_EXCHANGE_PROTOCOL_ID    protocol.ID = "/go-nitro/peerinfo/1.0.0"
```
bitwiseguy, line 45: I replaced the `go-nitro` part with just `nitro` in case there are other node types in the future that are written in different languages

node/engine/messageservice/p2p-message-service/service.go
```diff
@@ -143,19 +142,21 @@ func (ms *P2PMessageService) setupDht(bootPeers []string) {
...
 	ms.addBootPeers(bootPeers)
 
 	expectedPeers := len(bootPeers)
 	if expectedPeers > 0 {
+		ms.logger.Debug().Msgf("waiting for %d bootpeer connections", expectedPeers)
 		ticker := time.NewTicker(BOOTSTRAP_SLEEP_DURATION)
 		for range ticker.C {
-			peers := ms.p2pHost.Peerstore().Peers()
-			actualPeers := len(peers) - 1 // subtract one to ignore self
+			peers := ms.p2pHost.Network().Peers()
```
bitwiseguy, line 158: This line checks the actual number of peers this node is currently connected. `ms.p2pHost.Peerstore().Peers()` checks how many entries are in the local `Peerstore` which could be leftover data from a previous node session if the node is using a persistent store. Also an entry in the `Peerstore` does not mean there is an actual connection between the nodes.

node/engine/messageservice/p2p-message-service/service.go
```diff
@@ -323,44 +326,10 @@ func (ms *P2PMessageService) addBootPeers(peers []string) {
...
-		err = ms.p2pHost.Connect(context.Background(), *peer) // Adds peerInfo to Peerstore
+		err = ms.p2pHost.Connect(context.Background(), *peer) // Adds peerInfo to local Peerstore
 		ms.checkError(err)
 		ms.logger.Debug().Msgf("connected to boot peer: %v", p)
 
 		ms.sendPeerInfo(peer.ID, true)
 	}
 }
-
-func (ms *P2PMessageService) discoverPeers(ctx context.Context, topic string) {
```
bitwiseguy, line 334: We may want to modify and add this method back at some point to augment the libp2p connection notifications for faster, more robust peer discovery. However, the libp2p connection notifications replace this method for now.

node/engine/messageservice/p2p-message-service/notifier.go
```diff
@@ -0,0 +1,22 @@
+package p2pms
+
+import (
+	"github.com/libp2p/go-libp2p/core/network"
+	"github.com/multiformats/go-multiaddr"
+)
+
+type NetworkNotifiee struct {
```
lalexgap, line 8: nit but it could be nice to add some godoc around what this struct does

node/engine/messageservice/p2p-message-service/service.go
```diff
@@ -91,12 +91,11 @@ func NewMessageService(ip string, port int, me types.Address, pk []byte, useMdns
 	ms.checkError(err)
 
 	ms.key = messageKey
-
 	options := []libp2p.Option{
 		libp2p.Identity(messageKey),
-		libp2p.ListenAddrStrings(fmt.Sprintf("/ip4/%s/tcp/%d", ip, port)),
+		libp2p.ListenAddrStrings(fmt.Sprintf("/ip4/%s/tcp/%d", "0.0.0.0", port)),
```
lalexgap, line 96: Is there a reason we're switching to `0.0.0.0`? Is it because specifying the IP address doesn't really matter?

node/engine/messageservice/p2p-message-service/service.go
```diff
@@ -91,12 +91,11 @@ func NewMessageService(ip string, port int, me types.Address, pk []byte, useMdns
...
 		libp2p.Transport(tcp.NewTCPTransport),
-		libp2p.NoSecurity,
+		libp2p.NATPortMap(),
```
lalexgap, line 98: Does this help with NAT traversal?

node/engine/messageservice/p2p-message-service/service.go
```diff
@@ -128,9 +127,9 @@ func (ms *P2PMessageService) setupMdns() {
 func (ms *P2PMessageService) setupDht(bootPeers []string) {
 	ctx := context.Background()
 	var options []dht.Option
-	if len(bootPeers) == 0 {
-		options = append(options, dht.Mode(dht.ModeServer)) // Server mode: allows other peers to connect to this node
-	}
+	options = append(options, dht.BucketSize(20))
```
lalexgap, line 130: Is there any reason for a bucket size of 20?

bitwiseguy: The default value is 20. I just set this option so it is obvious which bucket size we are using and it is easy to change later if we want.

node/engine/messageservice/p2p-message-service/service.go
```diff
@@ -143,19 +142,21 @@ func (ms *P2PMessageService) setupDht(bootPeers []string) {
 	addrs, err := peer.AddrInfoToP2pAddrs(&peerInfo)
 	ms.checkError(err)
 	ms.MultiAddr = addrs[0].String()
-	ms.logger.Debug().Msgf("libp2p node address: %s", ms.MultiAddr)
+	ms.logger.Debug().Msgf("libp2p node multiaddrs: %v", addrs)
+
+	// Setup notifications so we exchange nitro signing addresses when connected
+	n := &NetworkNotifiee{ms: ms}
```
lalexgap, line 148: Instead of wrapping a message service in a `NetworkNotifiee` struct, the message service itself could just implement the `Notifee` interface. That would avoid the need for a new `NetworkNotifiee` struct; although it means the message service has more functions, so maybe not worth it?

I wonder if longer term we should consider refactoring out all the dht/peer discovery logic out of message service struct into it's own struct that can be owned by the message service?

lalexgap (approved):

node/engine/messageservice/p2p-message-service/service.go
bitwiseguy: > I wonder if longer term we should consider refactoring out all the dht/peer discovery logic out of message service struct into it's own struct that can be owned by the message service?

This seems like a good approach since the message service itself has grown pretty large to support all of the dht-specific methods.

node/engine/messageservice/p2p-message-service/service.go
bitwiseguy: I'm changing this back to read from the `ip` argument instead of hardcoding to `0.0.0.0`. I can't tell why but the `0.0.0.0` argument seems to cause frequent failures in the CI. I will look into this further when addressing this issue: https://github.com/statechannels/go-nitro/issues/1489

PR #1491, opened 2023-07-25
Updates the payment proxy to charge per byte of response

lalexgap:
This updates the `ReversePaymentProxy` to accept a `costPerByte` and check the voucher amount against `contentLength * costPerByte`. It does this by applying the payment check **after the proxy has received the response from the destination server** so it knows the length of the response. If the voucher amount is not large enough we throw out the response and return a 402 error response. 

The downside of this approach is that we only find out if a voucher is too small **after** the destination server has provided a response to the proxy.  This means the destination server still has to do the "work" of responding even if it's not getting paid enough. The proxy just throws that response out if the voucher isn't big enough.


node_test/reverseproxy_test.go
```diff
@@ -53,8 +53,8 @@ func TestReversePaymentProxy(t *testing.T) {
 		proxyAddress,
 		bobRPCUrl,
 		destinationServerUrl,
-		big.NewInt(5),
-		zerolog.New(logDestination).Level(zerolog.DebugLevel))
+		1,
+		zerolog.New(os.Stdout).Level(zerolog.DebugLevel))
```
lalexgap, line 57:
```suggestion
		zerolog.New(logDestination).Level(zerolog.DebugLevel))
```

reverseproxy/proxy.go
```diff
@@ -44,130 +51,119 @@ func NewReversePaymentProxy(proxyAddress string, nitroEndpoint string, destinati
...
 	}
 
-	return p.nitroClient.Close()
+	// We remove the voucher from the query params so it doesn't pollute the URL
+	queryParams := r.URL.Query()
+	removeVoucher(queryParams, "")
+	r.URL.RawQuery = queryParams.Encode()
+
+	// We add the voucher to the request header so we can easily check the cost in the response handler without polluting the URL
+	addVoucher(v, r.Header, HEADER_PREFIX)
```
lalexgap, line 91: This isn't strictly necessary, we could just keep the voucher on the query params and parse it out from that later. However that means the voucher params would get passed to the destination server, which it may not expect. By moving it to a custom header we still have access to it but it's less likely to cause issues with the destination server.

kerzhner (approved):
Looks great!

reverseproxy/proxy.go
lalexgap: Ooh good idea!

reverseproxy/proxy.go
```diff
@@ -44,130 +51,119 @@ func NewReversePaymentProxy(proxyAddress string, nitroEndpoint string, destinati
...
-		nitroClient:           nitroClient,
-		reverseProxy:          proxy,
-		expectedPaymentAmount: expectedPaymentAmount,
+	p := &ReversePaymentProxy{
+		server:         server,
+		logger:         logger,
+		nitroClient:    nitroClient,
+		costPerByte:    costPerByte,
+		destinationUrl: destinationUrl,
+
```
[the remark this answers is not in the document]
lalexgap: https://github.com/statechannels/go-nitro/pull/1491/commits/152984079863bf210d55ef180f91bf08a8e351b2

PR #1492, opened 2023-07-26
Custom dht namespace for state channel addresses

bitwiseguy:
## Motivation

In order to allow go-nitro nodes to look up peers using only the state-channel signing address I added a custom “namespace” to the dht called `scaddr` (state channel address). This seems necessary due to our messaging format and the way the message-service `Send` method relies on the “To” field in the message body.

By querying the new namespace we can access a global mapping of `scaddr --> {SCAddr, PeerId, Timestamp}` (the dht lookup is only necessary if there is not an entry for that address within the [node's local map](https://github.com/statechannels/go-nitro/blob/dht-namespace/node/engine/messageservice/p2p-message-service/service.go#L57). A distributed, global mapping will provide a reliable data store that does not rely on any individual node’s ability to maintain the data. However, the downside to using a dht for storing this map is that it opens our network up to additional Sybil and DoS attacks. To mitigate some of that risk, there is a data validator on the custom dht namespace which includes a signature verification. That verification ensures only the owner of the `PeerId` key pair can add a record. This will help identify an offending node if they attempt to pollute the dht with junk data. Here are the data structs used for those dht records:

```
type DhtRecord struct {
	Data      DhtData `json:"data"`
	Signature []byte  `json:"signature"`
}

type DhtData struct {
	SCAddr    string `json:"scaddr"` // state channel address
	PeerID    string `json:"peerid"`
	Timestamp int64  `json:"timestamp"` // Unix timestamp (seconds since January 1, 1970)
}
```

> Note: there are two default libp2p dht namespaces called `pk` and `ipns`. Similar to the new namespace design, the records in the `ipns` namespace must be signed by the peerId private key in order to be added to the dht.

## Open question

Do we need to add a signature generated by the `SCAddr` private key to the `DhtRecord`? This would help protect against a malicious node who tried to claim they own a `SCAddr` to confuse other nodes and disrupt their communication with each other.

Right now the public key for the state channel signing address is stored in the [`P2PMessageService.me`](https://github.com/statechannels/go-nitro/blob/dht-namespace/node/engine/messageservice/p2p-message-service/service.go#L86). However, the message-service does not have access to the private key (owned by the `store` service) so it can’t perform any signatures, which prevents it from adding its own information to the `scaddr` namespace in the dht. A few options to perform that signature:

1. Allow the message-service to call the [`store.GetSecretKey()`](https://github.com/statechannels/go-nitro/blob/dht-namespace/node/engine/store/store.go#L22) method so that the message-service can perform the signature itself
2. Sign the `DhtRecord.Data` when the node first initializes and pass the signed record to the message-service so that it can add it to the dht
3. Other?

## TODO
Need to add a test for the custom dht namespace. In the existing `TestKademliaDhtIntegration` tests the nodes all exchange state channel address info with each other directly so they only use their local `peers` map to lookup `scaddr --> peerId` instead of having to reach out to the custom dht. We need to add a test that mocks out the current `NetworkNotifiee` (specifically the `sendPeerInfo` call in the `Connected` notification handler). Perhaps we have to add a `NetworkNotifiee` argument to the `NewMessageService` function to allow us to pass a mock during testing?


lalexgap:
> Do we need to add a signature generated by the `SCAddr` private key to the `DhtRecord`? This would help protect against a malicious node who tried to claim they own a `SCAddr` to confuse other nodes and disrupt their communication with each other.
> 
> Right now the public key for the state channel signing address is stored in the [`P2PMessageService.me`](https://github.com/statechannels/go-nitro/blob/dht-namespace/node/engine/messageservice/p2p-message-service/service.go#L86). However, the message-service does not have access to the private key (owned by the `store` service) so it can’t perform any signatures, which prevents it from adding its own information to the `scaddr` namespace in the dht. A few options to perform that signature:

This is slightly related but we could also consider using the state channel key **as the message key** (in fact that's what we do in [some of our tests](https://github.com/statechannels/go-nitro/blob/1fe3fcf7381439274ba4e6e64b29c45457b34394/node_test/helpers_test.go#L76)).  The main advantage of this is that we[ can figure out someone's SC address based on their peer ID](https://github.com/statechannels/go-nitro/pull/1494). This means we could avoid the extra effort of sharing around our SC address and maintaining a mapping of SC address to peer id.

bitwiseguy:
> This is slightly related but we could also consider using the state channel key **as the message key** (in fact that's what we do in [some of our tests](https://github.com/statechannels/go-nitro/blob/1fe3fcf7381439274ba4e6e64b29c45457b34394/node_test/helpers_test.go#L76)). The main advantage of this is that we[ can figure out someone's SC address based on their peer ID](https://github.com/statechannels/go-nitro/pull/1494). This means we could avoid the extra effort of sharing around our SC address and maintaining a mapping of SC address to peer id.

Oh that's interesting. I didn't realize you could derive a secp256k1 address (the addresses used on Ethereum) from the peerId. It makes me wonder why libp2p uses the alternate format for peerId instead of just using the same public key format as Ethereum addresses.

Removing the additional dht namespace is desirable but it seems like that means our nitro node's peerId would be 1:1 with a state channel address. I was under the impression that we wanted to allow multiple state channel addresses to be controlled/used by a single peerId. If that's not the case and we are good with a 1:1 mapping then I agree your suggestion seems like the better approach.

lalexgap:
Based on stand up discussion we're going to think about and discuss #1495 before reviewing and merging this in.

node/engine/messageservice/p2p-message-service/service.go
```diff
@@ -170,19 +172,43 @@ func (ms *P2PMessageService) setupDht(bootPeers []string) {
...
 			if actualPeers >= expectedPeers {
-				ms.logger.Debug().Msgf("discovered all expected bootPeers")
+				ms.logger.Info().Msgf("initial threshold for peer connections has been met")
 				ticker.Stop()
 				break
 			}
 		}
+
+		// Add my state channel address to the custom dht
+		recordData := &DhtData{
```
lalexgap, line 184: Might be nice to encapsulate this logic into a function on the message service

node/engine/messageservice/p2p-message-service/service.go
```diff
@@ -170,19 +172,43 @@ func (ms *P2PMessageService) setupDht(bootPeers []string) {
...
+			Data:      *recordData,
+			Signature: signature,
+		}
+		fullRecordBytes, err := json.Marshal(fullRecord)
+		ms.checkError(err)
+
+		key := DHT_RECORD_PREFIX + ms.me.String()
+		err = ms.dht.PutValue(ctx, key, fullRecordBytes)
+		ms.checkError(err)
+		ms.logger.Info().Msgf("Added value to dht - [key: %v, value: %v]", key, ms.Id())
```
lalexgap, line 205: nit but I think if we do this it makes the logs easier to parse:
```suggestion
			ms.logger.Info().Str(key, ms.Id().Pretty()).Msg("Added value to dht")
```

node/engine/messageservice/p2p-message-service/service.go
```diff
@@ -269,28 +295,38 @@ func (ms *P2PMessageService) Send(msg protocols.Message) {
 	raw, err := msg.Serialize()
 	ms.checkError(err)
 
+	// First try to get peerId from local "peers" map. If the address is not found there,
+	// query the dht to retrieve the peerId, then store in local map for next time
 	peerId, ok := ms.peers.Load(msg.To.String())
 	if !ok {
-		panic(fmt.Errorf("could not load peer %s", msg.To.String()))
+		ms.logger.Info().Msgf("did not find address %s in local map, will query dht", msg.To.String())
```
lalexgap, line 302: This logic could also be in it's own function, that could help reduce the overall complexity of `Send`

node/engine/messageservice/p2p-message-service/dht-record.go
```diff
@@ -0,0 +1,85 @@
...
+		return errors.New("invalid signature")
+	}
+
+	return nil
+}
+
+// Simply return the first record as the best record.
+// In a more complex scenario, we could add logic to select the best record
+// based on some criteria.
+func (v stateChannelAddrToPeerIDValidator) Select(key string, values [][]byte) (int, error) {
```
lalexgap, line 83: Should we return an error in this function if there are multiple entries(IE: `len(values)>1`) for now? I think we only expect one record to be published per SC address, so if there are multiple records something has gone wrong.

node/engine/messageservice/p2p-message-service/dht-record.go
```diff
@@ -0,0 +1,85 @@
...
+)
+
+const (
+	DHT_RECORD_PREFIX = "/scaddr/"
+)
+
+type stateChannelAddrToPeerIDValidator struct{}
+
+// DhtRecord represents the data stored in the DHT record
+type DhtRecord struct {
```
lalexgap, line 20: Nit but since `DhtRecord` is only using in the `p2pms` package I think it could be a private struct. Same thing with `DhtData`
```suggestion
type dhtRecord struct {
```

node/engine/messageservice/p2p-message-service/dht-record.go
bitwiseguy: The `Select` method is used to choose between records if different ones are returned by different peers for the same key, which is possible because the dht is not updated instantly across all nodes so some peers can have stale data before a record update is fully propagated. I don't think it should cause an error if we receive two different values when we query for a given key. Rather, I think we should choose the record that has the latest timestamp to ensure we have the most recently updated entry.

node/engine/messageservice/p2p-message-service/dht-record.go
```diff
@@ -0,0 +1,85 @@
...
+	Data      DhtData `json:"data"`
+	Signature []byte  `json:"signature"`
+}
+
+type DhtData struct {
+	SCAddr    string `json:"scaddr"` // state channel address
```
lalexgap, line 26: Is it worth using these struct tags to lowercase the JSON ouput? Or could we just omit them to simplify things?

lalexgap (approved):
👍

kerzhner (approved):
This looks good in general. It would be really nice if dht offered something akin to `HandlePeerFound`...

One option is to run a dht node that is the bootstrap node. All other nodes would reference this node as a boot node and, then, would be able to discover all other nodes.

node/engine/messageservice/p2p-message-service/dht-record.go
bitwiseguy: I added logic to the `Select` method to choose the most recent record we receive

node/engine/messageservice/p2p-message-service/dht-record.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	// Parse the value into a RecordData object
+	var dhtRecord DhtRecord
+	if err := json.Unmarshal(value, &dhtRecord); err != nil {
+		return errors.New("malformed record value")
+	}
+
+	// Make sure the timestamp is not in the future or negative number
+	if dhtRecord.Data.Timestamp > time.Time.Unix(time.Now()) || dhtRecord.Data.Timestamp < 0 {
```
kerzhner, line 48: Could this cause problems due to differing clocks on different machines? The clock on my machine could be ahead of the clock on your machine.

node/engine/messageservice/p2p-message-service/dht-record.go
lalexgap: > Rather, I think we should choose the record that has the latest timestamp to ensure we have the most recently updated entry.

Should we make that change in this PR? I'm just worried that always returning the first record might introduce some subtle race conditions that could be difficult to debug.

node/engine/messageservice/p2p-message-service/dht-record.go
bitwiseguy: @lalexgap are you suggesting to omit the `json` tags so that the struct looks like this?
```
type dhtData struct {
	SCAddr    string // state channel address
	PeerID    string
	Timestamp int64  // Unix timestamp (seconds since January 1, 1970)
}
```

node/engine/messageservice/p2p-message-service/dht-record.go
bitwiseguy: Thats a good point. Do you think I should just remove this timestamp check from the `Validate` method? I'm not sure that its necessary.

node/engine/messageservice/p2p-message-service/dht-record.go
lalexgap: @bitwiseguy Yup! The fields will still be outputted to JSON (but as `ScAddr` instead of `scaddr`) and we save same text.

node/engine/messageservice/p2p-message-service/service.go
```diff
@@ -138,7 +138,9 @@ func (ms *P2PMessageService) setupDht(bootPeers []string) {
 	ctx := context.Background()
 	var options []dht.Option
 	options = append(options, dht.BucketSize(20))
-	options = append(options, dht.Mode(dht.ModeServer)) // allows other peers to connect to this node
+	options = append(options, dht.Mode(dht.ModeServer))                                               // allows other peers to connect to this node
+	options = append(options, dht.ProtocolPrefix(DHT_PROTOCOL_PREFIX))                                // need this to allow custom NamespacedValidator
+	options = append(options, dht.NamespacedValidator("scaddr", stateChannelAddrToPeerIDValidator{})) // all records prefixed with /scaddr/ will use this custom validator
```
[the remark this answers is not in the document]
bitwiseguy: `scaddr` is a namespace within the DHT, so it is at a lower level than the `DHT_PROTOCOL_PREFIX`. A record in the `scaddr` namespace will look like this (assuming `DHT_PROTOCOL_PREFIX = "/nitro"`: `/nitro/scaddr/0x12345...9089`

node/engine/messageservice/p2p-message-service/dht-record.go
bitwiseguy: Yes it has been added

node/engine/messageservice/p2p-message-service/dht-record.go
bitwiseguy: @kerzhner I wanted to add the `SCAddr` to that struct so that it is included in the `Signature`. So the signature directly ties the `PeerID` to the `SCAddr`. But maybe that is unnecessary?

node/engine/messageservice/p2p-message-service/service.go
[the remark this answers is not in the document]
bitwiseguy: Sorry, I misunderstood what you were saying. We want to remove the hardcoded "scaddr" on that line and use a `const` instead:
```
const (
	DHT_RECORD_PREFIX                             = "/" + DHT_NAMESPACE + "/"
	DHT_NAMESPACE                                    = "scaddr"
	DHT_PROTOCOL_PREFIX    protocol.ID = "/nitro" // use /nitro/kad/1.0.0 instead of /ipfs/kad/1.0.0
)
	

options = append(options, dht.ProtocolPrefix(DHT_PROTOCOL_PREFIX))                                     // need this to allow custom NamespacedValidator
	options = append(options, dht.NamespacedValidator(DHT_NAMESPACE, stateChannelAddrToPeerIDValidator{})) //
```

bitwiseguy:
> This looks good in general. It would be really nice if dht offered something akin to `HandlePeerFound`...
> 
> One option is to run a dht node that is the bootstrap node. All other nodes would reference this node as a boot node and, then, would be able to discover all other nodes.

I believe we are already doing this. The dht's `Notifee` interface provides the `Connected` method, which is triggered anytime we connect to a new peer. So it is basically serving the same purpose as our `HandlePeerFound` method:
```
	n.ConnectedF = func(n network.Network, conn network.Conn) {
		ms.logger.Debug().Msgf("notification: connected to peer %s", conn.RemotePeer().String())
		go ms.sendPeerInfo(conn.RemotePeer(), false)
	}
```
Our kad-dht integration tests currently spin up an intermediary (Irene) that acts as a bootpeer. Alice and Bob then connect to Irene (by using Irene's multiaddr as a bootpeer). Alice and Bob then get connected to each automatically through the kad-dht because Irene is able to share Alice's info with Bob and vice a versa.

node/engine/messageservice/p2p-message-service/dht-record.go
kerzhner: That's a good point, and we should keep `SCAddr` in the struct. If we do not include the `SCAddr`, then the record and signature can be reused by any peer. A malicious peer can then add the record and signature pair for any state channel address key.

kerzhner:
> Our kad-dht integration tests currently spin up an intermediary (Irene) that acts as a bootpeer. Alice and Bob then connect to Irene (by using Irene's multiaddr as a bootpeer). Alice and Bob then connect to each other automatically through the kad-dht peer discovery because Irene is able to share Alice's info with Bob and vice a versa.

It looks like there are a couple independent components:
1. _How do peers discover each other?_ With mdns, it seems like we do not define any boot peers. With dht, we do. [My comment](https://github.com/statechannels/go-nitro/pull/1492#pullrequestreview-1561634374) laments this.
2. _How do peers share information about each other?_ With mds, we rely on `sendPeerInfo`. With dht, we should rely on reading dht records and remove sendPeerInfo.

Thoughts?

node/engine/messageservice/p2p-message-service/dht-record.go
bitwiseguy: Added a comparison of the record's key to the `SCAddr` value

PR #1493, opened 2023-07-26
Add RANGE request support to the payment proxy

lalexgap:
⚠️ Based on #1491 ⚠️ 

This adds support to the payment proxy for the [http range requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests).

If the range header is set the payment proxy checks the voucher amount against the range value.


reverseproxy/keyval.go
```diff
@@ -59,3 +62,46 @@ func removeVoucher(col keyValCollection, prefix string) {
...
+		start, err := strconv.ParseUint(numVals[0], 10, 64)
+		if err != nil {
+			return 0, fmt.Errorf("could not parse start: %w", err)
+		}
+
+		end, err := strconv.ParseUint(numVals[1], 10, 64)
+		if err != nil {
+			return 0, fmt.Errorf("could not parse start: %w", err)
+		}
+
```
kerzhner, line 102: We might want to validate that `end > start`. Otherwise, could the following scenario arise:
- the caller requests a valid, positive range and an invalid negative range.
- the destination server services the valid range.
- the caller underpays for the valid range.

node_test/reverseproxy_test.go
```diff
@@ -22,17 +22,39 @@ import (
...
+	// Open the file for writing (create or truncate)
+	file, err := os.Create(testFileName)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer file.Close()
+
+	_, err = file.WriteString(testFileContent)
+	if err != nil {
+		t.Fatal(err)
```
kerzhner, line 48: Should the file get removed in this case?

kerzhner (approved):
This PR makes sense, and thanks for putting it together. It seems like this PR is an optimization. Is it true that the reverse proxy will function correctly even without this PR as long as the destination server returns the correct number of bytes for a range request? If so, then the tradeoffs are:
- This PR optimizes range requests so that the request is denied prior to contacting the destination server.
- However, this PR adds complexity to the reverse proxy.

reverseproxy/keyval.go
lalexgap: https://github.com/statechannels/go-nitro/pull/1493/commits/68f81a64d8f36670af6f8fd95ee7b0b618a170bf

lalexgap:
> Is it true that the reverse proxy will function correctly even without this PR as long as the destination server returns the correct number of bytes for a range request

Yup that's true and a good point! Since the proxy is forwarding the request and just charging based on the response* it should already work with range requests.

> This PR optimizes range requests so that the request is denied prior to contacting the destination server.

I think this is the main motivation behind this PR; the proxy will only forward requests to the destination server if they've been paid for. 

However, I'm not sure how important that actually is. A malicious actor could underpay for requests knowing it's going to make the `destination server`  do the work of serving those requests (even if the proxy never forwards the data to the actor), and so could be a DoS attack vector.  However it seems pretty easy to blacklist actors that attempt this, since the voucher they submit includes a `channelId` we can use to identify them. I think we'd still want to perform some validation of the voucher before forwarding the request to the `destination server` to make sure it has a valid channel id and signature. Otherwise you could use malformed vouchers that could not be traced back to a channel and actor to blacklist.

*One minor difference is that in this PR we charge based on the size of the requested data and previously we charged based on the size of the response body. Most of the times this is the same thing, however it could differ slightly in the case of [multipart ranges](https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests#multipart_ranges). With multipart ranges the response from the server has to label each different part, which means the response will be slightly larger than the requested data and technically cost more. Practically I can't see it mattering since the extra data/cost is negligible; just wanted to flag a slight difference. 

```
--3d6b6a416f9b5
Content-Type: text/html
Content-Range: bytes 0-50/1270**
```

node_test/reverseproxy_test.go
lalexgap: https://github.com/statechannels/go-nitro/pull/1493/commits/9b0343da18afb0e467c1107251b201f86a57a187

lalexgap:
Closing this in favour of just the test for now (in #1499) [based on this discussion](https://github.com/statechannels/go-nitro/pull/1493#issuecomment-1658976210).

Issue #1495, opened 2023-07-27
Consider using SC key in P2P message service


lalexgap:
Our p2p message service currently relies on either a DHT or mdns to discover peers and share them with the message service. Whenever the message service is alerted of a new peer we get shared the `Peer.ID` and a multiaddress. We then share our SC address by manually [sending a message to our peers](https://github.com/statechannels/go-nitro/blob/5284f0dc92c3da37266813319014e9a8c74a8945/node/engine/messageservice/p2p-message-service/service.go#L234).

However the libp2p peer.ID contains a function to[ extract the public key](https://pkg.go.dev/github.com/libp2p/go-libp2p/core/peer#ID.ExtractPublicKey) from the ID. This means if a secp256k1 key is used we can [derive a state channel address from the peer.ID](https://github.com/statechannels/go-nitro/pull/1494) 

That means if we use the SC key as our message key, we can derive a peer's SC address from the peer.ID shared with us. It removes any extra communication to our peers to share our SC address, greatly simplifying the message service.

This does mean we're now sharing our private key with the message service and the libraries it uses, which comes with increased security risks.

It also means that our message service identity would be directly tied to our state channel identity. This means our message service wouldn't work with a nitro node running multiple SC addresses.

## Possible Options

1. Keep the SC key and message key separate and rely on the existing peer sharing logic or [custom DHT namespaces](https://github.com/statechannels/go-nitro/pull/1492) to share the SC address
2. [Update our message service to rely on the SC key and recover a peer's SC address from their peer ID](https://github.com/statechannels/go-nitro/pull/1496). This ties the message identity to our SC identity but simplifies the logic of discovering peers.
3. Refactor the current message service into two message services? One that just attempts to get the SC address from the peer.ID and a more complicated service that handles sharing the SC address?

lalexgap:
I'm leaning towards 2 since it removes a bunch of complexity in the message service around discovering peers. In 2 when you discover a peer via dht/mdns you have all the info you need to message with them, including their SC address. That removes a bunch of logic around communicating your SC address to your peers after you discover them. Not only does this make things easier to understand, it's  also more efficient and less error prone as we remove a layer of communication.

[...]

PR #1496, opened 2023-07-27
Use the SC key as the message key and derive a SC address from peer.ID

lalexgap:
See #1495 for the context.

### DIFF
--- a/node/engine/messageservice/p2p-message-service/service.go
+++ b/node/engine/messageservice/p2p-message-service/service.go
@@ -3,12 +3,12 @@ package p2pms
 import (
 	"bufio"
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
 	"time"
 
+	ethcrypto "github.com/ethereum/go-ethereum/crypto"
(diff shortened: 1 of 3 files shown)


node/engine/messageservice/p2p-message-service/service.go
```diff
@@ -326,10 +274,33 @@ func (ms *P2PMessageService) addBootPeers(peers []string) {
 		peer, err := peer.AddrInfoFromP2pAddr(addr)
 		ms.checkError(err)
 
+		a, err := addressFromPeerID(peer.ID)
+		ms.checkError(err)
+		ms.peers.Store(a.String(), peer.ID)
```
lalexgap, line 279: My understanding is that we can derive the SC address from a peer.ID but we cannot derive a peer.ID from a SC address. This is because an address is derived by hashing the public key and taking the last 20 bytes of that hash. This means we still need to maintain a peers map, to map SC address to peer.IDs.

lalexgap:
Closing as we decided we didn't want to require the SC key

PR #1497, opened 2023-07-27
Remove the notifier struct in the p2p message service

lalexgap:
Poking around I discovered a [NotifyBundle](https://pkg.go.dev/github.com/libp2p/go-libp2p-core/network#NotifyBundle) in libp2p that lets us register notification functions without bothering to create our own notifier struct. This cleans things up slightly and let's us inline the notification logic into `service.go`.


bitwiseguy (approved):
Nice find!

PR #1498, opened 2023-07-28
Docker image for use in cloud

bitwiseguy:
Create a docker image so that we can deploy a nitro node into the cloud. This node should connect to the Filecoin calibration testnet and can act as a hub. Developers should be able to run local nodes that connect to the hub, which gives them access to a distributed nitro network.


docker/cloud/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM golang:1.20-bullseye AS builder
+WORKDIR /app
+COPY . .
+RUN go build -o nitro .
+
+FROM alpine:latest
```
bitwiseguy, line 6: Using build steps like this in docker reduces the image size from ~2.5GB to ~69MB. This works especially well for golang since it is a compiled language and we only need the executable to run the program after its been built so we can discard the source code from the docker image.

docker/cloud/config.toml
```diff
@@ -0,0 +1,18 @@
+usedurablestore = true
+
+msgport = 3005
+rpcport = 4005
+guiport = 5005
+
+pk = "1ea91a2724b40fb8fed6a3648d49e9431996c09744fb841b718377fb0700f3e7"
+chainpk = "47e179ec197488593b187f80a00eb0da91f1b9d0b13f8733639f19c30a34926a"
```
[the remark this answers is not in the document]
bitwiseguy: For now I just set these values to empty strings and expect the user to populate them manually before building the image

node/engine/chainservice/eth_chainservice.go
```diff
@@ -65,6 +74,15 @@ const MAX_QUERY_BLOCK_RANGE = 2000
 // This has been reduced to 15 seconds to support local devnets with much shorter timeouts.
 const RESUB_INTERVAL = 15 * time.Second
 
+// REQUIRED_BLOCK_CONFIRMATIONS is how many blocks must be mined before an emitted event is processed
+const REQUIRED_BLOCK_CONFIRMATIONS = 0
```
bitwiseguy, line 78: This need to be set to something > 0 to work against the Filecoin Calibration testnet. However, it needs to be set to 0 to pass tests that run against the simulated chain used in testing. In a follow-up PR we will address this by adding logic to set the value based on the chainId (i.e. against simulated test chains we will use 0 and against all other chainIds we will use 2)

PR #1499, opened 2023-07-31
Add tests for range requests to payment proxy

lalexgap:
This adds some test for the payment proxy that makes sure it works with RANGE requests (based on #1493). @kerzhner brought up the good point that the changes in #1493 weren't strictly necessary, so this PR just adds the tests.

### DIFF
--- a/node_test/reverseproxy_test.go
+++ b/node_test/reverseproxy_test.go
@@ -22,8 +22,8 @@ import (
 )
 
 const (
-	smallResponse              = "Hello"
-	largeResponse              = "HelloWorldHelloWorld"
+	smallResponse = "Hello"
+
 	parseErrorResponseBody     = "could not parse voucher"
 	signatureErrorResponseBody = "error processing voucher"
 	proxyAddress               = ":5511"
@@ -31,7 +31,30 @@ const (
 	destPort                   = 6622
 	otherParam                 = "otherParam"
 	otherParamValue            = "2"
+	testFileContent            = "This a simple test file used in the reverse payment proxy"
+	testFileName               = "test_file.txt"
 )
+
+func setupTestFile(t *testing.T) func() {
+	// Open the file for writing (create or truncate)
+	file, err := os.Create(testFileName)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer file.Close()
+
+	_, err = file.WriteString(testFileContent)
+	if err != nil {
+		os.Remove(testFileName)
+		t.Fatal(err)
+	}
+	return func() {
+		err := os.Remove(testFileName)
+		if err != nil {
+			t.Fatal(err)
+		}
+	}
+}
 
 func TestReversePaymentProxy(t *testing.T) {
 	logFile := "reverse_payment_proxy.log"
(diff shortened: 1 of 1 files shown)


node_test/reverseproxy_test.go
```diff
@@ -195,21 +255,21 @@ func setupNitroClients(t *testing.T, logDestination *os.File) (alice, irene, bob
 
 // createChannelData creates ledgers channels and a payment channel between Alice and Bob
 func createChannelData(t *testing.T, aliceClient, ireneClient, bobClient rpc.RpcClientApi) (paymentChannelId types.Destination) {
-	aliceLedgerRes, err := aliceClient.CreateLedgerChannel(ta.Irene.Address(), 100, simpleOutcome(ta.Alice.Address(), ta.Irene.Address(), 100, 100))
+	aliceLedgerRes, err := aliceClient.CreateLedgerChannel(ta.Irene.Address(), 100, simpleOutcome(ta.Alice.Address(), ta.Irene.Address(), 500, 500))
```
lalexgap, line 258: This was increased to afford the cost of the multipart range response

node_test/reverseproxy_test.go
```diff
@@ -90,28 +116,60 @@ func TestReversePaymentProxy(t *testing.T) {
...
+	if body != testFileContent[0:5] {
+		t.Fatalf("Expected response body to be %s, but got %s", testFileContent[0:5], body)
+	}
+	if statusCode != http.StatusPartialContent {
+		t.Fatalf("Expected status code %d, but got %d", http.StatusPartialContent, statusCode)
+	}
+
+	// It should handle a complex range request
+	// Multipart ranges include extra information about the ranges in the response body
+	const multiPartResponseSize = 346
```
lalexgap, line 156: The response will look something like this
```
--01a50f63ca7bbec85c50a0b8f753646dd4f33bf5cdae5e3d8d81b472bec5
Content-Range: bytes 0-1/57
Content-Type: text/plain; charset=utf-8

Th
--01a50f63ca7bbec85c50a0b8f753646dd4f33bf5cdae5e3d8d81b472bec5
Content-Range: bytes 3-4/57
Content-Type: text/plain; charset=utf-8

s
--01a50f63ca7bbec85c50a0b8f753646dd4f33bf5cdae5e3d8d81b472bec5--
```

kerzhner: Is it true that currently the requester pays for every byte of the response body? If so, does it make sense to consider an approach where the requester only pays for content (and not metadata)?

lalexgap: > Is it true that currently the requester pays for every byte of the response body? 

That's true! One thing to note is that only multipart range responses include metadata. If you request data with a single range (IE: `bytes=0-49`) the response body will only contain the 50 bytes you requested and no extraneous metadata. So single part ranges the size of the response body is the size of the requested data.

It's only when a multipart range is requested (IE: `bytes=0-10,40-49`) that the extra metadata is included, and thus you'd end up paying more.  If we wanted to address this I think the payment proxy would have to have parse the range information from the request (like we did in this [PR](https://github.com/statechannels/go-nitro/pull/1493)) so we'd know the size of the data being requested.

However I'm wondering if it's worth the effort since paying slightly more for multipart range requests doesn't seem like the end of the world.  In this test the extra metadata dominates the size but that's because we're requesting 4 bytes of data. In the real world the chunks of data you're requesting will probably be much larger than that so the extra cost won't be as painful. It also seems unlikely that people will use a multipart range request for fetching file data, either you request the file in one go or you single part ranges to request chunks of the file over different requests.

kerzhner: All good points. Sounds good and let's merge the code as is.

PR #1510, opened 2023-08-03
Add micropayment support to proxy client

### LINKED ISSUE #1503: Update boost demo to work with Range Requests
Add support for HTTP range requests in the boost react demo.

lalexgap:
Fixes #1503 

This adds support to the proxy client UI to use [HTTP range requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests) to request (and pay for) a file over multiple http requests. 

The client can specify the chunk size (in bytes) they want to use.
micropayment_v2

### DIFF
--- a/packages/payment-proxy-client/src/file.ts
+++ b/packages/payment-proxy-client/src/file.ts
@@ -4,13 +4,10 @@ import { Voucher } from "@statechannels/nitro-rpc-client/src/types";
 export async function fetchFile(
   url: string,
   paymentAmount: number,
-  selectedChannel: string,
+  channelId: string,
   nitroClient: NitroRpcClient
 ): Promise<File> {
-  const voucher = await nitroClient.CreateVoucher(
-    selectedChannel,
-    paymentAmount
-  );
+  const voucher = await nitroClient.CreateVoucher(channelId, paymentAmount);
 
   const response = await fetch(addVoucherToUrl(url, voucher));
 
@@ -22,7 +19,7 @@ export async function fetchFileInChunks(
   chunkSize: number,
   url: string,
   costPerByte: number,
-  selectedChannel: string,
+  channelId: string,
   nitroClient: NitroRpcClient,
   updateProgress: (progress: number) => void
 ): Promise<File> {
@@ -33,7 +30,7 @@ export async function fetchFileInChunks(
     chunkSize - 1,
     url,
     costPerByte,
-    selectedChannel,
+    channelId,
     nitroClient
   );
 
@@ -71,7 +68,7 @@ export async function fetchFileInChunks(
       stop,
       url,
       costPerByte,
-      selectedChannel,
+      channelId,
       nitroClient
     );
 
@@ -87,7 +84,7 @@ export async function fetchFileInChunks(
       stop,
       url,
       costPerByte,
-      selectedChannel,
+      channelId,
       nitroClient
     );
     fileContents.set(data, start);
@@ -104,13 +101,13 @@ async function fetchChunk(
   stop: number,
   url: string,
   costPerByte: number,
-  selectedChannel: string,
+  channelId: string,
   nitroClient: NitroRpcClient
 ): Promise<{ data: Uint8Array; contentLength: number; fileName: string }> {
   const dataLength = stop - start + 1; // +1 because stop is inclusive
   const chunkCost = dataLength * costPerByte;
 
-  const voucher = await nitroClient.CreateVoucher(selectedChannel, chunkCost);
+  const voucher = await nitroClient.CreateVoucher(channelId, chunkCost);
 
   const req = new Request(addVoucherToUrl(url, voucher));
   req.headers.set("Range", `bytes=${start}-${stop}`);
--- a/packages/payment-proxy-client/src/file.ts
+++ b/packages/payment-proxy-client/src/file.ts
@@ -42,6 +42,8 @@ export async function fetchFileInChunks(
   nitroClient: NitroRpcClient,
   updateProgress: (progress: number) => void
 ): Promise<File> {
+  updateProgress(0);
+
   const firstChunk = await fetchFileChunk(
     0,
     chunkSize - 1,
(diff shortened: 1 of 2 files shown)


packages/payment-proxy-client/src/file.ts
```diff
@@ -1,29 +1,158 @@
...
+  chunkSize: number,
+  url: string,
+  costPerByte: number,
+  selectedChannel: string,
+  nitroClient: NitroRpcClient,
+  updateProgress: (progress: number) => void
+): Promise<File> {
+  updateProgress(0);
+
+  const firstChunk = await fetchChunk(
```
lalexgap, line 31: The HTTP range response will include a `Content-Range` header that includes the size of the total file. So after we get the first chunk we know the total length of the file we're requesting, as thus our progress as we download it.

If the size of the first chunk is bigger than the whole file, we just end up receiving the whole first in the first request.

packages/payment-proxy-client/src/file.ts
```diff
@@ -1,29 +1,158 @@
...
+  const { contentLength, fileName } = firstChunk;
+  let remainingContentLength = contentLength - chunkSize;
+
+  console.log(
+    `Fetched the first chunk of the file using a chunk size of ${chunkSize} bytes`
+  );
+
+  console.log(`The file ${fileName} is ${contentLength} bytes in size`);
+
+  const fileContents = new Uint8Array(contentLength);
```
lalexgap, line 49: This is probably fine for demo purposes where files are small. However with massive files we probably would want to avoid storing the whole file in memory before writing it to disk.

packages/payment-proxy-client/src/file.ts
```diff
@@ -38,3 +33,91 @@ function parseFileNameFromUrl(url: string): string {
...
+export async function fetchFileChunk(
+  start: number,
+  stop: number,
+  baseUrl: string,
+  costPerByte: number,
+  selectedChannel: string,
+  nitroClient: NitroRpcClient
+): Promise<{ data: Uint8Array; contentLength: number; fileName: string }> {
+  const dataLength = stop - start + 1; // +1 because stop is inclusive
+  console.log(dataLength);
```
geoknee: ✂️

packages/payment-proxy-client/src/file.ts
```diff
@@ -38,3 +33,91 @@ function parseFileNameFromUrl(url: string): string {
...
+
+  return new File([fileContents], fileName);
+}
+
+export async function fetchFileChunk(
+  start: number,
+  stop: number,
+  baseUrl: string,
+  costPerByte: number,
+  selectedChannel: string,
```
geoknee, line 107: Nit: "selected" is a GUI concept which probably doesn't belong here. Maybe `paymentChannel` would do?

packages/payment-proxy-client/src/file.ts
```diff
@@ -38,3 +33,91 @@ function parseFileNameFromUrl(url: string): string {
...
+): Promise<{ data: Uint8Array; contentLength: number; fileName: string }> {
+  const dataLength = stop - start + 1; // +1 because stop is inclusive
+  console.log(dataLength);
+  const chunkCost = dataLength * costPerByte;
+
+  const voucher = await nitroClient.CreateVoucher(selectedChannel, chunkCost);
+
+  const req = new Request(
+    `${baseUrl}?channelId=${voucher.ChannelId}&amount=${voucher.Amount}&signature=${voucher.Signature}`
+  );
```
geoknee: 1. Consider factoring into a helper function that accepts `voucher` to aid future refactoring. 
2. Should we consider optimising the format of these requests somewhat? i.e. using `c`,`a`,`s` for the variables

packages/payment-proxy-client/src/file.ts
```diff
@@ -38,3 +33,91 @@ function parseFileNameFromUrl(url: string): string {
...
+  const response = await fetch(req);
+  if (!response.body) {
+    throw new Error("Response body is null");
+  }
+
+  if (!response.ok) {
+    throw new Error(`Response status ${response.status}`);
+  }
+  const result = await response.body.getReader().read();
+  console.log(...response.headers);
```
geoknee: ✂️

packages/payment-proxy-client/src/file.ts
```diff
@@ -38,3 +33,91 @@ function parseFileNameFromUrl(url: string): string {
...
+  return {
+    data: result.value || new Uint8Array(),
+    contentLength: parseTotalSizeFromContentRange(
+      response.headers.get("Content-Range") || ""
+    ),
+    fileName: parseFileNameFromUrl(response.url),
+  };
+}
+
+function parseTotalSizeFromContentRange(contentRange: string): number {
```
geoknee: Should we make a unit test for this?

packages/payment-proxy-client/src/file.ts
geoknee: Does this play into the problem of knowing how much the download is going to cost ahead of time?

geoknee (approved):

packages/payment-proxy-client/src/file.ts
geoknee: I see you basically already did 1, so ignore that (commented on an intermediate commit).

packages/payment-proxy-client/src/file.ts
lalexgap: It definitely helps! After we get the first response back we can parse the `Content-Range` header, and know the exact size of the file and how much it will cost (assuming you know the cost per byte being charged). We could probably even expand this by making a request for `0` bytes initially just to get the size of the file, so we know the cost in advance.

All of this relies on the use of the [HTTP range header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Range). If a server doesn't support range headers we won't be able to do any of this, we can just request the whole file in one chunk.

packages/payment-proxy-client/src/file.ts
lalexgap: > Should we consider optimising the format of these requests somewhat? i.e. using c,a,s for the variables

I personally prefer the longer terms as it makes the query string a lot more readable. What would we be optimizing for?

packages/payment-proxy-client/src/file.ts
lalexgap: Right now this is just part of the payment proxy demo, which lacks tests completely, so I think it's OK for now to leave it without a test.

packages/payment-proxy-client/src/file.ts
lalexgap: Updated in https://github.com/statechannels/go-nitro/pull/1510/commits/25ebabf68c212c38ab76ef8a61cfd2656fb6b35d

packages/payment-proxy-client/src/file.ts
geoknee: Just the size of the request. But we don't stand to trim it down by much and it also probably doesn't matter much either 🤷

PR #1514, opened 2023-08-03
Docker image for use in cloud

bitwiseguy:
Create a docker image so that we can deploy a nitro node into the cloud. This node should connect to the Filecoin calibration testnet and can act as a hub. Developers should be able to run local nodes that connect to the hub, which gives them access to a distributed nitro network.


node/engine/chainservice/eth_chainservice.go
```diff
@@ -65,6 +74,15 @@ const MAX_QUERY_BLOCK_RANGE = 2000
 // This has been reduced to 15 seconds to support local devnets with much shorter timeouts.
 const RESUB_INTERVAL = 15 * time.Second
 
+// REQUIRED_BLOCK_CONFIRMATIONS is how many blocks must be mined before an emitted event is processed
+const REQUIRED_BLOCK_CONFIRMATIONS = 0
```
bitwiseguy, line 78: This needs to be > 0 to work on Filecoin Calibration testnet. But it needs to set to 0 for our test suite to pass. We discussed a modification to the `AllocationUpdated` event that would allow the tests to pass when its set > 0.

node/engine/chainservice/eth_chainservice.go
```diff
@@ -271,12 +291,63 @@ out:
...
+			}
+		}
+	}
+}
+
+func (ecs *EthChainService) waitToProcessEvent(chainEvent ethTypes.Log, errorChan chan<- error) {
+	ecs.logger.Debug().Msgf("event emitted in tx %s (topic: %s)", chainEvent.TxHash.Hex(), chainEvent.Topics[0])
+	initialBlock := big.NewInt(int64(chainEvent.BlockNumber))
+
+	ticker := time.NewTicker(BLOCK_QUERY_DELAY_SECONDS * time.Second)
```
bitwiseguy, line 310: In an earlier implementation I created a `ecs.chain.NewHead` subscription that would trigger each time a new block was mined on the network. However, that implementation seemed to be more heavyweight because a single transaction could cause us to generate many equivalent subscriptions. This ticker implementation seems like it will be less of a strain on the node when we have many concurrent events because we just poll once every `BLOCK_QUERY_DELAY_SECONDS`

node/engine/messageservice/p2p-message-service/service.go
```diff
@@ -232,7 +233,10 @@ func (ms *P2PMessageService) msgStreamHandler(stream network.Stream) {
 // Triggered whenever node establishes a connection with a peer
 func (ms *P2PMessageService) sendPeerInfo(recipientId peer.ID, expectResponse bool) {
 	stream, err := ms.p2pHost.NewStream(context.Background(), recipientId, PEER_EXCHANGE_PROTOCOL_ID)
-	ms.checkError(err)
+	if err != nil {
+		ms.logger.Err(err).Msgf("failed to create stream for passing peerInfo with %s", recipientId.String())
```
bitwiseguy, line 237: Instead of panicking, this will just log an error

node/engine/chainservice/eth_chainservice.go
kerzhner: Did we explore the thought of having variable `REQUIRED_BLOCK_CONFIRMATIONS` that depend on the chain the chain service connects to?

node/engine/chainservice/eth_chainservice.go
kerzhner: A couple thoughts:
1. Chain events and transactions will be relatively rare events in a nitro network (compared to state updates).
2. Is there a way to have a single `NewHead` subscription that is shared? It would be great to explore this approach before resorting to polling.

kerzhner (approved):

Makefile
```diff
@@ -1,7 +1,22 @@
...
+	docker build -f docker/cloud/Dockerfile -t go-nitro-cloud .
+
+docker/cloud/push:
+	docker tag go-nitro-cloud:latest registry.digitalocean.com/magmo/go-nitro:latest
+	docker push registry.digitalocean.com/magmo/go-nitro:latest
+
+docker/local/build:
+	docker build -f docker/local/Dockerfile -t go-nitro-local .
+
+docker/local/start:
```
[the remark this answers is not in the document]
bitwiseguy: I built the `docker/local` image to act as a single go-nitro node. The dockerfile used in the `docker/start` commands runs 3 separate go-nitro nodes + Anvil all within the same docker container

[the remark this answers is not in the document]
kerzhner: Got it! We might want to rename `docker/start` for clarity.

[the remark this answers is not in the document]
bitwiseguy: Any suggestions for what to rename `docker/start`? I'm not entirely sure what that container is being used for at the moment so I'm not sure what is a good name for it

node/engine/chainservice/eth_chainservice.go
bitwiseguy: I think the better solution is to update the `AllocationUpdated` event so that we don't have to check chain state at a previous point in time (instead we can just parse the event for all the info we need). That would allow our tests to verify the block confirmation code is working as expected before we get to production. We discussed making the update to that event in the smart contract code, but are waiting to discuss the change with George/Colin next week

node/engine/chainservice/eth_chainservice.go
bitwiseguy: One approach that would only use a single `NewHead` subscription and eliminate polling:

1. Create a single `newBlocks` channel that is always written to whenever a new block is mined
```
newBlocks := make(chan ethTypes.Header)
sub, err := ecs.chain.SubscribeNewHead(ecs.ctx, newBlocks)
```
2. Create a priority queue (aka heap) to keep track of emitted logs that are waiting to be processed. Can use the golang "container/heap" package for this.
3. Modify our `listenForLogs` loop to check for newBlocks. When a new block is received, we can check the queue and asynchronously trigger the event processor for all events that have enough block confirmations.
```
case naChainEvent := <-logs:
	// When we get a log, add it to the heap
	heap.Push(logsHeap, LogEvent{
		Log:   naChainEvent,
		Block: big.NewInt(int64(naChainEvent.BlockNumber)),
	})

case newBlock := <-newBlocks:
	// When we get a new block, check the logs at the top of the queue.
	// If they're ready to be processed, process them and remove them from the queue.
	for logsHeap.Len() > 0 && newBlock.Number().Cmp(new(big.Int).Add(logsHeap[0].Block, big.NewInt(REQUIRED_CONFIRMATION_BLOCKS))) >= 0 {
		chainEvent := heap.Pop(logsHeap).(LogEvent)
		go ecs.dispatchChainEvents(chainEvent)
	}
}
```

docker/cloud/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+FROM debian:bullseye-slim
+RUN apt-get update
+RUN apt-get install -y ca-certificates
+RUN rm -rf /var/lib/apt/lists/*
+WORKDIR /app/
+COPY --from=builder /app/nitro .
+COPY --from=builder /app/config.toml .
+
+EXPOSE 3005 4005
```
geoknee, line 15: @bitwiseguy was it a conscious decision to not expose the gui port 5005?

bitwiseguy: Yes it was a conscious decision, but its one we can certainly change. I was trying to keep the image as small as possible so I excluded all of the react code. The main reason to exclude that code for now is that Digital Ocean free tier allows 1 container registry, 500MB of storage (next tier is $5/month for 5 repos, 5GB)

node/engine/chainservice/eth_chainservice.go
bitwiseguy: Thinking about this a bit more. Updating the `AllocationUpdated` event and mapping different chain IDs to different values for `REQUIRED_BLOCK_CONFIRMATIONS` are related, but are addressing slightly different issues. We could go ahead and make both of those updates

node/engine/chainservice/eth_chainservice.go
kerzhner: Logs are available for the latest run of a PR branch. The logs are under Summary -> Artifacts. For example:
Screenshot 2023-08-14 at 8 28 27 AM

Makefile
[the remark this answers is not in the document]
kerzhner: We could use `docker/start-network`. The container is used to spin up a 3 node network with 2 ledger channels and 1 virtual channel. The container allows a developer to try out a nitro network without setting up their machine for nitro.

docker/cloud/Dockerfile
geoknee: Ok good to know. Do we know how heavy the react part would be? Might be nice to have that available as a quick-and-dirty dashboard for the cloud deployment.

My hunch is that it adds a negligible amount of weight to the image.

geoknee:
```
~/s/go-nitro ❯❯❯ go build -o gonitro && go build -o gonitro-with-ui -tags embed_ui &&  ls -l gonitro*                                                                                                              main ✭ ◼
-rwxr-xr-x  1 georgeknee  staff  59115138 14 Aug 17:59 gonitro
-rwxr-xr-x  1 georgeknee  staff  59716818 14 Aug 17:59 gonitro-with-ui
```

Makefile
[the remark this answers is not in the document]
bitwiseguy: Done in https://github.com/statechannels/go-nitro/pull/1514/commits/9b7a4776438b75edb71c6c5e201b6c4124f35faf

docker/cloud/README.md
```diff
@@ -0,0 +1,42 @@
...
+Build the image:
+
+```
+make docker/cloud/build
+```
+
+## Push image to docker registry
+
+```
+docker login -u <api_key> -p <api_key> registry.digitalocean.com
```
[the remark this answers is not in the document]
bitwiseguy: AFAIK this is the simplest I could do for the make command:

makefile:
```
docker/login:
	docker login -u $(APIKEY) -p $(APIKEY) registry.digitalocean.com
```

Usage:
```
$ make docker/login APIKEY=xxxxx
```

Is that what you had in mind or is there a better way?

node/engine/chainservice/eth_chainservice.go
bitwiseguy: Created new issue to address variable `REQUIRED_BLOCK_CONFIRMATIONS` in a future PR: https://github.com/statechannels/go-nitro/issues/1535

docker/cloud/Dockerfile
bitwiseguy: I created a new issue to make this change. Since this PR has been dangling around for awhile and has grown quite large, I'd like to get it merged and address with a follow-up PR. https://github.com/statechannels/go-nitro/issues/1536

PR #1515, opened 2023-08-04
Refine dht peer discovery

bitwiseguy:
Makes the following changes:

1. Remove timestamp check from dht custom Validate
2. Add check in dht custom Validate to compare record key to `SCAddr` in record value
3. Remove `sendPeerInfo` message exchange when using dht. Instead pull that info directly from dht


node/engine/messageservice/p2p-message-service/service.go
```diff
@@ -132,6 +134,9 @@ func (ms *P2PMessageService) setupMdns() {
 	ms.mdns = mdns.NewMdnsService(ms.p2pHost, "", ms)
 	err := ms.mdns.Start()
 	ms.checkError(err)
+
+	ms.InitComplete <- true
```
geoknee, line 138: Did you consider simply closing the `chan` instead?

node_test/helpers_test.go
```diff
@@ -55,6 +55,7 @@ func waitForPeerInfoExchange(services ...*p2pms.P2PMessageService) {
 			peerInfo := <-s.PeerInfoReceived()
 			fmt.Printf("Service num: %d, peer num: %d, peerInfo: %v\n", sNum, i, peerInfo)
 		}
+		<-s.InitComplete
```
geoknee, line 58: I wonder if this should be encapsulated into a method on `s`? e.g. `WaitForInitComplete`.

Do we anticipate using this in production (and not just test) code?

node/engine/messageservice/p2p-message-service/service.go
bitwiseguy: I did not consider it but thats probably a better implementation in case there are multiple listeners on that channel. I will make that change

node_test/helpers_test.go
geoknee: I guess it's not a huge difference. Advantages are that it is the way the chan is being used is made more explicit, and that it may be more maintainable (in case e.g.  we end up needing to wait on anything else). 

Your call if it is worth it, here.

node/engine/messageservice/p2p-message-service/service.go
```diff
@@ -53,8 +53,9 @@ const (
 
 // P2PMessageService is a rudimentary message service that uses TCP to send and receive messages.
 type P2PMessageService struct {
-	toEngine chan protocols.Message // for forwarding processed messages to the engine
-	peers    *safesync.Map[peer.ID]
+	InitComplete chan bool
```
lalexgap, line 56: This might be nitpicking but we could use a private channel and expose it via a public getter. The advantage of that is we can expose `InitComplete` as a one-way read only chan to public consumers.  [Here's an example from the chain service.](https://github.com/statechannels/go-nitro/blob/18d7bd40d9ce4af0abaa2317d0e5a7ebb3b61bfe/node/engine/chainservice/eth_chainservice.go#L305)

node_test/helpers_test.go
lalexgap: Ha, I made a similar [comment](https://github.com/statechannels/go-nitro/pull/1515/files#r1293895933) before noticing this one. 

Another advantage of using a function is that it let's us specify the direction of the channel, which adds some guardrails. Right now anyone can read or **send** to `s.InitComplete`. If we used a function to return a one way chan, we ensure that only the message service can send on the chan.

node/engine/messageservice/p2p-message-service/service.go
```diff
@@ -80,11 +81,12 @@ func NewMessageService(ip string, port int, me types.Address, pk []byte, useMdns
 	logging.ConfigureZeroLogger()
 
 	ms := &P2PMessageService{
-		toEngine:    make(chan protocols.Message, BUFFER_SIZE),
-		newPeerInfo: make(chan basicPeerInfo, BUFFER_SIZE),
-		peers:       &safesync.Map[peer.ID]{},
-		me:          me,
-		logger:      logging.WithAddress(zerolog.New(logWriter).With().Timestamp(), &me).Caller().Logger(),
+		InitComplete: make(chan bool, 1),
```
lalexgap, line 84: Instead of a `bool` chan could we use a empty struct chan? I think we're just using the `chan` to signal, so we don't care about the value being sent.

I think this helps with understand-ability since it makes it clear that we don't care about the actual value and it doesn't have any special meaning(IE: Will `InitComplete` return `false` if init fails in some way?)

```suggestion
		InitComplete:  make(chan struct{}, 1),
```

node/engine/messageservice/p2p-message-service/service.go
```diff
@@ -146,51 +151,44 @@ func (ms *P2PMessageService) setupDht(bootPeers []string) {
...
 
-	expectedPeers := len(bootPeers)
-	if expectedPeers > 0 {
-		// Add bootpeers and wait for connections before proceeding
-		ms.addBootPeers(bootPeers)
-		ms.logger.Info().Msgf("waiting for %d bootpeer connections", expectedPeers)
+	err = ms.dht.Bootstrap(ctx) // Sends FIND_NODE queries periodically to populate dht routing table
+	ms.checkError(err)
+
+	// Must wait until dht RoutingTable has an entry before adding custom dht record
```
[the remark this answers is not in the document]
bitwiseguy: This is a restriction enforced by the libp2p library. When we try to put a value into the DHT, the node is not storing it locally but rather telling other peers to store it. The key-value pairs are stored on nodes with IDs closest to the key. If the RoutingTable is empty, the node has no peers to propagate this information to.

It seems like the node could just store the record locally if the RoutingTable is empty, but there's no configuration on the dht to allow that behavior AFAIK

node/engine/messageservice/p2p-message-service/service.go
bitwiseguy: Done in https://github.com/statechannels/go-nitro/pull/1515/commits/ecfd7dfb5ea14f594288e01681f05f3e50c71f76

node/engine/messageservice/p2p-message-service/service.go
bitwiseguy: Done in https://github.com/statechannels/go-nitro/pull/1515/commits/ecfd7dfb5ea14f594288e01681f05f3e50c71f76

node/engine/messageservice/p2p-message-service/service.go
```diff
@@ -146,51 +151,49 @@ func (ms *P2PMessageService) setupDht(bootPeers []string) {
...
-	}
-
-	err = ms.dht.Bootstrap(ctx) // Runs periodically to maintain a healthy routing table
-	ms.checkError(err)
+	}()
 
 	ms.logger.Info().Msgf("DHT setup complete")
 }
 
+// EventFeed returns the out chan, and narrows the type so that external consumers may only receive on it.
```
lalexgap, line 188:
```suggestion
// InitComplete returns a chan that gets closed once the message service is initialized.
```

lalexgap (approved):

node/engine/messageservice/p2p-message-service/service.go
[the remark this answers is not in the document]
kerzhner: Got it, thanks for the explanation. It would be great to include some of this context in the comment.

node/engine/messageservice/p2p-message-service/service.go
bitwiseguy: Done in https://github.com/statechannels/go-nitro/pull/1515/commits/b0c076b83b6f8f860bfd52ee7243b605ed734afd

PR #1517, opened 2023-08-08
Add EventQueue and new block sub to wait for block confirmations

bitwiseguy:


node/engine/chainservice/eth_chainservice.go
```diff
@@ -263,113 +265,102 @@ func (ecs *EthChainService) dispatchChainEvents(logs []ethTypes.Log) error {
...
-		confirmations := int(header.Number.Int64() - initialBlock.Int64())
-		ecs.logger.Trace().Msgf("event block confirmations: %d (goal: %d)", confirmations, REQUIRED_BLOCK_CONFIRMATIONS)
-
-		if confirmations >= REQUIRED_BLOCK_CONFIRMATIONS {
-			err := ecs.dispatchChainEvents([]ethTypes.Log{chainEvent})
-			if err != nil {
-				return fmt.Errorf("error in dispatchChainEvents: %w", err)
+		case newBlock := <-newBlockChan:
+			ecs.logger.Debug().Msgf("detected new block: %d", newBlock.Number.Uint64())
+			// When we get a new block, check the events at the top of the queue.
```
kerzhner, line 328: It doesn't seem like this takes into account uncle blocks. Probably ok for now, but we should file an issue for future work.

node_test/helpers_test.go
```diff
@@ -185,20 +185,39 @@ func setupLedgerChannel(t *testing.T, alpha node.Node, beta node.Node, asset com
...
 
-func closeLedgerChannel(t *testing.T, alpha node.Node, beta node.Node, channelId types.Destination) {
+func closeLedgerChannel(t *testing.T, sim chainservice.SimulatedChain, alpha node.Node, beta node.Node, channelId types.Destination) {
 	response, err := alpha.CloseLedgerChannel(channelId)
 	if err != nil {
 		t.Fatal(err)
 	}
 
+	if sim != nil {
+		time.Sleep(100 * time.Millisecond)
```
kerzhner, line 211: What is the reason for this delay?

node/engine/chainservice/eth_chainservice.go
bitwiseguy: Added new issue here: https://github.com/statechannels/go-nitro/issues/1525

lalexgap: I think with proof of stake [uncle blocks aren't a thing](https://ethereum.stackexchange.com/questions/139390/are-uncle-ommer-blocks-still-possible-after-the-merge).  However it probably still makes sense to consider this so we can run against chains where uncle blocks are a thing.

node/engine/chainservice/eth_chainservice.go
```diff
@@ -263,113 +265,102 @@ func (ecs *EthChainService) dispatchChainEvents(logs []ethTypes.Log) error {
...
-			sub, sErr = ecs.chain.SubscribeFilterLogs(ecs.ctx, query, logs)
+			eventSub, sErr = ecs.chain.SubscribeFilterLogs(ecs.ctx, eventQuery, eventChan)
 			if sErr != nil {
 				errorChan <- fmt.Errorf("subscribeFilterLogs failed on resubscribe: %w", err)
 				break out
 			}
-			ecs.logger.Trace().Msg("resubscribed to filtered logs")
+			ecs.logger.Trace().Msg("resubscribed to filtered event logs")
+
+		case err := <-newBlockSub.Err():
```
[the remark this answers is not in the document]
bitwiseguy: Yes that's a good call. Done in https://github.com/statechannels/go-nitro/pull/1517/commits/6decae21bb6cb7917a5bff2f55173822b6f036ee

node_test/helpers_test.go
bitwiseguy: This does feel prone to future errors because anytime we send a transaction when using the Simulated Backend we may need to add this delay. Perhaps it would be better to just add the delay inside the [`SendTransaction`](https://github.com/statechannels/go-nitro/blob/newBlock-listener/node/engine/chainservice/simulated_backend_chainservice.go#L86) method directly so we don't have to do it externally?

bitwiseguy: Made the change I mentioned above in this commit: https://github.com/statechannels/go-nitro/pull/1517/commits/a1b1afd1c3cfbd5ab7ea6aa04c2790aff0514d5c

PR #1522, opened 2023-08-14
Handle errors in p2p message service

### LINKED ISSUE #1474: Implement error handling in the p2p message service
Completely outside the scope of this PR but we should probably consider some kind of proper error handling for the p2p message service. Right now we just `panic`. I'll create an issue for this.

_Originally posted by @lalexgap in https://github.com/statechannels/go-nitro/pull/1462#discussion_r1268638797_

geoknee:
Closes #1474

### DIFF
--- a/node/engine/engine.go
+++ b/node/engine/engine.go
@@ -555,7 +555,10 @@ func (e *Engine) sendMessages(msgs []protocols.Message) {
 		message.From = *e.store.GetAddress()
 		e.logMessage(message, Outgoing)
 		e.recordMessageMetrics(message)
-		e.msg.Send(message)
+		err := e.msg.Send(message)
+		if err != nil {
+			e.logger.Err(err)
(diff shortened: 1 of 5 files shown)


lalexgap (approved):
Beyond the scope of this PR but it would be nice if there was a way we could report errors to the client when they happen. Right now we log the error but if you're using the RPC client you probably won't even know the error happened.

geoknee:
> Beyond the scope of this PR but it would be nice if there was a way we could report errors to the client when they happen. Right now we log the error but if you're using the RPC client you probably won't even know the error happened.

Good point. What do you think would be the correct behaviour? Could / should we wire it up so that the RPC client can stream the logs from the node?

It seems that the RPC client (or "library" client) shouldn't get an error (in the usual sense) when say a badly formatted message is received by the node?

PR #1523, opened 2023-08-14
Replace `handleEngineEvent` goroutine and chan with callback

### LINKED ISSUE #1340: Consider refactoring `handleEngineEvents` into a callback (not a goroutine)
Following the principle of "don't start a goroutine unless you really need to". The body of the function isn't really asynchronous, so is not a great candidate for a goroutine??

geoknee:
Closes #1340

### DIFF
--- a/node/engine/engine.go
+++ b/node/engine/engine.go
@@ -65,7 +65,7 @@ type Engine struct {
 	fromMsg    <-chan protocols.Message
 	fromLedger chan consensus_channel.Proposal
 
-	toApi chan EngineEvent
+	eventHandler func(EngineEvent)
 
 	msg   messageservice.MessageService
 	chain chainservice.ChainService
@@ -129,7 +129,7 @@ type CompletedObjectiveEvent struct {
 type Response struct{}
 
 // NewEngine is the constructor for an Engine
-func New(vm *payments.VoucherManager, msg messageservice.MessageService, chain chainservice.ChainService, store store.Store, logDestination io.Writer, policymaker PolicyMaker, metricsApi MetricsApi) Engine {
+func New(vm *payments.VoucherManager, msg messageservice.MessageService, chain chainservice.ChainService, store store.Store, logDestination io.Writer, policymaker PolicyMaker, eventHandler func(EngineEvent), metricsApi MetricsApi) Engine {
 	e := Engine{}
 
 	e.store = store
@@ -145,7 +145,7 @@ func New(vm *payments.VoucherManager, msg messageservice.MessageService, chain c
 	e.chain = chain
 	e.msg = msg
 
-	e.toApi = make(chan EngineEvent, 100)
+	e.eventHandler = eventHandler
 
 	logging.ConfigureZeroLogger()
 	e.logger = logging.WithAddress(zerolog.New(logDestination).With().Timestamp(), e.store.GetAddress()).Caller().Logger()
@@ -172,10 +172,6 @@ func New(vm *payments.VoucherManager, msg messageservice.MessageService, chain c
 	return e
 }
 
-func (e *Engine) ToApi() <-chan EngineEvent {
-	return e.toApi
-}
-
 func (e *Engine) Close() error {
 	e.cancel()
 	e.wg.Wait()
@@ -183,7 +179,6 @@ func (e *Engine) Close() error {
 		return err
 	}
 
-	close(e.toApi)
 	return e.chain.Close()
 }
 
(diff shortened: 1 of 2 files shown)


node/engine/engine.go
```diff
@@ -227,7 +222,7 @@ func (e *Engine) run(ctx context.Context) {
 				e.logger.Printf("Objective %s is complete & returned to API", obj.Id())
 				e.metrics.RecordObjectiveCompleted(obj.Id())
 			}
-			e.toApi <- res
+			e.eventHandler(res)
```
lalexgap, line 225: I think this means that the engine `run` loop now blocks on the event handler, which means there might be a slight performance hit with this change? `handleEngineEvent` doesn't do too much, so it's quite possible that performance hit is negligible. 

It's too bad we no longer have the testground test, so we could see if there was any noticeable performance difference.

node/node.go
```diff
@@ -76,60 +71,40 @@ func New(messageService messageservice.MessageService, chainservice chainservice
...
-// It parses events from the ToApi chan and then dispatches events to the necessary node chan.
-func (n *Node) handleEngineEvents(ctx context.Context) {
-	for {
+// handleEngineEvents dispatches events to the necessary node chan.
+func (n *Node) handleEngineEvent(update engine.EngineEvent) {
+	for _, completed := range update.CompletedObjectives {
+		d, _ := n.completedObjectives.LoadOrStore(string(completed.Id()), make(chan struct{}))
+		close(d)
+
+		// use a nonblocking send to the RPC Client in case no one is listening
```
lalexgap, line 83: Outside the scope of this PR, but a helper function to handle the nonblocking send might make things slightly neater.

lalexgap (approved):

node/engine/engine.go
geoknee: Yes too bad we don't have testground. Maybe we can measure performance in another way for this change. 

My hunch is that there will not be a negative effect on performance. 

Take the following thought experiment: take as much synchronous code as possible in the engine and run it in goroutines. Would that improve performance? I think only if something were blocking on I/O or network. Or (maybe) heavy crypto operations, but then we would likely need to be using multiple CPUs to notice a significant difference. If putting very fast operations in a goroutine won't improve performance, removing the goroutine as this PR does shouldn't have a negative impact.

geoknee: Quick and dirty performance test:
```
~/s/go-nitro ❯❯❯ go test ./node_test -run TestComplexIntegrationScenario                                                                                                                                         ✘ 1 main ✭
ok  	github.com/statechannels/go-nitro/node_test	0.665s
~/s/go-nitro ❯❯❯ go test ./node_test -run TestComplexIntegrationScenario                                                                                                                                             main ✭
ok  	github.com/statechannels/go-nitro/node_test	0.506s
~/s/go-nitro ❯❯❯ go test ./node_test -run TestComplexIntegrationScenario                                                                                                                                             main ✭
ok  	github.com/statechannels/go-nitro/node_test	0.508s
~/s/go-nitro ❯❯❯ go test ./node_test -run TestComplexIntegrationScenario                                                                                                                                             main ✭
ok  	github.com/statechannels/go-nitro/node_test	0.500s
~/s/go-nitro ❯❯❯ go test ./node_test -run TestComplexIntegrationScenario                                                                                                                                             main ✭
ok  	github.com/statechannels/go-nitro/node_test	0.513s
                                                                                                                                                     ✘ 1 main ✭
```
```
~/s/go-nitro ❯❯❯ git checkout engine-event-handler          
Switched to branch 'engine-event-handler'
Your branch is up to date with 'origin/engine-event-handler'.
~/s/go-nitro ❯❯❯ go test ./node_test -run TestComplexIntegrationScenario                                                                                                                             engine-event-handler ✭
ok  	github.com/statechannels/go-nitro/node_test	0.581s
~/s/go-nitro ❯❯❯ go test ./node_test -run TestComplexIntegrationScenario                                                                                                                             engine-event-handler ✭
ok  	github.com/statechannels/go-nitro/node_test	0.496s
~/s/go-nitro ❯❯❯ go test ./node_test -run TestComplexIntegrationScenario                                                                                                                             engine-event-handler ✭
ok  	github.com/statechannels/go-nitro/node_test	0.496s
~/s/go-nitro ❯❯❯ go test ./node_test -run TestComplexIntegrationScenario                                                                                                                             engine-event-handler ✭
ok  	github.com/statechannels/go-nitro/node_test	0.506s
~/s/go-nitro ❯❯❯ go test ./node_test -run TestComplexIntegrationScenario                                                                                                                             engine-event-handler ✭
ok  	github.com/statechannels/go-nitro/node_test	0.488s
~/s/go-nitro ❯❯❯
```

geoknee: > I might be nitpicking but I don't think you need multiple CPUs, just multiple cores, which is pretty common. I think it would be rare to encounter a cpu with only a single core these days.

Yes I meant multiple processors / cores / logical CPUs. You're right that most machines have multiple cores and it looks like these can all be used by default since Go 1.15 https://medium.com/a-journey-with-go/go-gomaxprocs-live-updates-407ad08624e1.

PR #1534, opened 2023-08-15
Remove mdns peer discovery option from the p2p message service

### LINKED ISSUE #1507: Remove mDNS from p2p message service
Now that we're relying more on DHT it probably makes sense to use that as a the sole peer discovery mechanism. `mdns` does not work when nodes are running on external network so it's not suitable for a production setting. Right now most of are tests use `mdns` meaning we could not be adequately testing our DHT solution.

We should remove the `mdns` peer discovery and rely solely on DHT in our message service. It will force us to deal with any issues or paint points around using DHT.

lalexgap:
Fixes #1507 

This removes the option of `mdns` from the message service. This helps simplify the p2p message service and also forces us to rely on DHT, which is what we expect to run in production.

### DIFF
--- a/cmd/start-rpc-servers/main.go
+++ b/cmd/start-rpc-servers/main.go
@@ -38,10 +38,7 @@ const (
 	gray    color = "[90m"
 )
 
-var (
-	participants     = []participant{alice, bob, irene, ivan}
-	participantColor = map[participant]color{alice: blue, irene: green, ivan: cyan, bob: yellow}
-)
+var participantColor = map[participant]color{alice: blue, irene: green, ivan: cyan, bob: yellow}
 
 const (
 	FUNDED_TEST_PK  = "ac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80"
--- a/cmd/start-rpc-servers/main.go
+++ b/cmd/start-rpc-servers/main.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"os"
 	"os/exec"
+	"time"
 
 	"github.com/statechannels/go-nitro/cmd/utils"
 	"github.com/statechannels/go-nitro/internal/chain"
@@ -120,7 +121,21 @@ func main() {
 			}
 
 			hostUI := cCtx.Bool(HOST_UI)
-			for _, p := range participants {
+
+			for _, p := range []participant{ivan} {
+				client, err := setupRPCServer(p, participantColor[p], naAddress, vpaAddress, caAddress, chainUrl, chainAuthToken, dataFolder, hostUI)
+				if err != nil {
+					utils.StopCommands(running...)
+					panic(err)
+				}
+				running = append(running, client)
+			}
+
+			// TODO: There's not an easy way to detect when the server is up and running so we just sleep for now
+			time.Sleep(5 * time.Second)
+
+			for _, p := range []participant{alice, bob, irene} {
+
 				client, err := setupRPCServer(p, participantColor[p], naAddress, vpaAddress, caAddress, chainUrl, chainAuthToken, dataFolder, hostUI)
 				if err != nil {
 					utils.StopCommands(running...)
--- a/cmd/start-rpc-servers/main.go
+++ b/cmd/start-rpc-servers/main.go
@@ -121,14 +121,13 @@ func main() {
 
 			hostUI := cCtx.Bool(HOST_UI)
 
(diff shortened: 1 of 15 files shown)


cmd/test-configs/bob.toml
```diff
@@ -2,19 +2,18 @@
 
 usedurablestore = true
 
+guiport = 5007
```
lalexgap, line 5: I think this formatting was done by this [toml extension](https://marketplace.visualstudio.com/items?itemName=be5invis.toml) in vs code. Seems pretty harmless so I'm going to include it.

node/engine/messageservice/p2p-message-service/service.go
```diff
@@ -41,8 +39,8 @@ const (
 	BUFFER_SIZE                              = 1_000
 	NUM_CONNECT_ATTEMPTS                     = 10
 	RETRY_SLEEP_DURATION                     = 5 * time.Second
-	PEER_EXCHANGE_SLEEP_DURATION             = 10 * time.Second // how often we attempt FindPeers
-	BOOTSTRAP_SLEEP_DURATION                 = 1 * time.Second  // how often we check for bootpeers in Peerstore
+	PEER_EXCHANGE_SLEEP_DURATION             = 10 * time.Second       // how often we attempt FindPeers
+	BOOTSTRAP_SLEEP_DURATION                 = 100 * time.Millisecond // how often we check for bootpeers in Peerstore
```
lalexgap, line 43: This was shortened to avoid timeouts in some of our integration tests.

cmd/test-configs/irene.toml
```diff
@@ -2,18 +2,18 @@
...
 chainpk = "7c852118294e51e653712a81e05800f419141751be58f605c371e15141b007a6"
 
-chainurl = "ws://127.0.0.1:8545"
 chainauthtoken = ""
+chainurl = "ws://127.0.0.1:8545"
 
-bootpeers = ""
-usemdns = true
+# bootpeer = Ivan
+bootpeers = "/ip4/127.0.0.1/tcp/3008/p2p/16Uiu2HAm1hgN2MkrhGen8JPrBBYyXACbZtfqJmraN53XiHYCeoFi"
```
lalexgap, line 19: Previously Irene was no one's boot peer and didn't have a boot peer to connect to; so she was left out. Poor Irene 😢

cmd/start-rpc-servers/main.go
```diff
@@ -120,7 +120,24 @@ func main() {
 			}
 
 			hostUI := cCtx.Bool(HOST_UI)
-			for _, p := range participants {
+
+			for _, p := range []participant{ivan} {
```
bitwiseguy, line 124: Since there is only a single element in this `participant` array, could we remove this `for` loop?

node_test/integration_test.go
```diff
@@ -96,7 +76,7 @@ func RunIntegrationTestCase(tc TestCase, t *testing.T) {
 		intermediaries := make([]node.Node, 0)
 		bootPeers := make([]string, 0)
 		for _, intermediary := range tc.Participants[2:] {
```
bitwiseguy, line 78: This seems pretty brittle since we have to put the intermediaries at the end of the Participants array. Would it be cleaner to add another field in our `TestCase` struct to separate the Intermediaries like this?
```
TestCase{
  ...
  Participants: []TestParticipant{
    {StoreType: DurableStore, Actor: testactors.Alice},
    {StoreType: DurableStore, Actor: testactors.Bob},
  },
  Intermediaries: []TestParticpant{
    {StoreType: DurableStore, Actor: testactors.Irene},
    {StoreType: DurableStore, Actor: testactors.Ivan},
  },
}
```
Then instead of having to know where the intermediaries fall in the `tc.Participants` array, we could just loop over the separate `tc.Intermediaries` array:
```
		for _, intermediary := range tc.Intermediaries {
```

bitwiseguy: I know this isn't really part of this PR, just caught my eye when reviewing

bitwiseguy (approved):

node_test/integration_test.go
lalexgap: That's a good idea! Although we do have some test case validation based on the [amount of participants](https://github.com/statechannels/go-nitro/blob/0c9ccee6ddaddc0ea977657c7494024d4702a147/node_test/types_test.go#L68) so we could also consider a flag on the `TestParticipant` that can be set instead of a seperate slice.

PR #1537, opened 2023-08-15
Use a mutex to protect event queue

kerzhner:
Currently, we have a goroutine that adds to the event queue. We have another goroutine that reads from the event queue. This change:
- Protects the event queue with a mutex.
- Refactors notification handling so that both new events and new blocks trigger queue processing. This allows us to remove the simulated chain sleep that was added to make it more likely that an event update arrives before a new block update.

### DIFF
--- a/node/engine/chainservice/eth_chainservice.go
+++ b/node/engine/chainservice/eth_chainservice.go
@@ -48,6 +48,7 @@ type ethChain interface {
 	ChainID(ctx context.Context) (*big.Int, error)
 }
 
+// eventTracker holds on to events in memory and dispatches an event after required number of confirmations
 type eventTracker struct {
 	latestBlockNum uint64
 	events         EventQueue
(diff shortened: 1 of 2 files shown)


node/engine/chainservice/eth_chainservice.go
```diff
@@ -335,22 +344,42 @@ out:
...
+
+			ecs.logger.Debug().Msgf("detected new block: %d", newBlockNum)
 		}
 	}
 }
 
+// updateEventTracker accepts a new block number and/or new event and dispatches a chain event if there are enough block confirmations
+func (ecs *EthChainService) updateEventTracker(errorChan chan<- error, blockNumber *uint64, chainEvent *ethTypes.Log) {
+	// lock the mutex for the shortest amount of time. The mutex only need to be locked to update the eventTracker data structure
+	ecs.eventTracker.mu.Lock()
```
lalexgap, line 359: Instead of protecting calls to the `EventQueue` with a mutex in the chain service, could the `EventQueue` use mutexs (or some other `sync` feature) internally to protect against concurrent calls to `Pop`/`Push`? Kind of like [our safesync.Map](https://github.com/statechannels/go-nitro/blob/638e0447c56448cdcfe6fc1ff2a8b2f255963964/internal/safesync/safesync.go#L7)

kerzhner: Definitely an option! One nice feature of the current implementation is that the mutex also protects latest block variable read and writes.

bitwiseguy (approved):

PR #1538, opened 2023-08-16
Upgrade abigen

### LINKED ISSUE #1532: Upgrade geth/abigen version on CI
`abigen` is a tool within the `geth` suite that we use to automatically generate smart contract bindings used throughout go-nitro. Currently we are using v1.10.8 in the CI. The latest is v1.12.2. We should consider updating to the latest version so that we have full functionality and the latest security patches. Additionally, go-nitro relies on [v1.12.0](https://github.com/statechannels/go-nitro/blob/9e6d4281de72f7a427efd23aa6d06c2df03e5dc9/go.mod#L6)  of the go-etheurem library so it probably makes sense for our CI to use the geth tools for that version.

Along with upgrading abigen used by the CI, we will need to download v1.12.2 locally and run the `generate-adjudicator-bindings.sh` script so that the output matches what CI produced during its build checks.

lalexgap:
Fixes #1532 

⚠️⚠️⚠️ If you don't upgrade your local version of `abigen` to `1.12.2` you will generate different bindings when running `./generate-adjudicator-bindings.sh` . [Instructions on how to update geth (which `abigen` is part of)](https://geth.ethereum.org/docs/getting-started/installing-geth) ⚠️⚠️⚠️

This PR updates the version of `abigen` we use to `1.12.2`, which is the latest version. It also upgrades the go-etheurem library to `1.12.2` so we're using the same version of the library and `abigen`.


geoknee:
> ⚠️⚠️⚠️ If you don't upgrade your local version of `abigen` to `1.12.2` you will generate different bindings when running `./generate-adjudicator-bindings.sh` . [Instructions on how to update geth (which `abigen` is part of)](https://geth.ethereum.org/docs/getting-started/installing-geth) ⚠️⚠️⚠️

The amount of warning triangles caused me to dig into this #1539

.github/workflows/bindings-check.yml
```diff
@@ -4,7 +4,12 @@ on:
   push:
     branches: [main]
   pull_request:
-    paths: ["nitro-protocol/contracts/**", generate-adjudicator-bindings.sh]
+    paths:
+      [
+        "nitro-protocol/contracts/**",
+        generate-adjudicator-bindings.sh,
+        node/engine/chainservice/adjudicator/**,
```
bitwiseguy, line 11: What's the reason for adding this single `chainservice/adjudicator` directory but not the other similar directories for other contracts?
```
chainservice/consensusapp
chainservice/erc20
chainservice/virtualpaymentapp
```

bitwiseguy (approved):

PR #1539, opened 2023-08-16
Use local `abigen` (dependency of `go-nitro`) to generate bindings

geoknee:
In our github action to regenerate and validate bindings, we install abigen from a hardcoded URL. These means we are tracking another dependency on geth (which abigen is part of) alongside the one specified in our `go.mod`.

This PR changes the github action to use the version specified in `go.mod`. This means developers do not need to worry about another way to install `abigen`.

### DIFF
--- a/generate-adjudicator-bindings.sh
+++ b/generate-adjudicator-bindings.sh
@@ -31,7 +31,6 @@ runAbigen() {
     --bin=$TEMP_DIR/${1}.bin \
     --pkg=${1} \
     --out=$GONITRO_DIR/node/engine/chainservice/${2}/${1}.go 
-  cd ~-
 }
 
 echo "Using abigen from $GETH_DIR..."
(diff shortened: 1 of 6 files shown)


generate-adjudicator-bindings.sh
```diff
@@ -1,29 +1,41 @@
 set -e
 
+GONITRO_DIR=$(pwd)
+NITRO_PROTOCOL_DIR=$GONITRO_DIR/nitro-protocol
+TEMP_DIR=$NITRO_PROTOCOL_DIR/tmp-build
+GETH_DIR=$(go list -m -f '{{.Dir}}' github.com/ethereum/go-ethereum)
```
geoknee, line 6: This gets us the path to the version specified in `go.mod`.

bitwiseguy (approved):

Issue #1531, opened 2023-08-15
Enrich the stored "state" of channels


NiloCK:
Existing channel struct has a member only for on chain funding

```
type Channel struct {
    // ...
    OnChainFunding types.Funds
    // ...
}
```

But this leaves out whatever data was used by the chain to generate the currently stored channel fingerprint.

Probably we want all of the following:
```
type Channel struct {
    // ...
    OnChainFunding types.Funds
    OnChainState   state.VariablePart
    OnChainTurnNum int64
    OnChainOutcome outcome.Exit     
    // ...
}
```

or for the on-chain data to be encapsulated into its own new struct.

The stored data for channel `0x123` should update whenever a chain event is observed for `0x123`. At least one utility function for performing this update is [already written](https://github.com/statechannels/go-nitro/blob/18d7bd40d9ce4af0abaa2317d0e5a7ebb3b61bfe/channel/state/outcome/transfer.go#L10).

**Context:**

`NitroAdjudicator` emits channel updated events which aim to give the smallest necessary information for observers to compute the updated state. It assumes that observers know the state prior to the update.

**Related:**

Possibly related to #1525 and the block confirmations work done in #1517.

geoknee:
> The stored data for channel `0x123` should update whenever a chain event is observed for `0x123`

This poses a bit of a challenge for our current architecture. I believe chain events are only currently consumed by objectives. 

chain -> objective -> channel

This means we can fail to handle the event if there is not currently an objective running governing a particular channel. That may be unlikely but it doesn't seem impossible.

PR #1540, opened 2023-08-16
Allow `Channels` to handle `ChainEvents` without help of `Objectives`

geoknee:
This has the advantage of centralising our logic around chain event handling / block number checks and so on. 

It also means that we don't rely on a particular objective being active / approved / to catch information from the chain. 

Towards #1531

### DIFF
--- a/channel/channel.go
+++ b/channel/channel.go
@@ -327,7 +327,7 @@ func (c *Channel) SignAndAddState(s state.State, sk *[]byte) (state.SignedState,
 
 // UpdateWithChainEvent mutates the receiver if provided with a "new" chain event (with a greater block number than previously seen)
 func (c *Channel) UpdateWithChainEvent(event chainservice.Event) (*Channel, error) {
-	if event.BlockNum() <= c.latestBlockNumber {
+	if event.BlockNum() < c.latestBlockNumber {
 		return c, nil // ignore stale information TODO: is this reorg safe?
 	}
 	c.latestBlockNumber = event.BlockNum()
--- a/channel/channel.go
+++ b/channel/channel.go
@@ -9,6 +9,7 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/statechannels/go-nitro/channel/state"
 	"github.com/statechannels/go-nitro/channel/state/outcome"
+	"github.com/statechannels/go-nitro/node/engine/chainservice"
 	"github.com/statechannels/go-nitro/types"
 )
 
@@ -17,7 +18,8 @@ type Channel struct {
 	Id      types.Destination
 	MyIndex uint
 
-	OnChainFunding types.Funds
+	OnChainFunding    types.Funds
+	latestBlockNumber uint64 // the latest block number we've seen
 
 	state.FixedPart
 	// Support []uint64 // TODO: this property will be important, and allow the Channel to store the necessary data to close out the channel on chain
@@ -128,6 +130,7 @@ func (c *Channel) Clone() *Channel {
 		d.SignedStateForTurnNum[i] = ss.Clone()
 	}
 	d.OnChainFunding = c.OnChainFunding.Clone()
+	d.latestBlockNumber = c.latestBlockNumber
 	d.FixedPart = c.FixedPart.Clone()
 	return d
 }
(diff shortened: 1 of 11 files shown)


protocols/directfund/directfund.go
```diff
@@ -276,21 +275,13 @@ func (o *Objective) Update(p protocols.ObjectivePayload) (protocols.Objective, e
...
-	if de.BlockNum > updated.latestBlockNumber {
-		updated.C.OnChainFunding[de.Asset] = de.NowHeld
-		updated.latestBlockNumber = de.BlockNum
+	_, err := updated.C.UpdateWithChainEvent(event)
+	if err != nil {
+		return &Objective{}, err
 	}
-
 	return &updated, nil
 }
```
geoknee, line 287: There's an argument for removing this entirely. It is useful for unit testing though, which is why I left it in place for now.

channel/channel.go
```diff
@@ -321,3 +324,22 @@ func (c *Channel) SignAndAddState(s state.State, sk *[]byte) (state.SignedState,
 	}
 	return ss, nil
 }
+
+// UpdateWithChainEvent mutates the receiver if provided with a "new" chain event (with a greater block number than previously seen)
+func (c *Channel) UpdateWithChainEvent(event chainservice.Event) (*Channel, error) {
+	if event.BlockNum() <= c.latestBlockNumber {
+		return c, nil // ignore stale information TODO: is this reorg safe?
```
NiloCK, line 331: Without too much thought: a `<` instead of `<=` makes it reorg safe.

As is, two events fired in the same block won't both get processed.

channel/channel.go
```diff
@@ -321,3 +324,22 @@ func (c *Channel) SignAndAddState(s state.State, sk *[]byte) (state.SignedState,
...
+	}
+	c.latestBlockNumber = event.BlockNum()
+	switch e := event.(type) {
+	case chainservice.AllocationUpdatedEvent:
+		c.OnChainFunding[e.AssetAddress] = e.AssetAmount
+	case chainservice.DepositedEvent:
+		c.OnChainFunding[e.Asset] = e.NowHeld
```
NiloCK, line 338: out of PR scope: `Asset` and `AssetAddress` are a bit jarring when seen this close together!

node/engine/chainservice/chainservice.go
```diff
@@ -13,18 +13,23 @@ import (
...
 }
 
 func (ce commonEvent) ChannelID() types.Destination {
 	return ce.channelID
 }
 
+func (ce commonEvent) BlockNum() uint64 {
+	return ce.blockNum
+}
+
```
NiloCK, line 32: :+1:

protocols/directfund/directfund.go
NiloCK: Is the unit-testing utility more or less that existing tests hit this method, and refactoring them to achieve their goals differently is low-reward work?

NiloCK (approved):
This LGTM.

I don't know if we possibly might want to store a (small) tree of channel states as a robust re-org protection mechanism. EG, instead of the chain service waiting `n` blocks before notifying the engine, the engine just consumes and stores potentially relevant channel updates, and prunes all "non-canonical" items away after the best one is `n` blocks deep.

protocols/directfund/directfund.go
geoknee: Yes that's about it. I think it's worth doing that work though, so perhaps not a low reward in the end.

channel/channel.go
```diff
@@ -321,3 +324,22 @@ func (c *Channel) SignAndAddState(s state.State, sk *[]byte) (state.SignedState,
 	}
 	return ss, nil
 }
+
+// UpdateWithChainEvent mutates the receiver if provided with a "new" chain event (with a greater block number than previously seen)
+func (c *Channel) UpdateWithChainEvent(event chainservice.Event) (*Channel, error) {
+	if event.BlockNum() <= c.latestBlockNumber {
```
geoknee, line 330:
```suggestion
	if event.BlockNum() < c.latestBlockNumber {
```

channel/channel.go
geoknee: So true :-P

geoknee:
> I don't know if we possibly might want to store a (small) tree of channel states as a robust re-org protection mechanism. EG, instead of the chain service waiting `n` blocks before notifying the engine, the engine just consumes and stores potentially relevant channel updates, and prunes all "non-canonical" items away after the best one is `n` blocks deep.

Interesting. I think we already have some logic which requires the chain service to feed the engine a monotonic stream of blocks. So perhaps all that complexity you mention ought to go into the chain service so that the engine can be mostly ignorant of reorgs?

protocols/directfund/directfund.go
geoknee: [b840a98](https://github.com/statechannels/go-nitro/pull/1540/commits/b840a98ba67c0748434a6e6c1ebffb8953fb4652) and [d174c0f](https://github.com/statechannels/go-nitro/pull/1540/commits/d174c0f7f4a5b01fadcc9cb4709c6a1b98469f86)

NiloCK:
> Interesting. I think we already have some logic which requires the chain service to feed the engine a monotonic stream of blocks. So perhaps all that complexity you mention ought to go into the chain service so that the engine can be mostly ignorant of reorgs?

This is probably correct - the niggling thing is the possibility that an on-chain event might call for wallet actions that carry some urgency, and potentially some round trips. Keeping this info from the wallet until the chainservice itself is satisfied about the finality might interfere with the wallet's ability to effectively respond.

A concocted scenario:

- Alice, Bob, and Irene are in some application channel where any two persons can combine to force a final state
- Alice launches a malicious challenge
- Irene and Bob must do some network round trips to construct their response

The networking steps here don't get started until I and B's chainservices return the event. With tight timelines, Irene and Bob might be better off "pessimistically" preparing their response at the first indication, and then submitting that prepared response after subjective finality is achieved.


Probably this can be sufficiently mitigated by a policy mechanism: refuse to enter channels with timeouts lower than (some function of) the chainservice's block finality decision time.

PR #1547, opened 2023-08-16
Add gui to cloud docker image

bitwiseguy:
Updates the `docker/cloud/Dockerfile` to build the `nitro-gui` package. The gui then gets included in the `go build` and is exposed on port 5005 of the docker image.

This change increased the docker image size from 145MB --> 164MB.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,8 +1,6 @@
 artifacts
 data
+docs
 
 .adr
-.yarn
-docs
-node_modules
-packages
+.github
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,6 @@
 artifacts
 data
+node_modules
 docs
 
 .adr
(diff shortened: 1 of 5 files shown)


.dockerignore
```diff
@@ -1,8 +1,6 @@
 artifacts
 data
+docs
 
 .adr
-.yarn
-docs
-node_modules
-packages
```
geoknee, line 8: Are we sure we no longer want to dockerignore these directories?

geoknee (approved):

.dockerignore
bitwiseguy: Good catch. I removed `node_modules`. Explanation on others below:

* `.yarn`: needed in order to build the gui
* `docs`: still in .dockerignore file just got moved to another line
* `node_modules`: not needed. Removed
* `packages`: needed in order to build the gui

geoknee: Great. I wonder if that was responsible for the 19MB increase in image size (the GUI itself is much smaller than that so I was a little surprised).

PR #1557, opened 2023-08-17
Create github action to deploy cloud node to Digital Ocean

bitwiseguy:
In order to securely configure our cloud node, we need to be able to pass the `pk` and `chainpk` values as env vars instead of reading them from a config file (which is stored on github and is therefore visible to anyone who views this repo). I added this comment to `main.go` to remind us about the default order of precedence:
```
	// urfave default precedence for flag value sources (highest to lowest):
	// 1. Command line flag value
	// 2. Environment variable (if specified)
	// 3. Configuration file (if specified)
	// 4. Default defined on the flag
```

This PR also adds a github action .yml file to automatically deploy the nitro node to Digital Ocean. Complementary work for this PR was to store the following values in in github secrets. 

* `DO_API_KEY`
* `DO_SSH_PRIVATE_KEY`
* `DO_SC_PK`
* `DO_CHAIN_PK`

`DO_SC_PK` and `DO_CHAIN_PK` are passed as env vars to the docker container running inside our Digital Ocean droplet.

### DIFF
--- a/.github/workflows/deploy-cloud-node.yml
+++ b/.github/workflows/deploy-cloud-node.yml
@@ -24,10 +24,7 @@ jobs:
         env:
           SSH_PRIVATE_KEY: ${{ secrets.DO_SSH_PRIVATE_KEY }}
         run: |
-          echo "$SSH_PRIVATE_KEY" > private_key.pem
-          chmod 600 private_key.pem
-          scp -o StrictHostKeyChecking=no -i private_key.pem docker-compose.yml root@67.207.88.72:~/
-          ssh -o StrictHostKeyChecking=no -i private_key.pem root@67.207.88.72 <<'ENDSSH'
+          echo "$SSH_PRIVATE_KEY" | ssh -o StrictHostKeyChecking=no -i /dev/stdin root@67.207.88.72 <<'ENDSSH'
             docker pull registry.digitalocean.com/magmo/go-nitro:latest
             docker run -it -d -p 3005:3005 -p 4005:4005 -p 5005:5005 -e SC_PK=$DO_SC_PK -e CHAIN_PK=$DO_CHAIN_PK registry.digitalocean.com/magmo/go-nitro:latest
           ENDSSH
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-cloud-node.yml
```diff
@@ -0,0 +1,33 @@
...
+        run: docker login -u ${{ secrets.DO_API_KEY }} -p ${{ secrets.DO_API_KEY }} registry.digitalocean.com
+
+      - name: Push Docker image to DigitalOcean
+        run: make docker/cloud/push
+
+      - name: Deploy to Droplet
+        env:
+          SSH_PRIVATE_KEY: ${{ secrets.DO_SSH_PRIVATE_KEY }}
+        run: |
+          echo "$SSH_PRIVATE_KEY" > private_key.pem
```
kerzhner, line 27: Is this a standard approach? It would be nice to avoid storing secrets on disk even if the runner is ephemeral.

kerzhner (approved):

.github/workflows/deploy-cloud-node.yml
bitwiseguy: I'm not sure what the standard approach is. We might be able to read from stdin instead like this:
```
run: |
  echo "$SSH_PRIVATE_KEY" | ssh -o StrictHostKeyChecking=no -i /dev/stdin root@67.207.88.72 <<'ENDSSH'
    docker pull registry.digitalocean.com/magmo/go-nitro:latest
    docker run -it -d -p 3005:3005 -p 4005:4005 -p 5005:5005 -e SC_PK=$DO_SC_PK -e CHAIN_PK=$DO_CHAIN_PK registry.digitalocean.com/magmo/go-nitro:latest
  ENDSSH
```
I will try that instead.

bitwiseguy: I couldn't get the stdin approach to work. I was getting permissions errors that made it seem like the ssh client would not allow the private key to be fed that way. For now, I am going to use a temporary .pem file that gets deleted as soon as the ssh session ends.

PR #1560, opened 2023-08-17
Remove comments about adding a support to channel struct

lalexgap:
Currently the `channel`s only concept of a supported state is a [state signed by all participants](https://github.com/statechannels/go-nitro/blob/73c4f859607e2971714df6da804e2aff2f2f0333/channel/channel.go#L290). So we already expose the `support` via [LatestSupportedState](https://github.com/statechannels/go-nitro/blob/2acf498b7d702659a5dc975fc8f9499959298eb2/channel/channel.go#L211), there's no need to add another field.

### DIFF
--- a/channel/channel.go
+++ b/channel/channel.go
@@ -22,8 +22,6 @@ type Channel struct {
 	latestBlockNumber uint64 // the latest block number we've seen
 
 	state.FixedPart
-	// Support []uint64 // TODO: this property will be important, and allow the Channel to store the necessary data to close out the channel on chain
-	// It could be an array of turnNums, which can be used to slice into Channel.SignedStateForTurnNum
 
 	SignedStateForTurnNum map[uint64]state.SignedState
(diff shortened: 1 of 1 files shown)


geoknee:
> So we already expose the `support` via [LatestSupportedState](https://github.com/statechannels/go-nitro/blob/2acf498b7d702659a5dc975fc8f9499959298eb2/channel/channel.go#L211),

Nit: this actually only returns a state (no signatures) so not quite the `support`.

geoknee (approved):
I think you're right we should remove this. 

I think we probably will just hand roll the logic to challenge each kind of channel for now, without using general methods for exposing the support (which as you say is difficult to do in full generality.

lalexgap:
> > So we already expose the `support` via [LatestSupportedState](https://github.com/statechannels/go-nitro/blob/2acf498b7d702659a5dc975fc8f9499959298eb2/channel/channel.go#L211),
> 
> Nit: this actually only returns a state (no signatures) so not quite the `support`.

Good point! We'd need to add a `LatestSignedSupportedState` getter.

PR #1561, opened 2023-08-18
Fix github action for deploying cloud node

bitwiseguy:
We need to stop the previous docker container (if there is one running) before starting the new one.


geoknee (approved):

geoknee:
👁️

PR #1564, opened 2023-08-18
use a mutex in  SubmitTransaction for mock chain

lalexgap:
Based on this [data race failure](https://github.com/statechannels/go-nitro/actions/runs/5905501765) it was revealed that there could a data race around `blockNum`.

This PR resolves this by locking on a mutex for the duration of `SubmitTransaction`, ensuring that it is not run concurrently.


lalexgap:
Closing in favour of #1565

PR #1565, opened 2023-08-18
Mock chain protects state with a mutex

kerzhner:
This PR adds a mutex that protects updating chain state. The mutex is released before chain events are sent to avoid the following deadlock:
1. Transaction is submitted.
2. Mutex is locked.
3. A chain event is broadcast.
4. A listener to the event submits another transaction.
5. A deadlock occurs since the mutex is already locked.

Given that step 3 (event broadcast) is a send on a channel, this deadlock will not occur with the current code base. This mutex approach protects against implementation changes to event broadcasting.

https://github.com/statechannels/go-nitro/actions/runs/5905501765/job/16019700856 failure is addressed with this PR.

### DIFF
--- a/node/engine/chainservice/mock_chain.go
+++ b/node/engine/chainservice/mock_chain.go
@@ -2,6 +2,7 @@ package chainservice
 
 import (
 	"fmt"
+	"sync"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/statechannels/go-nitro/internal/safesync"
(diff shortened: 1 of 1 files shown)


lalexgap (approved):

lalexgap:
> This PR adds a mutex that protects updating chain state. The mutex is released before chain events are sent to avoid the following deadlock:
> 
>     1. Transaction is submitted.
> 
>     2. Mutex is locked.
> 
>     3. A chain event is broadcast.
> 
>     4. A listener to the event submits another transaction.
> 
>     5. A deadlock occurs since the mutex is already locked.

I think for a deadlock like this to be possible step 4 must be executed within the same `SubmitTransaction` call inside the `MockChain`?   Or step 3 must block until step 4 completes by some other manner, which seems kinda strange?

PR #1574, opened 2023-08-21
Batch deposits

NiloCK:
PR introduces a proxy contract with two utility functions for batched deposits: `batch_deposit_eth` and `batch_deposit_erc`, as well as a correctness test and additions to the gas benchmarks to record performance (eg, relative to "naive" deposits).

Primary motivation for proxy contract rather than direct implementation in MultiAssetHolder is security / complexity / API surface area.

Results (cost multiplier relative to existing deposit method) :

| deposit type \ N | 1    | 5     | 10    | 20    | 50    | 100   |
|------------------|------|-------|-------|-------|-------|-------|
| ETH (1st)        | 1.29 | 0.829 | 0.769 | 0.740 | 0.722 | 0.716 |
| ETH (2nd)        | 1.46 | 0.728 | 0.636 | 0.590 | 0.563 | 0.553 |
| ERC (1st)        | 1.51 | 0.638 | 0.553 | 0.531 | 0.485 | 0.476 |
| ERC (2nd)        | 1.73 | 0.537 | 0.409 | 0.353 | 0.319 | 0.308 |


nitro-protocol/contracts/auxiliary/BatchOperator.sol
```diff
@@ -0,0 +1,64 @@
+pragma solidity 0.8.17;
+import '../MultiAssetHolder.sol';
+
+import '@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol';
+import '@openzeppelin/contracts/token/ERC20/IERC20.sol';
+
+string constant lengthsErr = 'Array lengths must match';
+
+/**
+@dev This contract is used to batch deposit ERC20 tokens into .
```
geoknee, line 10: Comment trails off?

nitro-protocol/contracts/auxiliary/BatchOperator.sol
```diff
@@ -0,0 +1,64 @@
...
+                expectedHelds[i],
+                amounts[i]
+            );
+        }
+    }
+
+    /**
+     * @dev Deposits ERC20 tokens into the adjudicator for multiple channels.
+     */
+    function deposit_batch_erc(
```
geoknee, line 45: Probably worth using `erc20` in the name?

nitro-protocol/contracts/auxiliary/BatchOperator.sol
```diff
@@ -0,0 +1,64 @@
...
+        uint256[] calldata amounts,
+        uint256 totalAmount
+    ) external payable virtual {
+        require(
+            channelIds.length == expectedHelds.length && expectedHelds.length == amounts.length,
+            lengthsErr
+        );
+
+        IERC20(asset).safeTransferFrom(msg.sender, address(this), totalAmount);
+        IERC20(asset).safeApprove(address(adjudicator), totalAmount);
```
geoknee, line 58: I'm wondering if there is a cheaper (but still safe) way of doing this. What if the adjudicator could pull tokens from a supplied address, rather than `msg.sender`? Then, the tokens wouldn't need to go to the batch depositor before going to the adjudicator?

geoknee: Tempting to consider `tx.origin` but that is not recommended https://consensys.github.io/smart-contract-best-practices/development-recommendations/solidity-specific/tx-origin/

nitro-protocol/test/contracts/BatchOperator/batchDeposits.test.ts
```diff
@@ -0,0 +1,216 @@
...
+import TokenArtifact from '../../../artifacts/contracts/Token.sol/Token.json';
+import BatchOperatorArtifact from '../../../artifacts/contracts/auxiliary/BatchOperator.sol/BatchOperator.json';
+import BadTokenArtifact from '../../../artifacts/contracts/test/BadToken.sol/BadToken.json';
+import {BadToken, BatchOperator, NitroAdjudicator, Token} from '../../../typechain-types';
+const provider = getTestProvider();
+
+const batchOperator = setupContract(
+  provider,
+  BatchOperatorArtifact,
+  process.env.BATCH_OPERATOR_ADDRESS!
```
geoknee, line 18: I think if you make the right edits to this file https://github.com/statechannels/go-nitro/blob/2b0bc7c5998abe021970704f448d030844a05928/nitro-protocol/test/environment.d.ts#L23 you woun't need these assertions.

geoknee (approved):
How much of a demand do you anticipate for this functionality? Does it justify us putting the batch operator into our set of standard contracts to deploy?

nitro-protocol/contracts/auxiliary/BatchOperator.sol
NiloCK: I was somewhat unsatisfied with this, but in the end the cost is amortized against the entirety of the batch so even at n=2 the total number of approve / transfer operations reaches parity against existing deposit flow.

Your first option is neat, but requires an addition to `deposit`'s fields list which in all other cases would be wasted space.

In general, through various drafts I found that there are a large number of public facing api methods one *could* write in the name of optimizing costs of particular deposit workflows. EG, why not have a `depositETH` function with no asset qualifier?

In the end, I think I'm happy enough to take reasonable gains via methods that don't alter the existing API.

nitro-protocol/contracts/auxiliary/BatchOperator.sol
NiloCK: 70999c61

nitro-protocol/contracts/auxiliary/BatchOperator.sol
geoknee: That's fair!

PR #1575, opened 2023-08-22
Use correct syntax in GHA

geoknee:


geoknee:
@bitwiseguy I am wondering if we can skip building the docker image in case we already pushed it to the registry? I'm not sure if there is any easy way but rerunning the action often we spend a couple of minutes building and then docker detects it has already pushed and aborts that step.

bitwiseguy (approved):

geoknee:
> When you say that docker "aborts that step" do you mean it fails or it just realizes it doesn't need to push the image and the GHA can still continue?

It realises it doesn't need to push and the GHA can continue. 

Probably not worth a lot of time doing this optimization, it's only really a pain when developing / modifying the GHA itself.

Issue #1576, opened 2023-08-22
Persist cloud node data between container/droplet restarts


bitwiseguy:
Currently, every time we restart the docker container running in the cloud, we lose its `store` data. We should mount a volume in the docker container and on the droplet to allow us to restart the container without wiping all of its `store` data.

PR #1579, opened 2023-08-22
Persist store data in cloud node

bitwiseguy:
Closes https://github.com/statechannels/go-nitro/issues/1576


bitwiseguy:
> How can we test this?
I ran through this sequence:

1. Ran the Github Action against this branch. 
2. ssh'ed into the Droplet and saw these files in the VM:
```
/var/
  |_ nitro_iris_store/
    |_ 0x89825D58A7E2C198a06125be9CD0631317f9A07B/
      |_ channel_to_objective_89825.db
      |_ channels_89825.db
      |_ consensus_channels_89825.db
      |_ objectives_89825.db
      |_ vouchers_89825.db
```
3. Ran the following sequence of commands:
```
docker stop nitro_iris
docker rm nitro_iris
```
4. Confirmed that the db files listed above were still present on the VM after the container was removed

PR #1581, opened 2023-08-23
Store lastBlockSeen value

bitwiseguy:
This is a precursor to adding initialization logic to the `chainservice` to allow it to process events from blocks since the node last started. Currently, our node always assumes we are starting for the first time, therefore it doesn't look back in time to see if it's missed anything.

This PR makes the following changes:

1. Adds store methods to get/set the last block processed. Also adds unit tests for these new methods.
2. Adds code to the engine to call `store.SetLastBlockSeen` every time an on-chain event is received from `chainservice`

### DIFF
--- a/node/engine/engine.go
+++ b/node/engine/engine.go
@@ -189,7 +189,7 @@ func (e *Engine) run(ctx context.Context) {
 		case pr := <-e.PaymentRequestsFromAPI:
 			res, err = e.handlePaymentRequest(pr)
 		case chainEvent := <-e.fromChain:
-			err = e.store.SetLastBlockSeen(chainEvent.BlockNum())
+			err = e.store.SetLastBlockNumSeen(chainEvent.BlockNum())
 			e.checkError(err)
 			res, err = e.handleChainEvent(chainEvent)
 		case message := <-e.fromMsg:
--- a/node/engine/engine.go
+++ b/node/engine/engine.go
@@ -189,7 +189,7 @@ func (e *Engine) run(ctx context.Context) {
 		case pr := <-e.PaymentRequestsFromAPI:
 			res, err = e.handlePaymentRequest(pr)
 		case chainEvent := <-e.fromChain:
-			err = e.store.SetLastBlockProcessed(chainEvent.BlockNum())
+			err = e.store.SetLastBlockSeen(chainEvent.BlockNum())
 			e.checkError(err)
 			res, err = e.handleChainEvent(chainEvent)
 		case message := <-e.fromMsg:
--- a/node/engine/engine.go
+++ b/node/engine/engine.go
@@ -189,6 +189,8 @@ func (e *Engine) run(ctx context.Context) {
 		case pr := <-e.PaymentRequestsFromAPI:
 			res, err = e.handlePaymentRequest(pr)
 		case chainEvent := <-e.fromChain:
+			err = e.store.SetLastBlockProcessed(chainEvent.BlockNum())
+			e.checkError(err)
 			res, err = e.handleChainEvent(chainEvent)
 		case message := <-e.fromMsg:
 			res, err = e.handleMessage(message)
--- a/node/engine/store/durablestore.go
+++ b/node/engine/store/durablestore.go
@@ -27,7 +27,7 @@ type DurableStore struct {
 	consensusChannels  *buntdb.DB
 	channelToObjective *buntdb.DB
 	vouchers           *buntdb.DB
-	blocks             *buntdb.DB
+	lastBlockSeen      *buntdb.DB
 
 	key     string // the signing key of the store's engine
 	address string // the (Ethereum) address associated to the signing key
@@ -68,7 +68,7 @@ func NewDurableStore(key []byte, folder string, config buntdb.Config) (Store, er
 		return nil, err
 	}
 
-	ps.blocks, err = ps.openDB("blocks", config)
+	ps.lastBlockSeen, err = ps.openDB("lastBlockSeen", config)
 	if err != nil {
 		return nil, err
 	}
@@ -217,10 +217,10 @@ func (ds *DurableStore) SetObjective(obj protocols.Objective) error {
 	return nil
 }
 
-// GetLastBlockSeen retrieves the last blockchain block processed by this node
-func (ds *DurableStore) GetLastBlockSeen() (uint64, error) {
+// GetLastBlockNumSeen retrieves the last blockchain block processed by this node
+func (ds *DurableStore) GetLastBlockNumSeen() (uint64, error) {
 	var result uint64
-	err := ds.blocks.View(func(tx *buntdb.Tx) error {
+	err := ds.lastBlockSeen.View(func(tx *buntdb.Tx) error {
 		val, err := tx.Get(lastBlockSeenKey)
 		if err != nil {
 			return err
@@ -231,9 +231,9 @@ func (ds *DurableStore) GetLastBlockSeen() (uint64, error) {
 	return result, err
 }
 
-// SetLastBlockSeen sets the last blockchain block processed by this node
-func (ds *DurableStore) SetLastBlockSeen(blockNumber uint64) error {
-	return ds.blocks.Update(func(tx *buntdb.Tx) error {
+// SetLastBlockNumSeen sets the last blockchain block processed by this node
+func (ds *DurableStore) SetLastBlockNumSeen(blockNumber uint64) error {
+	return ds.lastBlockSeen.Update(func(tx *buntdb.Tx) error {
 		_, _, err := tx.Set(lastBlockSeenKey, strconv.FormatUint(blockNumber, 10), nil)
 		return err
 	})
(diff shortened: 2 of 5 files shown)


node/engine/store/memstore.go
```diff
@@ -24,6 +24,7 @@ type MemStore struct {
 	consensusChannels  safesync.Map[[]byte]
 	channelToObjective safesync.Map[protocols.ObjectiveId]
 	vouchers           safesync.Map[[]byte]
+	blocks             safesync.Map[uint64]
```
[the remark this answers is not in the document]
bitwiseguy: I agree that's overkill in this situation. Changed in https://github.com/statechannels/go-nitro/pull/1581/commits/ca4f034bbff090d1d34ddf1c90a970e36d318593

node/engine/engine.go
```diff
@@ -189,6 +189,8 @@ func (e *Engine) run(ctx context.Context) {
 		case pr := <-e.PaymentRequestsFromAPI:
 			res, err = e.handlePaymentRequest(pr)
 		case chainEvent := <-e.fromChain:
+			err = e.store.SetLastBlockProcessed(chainEvent.BlockNum())
```
bitwiseguy, line 192: In this branch this is the only time we call `store.SetLastBlockProcessed`. We should improve this so that we are updating the `lastBlockProcessed` even when no events are emitted by the NitroAdjudicator contractor. This optimization would reduce the amount of processing the chainservice has to do when it is initialized and checks for any events that were emitted since the node was last online.

lalexgap, line 192: Since we're setting `lastBlockProcessed` when the engine receives an event with that block number, it means `lastBlockProcessed` gets set when the engine **handles the first event in a block**, not when all events in a block are handled. 

For example lets say we have block which contains two events: 1)`Deposited 0x0, 0x0AAA... Block Num:55` 2)`Deposited 0x0, 0xBBBB...Block Num:55`

Let's say our engine receives the first event and sets `lastBlockProcessed=55`. Now let's say the engine crashes. Based on the `lastBlockProcessed` it has no way of knowing that we never handled the second event.

geoknee: Shouldn't the chainservice have a pretty smart way of checking historical events, though? If we are filtering our queries using the NitroAdjudicator address https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_getlogs , we could potentially get away with only changing `lastBlockProcessed` when the adjudicator emits an event.

node/engine/store/memstore.go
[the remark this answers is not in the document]
geoknee: Cool. I suppose by that change does render that variable "concurrency unsafe" in theory, since there is no longer any `sync.Mutex` under the hood preventing concurrent reads/writes. Might be worth running the race condition checker on this branch and/or adding a mutex if we think that is necessary.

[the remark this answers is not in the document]
geoknee: Also, do we want to make the similar change in the `durablestore`, too?

node/engine/engine.go
geoknee: See #1591

bitwiseguy: Yes I think we could get away with only changing `lastBlockSeen` when the adjudicator emits an event. However, I was trying to think of a way to reduce processing time during this scenario:

** nitro node is running ***
1. Block 100 mined on-chain
2. node detects adjudicator event (AE1) in block 100
3. Block 101 mined on-chain
4. Block 102 mined on-chain
5. node processes AE1
6. node updates `lastBlockSeen` to 100
7. Block 103-20,000 mined on-chain, but no adjudicator events are emitted
8. node is turns off
9. Block 20,001-30,000 mined on-chain
10. node is restarted and begins init routines
11. node looks through blocks from `lastBlockSeen` (100) to current block (30,000) for adjudicator events

The query in step [11] to search through many blocks seems wasteful since our nitro node was actually running up until block 20,000 was mined, but `lastBlockSeen` was not updated because no adjudicator events were included in any of those blocks. Instead of searching through the block range 100-20,000, we should just be able to search through the range 20,000-30,000 since those were the only blocks mined while the node was offline.

lalexgap: That's a good point, I didn't think about that scenario!

One thing to note is that ethereum makes use of [bloom filters](https://en.wikipedia.org/wiki/Bloom_filter) for logs(aka events) on [the block header.](https://www.educative.io/answers/what-is-ethereums-block-structure)  This means it's very efficient to check if a block contains an event by checking the `logsBloom` on the block header. Instead of having to parse through the a block you can just check your event topics against the bloom filter to know if the events are in the block. This means it's relatively efficient to query for events over a large range of blocks.  The [JSON-RPC API allows you to specify a from and to block](https://docs.infura.io/networks/ethereum/json-rpc-methods/eth_getlogs) so it's pretty easy to query a large range of blocks for event's we'd be interested in.

Based on that I think we can avoid worrying about this scenario too much for now. If it's something we start noticing we can revisit and implement a solution.

node/engine/store/durablestore.go
```diff
@@ -26,6 +27,7 @@ type DurableStore struct {
 	consensusChannels  *buntdb.DB
 	channelToObjective *buntdb.DB
 	vouchers           *buntdb.DB
+	blocks             *buntdb.DB
```
lalexgap, line 30: My preference would be something like `lastBlockSeen` instead of `blocks`, since that's all we're really storing in `blocks`. Not a huge deal either way.

node/engine/store/durablestore.go
```diff
@@ -210,6 +217,28 @@ func (ds *DurableStore) SetObjective(obj protocols.Objective) error {
 	return nil
 }
 
+// GetLastBlockSeen retrieves the last blockchain block processed by this node
+func (ds *DurableStore) GetLastBlockSeen() (uint64, error) {
```
lalexgap, line 221:
```suggestion
func (ds *DurableStore) GetLastSeenBlockNum() (uint64, error) {
```
This might be nitpicking but I think we should use the term `BlockNum` instead of `Block`, since we're just returning a block number, not an actual block.

lalexgap (approved):

node/engine/store/durablestore.go
bitwiseguy: Updated in https://github.com/statechannels/go-nitro/pull/1581/commits/1185ea1f6ba21612424ee2678b86ee7a4bff284e

node/engine/store/durablestore.go
bitwiseguy: Updated in https://github.com/statechannels/go-nitro/pull/1581/commits/1185ea1f6ba21612424ee2678b86ee7a4bff284e

node/engine/engine.go
bitwiseguy: That's good info on the bloom filters. I will not worry about this scenario for now but I made this issue to remind us to revisit if we see any related problems in the future: https://github.com/statechannels/go-nitro/issues/1600

geoknee: > If we already have protection in the Channel class against processing the same event twice

I think we need an ADR for this - a well-thought-through pattern which involves some "contract" between a chain service and the `Channel` class, such that we don't get any bugs around stale or incorrect information.

Example: the `Channel` class promises to have an idempotent function for accepting events from the chain service, and the chain service promises to pass events in order. 

☝️ I'm not 100% sure about this example though. It seems like we might want to avoid (re)playing a `ChallengeRegistered` event if the challenge has long since expired or been cleared. 

Not a blocker for merging this PR, because I think in mose conceivable patterns we will want to store `lastBlockNumberSeen`.

node/engine/store/memstore.go
```diff
@@ -125,6 +127,17 @@ func (ms *MemStore) SetObjective(obj protocols.Objective) error {
...
+// SetLastBlockNumSeen
+func (ms *MemStore) SetLastBlockNumSeen(blockNumber uint64) error {
+	ms.lastBlockSeen = blockNumber
+	return nil
+}
+
+// GetLastBlockNumSeen
+func (ms *MemStore) GetLastBlockNumSeen() (uint64, error) {
+	return ms.lastBlockSeen, nil
+}
```
geoknee, line 139: Let's tidy this up following @lalexgap's comment on block vs block num. It's a pain but when we fix one store implementation, we should fix the other :-P

node/engine/store/store.go
```diff
@@ -15,6 +15,7 @@ const (
 	ErrNoSuchObjective = types.ConstError("store: no such objective")
 	ErrNoSuchChannel   = types.ConstError("store: failed to find required channel data")
 	ErrLoadVouchers    = types.ConstError("store: could not load vouchers")
+	lastBlockSeenKey   = "lastBlockSeen"
```
geoknee, line 18: Another instance of "block" vs "blocknum"

geoknee (approved):
Still a few small things to tidy up, but this LGTM.

node/engine/store/store.go
bitwiseguy: Changed in https://github.com/statechannels/go-nitro/pull/1581/commits/5d4a6161745e089a900cf5dc42e642b1116a3a44

PR #1587, opened 2023-08-24
Fix the availability of type information on `process.env` (nitro-protocol tests)

### LINKED ISSUE #1475: Fix type check errors on `process.env` data
Existing test suites for nitro adjudicator involve:
- running a contract deployment [script](https://github.com/statechannels/go-nitro/blob/main/nitro-protocol/deployment/deploy-test-contracts.ts#L62) 
- [writing](https://github.com/statechannels/go-nitro/blob/main/nitro-protocol/jest/contract-test-setup.ts#L13) resultant contract addresses onto the `process.env` variable
- creating nicely typed contract interface objects [hydrated](https://github.com/statechannels/go-nitro/blob/main/nitro-protocol/test/contracts/MultiAssetHolder/deposit.test.ts#L19) with this address data

**Issue**: vscode (via typescript compiler) complains about accessing data from the non-typed `process.env`.

**Fix**: one viable approach is [here](https://github.com/NiloCK/vue-skuilder/blob/master/packages/express/src/utils/env.ts)
- create a type
- hydrate an instance of it from existing `process.env` & deploy script addresses, fail-fast if data is missing
- export the typed `ENV` for usage elsewhere

This could be done inside https://github.com/statechannels/go-nitro/blob/main/nitro-protocol/jest/contract-test-setup.ts

NiloCK:
`process.env` has declared values for, eg, addresses of various contracts.

On my machine, at least, these types were not found by VSCode, so usage of those environment variables involved either:
- squiggly underlines (`string or undefined not assignable to type string`)
- using `!` assertions (which a linter complains about)
- explicitly checking the value before using
- manually locating the types with `import '../environment.d.ts'`

... all of which are annoying.

PR moves the declared types into a `@types` dir, which fixes the issue - VSCode now finds these types and provides correct type inference / autocomplete on `process.env`.

Closes #1475

### DIFF
--- a/nitro-protocol/test/tsconfig.json
+++ b/nitro-protocol/test/tsconfig.json
@@ -11,5 +11,5 @@
     "skipLibCheck": true,
     "noEmit": true
   },
-  "include": ["environment.d.ts"]
+  "include": ["@types/environment.d.ts"]
 }


geoknee (approved):
Does this have any impact on things like `tsc`? Or is it just a vscode thing?

NiloCK:
> Does this have any impact on things like `tsc`? Or is it just a vscode thing?

No visible impact. `tsc`, `npm test`, similar commands all work as expected locally.

PR #1590, opened 2023-08-24
Add benchmark for clearing a challenge

### LINKED ISSUE #1585: Implement gas benchmarks for clearing a challenge
Our [current gas benchmarks](https://github.com/statechannels/go-nitro/blob/dd7dcf368e977d944162a081c4673a25fe158c2f/nitro-protocol/gas-benchmarks/benchmark.test.ts#L1) currently do not have a scenario of responding to a challenge. We should add a scenario where we respond to a challenge by checkpoint or challenge.

lalexgap:
Fixes #1585 

This implements gas benchmarks for clearing a challenge via `checkpoint` and `challenge`.

### DIFF
--- a/nitro-protocol/gas-benchmarks/gasResults.json
+++ b/nitro-protocol/gas-benchmarks/gasResults.json
@@ -98,7 +98,8 @@
       "checkpointX": 73605,
       "challengeResponseX": 99610,
       "checkpointL": 69601,
-      "challengeResponseL": 92781
+      "challengeResponseL": 92781,
+      "challengeResponseV": 110145
     }
   }
 }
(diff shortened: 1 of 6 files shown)


nitro-protocol/gas-benchmarks/gasResults.json
```diff
@@ -92,5 +92,14 @@
       "transferAllAssetsL": 113759,
       "total": 477034
     }
+  },
+  "ETHClearChallenge": {
```
geoknee, line 96: Do we expect it to matter about the `ETH` part?

geoknee (approved):
Out of scope here, but I think it would be nice for us to be able to split up the benchmark code a bit into several files. That could be tackled as a part of #809 .

nitro-protocol/gas-benchmarks/gasResults.json
geoknee: Let's try with an ERC20 and see.

PR #1593, opened 2023-08-25
`Channel` class: store `stateHash`, `candidate` signed state and `Outcome` following `ChallengeRegistered`

geoknee:
https://docs.statechannels.org/protocol-tutorial/0080-defunding-a-channel/#tracking-on-chain-storage

This should be sufficient to allow us to call `transfer` at some later stage. 

It should also trigger an appropriate notification for the client (same behaviour as receiving the state off chain). 

Towards #1531

### DIFF
--- a/node/engine/chainservice/chainservice.go
+++ b/node/engine/chainservice/chainservice.go
@@ -78,7 +78,7 @@ type ChallengeRegisteredEvent struct {
 	candidateSignatures []state.Signature
 }
 
-// StateHash returns the statehash which will have been stored on chain in the adjudicator after the ChallengeRegistered Event fires.
+// StateHash returns the statehash stored on chain at the time of the ChallengeRegistered Event firing.
 func (cr ChallengeRegisteredEvent) StateHash(fp state.FixedPart) (common.Hash, error) {
 	return state.StateFromFixedAndVariablePart(fp, cr.candidate).Hash()
 }
@@ -88,7 +88,7 @@ func (cr ChallengeRegisteredEvent) Outcome() outcome.Exit {
 	return cr.candidate.Outcome
 }
 
-// SignedState returns the signe state  which will have been stored on chain in the adjudicator after the ChallengeRegistered Event fires.
+// SignedState returns the signed state which will have been stored on chain in the adjudicator after the ChallengeRegistered Event fires.
 func (cr ChallengeRegisteredEvent) SignedState(fp state.FixedPart) (state.SignedState, error) {
 	s := state.StateFromFixedAndVariablePart(fp, cr.candidate)
 	ss := state.NewSignedState(s)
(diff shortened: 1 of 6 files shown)


node/engine/chainservice/chainservice.go
```diff
@@ -66,13 +68,41 @@ type ConcludedEvent struct {
...
+func (cr ChallengeRegisteredEvent) StateHash(fp state.FixedPart) (common.Hash, error) {
+	return state.StateFromFixedAndVariablePart(fp, cr.candidate).Hash()
+}
+
+// Outcome returns the outcome which will have been stored on chain in the adjudicator after the ChallengeRegistered Event fires.
+func (cr ChallengeRegisteredEvent) Outcome() outcome.Exit {
+	return cr.candidate.Outcome
+}
+
+// SignedState returns the signe state  which will have been stored on chain in the adjudicator after the ChallengeRegistered Event fires.
```
NiloCK, line 91:
```suggestion
// SignedState returns the signed state which will have been stored on chain in the adjudicator after the ChallengeRegistered Event fires.
```

node/engine/chainservice/chainservice.go
```diff
@@ -66,13 +68,41 @@ type ConcludedEvent struct {
...
+type ChallengeRegisteredEvent struct {
 	commonEvent
-	// TODO fill out other fields
+	candidate           state.VariablePart
+	candidateSignatures []state.Signature
 }
 
-func (ce ConcludedEvent) String() string {
-	return "Channel " + ce.channelID.String() + " concluded at Block " + fmt.Sprint(ce.blockNum)
+// StateHash returns the statehash which will have been stored on chain in the adjudicator after the ChallengeRegistered Event fires.
```
NiloCK, line 81: The tense of these messages is a little hard to parse. Not sure if the below is better.

```suggestion
// StateHash returns the statehash stored on chain at the time of the ChallengeRegistered Event firing.
```

NiloCK (approved):
This LGTM, but wanting for a follow-up with tests that consume the same.

(which currently are probably blocked by go-nitro not being able to launch challenges).

geoknee:
> This LGTM, but wanting for a follow-up with tests that consume the same.

I added a test here [add estUpdateWithChallengeRegisteredEvent](https://github.com/statechannels/go-nitro/pull/1593/commits/f68bdd7c87ceb7277e193661479c54f935dae814) but as you say it doesn't cover that much code at present. Be nice to get an integration test going with a chain and engine and then we could assert on some notification perhaps?

lalexgap:
> I added a test here [add estUpdateWithChallengeRegisteredEvent](https://github.com/statechannels/go-nitro/pull/1593/commits/f68bdd7c87ceb7277e193661479c54f935dae814) but as you say it doesn't cover that much code at present. Be nice to get an integration test going with a chain and engine and then we could assert on some notification perhaps?

We could consider expanding the [simulated chain service test as well](https://github.com/statechannels/go-nitro/blob/9888d195fad804e9480dfe3441f8b8df310c95f4/node/engine/chainservice/simulated_backend_chainservice_test.go#L80). Currently we submit a deposit transaction and wait for the deposited event, so we could do the same for challenging.

PR #1598, opened 2023-08-28
Add dedicated `deposit_eth` function and event

NiloCK:
Following [discussion](https://github.com/statechannels/go-nitro/pull/1574#discussion_r1303379537) in the batch deposits PR, I decided to test out separating eth vs erc20 deposits. Main motivation: we expect to pass and emit a lot of `0x000000...`s, which seems wasteful.

Gas savings are moderate - 571 gas on a single ETH deposit, which is somewhere between 1 and 2% of the prior cost.


nitro-protocol/contracts/MultiAssetHolder.sol
```diff
@@ -49,12 +78,7 @@ contract MultiAssetHolder is IMultiAssetHolder, StatusManager {
         uint256 held = holdings[asset][channelId];
         require(held == expectedHeld, 'held != expectedHeld');
 
-        // require successful deposit before updating holdings (protect against reentrancy)
-        if (asset == address(0)) {
-            require(msg.value == amount, 'Incorrect msg.value for deposit');
-        } else {
-            IERC20(asset).safeTransferFrom(msg.sender, address(this), amount);
-        }
+        IERC20(asset).safeTransferFrom(msg.sender, address(this), amount);
```
lalexgap, line 81: What happens if a zero address is passed is now? Does it revert?

nitro-protocol/gas-benchmarks/gasResults.json
```diff
@@ -1,105 +1,105 @@
...
     }
   },
   "ETHClearChallenge": {
     "satp": {
-      "checkpointX": 73605,
-      "challengeResponseX": 99610,
-      "checkpointL": 69601,
-      "challengeResponseL": 92781,
-      "challengeResponseV": 110145
+      "checkpointX": 73857,
```
lalexgap, line 98: It's interesting that some of these gas values changed, since they have nothing to do with the contract that was changed.

nitro-protocol/contracts/MultiAssetHolder.sol
NiloCK: It does revert: `reverted with reason string 'Address: call to non-contract'`

nitro-protocol/gas-benchmarks/gasResults.json
NiloCK: Agreed. Weird updates to the gas estimates happen more often than not.

nitro-protocol/contracts/interfaces/IMultiAssetHolder.sol
```diff
@@ -69,13 +69,20 @@ interface IMultiAssetHolder {
...
      * @param destinationHoldings The new holdings for `destination`.
      */
     event Deposited(bytes32 indexed destination, address asset, uint256 destinationHoldings);
 
+    /**
+     * @dev Indicates that an ETH deposit has been made.
+     * @param destination The channel being deposited into.
+     * @param destinationHoldings The new holdings for `destination`.
+     */
+    event DepositedEth(bytes32 indexed destination, uint256 destinationHoldings);
```
lalexgap, line 84: If we didn't use a new slimmer `DepositedEth` event but instead used `Deposited` (with a zero asset) when we deposited eth I wonder what the gas difference would be.

NiloCK: I had done this in stages, and the slimmer emit is a similar cost saving to the leaner function args list from `deposit_eth`.

(rummaging around for numbers...)

With `deposit_eth` emitting the existing event, the gas reduction for eth deposits was 303. When emitting the slimmer event, the reduction is 571.

NiloCK: > Leaving the existing `deposit` function as is and use the existing `Deposited` event would mean the downstream changes from this are pretty minimal. All our existing code would continue to work but we would have the option of optimizing our code by calling `depositEth` when depositing eth.

This is a great point. Will undo the alteration to `deposit`.

PR #1599, opened 2023-08-28
Implement some protective policies

lalexgap:
This PR makes some changes to our policy infrastructure to make it easier to implement some protective policies.

## Policies can now be combined

Instead of the engine holding a reference to a single `PolicyMaker` it now uses a `Policies` struct that contains multiple `PolicyMaker` implementations.  A new objective will be approved by `Policies` if every `PolicyMaker` it contains approves it.

This allows us to build individual policies that are narrow in scope and can be combined to get the desired effect. 

## Introduction of new Policies

I've added a few basic policies that, that when combined, should be able to provide a some basic protections for a nitro node.

- Allow list policy
- Deny list policy
- Fair outcome policy
- Max Spend policy


node/engine/policy/policy.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"fmt"
+	"log/slog"
+
+	"github.com/statechannels/go-nitro/internal/logging"
+	"github.com/statechannels/go-nitro/protocols"
+)
+
+// PolicyMaker is used to decide whether to approve or reject an objective
+type PolicyMaker interface {
```
geoknee, line 12: Do we know why we chose the name `PolicyMaker`? Why not just `Policy`?

node/engine/policy/policy.go
```diff
@@ -0,0 +1,37 @@
...
+	ShouldApprove(o protocols.Objective) bool
+}
+
+type Policies struct {
+	policies []PolicyMaker
+}
+
+func NewPolicies(policies ...PolicyMaker) *Policies {
+	return &Policies{policies: policies}
+}
```
geoknee, line 22: Maybe doesn't bring much/any benefit, but wanted to suggest:
```suggestion
type Policies []PolicyMaker

func NewPolicies(policies ...PolicyMaker) *Policies {
	p := Policies(policies)
	return &p
}
```

node/engine/policy/fairoutcome.go
```diff
@@ -0,0 +1,63 @@
...
+		}
+
+		for _, e := range obj.V.PreFundState().Outcome {
+			total := e.TotalAllocated()
+			for i, a := range e.Allocations {
+				if i == payments.PAYER_INDEX && isNotEqual(a.Amount, total) {
+					slog.Warn("FairOutcomePolicy: rejecting virtualfund objective, expected payer to start with full amount", "allocations", e.Allocations, logging.WithObjectiveIdAttribute(o.Id()))
+					return false
+				} else if i != payments.PAYER_INDEX && isNotEqual(a.Amount, big.NewInt(0)) {
+					slog.Warn("FairOutcomePolicy: rejecting virtualfund objective, expected payee to start with 0", "allocations", e.Allocations, logging.WithObjectiveIdAttribute(o.Id()))
```
geoknee, line 51: What makes the [fullAmount, 0] outcome "fair"? Is it likely anyone would want to _not_ have this policy protecting them? An alternative is to bake the check in to the protocol itself so it runs automatically (if it is that important)?

geoknee:
This is looking like a big improvement over what we have now 👍 . Left some initial questions. 

Also

> A new objective will be approved by Policies if every PolicyMaker it contains approves it.

Do you think there is any more complex logic we might want to support? As in, a policy that approves if at least one child approves the objective? (As an example, consider concatenating two allow-lists).

node/engine/policy/policy.go
lalexgap: I suspect this was based on the naming convention that [interfaces should be an "agent" noun ending with the er suffix.](https://go.dev/doc/effective_go#interface-names)

lalexgap:
@NiloCK also brought up the fact that `AllowList` and `DenyList` may want to be updated over time, so we may want to consider API functions that allow these to be modified in the future.

node/engine/policy/policy.go
lalexgap: https://github.com/statechannels/go-nitro/pull/1599/commits/73192c47e719edf70493dce900e8b2ebed144d39

node/engine/policy/fairoutcome.go
lalexgap: I agree, these checks should probably just be baked into the protocol themselves.

lalexgap:
Based on standup discussion I'm going to try and put together an RFC that thinks about what "default" protections we want and how that changes by the role by we play.

lalexgap:
Closing for now, will reopen based on results of RFC discussion.

PR #1603, opened 2023-08-28
Add authentication tokens

kerzhner:
This PR adds authentication tokens for the JSON RPC api. Every API method either requires no authentication token or an authentication token with correct permissions. If the direction of this work makes sense, the follow up work is:

- [ ] Do not store the token signing key in code.
- [ ] Wire up tokens in the Typescript RPC client
- [ ] Define an authorization strategy. At the moment, the server will issue a valid token to any requester.

### DIFF
--- a/rpc/auth.go
+++ b/rpc/auth.go
@@ -34,13 +34,14 @@ var (
 var invalidIAtFormat = "invalid issued at: %w"
 
 // generateAuthToken generates a JWT token that a client uses to authenticate with the server for restricted endpoints
-func generateAuthToken(p []permission) (string, error) {
+// subject is the itentifier of the client for which the token is generated
+func generateAuthToken(subject string, p []permission) (string, error) {
 	token := jwt.New(jwt.SigningMethodHS256)
 	claims := token.Claims.(jwt.MapClaims)
 	claims[permissionKey] = p
 	// the keys are defined by https://datatracker.ietf.org/doc/html/rfc7519
 	claims["iat"] = time.Now().Unix()
-	claims["sub"] = "client"
+	claims["sub"] = subject
 	return token.SignedString(rpcPK)
 }
 
--- a/rpc/auth.go
+++ b/rpc/auth.go
@@ -34,7 +34,7 @@ var (
 var invalidIAtFormat = "invalid issued at: %w"
 
 // generateAuthToken generates a JWT token that a client uses to authenticate with the server for restricted endpoints
-// subject is the itentifier of the client for which the token is generated
+// subject is the identifier of the client for which the token is generated
 func generateAuthToken(subject string, p []permission) (string, error) {
 	token := jwt.New(jwt.SigningMethodHS256)
 	claims := token.Claims.(jwt.MapClaims)
--- a/rpc/auth.go
+++ b/rpc/auth.go
@@ -12,6 +12,7 @@ var rpcPK = []byte("rpcPK")
 type permission string
 
 const (
+	permNone permission = "none"
 	permRead permission = "read"
 	permSign permission = "sign"
 )
--- a/rpc/auth.go
+++ b/rpc/auth.go
@@ -46,7 +46,7 @@ func generateAuthToken(subject string, p []permission) (string, error) {
 }
 
 // checkTokenValidity takes a JWT token, verifies that the token is valid and that the token contains the required permission
-func checkTokenValidity(tokenString string, requiredPermission permission, validDuration *time.Duration) error {
+func checkTokenValidity(tokenString string, requiredPermission permission, validDuration time.Duration) error {
 	if requiredPermission == permNone {
 		return nil
 	}
@@ -74,13 +74,7 @@ func checkTokenValidity(tokenString string, requiredPermission permission, valid
 		return fmt.Errorf(invalidIAtFormat, err)
 	}
 
-	var duration time.Duration
-	if validDuration == nil {
-		duration = time.Hour
-	} else {
-		duration = *validDuration
-	}
-	if time.Now().After(iAt.Add(duration)) {
+	if time.Now().After(iAt.Add(validDuration)) {
 		return errExpiredToken
 	}
 
--- a/rpc/auth.go
+++ b/rpc/auth.go
@@ -2,6 +2,7 @@ package rpc
 
 import (
 	"errors"
+	"fmt"
 	"time"
 
 	"github.com/golang-jwt/jwt/v5"
@@ -24,10 +25,13 @@ var allPermissions = []permission{permRead, permSign}
 var (
 	errInvalidSigningMethod = errors.New("invalid signing method")
 	errInvalidToken         = errors.New("invalid token")
+	errExpiredToken         = errors.New("token has expired")
 	errInvalidPermissions   = errors.New("token has invalid permissions")
 	errInvalidPermission    = errors.New("token has an invalid permission")
 	errMissingPermission    = errors.New("token is missing permission")
 )
+
+var invalidIAtFormat = "invalid issued at: %w"
 
 // generateAuthToken generates a JWT token that a client uses to authenticate with the server for restricted endpoints
 func generateAuthToken(p []permission) (string, error) {
@@ -41,7 +45,7 @@ func generateAuthToken(p []permission) (string, error) {
 }
 
 // verifyPermission takes a JWT token, verifies that the token is valid and that the token contains the required permission
-func checkTokenValidity(tokenString string, requiredPermission permission) error {
+func checkTokenValidity(tokenString string, requiredPermission permission, validDuration *time.Duration) error {
 	if requiredPermission == permNone {
 		return nil
 	}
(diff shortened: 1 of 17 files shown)


lalexgap:
> Do not store the token signing key in code.

Do we know what we'll use for the signing key? Could we derive a new key based on our SC address so we don't have to store another key?

rpc/auth.go
```diff
@@ -0,0 +1,107 @@
...
+	claims := token.Claims.(jwt.MapClaims)
+
+	// Check expiration
+	iAt, err := claims.GetIssuedAt()
+	if err != nil {
+		return fmt.Errorf(invalidIAtFormat, err)
+	}
+
+	var duration time.Duration
+	if validDuration == nil {
```
lalexgap, line 78: I'd prefer to just return an error if no `validDuration` is provided instead of using a default of 1 hour . Looking at the function signature it's not obvious what calling `checkTokenValidity` with a nil `validDuration` would without inspecting the implementation.

lalexgap:
Generally this looks good and makes sense to me! 🚀

rpc/auth.go
```diff
@@ -0,0 +1,107 @@
...
+	claims := token.Claims.(jwt.MapClaims)
+	claims[permissionKey] = p
+	// the keys are defined by https://datatracker.ietf.org/doc/html/rfc7519
+	claims["iat"] = time.Now().Unix()
+	claims["sub"] = subject
+	return token.SignedString(rpcPK)
+}
+
+// verifyPermission takes a JWT token, verifies that the token is valid and that the token contains the required permission
+func checkTokenValidity(tokenString string, requiredPermission permission, validDuration *time.Duration) error {
```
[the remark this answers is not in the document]
kerzhner: Good catch!

kerzhner:
> Do we know what we'll use for the signing key? Could we derive a new key based on our SC address so we don't have to store another key?

That's a neat idea. Invoking RPC apis that sign new states is equivalent to controlling the state channel address. So the auth token key should have the same level of security as the state channel private key.

rpc/auth.go
kerzhner: Good point. Refactored in e2475ec0 to remove default values.

rpc/auth.go
```diff
@@ -0,0 +1,101 @@
...
+	errExpiredToken         = errors.New("token has expired")
+	errInvalidPermissions   = errors.New("token has invalid permissions")
+	errInvalidPermission    = errors.New("token has an invalid permission")
+	errMissingPermission    = errors.New("token is missing permission")
+)
+
+var invalidIAtFormat = "invalid issued at: %w"
+
+// generateAuthToken generates a JWT token that a client uses to authenticate with the server for restricted endpoints
+// subject is the itentifier of the client for which the token is generated
```
geoknee, line 37:
```suggestion
// subject is the identifier of the client for which the token is generated
```

rpc/client.go
```diff
@@ -292,11 +314,22 @@ func (rc *rpcClient) PaymentChannelUpdatesChan(paymentChannelId types.Destinatio
 	return c
 }
 
-func waitForRequest[T serde.RequestPayload, U serde.ResponsePayload](rc *rpcClient, method serde.RequestMethod, requestData T) (U, error) {
+// unauthenticatedWaitForRequest calls waitForRequest with an empty auth token
+func unauthenticatedWaitForRequest[T serde.RequestPayload, U serde.ResponsePayload](rc *rpcClient, method serde.RequestMethod, requestData T) (U, error) {
+	return waitForRequest[T, U](rc, method, requestData, "")
+}
+
+// authenticatedWaitForRequest calls waitForRequest with an empty auth token
```
geoknee, line 322: I think there must be a typo on this line?

rpc/client.go
```diff
@@ -101,11 +103,20 @@ type response[T serde.ResponsePayload] struct {
...
 	ctx, cancel := context.WithCancel(context.Background())
-
-	c := &rpcClient{trans, &safesync.Map[chan struct{}]{}, &safesync.Map[chan query.LedgerChannelInfo]{}, &safesync.Map[chan query.PaymentChannelInfo]{}, cancel, &sync.WaitGroup{}, common.Address{}, slog.Default()}
+	c := &rpcClient{
+		transport:             trans,
+		completedObjectives:   &safesync.Map[chan struct{}]{},
+		ledgerChannelUpdates:  &safesync.Map[chan query.LedgerChannelInfo]{},
+		paymentChannelUpdates: &safesync.Map[chan query.PaymentChannelInfo]{},
+		cancel:                cancel,
+		wg:                    &sync.WaitGroup{},
```
geoknee, line 112: Slightly out of scope, but I wonder if we should rename this wait group (the one you added has a good name).

rpc/client.go
```diff
@@ -101,11 +103,20 @@ type response[T serde.ResponsePayload] struct {
...
+		nodeAddress:           common.Address{},
+		authTokenReady:        &sync.WaitGroup{},
+		logger:                slog.Default(),
+	}
 
 	// Retrieve the address and set it on the rpcClient
-	res, err := waitForRequest[serde.NoPayloadRequest, common.Address](c, serde.GetAddressMethod, serde.NoPayloadRequest{})
+	res, err := authenticatedWaitForRequest[serde.NoPayloadRequest, common.Address](c, serde.GetAddressMethod, serde.NoPayloadRequest{})
```
geoknee, line 119: Nit: slightly confusing to call an authorized request method here before getting the auth token. Inside, it will wait on the `c.authToken` wait group which has not yet been "primed" by adding to it. Could/should we replace this call with an `unauthorizedWaitForRequest` to cut out that step?

geoknee (approved):

rpc/client.go
kerzhner: Good catch. Fixed in cf18e248.

rpc/client.go
```diff
@@ -292,11 +314,22 @@ func (rc *rpcClient) PaymentChannelUpdatesChan(paymentChannelId types.Destinatio
...
 }
 
-func waitForRequest[T serde.RequestPayload, U serde.ResponsePayload](rc *rpcClient, method serde.RequestMethod, requestData T) (U, error) {
+// unauthenticatedWaitForRequest calls waitForRequest with an empty auth token
+func unauthenticatedWaitForRequest[T serde.RequestPayload, U serde.ResponsePayload](rc *rpcClient, method serde.RequestMethod, requestData T) (U, error) {
+	return waitForRequest[T, U](rc, method, requestData, "")
+}
+
+// authenticatedWaitForRequest calls waitForRequest with an empty auth token
+func authenticatedWaitForRequest[T serde.RequestPayload, U serde.ResponsePayload](rc *rpcClient, method serde.RequestMethod, requestData T) (U, error) {
```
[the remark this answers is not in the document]
kerzhner: Good point! Done in cf18e248.

rpc/client.go
kerzhner: This is a good point. Switched to unauthenticated request in 7df7f7ea.

kerzhner:
With typescript client changes, this PR passes our CI. Even though this PR is approved, it could be good to get a second pair of eyes on the typescript changes.

rpc/serde/jsonrpc.go
```diff
@@ -82,11 +87,16 @@ type NotificationPayload interface {
 		query.LedgerChannelInfo
 }
 
+type Params[T RequestPayload | NotificationPayload] struct {
+	AuthToken string `json:"authtoken"`
```
lalexgap, line 91: Have you considered the RPC server accepting  the JWT token in using the `Authorization` header ([like infura does](https://docs.infura.io/learn/json-web-token-jwt))? I think that's the "standard" way of including a token in an http request.  Although I guess we do also have to support NATS, so that couldn't be the only way to accept an auth token.

rpc/auth.go
```diff
@@ -0,0 +1,101 @@
...
+	if err != nil {
+		return fmt.Errorf(invalidIAtFormat, err)
+	}
+
+	if time.Now().After(iAt.Add(validDuration)) {
+		return errExpiredToken
+	}
+
+	// Check permissions
+	permissions, ok := claims[permissionKey].([]interface{})
```
lalexgap, line 82: It I think we could avoid a second cast on line 88 if we just cast this to a slice of strings here?
```suggestion
	permissions, ok := claims[permissionKey].([]string)
```

rpc/client.go
```diff
@@ -138,24 +152,32 @@ func (rc *rpcClient) Address() (common.Address, error) {
 	return rc.nodeAddress, nil
 }
 
+// AuthToken fetches an authentication token from the nitro node
+func (rc *rpcClient) AuthToken() (string, error) {
+	if rc.authToken == "" {
```
lalexgap, line 157: Could we follow the same pattern as [Address](https://github.com/statechannels/go-nitro/blob/7df7f7ea59f544510b015afdcb2a3564ecb4add6/rpc/client.go#L152C16-L152C16) where `AuthToken` just returns the private variable, and we rely on the constructor to set it?

rpc/serde/jsonrpc.go
kerzhner: HTTP auth headers were the initial implementation. I decided to include the tokens in the RPC params to unify how tokens are sent for all transports.

rpc/auth.go
kerzhner: This is a good idea, but fails in practice. `claims` map values are of `interface{}` type. First, we have to assert the slice type for a value. Then we have to convert each element of the slice to a string.

rpc/client.go
```diff
@@ -118,10 +129,13 @@ func NewRpcClient(trans transport.Requester) (RpcClientApi, error) {
 	if err != nil {
 		return nil, err
 	}
-	c.wg.Add(1)
+	c.routineTracker.Add(1)
 	go c.subscribeToNotifications(ctx, notificationChan)
 
-	return c, nil
+	c.authTokenReady.Add(1)
```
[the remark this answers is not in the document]
kerzhner: Good catch. The wait group was needed in an earlier implementation. But it's no longer necessary. Removed in 85fb634b.

rpc/client.go
kerzhner: Agreed. Cleaned up in 38a65469.

PR #1607, opened 2023-08-29
Fix rpcClient.Address() race condition

bitwiseguy:
Fix `rpcClient.Address()` race condition that caused a CI failure when changes were merged into `main`: https://github.com/statechannels/go-nitro/actions/runs/6012869180/job/16309076480

### DIFF
--- a/main.go
+++ b/main.go
@@ -178,7 +178,7 @@ func main() {
 
 			logging.SetupDefaultLogger(os.Stdout, slog.LevelDebug)
 
-			node, _, _, _, err := node.InitializeNode(pkString, chainOpts, useDurableStore, durableStoreFolder, msgPort, os.Stdout, peerSlice)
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -178,7 +178,7 @@ func main() {
 
 			logging.SetupDefaultLogger(os.Stdout, slog.LevelDebug)
 
-			node, _, _, _, err := node.InitializeNode(pkString, chainOpts, useDurableStore, durableStoreFolder, msgPort, os.Stdout, peerSlice)
+			node, _, _, _, err := node.InitializeNode(pkString, chainOpts, useDurableStore, durableStoreFolder, msgPort, peerSlice)
```
kerzhner, line 181: Just checking, but it seems like this change is unrelated to the race condition. Is that correct?

kerzhner (approved):
Nicely done!

main.go
bitwiseguy: Yes it is unrelated to the race condition. Just something I noticed when scanning through the code. Its related to the PR that was merged this morning

PR #1609, opened 2023-08-29
Check for missed events when `chainservice` initializes

bitwiseguy:
The `chainservice` should check for any `chainEvents` that occurred since the last time it was online. This will allow us to restart a nitro node without risk of the node's state getting permanently out-of-sync with the on-chain state.


node/engine/chainservice/eth_chainservice.go
```diff
@@ -387,6 +420,7 @@ func (ecs *EthChainService) subscribeForLogs() (chan error, ethereum.Subscriptio
 	// Subscribe to Adjudicator events
 	eventQuery := ethereum.FilterQuery{
 		Addresses: []common.Address{ecs.naAddress},
+		Topics:    [][]common.Hash{topicsToWatch},
```
bitwiseguy, line 423: The addition of `Topics` creates a more targeted event subscription so we don't have to filter out unwanted topics later.

internal/chain/chain.go
```diff
@@ -20,28 +19,13 @@ import (
...
 type ChainOpts struct {
-	ChainUrl       string
-	ChainAuthToken string
-	ChainPk        string
-	NaAddress      common.Address
-	VpaAddress     common.Address
-	CaAddress      common.Address
-}
-
-func InitializeEthChainService(chainOpts ChainOpts) (*chainservice.EthChainService, error) {
```
bitwiseguy, line 31: This function is no longer needed since the `chainservice.NewEthChainService` now accepts `chainOpts` as input.

node/engine/chainservice/simulated_backend_chainservice_test.go
```diff
@@ -147,6 +150,34 @@ func TestSimulatedBackendChainService(t *testing.T) {
...
+			t.Fatalf("Expected chain event to be DepositedEvent")
+		}
+	}
+
+	receivedEvent = <-cs2.EventFeed()
+	if diff := cmp.Diff(expectedConcludedEvent, receivedEvent, cmp.AllowUnexported(ConcludedEvent{}, commonEvent{})); diff != "" {
+		t.Fatalf("Received event did not match expectation; (-want +got):\n%s", diff)
+	}
+
+	receivedEvent = <-cs2.EventFeed()
```
lalexgap, line 177: Would it be worth dispatching a new event and making sure that our `cs2` also successfully picks up new blocks after starting late?

node/engine/chainservice/eth_chainservice.go
```diff
@@ -123,7 +132,12 @@ func newEthChainService(chain ethChain, na *NitroAdjudicator.NitroAdjudicator,
 		return nil, err
 	}
 
-	// TODO: Return error from chain service instead of panicking
+	// Search for any missed events emitted while this node was offline
+	err = ecs.checkForMissedEvents(startBlock)
```
lalexgap, line 136: I think in theory there's a risk here that a block (or blocks) could get mined while `checkForMissedEvents` is executing, before we start listening for new events in `ecs.listenForEventLogs`. This means the events in those blocks would be "ignored". In practice this is incredibly unlikely since `checkForMissedEvents` is just pushing events into the `event-tracker` and thus is really fast. Blocks are mined at a pretty slow pace in comparison, so it's very unlikely a block will be mined in between `checkForMissedEvents` and when we start the listener.

Maybe it's worth starting the listeners before calling `checkMissedEvents`  just to be on the safe side? Once we know we're listening to any new events we can call `checkMissedEvents` to fetch the missed ones.

node/engine/chainservice/eth_chainservice.go
```diff
@@ -132,6 +146,31 @@ func newEthChainService(chain ethChain, na *NitroAdjudicator.NitroAdjudicator,
 	return &ecs, nil
 }
 
+func (ecs *EthChainService) checkForMissedEvents(startBlock uint64) error {
+	ecs.logger.Info("checking for missed chainservice events", "startBlock", startBlock)
+	query := ethereum.FilterQuery{
```
lalexgap, line 151: Totally optional but now that we're constructing a `FilterQuery` in two places we could consider adding a helper function to create a `FilterQuery`

node/engine/chainservice/eth_chainservice.go
bitwiseguy: Yes thats a valid concern. I was thinking about it as part of [this adr](https://github.com/statechannels/go-nitro/blob/b82b0dd1aef34b0fa84cf6ed02b8ef99f46be605/.adr/0017-missed-chain-events.md#future-considerations)). 

My counter-concern if we start `go ecs.listenForEventLogs` before `ecs.checkForMissedEvents` is this scenario (which is probably less likely to occur, but ):

1. Block 100 mined
2. `listenForEventLogs` queues events `AE25`
3. `checkForMissedEvents` starts queuing events `AE1`, `AE2`, etc.
4. Block 101 mined
5. `checkForMissedEvents` queues events `AE3`, `AE4`
6. Block 102 mined
7. `dispatchChainEvents([AE1, AE2, AE3, AE4, AE25])` is triggered
8. `checkForMissedEvents` queues events `AE5`, `AE6` and completes
9. `dispatchChainEvents([AE5, AE6])` is triggered
10. `AE5` and `AE6` can't be processed if they're related to same channel as `AE25`

This is a very low concern if the time between blocks is long because `checkForMissedEvents` should be fast. But I just wanted to mention it in case there is a scenario like this:

* blockchain with fast block times
* `checkForMissedEvents` takes a long time to complete because it has to search through many blocks (nitro node has been off for a long time)

node/engine/chainservice/simulated_backend_chainservice_test.go
bitwiseguy: Yes that would be a more fully-featured test. Will add that

node/engine/chainservice/eth_chainservice.go
lalexgap: > This is a very low concern if the time between blocks is long because `checkForMissedEvents` should be fast. But I just wanted to mention it in case there is a scenario like this:
> 
>     * blockchain with fast block times
> 
>     * `checkForMissedEvents` takes a long time to complete because it has to search through many blocks (nitro node has been off for a long time)

Unfortunately this is also the exact scenario where we would expect to miss events when calling `checkForMissedEvents` before starting `ecs.listenForEventLog`.

I think the "full solution" may involve listening for events before `checkForMissedEvents` but only dispatching events (in order) after `checkForMissedEvents` completes. That way we're protected against missing events but enforce that events are dispatched in the correct order.

node/engine/chainservice/simulated_backend_chainservice_test.go
bitwiseguy: Added in https://github.com/statechannels/go-nitro/pull/1609/commits/c867cef4d4b2e791b9430b93e4aed52f84406252

node/engine/chainservice/eth_chainservice.go
bitwiseguy: Could we use the mutex on the eventTracker as a way to ensure events are NOT dispatched before `checkForMissedEvents` finishes queuing old events? Something like this:

```
	// Prevent go routines from processing events before checkForMissedEvents completes
	ecs.eventTracker.mu.Lock()
	
	ecs.wg.Add(3)
	go ecs.listenForEventLogs(errChan, eventSub, eventChan, eventQuery)
	go ecs.listenForNewBlocks(errChan, newBlockSub, newBlockChan)
	go ecs.listenForErrors(errChan)

	// Search for any missed events emitted while this node was offline
	err = ecs.checkForMissedEvents(startBlock)
	if err != nil {
		return nil, err
	}
	ecs.eventTracker.mu.Unlock()
```

`listenForEventLogs` calls `updateEventTracker`, which also tries to acquire the `mu.Lock()`. That should be blocked until the `mu.Unlock()` is called in the init routine code shown above.

bitwiseguy: Added above implementation in https://github.com/statechannels/go-nitro/pull/1609/commits/edade6ed43a5fce06386c6345afff7e81d7917f5

node/engine/chainservice/eth_chainservice.go
bitwiseguy: The two instances of `FilterQuery` that we currently use are different enough that I don't think a helper function accomplishes much. I propose we add a helper function in the future if we have more instances of `FilterQuery` and the helper function saves us lines of code

node/engine/chainservice/eth_chainservice.go
lalexgap: Nice solution 🚀 ! By locking on the mutex we block the listeners from calling `updateEventTracker`, but since the `checkForMissedEvents` doesn't acquire a mutex it will still be able to write to the event tracker.

I think it's beyond the scope of this PR but we may want to consider refactoring the `eventTracker` so it exposes functions (`AddEvent`,`StartDispatchingEvents`) instead of requiring the chain service to deal with the event heap and acquiring mutexes. The chain service is already quite complex, so I think it would be a win to refactor some of the complexity into another struct.

lalexgap (approved):

node/engine/chainservice/eth_chainservice.go
bitwiseguy: That's a good refactor suggestion. I'll address in a follow-up PR

PR #1610, opened 2023-08-29
Enforce fair outcome in direct and virtual fund objectives

### LINKED ISSUE #1608: Add "fair" outcome checks to funding protocols
What makes the [fullAmount, 0] outcome "fair"? Is it likely anyone would want to _not_ have this policy protecting them? An alternative is to bake the check in to the protocol itself so it runs automatically (if it is that important)?

_Originally posted by @geoknee in https://github.com/statechannels/go-nitro/pull/1599#discussion_r1308475804_

lalexgap:
Fixes #1608 
This updates the protocols to enforce a "fair" outcome based on the some of the [policy work](https://github.com/statechannels/go-nitro/pull/1599#discussion_r1308475804).

- Direct fund objectives now check that the each participant in the outcome provides the same amount.
- Virtual fund objectives now checks that the `payer` starts with the full balance.

By enforcing these outcome restrictions in our protocols our nitro clients are protected from accepting an objective with an unfavourable or less efficient outcome.

### DIFF
--- a/protocols/directfund/directfund.go
+++ b/protocols/directfund/directfund.go
@@ -110,6 +110,19 @@ func ChannelsExistWithCounterparty(counterparty types.Address, getChannels GetCh
 	return ok, nil
 }
 
+// hasEqualOutcome returns true if the outcome allocates an equal amount to each participant
+func hasEqualOutcome(s state.State, me types.Address) bool {
+	for _, e := range s.Outcome {
+		forMe := e.TotalAllocatedFor(types.AddressToDestination(me))
+		for _, a := range e.Allocations {
+			if a.Amount.Cmp(forMe) != 0 {
+				return false
+			}
+		}
+	}
+	return true
+}
+
 // ConstructFromPayload initiates a Objective with data calculated from
 // the supplied initialState and client address
 func ConstructFromPayload(
@@ -124,6 +137,7 @@ func ConstructFromPayload(
 		return Objective{}, fmt.Errorf("could not get signed state payload: %w", err)
 	}
 	initialState := initialSignedState.State()
+
 	err = initialState.FixedPart().Validate()
 	if err != nil {
 		return Objective{}, err
@@ -135,6 +149,9 @@ func ConstructFromPayload(
 		return Objective{}, errors.New("attempted to initiate new direct-funding objective with IsFinal == true")
 	}
 
+	if !hasEqualOutcome(initialState, myAddress) {
+		return Objective{}, errors.New("attempted to initiate new direct-funding objective with non-equal outcome")
+	}
 	init := Objective{}
 
 	if preApprove {
--- a/protocols/directfund/directfund_test.go
+++ b/protocols/directfund/directfund_test.go
@@ -59,7 +59,7 @@ func TestNew(t *testing.T) {
 	request := NewObjectiveRequest(
 		testState.Participants[1],
 		testState.ChallengeDuration,
-		testState.Outcome,
+		testState.Outcome.Clone(),
 		0,
 		testState.AppDefinition,
 	)
@@ -69,7 +69,7 @@ func TestNew(t *testing.T) {
 	}
 
 	getByParticipantHasChannel := func(id types.Address) ([]*channel.Channel, error) {
-		c, _ := channel.New(testState, 0)
+		c, _ := channel.New(testState.Clone(), 0)
 		return []*channel.Channel{c}, nil
 	}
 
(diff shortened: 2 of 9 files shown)


protocols/directfund/directfund.go
```diff
@@ -110,6 +110,19 @@ func ChannelsExistWithCounterparty(counterparty types.Address, getChannels GetCh
 	return ok, nil
 }
 
+// hasEqualOutcome returns true if the outcome allocates an equal amount to each participant
+func hasEqualOutcome(s state.State, me types.Address) bool {
```
geoknee, line 114: `hasUniformOutcome` is better, I think?

protocols/directfund/directfund.go
```diff
@@ -135,6 +149,9 @@ func ConstructFromPayload(
 		return Objective{}, errors.New("attempted to initiate new direct-funding objective with IsFinal == true")
 	}
 
+	if !hasEqualOutcome(initialState, myAddress) {
+		return Objective{}, errors.New("attempted to initiate new direct-funding objective with non-equal outcome")
+	}
```
geoknee, line 154: Follow on changes here if you like "uniform outcome".

protocols/directfund/directfund_test.go
```diff
@@ -83,10 +83,15 @@ func TestNew(t *testing.T) {
 	if _, err := NewObjective(request, false, testState.Participants[0], big.NewInt(TEST_CHAIN_ID), getByParticipant, getByConsensusHasChannel); err == nil {
 		t.Errorf("Expected an error when constructing with an objective when an existing channel consensus channel exists")
 	}
+
+	request.Outcome[0].Allocations[0].Amount = big.NewInt(10)
+	if _, err := NewObjective(request, false, testState.Participants[0], big.NewInt(TEST_CHAIN_ID), getByParticipant, getByConsensus); err == nil {
+		t.Errorf("Expected an error when constructing a direct fund objective with an unfair outcome")
```
geoknee, line 89:
```suggestion
		t.Errorf("Expected an error when constructing a direct fund objective with an non-uniform outcome")
```

protocols/virtualfund/virtualfund.go
```diff
@@ -132,6 +133,21 @@ type Objective struct {
...
 }
 
+// onlyPayerHasFunds checks that the outcome only has funds allocated to the payer
+func onlyPayerHasFunds(s state.State) bool {
+	for _, e := range s.Outcome {
+		total := e.TotalAllocated()
+		for i, a := range e.Allocations {
+			if i == payments.PAYER_INDEX && a.Amount.Cmp(total) != 0 {
+				return false
+			} else if i != payments.PAYER_INDEX && a.Amount.Cmp(big.NewInt(0)) != 0 {
```
geoknee, line 143: Super nit but you don't need the `else` here, and I believe it is encouraged to omit it (saves some indentation).

protocols/virtualfund/helpers_test.go
```diff
@@ -10,16 +10,16 @@ import (
...
+	return prepareConsensusChannelHelper(role, leader, follower, leftActor, 5, 5, 1, guarantees...)
 }
 
 // consensusStateSignatures prepares a consensus channel with a consensus outcome and returns the signatures on the consensus state
 func consensusStateSignatures(leader, follower testactors.Actor, guarantees ...consensus_channel.Guarantee) [2]state.Signature {
-	return prepareConsensusChannelHelper(0, leader, follower, leader, 0, 0, 2, guarantees...).Signatures()
+	return prepareConsensusChannelHelper(0, leader, follower, leader, 0, 5, 2, guarantees...).Signatures()
 }
 
 func prepareConsensusChannelHelper(role uint, leader, follower, leftActor testactors.Actor, leftBalance, rightBalance, turnNum int, guarantees ...consensus_channel.Guarantee) *consensus_channel.ConsensusChannel {
```
geoknee, line 25: Out of scope nit: I wonder if there's a more useful way to name this function. "Helper" doesn't really tell us anything...

geoknee (commented):
I think I'm on board with restricting virtual channels in the way you have here. 

I am far less sure about the ledger channel restriction. Why do you think it is important for each party in the ledger channel to have the same initial allocation (i.e. for there to be a uniform distribution)?

lalexgap:
> I am far less sure about the ledger channel restriction. Why do you think it is important for each party in the ledger channel to have the same initial allocation (i.e. for there to be a uniform distribution)?

By enforcing this rule we rule out some ledger scenarios that we know we're not interested in.For example, we wouldn't want to join a ledger channel where we're the only one contributing to the balance, since that ledger channel won't be that useful for us.

Is there a use case where we would want to open a ledger a channel with a non-uniform balance?

lalexgap:
> > I am far less sure about the ledger channel restriction. Why do you think it is important for each party in the ledger channel to have the same initial allocation (i.e. for there to be a uniform distribution)?
> 
> By enforcing this rule we rule out some ledger scenarios that we know we're not interested in.For example, we wouldn't want to join a ledger channel where we're the only one contributing to the balance, since that ledger channel won't be that useful for us.
> 
> Is there a use case where we would want to open a ledger a channel with a non-uniform balance?

Stand-up discussion brought up the point that depending on our role we may be interested in opening ledger channels with unequal balances. Enforcing a uniform balance seems more like an optional policy and not something that should be baked into the protcol.

lalexgap:
Closing for now, will re-open when ready.

PR #1621, opened 2023-08-31
Reduce log noise for start-rpc-servers

kerzhner:
- Introduce trace log level
- Use trace log level in the ethereum chain service for new block logs and resubscriptions
- Run Anvil in silent mode

### DIFF
--- a/internal/logging/logging.go
+++ b/internal/logging/logging.go
@@ -11,7 +11,7 @@ import (
 	"github.com/statechannels/go-nitro/types"
 )
 
-const TraceLogLevel = -5
+const LevelTrace slog.Level = -5
 
 // newLogWriter returns a writer for the given logDir and logFile
 // If the log file already exists it will be removed and a fresh file will be created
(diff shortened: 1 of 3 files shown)


internal/logging/logging.go
```diff
@@ -11,6 +11,8 @@ import (
 	"github.com/statechannels/go-nitro/types"
 )
 
+const TraceLogLevel = -5
```
lalexgap, line 14: Nitpicks:
1. We could rename this to `LevelTrace` to follow the naming pattern of the [existing levels](https://pkg.go.dev/log/slog#Level)
2. We could define this as `slog.Level` const instead of `int`
```suggestion
const LevelTrace slog.Level = -5
```

lalexgap (approved):
👍

internal/logging/logging.go
kerzhner: Very good point. Done in cda2af8a.

PR #1623, opened 2023-08-31
Ensure rpc servers are fully initialized before sending requests

bitwiseguy:
While troubleshooting the recent CI failures (which we eventually attributed to a faulty version of `anvil`), I noticed there are a few spots where we could improve the resiliency of our scripts by ensuring the rpc clients/servers are fully initialized before trying to send requests.

### DIFF
--- a/packages/nitro-rpc-client/scripts/client-runner.ts
+++ b/packages/nitro-rpc-client/scripts/client-runner.ts
@@ -14,6 +14,13 @@ import {
 } from "../src/utils";
 
 const clientNames = ["alice", "irene", "bob", "ivan"] as const;
+const clientPortMap: Record<ClientNames, number> = {
+  alice: 4005,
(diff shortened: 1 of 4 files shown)


packages/nitro-rpc-client/scripts/client-runner.ts
```diff
@@ -139,6 +139,7 @@ yargs(hideBin(process.argv))
           waitForRPCServer(4005, yargs.waitduration),
           waitForRPCServer(4006, yargs.waitduration),
           waitForRPCServer(4007, yargs.waitduration),
+          waitForRPCServer(4008, yargs.waitduration),
```
geoknee, line 142: Nit: could make this more robust to future refactors by pushing the ports we used into an array that is read from here?

geoknee (approved):

PR #1624, opened 2023-08-31
Add 0017-missed-chain-events adr

bitwiseguy:
Add adr to document design and thought-process around restart-capable chain event tracking


.adr/0017-missed-chain-events.md
```diff
@@ -0,0 +1,75 @@
+# 0017 Missed Chain Events
+
+## Status
+
+Review
+
+## Context
+
+A user should be able to restart a nitro node without concern that it has missed chain events that caused the node's data to get out of sync with other nodes and the chain state. Similarly, a nitro node should be able to start for the first time and catch up on any relevant chain events that occurred before the node started. To address these concerns, a nitro node should check for any chain events it might have missed since it was last online, and queue those events to be processed as part of the node's initialization routines.
```
geoknee, line 9:
```suggestion

It is critical that a go-nitro node can synchronise pertinent blockchain state, in order to successfully perform off chain protocols and protect against faulty challenges. For efficiency reasons, a node synchronises that state by listening to blockchain events. Missing those events can therefore cause the node's data to get out of sync with other nodes and the chain state. Events might be missed if the node has not been continuously running throughout the lifetime of any particular channel it is a participant in. For example, there may have been a period of downtime or loss of internet due to power loss or maintenance.

To address these concerns, a nitro node should check for any chain events it might have missed since it was last online, and queue those events to be processed as part of the node's initialization routines.
```

geoknee (commented):
Forgot to click "submit" on this review!

.adr/0017-missed-chain-events.md
```diff
@@ -0,0 +1,75 @@
...
+
+If the `chainservice` communicates with the `engine` when new blocks are processed, the node can store the `lastBlockSeen` in the `durablestore`. Then, when the node is initialized it can read the `lastBlockSeen` from the `store`, and search for any chain events that occurred between that block and the current block.
```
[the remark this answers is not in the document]
bitwiseguy: Yes the `memstore` just checks from genesis every time when restarting

.adr/0017-missed-chain-events.md
```diff
@@ -0,0 +1,75 @@
...
+	}
+	return nil, fmt.Errorf("chain event older than channels last update")
+}
+```
+
+## Future considerations
+
+1. Is there any situation where an older chain event SHOULD be allowed to update a `Channel` (i.e. chain re-org)?
+
+2. Is there a chance any events are missed between `ecs.checkForMissedEvents(startBlock)` and `go ecs.listenForEventLogs(errChan, eventSub, eventChan, eventQuery)` during the `chainservice` initialization?
```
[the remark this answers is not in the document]
bitwiseguy: This bullet has now been removed and the design details section has been updated to explain how we are doing what you suggest: start listening for event logs first, then go back and get historical logs (using the `eventTracker` mutex to safely allow these things to happen simultaneously). https://github.com/statechannels/go-nitro/blob/missed-chain-events-adr/.adr/0017-missed-chain-events.md?plain=1#L73

.adr/0017-missed-chain-events.md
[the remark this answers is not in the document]
geoknee: Ok, perhaps let us put that in the ADR then. Does the `engine` need to know about what kind of store it is dealing with? It only has an `interface` right?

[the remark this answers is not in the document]
bitwiseguy: That's correct, the `engine` will just call the `store`'s interface, regardless of which `store` type is being used. During the `InitializeNode` function, we call `store.GetLastBlockNumSeen()`. For the `memstore` this will always return `0`. https://github.com/statechannels/go-nitro/blob/main/internal/node/node.go#L36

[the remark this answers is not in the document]
bitwiseguy: Added sentence to explain `memstore` init: https://github.com/statechannels/go-nitro/pull/1624/commits/201969ba1cf3f96300a2affdfcd4bcd0b0ea2d61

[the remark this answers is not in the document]
geoknee: Thanks!

.adr/0017-missed-chain-events.md
```diff
@@ -0,0 +1,127 @@
...
+			res, err = e.handleChainEvent(chainEvent)
+		case <-blockTicker.C:
+			blockNum := e.chain.GetLastConfirmedBlockNum()
+			err = e.store.SetLastBlockNumSeen(blockNum)
+	}
+```
+
+## Future considerations
+
+1. Is there any situation where an older chain event SHOULD be allowed to update a `Channel` (i.e. chain re-org)? We could allow a way to forcibly process an old chain event, or we could assume that the node is fully protected against any situation that would necessitate that action.
```
lalexgap, line 127: > We could allow a way to forcibly process an old chain event,
I'm not sure if that would be enough to handle chain re-orgs, since we'd also have to handle a case where an event happens and then after the chain-org the event didn't happen. So we'd need a mechanism to detect an event no longer exists and "undo" it.

I think there may also be an inherit risk to chain re-orgs that we can really only solve by waiting? That's because chain events can trigger nitro nodes to exchanged signed states, and once these are sent out we can't really undo them. For example if Bob signs a state with outcome `[A:5,B:5]`  based on-chain channel balance of `10`, he's at risk if a chain re-org occurs and the on-chain channel balance  is actually `5` now.

.adr/0017-missed-chain-events.md
```diff
@@ -0,0 +1,127 @@
...
+
+### 2. Updating `lastBlockNum`: `chainservice` sends write on chan to `engine`
+
+If the `chainservice` sends a new `blockNum` on a chan the `engine` each time a new block is mined, the `engine` can trigger a write to store to update the `lastBlockNum`. However, in environments where blocks are mined fast, writing to the `store` for every block could have a performance impact especially if the `store` updates are part of the `engine`'s main `run` loop.
+
+## Decision
+
+After considering the aforementioned alternatives, the following design decisions were made.
+
+### Avoid init race conditions that would cause missed events
```
lalexgap, line 45: Should this be numbered like the rest of the decisions?

.adr/0017-missed-chain-events.md
```diff
@@ -0,0 +1,127 @@
...
+
+### 1. Put `Channel` data protection logic in `Channel` class
+
+Add logic to the `Channel` class to protect its data from stale updates rather than relying on an external component (`chainservice`) to only feed it sanitized data. This seems less error-prone and it could be useful to add a `Channel.LastUpdated` struct in case a user wants to know when the channel was last changed. The `chainservice` will still be expected to feed chain events in order, but the added `Channel` logic will act as another layer of protection against faulty data updates.
+
+The `Channel` class should keep track of the `TxIndex` associated with each chain event in case there are multiple events related to a single channel within the same block. Also, an `ethereum.FilterQuery` cannot specify a `TxIndex` so it will be more computationally efficient to check the `TxIndex` in the `Channel` class rather than in the `chainservice` logic.
```
lalexgap, line 73: >Also, an `ethereum.FilterQuery` cannot specify a `TxIndex` so it will be more computationally efficient to check the `TxIndex` in the `Channel` class rather than in the `chainservice` logic.

While the `ethereum.FilterQuery` doesn't support querying by `TxIndex` I'm not sure why this makes the chain service less computationally efficient? It seems unlikely that you'd be querying by a specific `TxIndex`.

.adr/0017-missed-chain-events.md
```diff
@@ -0,0 +1,127 @@
...
+  LastChainUpdate ChainUpdateData
+}
+
+type ChainUpdateData struct {
+        BlockNum uint64
+        TxIndex uint64
+}
+
+func (c *Channel) UpdateWithChainEvent(event chainservice.Event) (*Channel, error) {
+	if event.BlockNum() > c.LastChainUpdate.BlockNum && event.TxIndex() > c.LastChainUpdate.TxIndex {
```
lalexgap, line 93: I think this check might have to be tweaked a bit.

- `event.BlockNum() > c.LastChainUpdate.BlockNum` we're ignoring the case where `event.BlockNum==c.LastChainUpdate.BlockNum` which is where `TxIndex` comes into play
- We require the `TxIndex` always be larger then the last, even if the events are in different blocks.
I think we want something like:

```suggestion
	if event.BlockNum() > c.LastChainUpdate.BlockNum || (event.BlockNum() == c.LastChainUpdate.BlockNum) && event.TxIndex() > c.LastChainUpdate.TxIndex {
```

.adr/0017-missed-chain-events.md
```diff
@@ -0,0 +1,127 @@
...
+	err = ecs.checkForMissedEvents(startBlock)
+	if err != nil {
+		return nil, err
+	}
+```
+
+There are a few decisions to make within this design:
+
+1. How does the node ensure it processes chain events idempotently (i.e. how does it ensure the same chain event does not make changes to a `Channel`'s data multiple times?). This situation could occur if the node restarts, scans through blocks for events during initialization, and detects events that it has already processed.
+2. When/how should the `lastBlockNum` be updated in the `store`? When the node is initialized it can read the `lastBlockNumSeen` from the `store`, and search for any chain events that occurred between that block and the current block. If the node uses a `memstore`, then the `lastBlockNumSeen` will always be set to `0` when the node is first initialized. By updating the `lastBlockNum` in `store`, the `chainservice` will not have to scan through as many old blocks when it first initializes and calls `checkForMissedEvents`.
```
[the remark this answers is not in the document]
bitwiseguy: Yes I agree that the `memstore` should only be used in a test setting. I added the sentence about `memstore` based on @geoknee feedback in this comment: https://github.com/statechannels/go-nitro/pull/1624#discussion_r1312902801

.adr/0017-missed-chain-events.md
bitwiseguy: Yes that makes sense about having to "undo" off-chain state changes. I am happy to remove the "Future considerations" section to make this adr cleaner. Would that be preferable or do you think this adr needs more thought and detail about how the node should handle chain re-orgs that defeat the `REQUIRED_BLOCK_CONFIRMATIONS` limit?

.adr/0017-missed-chain-events.md
bitwiseguy: Would you be in favor of removing the one sentence you quoted above?

.adr/0017-missed-chain-events.md
bitwiseguy: Great catch! I will work on a PR that implements the `Channel` class data protections mentioned above and link that here before merging this adr PR

.adr/0017-missed-chain-events.md
lalexgap: I'm happy to remove this section (or even leave it) but I think further thought about chain re-orgs should probably be handled outside of this PR/ADR (due to the complexity).

.adr/0017-missed-chain-events.md
lalexgap: I'd be happy to remove it or reword it? I just don't quite follow why checking the `TxIndex` in the `chainservice` is less efficient than if we do so in the channel.

.adr/0017-missed-chain-events.md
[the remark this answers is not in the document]
geoknee: Agreed that `memstore` should be deprecated, since I think we can recover `memstore`-like functionality by configuring the `durablestore` a certain way anyway?

In my comment I was mostly concerned that the ADR was describing how the `durablestore` (implementation) should work, whereas it would be nice if it just described how a general `store` (interface) should work. I think not a huge deal either way.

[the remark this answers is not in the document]
bitwiseguy: Would you guys prefer that I remove this sentence:
```
If the node uses a `memstore`, then the `lastBlockNumSeen` will always be set to `0` when the node is first initialized.
```
That is the only reference to `memstore` in this adr.

.adr/0017-missed-chain-events.md
bitwiseguy: Removed that section: https://github.com/statechannels/go-nitro/pull/1624/commits/f69ab0eb34a0dcbc8c65c70ac2edfd051a682047

.adr/0017-missed-chain-events.md
bitwiseguy: PR to make the above modifications to the `Channel` class: https://github.com/statechannels/go-nitro/pull/1659

lalexgap (approved):
🌋

.adr/0017-missed-chain-events.md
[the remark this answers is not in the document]
geoknee: I don't feel strongly. Your call @bitwiseguy !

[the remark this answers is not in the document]
bitwiseguy: Removed in https://github.com/statechannels/go-nitro/pull/1624/commits/f69ab0eb34a0dcbc8c65c70ac2edfd051a682047

.adr/0017-missed-chain-events.md
```diff
@@ -0,0 +1,131 @@
...
+	if err != nil {
+		return nil, err
+	}
+```
+
+There are a few decisions to make within this design:
+
+1. How does the node protect against race conditions that could caused missed chain events during initialization?
+2. How does the node ensure it processes chain events idempotently (i.e. how does it ensure the same chain event does not make changes to a `Channel`'s data multiple times?). This situation could occur if the node restarts, scans through blocks for events during initialization, and detects events that it has already processed.
+3. When/how should the `lastBlockNum` be updated in the `store`? When the node is initialized it can read the `lastBlockNumSeen` from the `store`, and search for any chain events that occurred between that block and the current block. If the node uses a `memstore`, then the `lastBlockNumSeen` will always be set to `0` when the node is first initialized. By updating the `lastBlockNum` in `store`, the `chainservice` will not have to scan through as many old blocks when it first initializes and calls `checkForMissedEvents`.
```
geoknee, line 30:
```suggestion
3. By storing a `lastBlockNum` variable in the `store`,  the `chainservice` will not have to scan through as many old blocks when it first initializes and calls `checkForMissedEvents`. When the node is initialized it can read the `lastBlockNumSeen` from the `store`, and search for any chain events that occurred between that block and the current block. If the node uses a `memstore`, then the `lastBlockNumSeen` will always be set to `0` when the node is first initialized. When/how should the `lastBlockNumSeen` be updated?
```

.adr/0017-missed-chain-events.md
```diff
@@ -0,0 +1,131 @@
...
+	// Update Channel.LastChainUpdate
+	c.LastChainUpdate.BlockNum = event.BlockNum()
+	c.LastChainUpdate.TxIndex  = event.TxIndex()
+	return c, nil
+}
+```
+
+### [3] Updating `lastBlockNum`: `engine` periodically reads from `chainservice`
+
+Instead of the `chainservice` alerting the `engine`, the `engine` can request the block number from the `chainservice`. This means the `lastBlockNumSeen` wouldn't get updated every block, but that's not a major concern since the node
```
geoknee, line 116: This sentence trails off?

bitwiseguy: Fixed in https://github.com/statechannels/go-nitro/pull/1624/commits/06b800ff68d376a29ee776e21e9b8a074fcd8ff6

.adr/0017-missed-chain-events.md
bitwiseguy: Change made in https://github.com/statechannels/go-nitro/pull/1624/commits/06b800ff68d376a29ee776e21e9b8a074fcd8ff6

PR #1627, opened 2023-09-02
Rename rpc websocket transport to http

kerzhner:
When rpc was built initially, the websocket protocol was used for requests, responses and notifications. The websocket transport evolved to use http for requests and responses and websockets for notification. This PR renames the transport to http for clarity. A websocket handshake starts with an http requests, so http naming still accommodates our websocket usage.

### DIFF
--- a/rpc/transport/http/client.go
+++ b/rpc/transport/http/client.go
@@ -11,7 +11,7 @@ import (
 	"github.com/gorilla/websocket"
 )
 
-type clientWebSocketTransport struct {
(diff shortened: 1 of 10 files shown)


rpc/transport/http/client.go
```diff
@@ -0,0 +1,87 @@
+package http
```
kerzhner, line 1: This file is unchanged from `ws/client.go` except for renames.

lalexgap (approved):

PR #1631, opened 2023-09-05
Refactor chainservice. Refine logs

bitwiseguy:
Reorganizes some of the `chainservice` code. Adds more relevant info to some logs to give user better idea of `chainservice`'s state

### DIFF
--- a/cmd/deploy-contracts/main.go
+++ b/cmd/deploy-contracts/main.go
@@ -2,11 +2,10 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"log"
 	"os"
-	"os/exec"
 
-	"github.com/statechannels/go-nitro/cmd/utils"
 	"github.com/statechannels/go-nitro/internal/chain"
 	"github.com/urfave/cli/v2"
 )
@@ -20,17 +19,10 @@ const (
 	CHAIN_AUTH_TOKEN = "chainauthtoken"
 	CHAIN_URL        = "chainurl"
 	DEPLOYER_PK      = "chainpk"
-	START_ANVIL      = "startanvil"
 )
 
 func main() {
 	flags := []cli.Flag{
-		&cli.BoolFlag{
-			Name:    START_ANVIL,
-			Usage:   "Specifies whether to start a local anvil instance",
-			Value:   false,
-			Aliases: []string{"a"},
-		},
 		&cli.StringFlag{
 			Name:    CHAIN_AUTH_TOKEN,
 			Usage:   "Specifies the auth token for the chain",
@@ -58,27 +50,16 @@ func main() {
 		Flags: flags,
 
 		Action: func(cCtx *cli.Context) error {
-			running := []*exec.Cmd{}
-			if cCtx.Bool(START_ANVIL) {
-				anvilCmd, err := chain.StartAnvil()
-				if err != nil {
-					utils.StopCommands(running...)
-					panic(err)
-				}
-				running = append(running, anvilCmd)
-			}
-
 			chainAuthToken := cCtx.String(CHAIN_AUTH_TOKEN)
 			chainUrl := cCtx.String(CHAIN_URL)
 			chainPk := cCtx.String(DEPLOYER_PK)
 
 			_, _, _, err := chain.DeployContracts(context.Background(), chainUrl, chainAuthToken, chainPk)
 			if err != nil {
-				utils.StopCommands(running...)
 				panic(err)
 			}
 
-			utils.StopCommands(running...)
+			fmt.Println("Successfully completed!")
 			return nil
 		},
 	}
--- a/cmd/deploy-contracts/main.go
+++ b/cmd/deploy-contracts/main.go
@@ -28,7 +28,7 @@ func main() {
 		&cli.BoolFlag{
 			Name:    START_ANVIL,
 			Usage:   "Specifies whether to start a local anvil instance",
-			Value:   true,
+			Value:   false,
 			Aliases: []string{"a"},
 		},
 		&cli.StringFlag{
@@ -78,7 +78,6 @@ func main() {
 				panic(err)
 			}
 
-			utils.WaitForKillSignal()
 			utils.StopCommands(running...)
 			return nil
 		},
--- a/docker/cloud/config.toml
+++ b/docker/cloud/config.toml
@@ -17,6 +17,7 @@ caaddress = "0x0C9D79725afc344A388045235CD0B23eA4f0E838"
 
 # RPC provider docs: https://lotus.filecoin.io/lotus/developers/glif-nodes/#testnet-endpoint
 chainurl = "wss://wss.calibration.node.glif.io/apigw/lotus/rpc/v0"
+chainstartblock = 886696
 chainauthtoken = ""
 
 bootpeers = ""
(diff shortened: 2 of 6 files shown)


node/engine/chainservice/eth_chainservice.go
```diff
@@ -135,39 +126,46 @@ func newEthChainService(chain ethChain, startBlock uint64, na *NitroAdjudicator.
...
+		ecs.wg.Add(3)
+		go ecs.listenForEventLogs(errChan, eventSub, eventChan, eventQuery)
+		go ecs.listenForNewBlocks(errChan, newBlockSub, newBlockChan)
+		go ecs.listenForErrors(errChan)
+
+		// Search for any missed events emitted while this node was offline
+		err = ecs.checkForMissedEvents(startBlock)
+		if err != nil {
+			return nil, err
+		}
```
geoknee, line 139: Do these braces actually do anything, or are they there just as a guide for the eye?

I removed them and things still seemed to work. I might suggest a function if you want to encapsulate this logic.

docker/cloud/config.toml
```diff
@@ -11,12 +11,13 @@ pk = ""
 # Chain Public Address: 0x16e43DDc5FE417046Da7A0C469cCD4DE6b4764fA
 chainpk = ""
 
-naaddress = "0x86dbCA138C8aF6f9710d17F892bA788e87BeF000"
-vpaaddress = "0x9B955E99c325Be47e4De330F90d36D2E3D4778a9"
-caaddress = "0xC0FA32D0ca8067C9B5b4471Eb41299537E7B6a33"
+naaddress = "0xe32d4B5C5a80660710f6a2aD3cB1c11664138F34"
+vpaaddress = "0x4D1a804e1cE383D75032C52878553212992C98D7"
+caaddress = "0x0C9D79725afc344A388045235CD0B23eA4f0E838"
```
geoknee, line 16: Did we redeploy the contracts?

docker/cloud/config.toml
```diff
@@ -11,12 +11,13 @@ pk = ""
...
 
 # RPC provider docs: https://lotus.filecoin.io/lotus/developers/glif-nodes/#testnet-endpoint
 chainurl = "wss://wss.calibration.node.glif.io/apigw/lotus/rpc/v0"
+chainstartblock = 886696
```
geoknee, line 20: Is this the block where the contracts were deployed?

cmd/deploy-contracts/main.go
```diff
@@ -28,7 +28,7 @@ func main() {
 		&cli.BoolFlag{
 			Name:    START_ANVIL,
 			Usage:   "Specifies whether to start a local anvil instance",
-			Value:   true,
+			Value:   false,
```
geoknee, line 31: Can you explain the reasoning for this change?

cmd/deploy-contracts/main.go
```diff
@@ -78,7 +78,6 @@ func main() {
 				panic(err)
 			}
 
-			utils.WaitForKillSignal()
```
geoknee, line 81: Can you explain the reasoning for this change?

docker/local/config.toml
```diff
@@ -10,12 +10,13 @@ pk = ""
...
-vpaaddress = "0x9B955E99c325Be47e4De330F90d36D2E3D4778a9"
-caaddress = "0xC0FA32D0ca8067C9B5b4471Eb41299537E7B6a33"
+naaddress = "0xe32d4B5C5a80660710f6a2aD3cB1c11664138F34"
+vpaaddress = "0x4D1a804e1cE383D75032C52878553212992C98D7"
+caaddress = "0x0C9D79725afc344A388045235CD0B23eA4f0E838"
 
 # RPC provider docs: https://lotus.filecoin.io/lotus/developers/glif-nodes/#testnet-endpoint
 chainurl = "wss://wss.calibration.node.glif.io/apigw/lotus/rpc/v0"
+chainstartblock = 886696
 chainauthtoken = ""
```
geoknee, line 20: Out of scope, but do we need to specify these keys if they are just empty?

geoknee (approved):

node/engine/chainservice/eth_chainservice.go
bitwiseguy: I just used them to help make the code more readable and easier to understand what is enclosed by the `mutex.Lock()` and `mutex.Unlock()`. I will look at refactoring into a function. I was just hesitant to do that because its a one-time use function that will require 7 arguments

docker/cloud/config.toml
bitwiseguy: Yes

docker/local/config.toml
bitwiseguy: No its not necessary. We could remove `chainauthtoken` from this config file and it should not affect the node at all

node/engine/chainservice/eth_chainservice.go
```diff
@@ -286,7 +281,7 @@ func (ecs *EthChainService) dispatchChainEvents(logs []ethTypes.Log) error {
 				return fmt.Errorf("error in ParseAllocationUpdated: %w", err)
 			}
 
-			tx, pending, err := ecs.chain.TransactionByHash(context.Background(), l.TxHash)
+			tx, pending, err := ecs.chain.TransactionByHash(ecs.ctx, l.TxHash)
```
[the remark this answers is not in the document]
bitwiseguy: On L93 the `ecs` has not been created yet so `ecs.ctx` does not exist at that point

cmd/deploy-contracts/main.go
bitwiseguy: The `anvil` network was starting even when the `CHAIN_URL` was set to Filecoin. It would probably make more sense just to remove the `START_ANVIL` option completely from this script. There are no instances where this script is used with that option set right now. We should just be able to use `CHAIN_URL` to determine where the contracts are deployed, and the user can start `anvil` externally (and keep it running) if thats the network they want to use.

node/engine/chainservice/eth_chainservice.go
geoknee: I see. Not opposed to keeping them for readability. My editor automatically removed those braces when I was tinkering on a branch based on this :-/

cmd/deploy-contracts/main.go
geoknee: Sounds like good cleanup to me, perhaps lets give a chance for folks to object at today's standup and then do that.

node/engine/chainservice/eth_chainservice.go
bitwiseguy: I'll just leave them removed as done by your PR

cmd/deploy-contracts/main.go
bitwiseguy: I went with the more modular approach in https://github.com/statechannels/go-nitro/pull/1631/commits/913eb225e4e3954ef0b18c99bf00591059e33669. That commit removes the code to start an `anvil` network as part of the `deploy-contracts` script. Instead, the user can first start an `anvil` network (or any local network), and pass that network's CHAIN_URL to this script if they want the contracts deployed there.

cmd/deploy-contracts/main.go
bitwiseguy: Removed `anvil` startup from this script altogether: https://github.com/statechannels/go-nitro/pull/1631/commits/913eb225e4e3954ef0b18c99bf00591059e33669

Issue #1626, opened 2023-09-01
nitro-protocol developer friendliness 


kerzhner:
We have the following pain points:
- nitro-protocol is in its own directory instead of in the `packages` directory.
- nitro-protocol uses npm while other typescript parts of the project use yarn.
- nitro-protocol relies on the `--legacy-peer-deps` npm flag.

PR #1637, opened 2023-09-06
Absorb `nitro-protocol` as a yarn workspace

geoknee:
Towards #1626.

I decided to split the work up into
- [x] switch to yarn
- [x] add as a workspace
- [ ] relocate directory ~(we don't even necessarily need to do this one. If we want to, it can be a separate PR)~ #1638


nitro-protocol/package.json
```diff
@@ -23,9 +23,9 @@
     "test": "yarn test:contracts && yarn test:app --all",
     "test:app": "yarn hardhat compile && jest -c ./config/jest/jest.config.js",
     "test:contracts": "yarn hardhat compile && jest -c ./config/jest/jest.contracts.config.js",
-    "lint:check": "yarn eslint \"./\" --cache && yarn solhint:check && yarn prettier:check",
+    "lint:check": "yarn eslint \"./\" --cache && yarn prettier:check",
```
geoknee: If we merge this, let's make an issue to reenable this.

kerzhner:
> * [x]  switch to yarn
> * [x]  add as a workspace
> * [ ]  relocate directory ~(we don't even necessarily need to do this one. If we want to, it can be a separate PR)~ [Move `nitro-protocol` into `packages` directory #1638](https://github.com/statechannels/go-nitro/pull/1638)

We should remember to update the appropriate readme's as well.

kerzhner (approved):

PR #1638, opened 2023-09-06
Move `nitro-protocol` into `packages` directory

### LINKED ISSUE #1626: nitro-protocol developer friendliness

geoknee:
Closes #1626


geoknee:
In favour of #1640

PR #1641, opened 2023-09-06
Continue working on moving nitro protocol

lalexgap:
This PR makes some further changes on #1640 to get things working.

- Updates `nitro-protocol `to use the same version of `jest` as other packages. Previously `nitro-protocol` was built against and older version of `jest`. When running the tests from within a `yarn` workspace` however it looks like a newer version of jest is used. I couldn't find an easy way to use an older version specifically for nitro-protocol so I decided to update nitro-protocol to use the newer version. 
- Adding a `yarn lint` command to `nitro-protocol` so linting is triggered when running `yarn lint` from the root folder.
- Cleaned up a bunch of lint errors in `nitro-protocol`. Not sure how we weren't running into these before...
- Updated the `nitro-protocol` eslint config file to inherit from the workspace config file. This resolved an issue with eslint complaining about two versions of a plugin.
- The "node" github workflow as been renamed and tweaked to only deal with contracts. It will compile the contracts, run the gas benmark, and run the contract tests. The regular yarn workflow will take care of running other tests and linting.


packages/nitro-protocol/test/contracts/BatchOperator/batchDeposits.test.ts
```diff
@@ -16,25 +16,25 @@ const provider = getTestProvider();
 const batchOperator = setupContract(
   provider,
   BatchOperatorArtifact,
-  process.env.BATCH_OPERATOR_ADDRESS
+  process.env.BATCH_OPERATOR_ADDRESS || ''
```
geoknee, line 19:
```suggestion
  process.env.BATCH_OPERATOR_ADDRESS!
```
Works too, and is possibly a bit nicer?

I think somewhere we have some type file to force these as global variables... tagging @NiloCK for input.

lalexgap: I think that's going to violate the [no-null-assertion](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-non-null-assertion.md) linting rule.

packages/nitro-protocol/package.json
```diff
@@ -57,9 +57,6 @@
     "chai": "^4.3.6",
     "dotenv": "^14.3.2",
     "eslint": "7.17.0",
-    "eslint-config-prettier": "^8.5.0",
-    "eslint-plugin-import": "^2.26.0",
-    "eslint-plugin-prettier": "3.3.1",
```
geoknee, line 61: Why remove these?

geoknee (approved):

packages/nitro-protocol/test/contracts/BatchOperator/batchDeposits.test.ts
geoknee: Ah ok. Not a huge deal I think.

packages/nitro-protocol/package.json
lalexgap: Not sure this was actually needed. Reverted in https://github.com/statechannels/go-nitro/pull/1641/commits/c11c846221f5bddc6aad77cee5be1a5dc2af643d

PR #1646, opened 2023-09-07
Reenable solhint

### LINKED ISSUE #1639: Re-enable solhint
If we merge this, let's make an issue to reenable this.

_Originally posted by @geoknee in https://github.com/statechannels/go-nitro/pull/1637#discussion_r1317224695_

geoknee:
Closes #1639


packages/nitro-protocol/package.json
lalexgap, line 1: Could we update the `solhint:write` command? Right now it echoes out it doesn't work:
```
"solhint:write": "echo 'This command does not seem to work. Use yarn prettier:write ' && yarn solhint --fix './contracts/**/*.sol'",
```

lalexgap (approved):

Issue #1600, opened 2023-08-28
Optimize chainservice init function


bitwiseguy:
We should consider updating `lastBlockNumSeen` even when we don't receive adjudicator events in blocks so that our `chainservice` init routine does not have to search all blocks since the last adjudicator event was received.

https://github.com/statechannels/go-nitro/pull/1581#discussion_r1307488870

On the `Calibration` testnet I am running into an issue where this node provider `wss://wss.calibration.node.glif.io/apigw/lotus/rpc/v0` can only look back in time ~16 hours. Updating the `lastBlockNumSeen` for each block would allow us to have a downtime of up to 16 hours using that chain url, regardless of when the last NitroAdjudicator event was processed.

bitwiseguy:
Possible solutions:

# Add a `NewBlock` custom event type
We could add this custom event topic:
```go
var newBlockTopic = crypto.Keccak256Hash([]byte("NewBlockMined"))
```
Then could transform a `newBlock` header into our own custom `ethTypes.Log`:
```go
		case newBlock := <-newBlockChan:
			newBlockNum := newBlock.Number.Uint64()
			ecs.logger.Debug("detected new block", "block-num", newBlockNum)
			newBlockEvent := &ethTypes.Log{
				BlockNumber: newBlockNum,
				Topics:      []common.Hash{newBlockTopic},
			}
			ecs.updateEventTracker(errorChan, newBlockEvent)
```
Then within `dispatchChainEvents` we could add this addition `case` statement:
```go
		case newBlockTopic:
			ecs.logger.Debug("Processing New Block event")
			event := NewBlockEvent{commonEvent{blockNum: l.BlockNumber}}
			ecs.out <- event
```
This would utilize the same `ecs.out` chan to communicate with the engine. We would then have to add additional logic to the engine (likely in the `handleChainEvent` function) to process this event type by simply calling `store.SetLastBlockSeen(event.BlockNum)`. The downside to this approach is we have more noise on the `ecs.out` chan. Not sure if that is a huge concern but it does make it more difficult to test because we have to filter through more events to find the ones we want to check.

# Add a `NewBlock` chan
Instead of adding `NewBlock` events to the existing `ecs.out` chan, we could create a new, dedicated `ecs.newBlock` chan for the `chainservice` to communicate with the `engine`. This means we would probably have to add another `case` statement to the `engine.run` loop to handle info passed on that channel:
```go
		case chainEvent := <-e.fromChain:
[...]

PR #1648, opened 2023-09-07
Get last confirmed block num

bitwiseguy:
Implements the following solution to periodically update the `lastBlockNumSeen` in the store even when no NitroAdjudicator events are being emitted:

> Instead of the chain service alerting the engine, the engine could request the block number from the chain service. Since the engine wouldn't know when a new block is created this would mostly likely be done by the engine on a timer. The engine would also do this when close is called.

> This does means the lastBlockNumSeen wouldn't get updated every block, but I think that's ok. It doesn't matter if the lastBlockNumSeen is slightly stale, we just want to prevent it from being waaaaay out of date.

> I think the benefit of this approach would be the simplicity. Instead of adding a new chan or event we just need to add a BlockNum function to the chain service, and call that in the engine. It also means we wouldn't write to the store every block but every x minutes (and when we close).

Further details and discussion about alternative solutions are part of the following issue: https://github.com/statechannels/go-nitro/issues/1600


lalexgap (approved):
LGTM!

PR #1650, opened 2023-09-08
Add mutex to protect memstore lastBlockNumSeen read/write

bitwiseguy:
Fixes race condition that caused failure on `main`: https://github.com/statechannels/go-nitro/actions/runs/6122226071/job/16617533727

### DIFF
--- a/node/engine/store/memstore.go
+++ b/node/engine/store/memstore.go
@@ -20,8 +20,8 @@ import (
 )
 
 type blockData struct {
-	lastBlockNumSeen uint64
-	mu               sync.Mutex
+	blockNum uint64
+	mu       sync.Mutex
 }
 
 type MemStore struct {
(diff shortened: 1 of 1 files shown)


node/engine/store/memstore.go
```diff
@@ -18,13 +19,18 @@ import (
...
+}
+
 type MemStore struct {
 	objectives         safesync.Map[[]byte]
 	channels           safesync.Map[[]byte]
 	consensusChannels  safesync.Map[[]byte]
 	channelToObjective safesync.Map[protocols.ObjectiveId]
 	vouchers           safesync.Map[[]byte]
-	lastBlockNumSeen   uint64
+	blocks             blockData
```
kerzhner, line 33: Is it more clear to name this `lastBlockSeen` or something like that? `blocks` is somewhat ambiguous...

kerzhner (approved):

node/engine/store/memstore.go
bitwiseguy: Yea I think that's a better name. Changed in https://github.com/statechannels/go-nitro/pull/1650/commits/e854ac45f9cddbe242ec43ba5490460e6e36c139

PR #1651, opened 2023-09-08
Fix `GetLastConfirmedBlockNum` test failures

bitwiseguy:
* Add informative log for `chainstartblock` work-around when `chainservice` init fails
* Add mutex (and use in tests) for `MockChain.BlockNum` to fix `GetLastConfirmedBlockNum` test failures

### DIFF
--- a/node/engine/chainservice/mock_chain.go
+++ b/node/engine/chainservice/mock_chain.go
@@ -13,7 +13,8 @@ import (
 // MockChain mimics the Ethereum blockchain by keeping track of block numbers and account balances in memory.
 // MockChain accepts transactions and broadcasts events.
 type MockChain struct {
-	BlockNum uint64
+	BlockNum   uint64
+	BlockNumMu sync.Mutex
 	// holdings tracks funds for each channel.
 	holdings map[types.Destination]types.Funds
 	// out maps addresses to an Event channel. Given that MockChainServices only subscribe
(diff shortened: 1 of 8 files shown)


node_test/integration_test.go
```diff
@@ -216,7 +216,9 @@ func RunIntegrationTestCase(tc TestCase, t *testing.T) {
 
 		var chainLastConfirmedBlockNum uint64
 		if infra.mockChain != nil {
+			infra.mockChain.BlockNumMu.Lock()
```
kerzhner, line 219: This is a code smell that an abstraction is broken since the test is reaching into the chain and manipulating the mutex.

node/engine/chainservice/mock_chain.go
```diff
@@ -37,6 +38,7 @@ func NewMockChain() *MockChain {
 func (mc *MockChain) SubmitTransaction(tx protocols.ChainTransaction) error {
 	eventsToBroadcast := []Event{}
 	mc.txMutex.Lock()
```
kerzhner, line 40: Is the txMutex needed at this point? `BlockNumMu` seems to protect the same section of code.

kerzhner (approved):

node_test/integration_test.go
bitwiseguy: Removed the direct use of this mutex from the tests. I believe https://github.com/statechannels/go-nitro/pull/1654 solved the issue that this mutex was trying to fix.

node/engine/chainservice/mock_chain.go
bitwiseguy: Removed `txMutex` in favor of `blockNumMu` here: https://github.com/statechannels/go-nitro/pull/1651/commits/be28326341fd26cf4420ee82f7921f4d319c2840

PR #1652, opened 2023-09-08
Introduce health endpoint to the HTTP server

kerzhner:
Use the health endpoint for initializing HttpTransportAsServer.

This is work is needed for introducing TLS. With TLS the HTTP RPC server takes longer to initialize. As a result, rpc clients can start to send requests to the server before the server is ready. In theory, even without adding TLS, `node_test` tests can fail due to clients not waiting for server readiness.

To make matters worse, `node_test` failures due to this issue have symptoms that do not immediately point to the underlying cause:
```
 go-nitro git:(wss) ✗ go test -run TestReversePaymentProxy ./node_test
--- FAIL: TestReversePaymentProxy (0.02s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x88 pc=0x104d45248]

goroutine 69 [running]:
testing.tRunner.func1.2({0x1057d50e0, 0x1066709d0})
        /usr/local/go/src/testing/testing.go:1545 +0x1c8
testing.tRunner.func1()
        /usr/local/go/src/testing/testing.go:1548 +0x360
panic({0x1057d50e0?, 0x1066709d0?})
        /usr/local/go/src/runtime/panic.go:914 +0x218
github.com/statechannels/go-nitro/rpc/transport/http.NewHttpTransportAsClient({0x140001411d0, 0x15})
        go-nitro/rpc/transport/http/client.go:30 +0x78
github.com/statechannels/go-nitro/node_test_test.setupNitroNodeWithRPCClient(0x14000803860, {0x14000046b80, 0x20, 0x40}, 0x14000088b18?, 0x105a72248?, 0x1059766e0?, {0x1050d5ff6, 0x4}, {0x14000439b18, ...})
        go-nitro/node_test/rpc_test.go:440 +0x2fc
github.com/statechannels/go-nitro/node_test_test.setupNitroClients(0x0?, {0x1052588f6, 0x19})
        go-nitro/node_test/reverseproxy_test.go:241 +0x3d4
github.com/statechannels/go-nitro/node_test_test.TestReversePaymentProxy(0x14000803860)
        go-nitro/node_test/reverseproxy_test.go:66 +0x54
testing.tRunner(0x14000803860, 0x105a3b120)
        /usr/local/go/src/testing/testing.go:1595 +0xe8
created by testing.(*T).Run in goroutine 1
        /usr/local/go/src/testing/testing.go:1648 +0x33c
FAIL    github.com/statechannels/go-nitro/node_test     0.406s
FAIL
```

### DIFF
--- a/internal/rpc/rpc.go
+++ b/internal/rpc/rpc.go
@@ -49,7 +49,7 @@ func blockUntilHttpServerIsReady(rpcPort int) error {
 
 	numAttempts := 10
 	for i := 0; i < numAttempts; i++ {
-		resp, err := http.Get(fmt.Sprintf("http://127.0.0.1:%d/health", rpcPort))
+		resp, err := http.Get(fmt.Sprintf("http://:%d/health", rpcPort))
 		if err != nil {
 			waitForServer()
 			continue
--- a/internal/rpc/rpc.go
+++ b/internal/rpc/rpc.go
@@ -3,8 +3,6 @@ package rpc
 import (
 	"fmt"
 	"log/slog"
-	"net/http"
-	"time"
 
 	"github.com/statechannels/go-nitro/node"
 	"github.com/statechannels/go-nitro/rpc"
@@ -23,10 +21,6 @@ func InitializeRpcServer(node *node.Node, rpcPort int, useNats bool) (*rpc.RpcSe
 	} else {
 		slog.Info("Initializing Http RPC transport...")
 		transport, err = httpTransport.NewHttpTransportAsServer(fmt.Sprint(rpcPort))
-		if err != nil {
-			return nil, err
-		}
-		err = blockUntilHttpServerIsReady(rpcPort)
 	}
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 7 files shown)


bitwiseguy (approved):
LGTM! You should be able to get the CI test suite to pass if you merge `main` into this branch. Some test fixes were recently merged into `main`

internal/rpc/rpc.go
```diff
@@ -20,7 +22,11 @@ func InitializeRpcServer(node *node.Node, rpcPort int, useNats bool) (*rpc.RpcSe
 		transport, err = nats.NewNatsTransportAsServer(rpcPort)
 	} else {
 		slog.Info("Initializing Http RPC transport...")
-		transport, err = http.NewHttpTransportAsServer(fmt.Sprint(rpcPort))
+		transport, err = httpTransport.NewHttpTransportAsServer(fmt.Sprint(rpcPort))
+		if err != nil {
+			return nil, err
+		}
+		err = blockUntilHttpServerIsReady(rpcPort)
```
geoknee, line 29: Is it a bit smelly to have the RPC server send GET requests to itself to decide when some part of it is ready? I would have thought the `blockUntilReady` helper would be more naturally used by an RPC Client?

geoknee:
Do you think we should be spreading this around / sharing it with things like this 

https://github.com/statechannels/go-nitro/blob/f093dcb96732bf09dedf7034a951bdc11d1b4880/cmd/utils/utils.go#L62

internal/rpc/rpc.go
kerzhner: This a good thought. The RPC client now checks for server readiness. Take a look at https://github.com/statechannels/go-nitro/pull/1652/commits/c9078d9dfc1777c2510111b0aa95aa3f0b5ecc45.

kerzhner:
> > Do you think we should be spreading this around / sharing it with things like this
> 
> Also a good point. I think it makes sense to eliminate bespoke `WaitForRpcClient`. This is done in [ad2ad7c](https://github.com/statechannels/go-nitro/commit/ad2ad7c1be53660ed6a26026220cbf77d904cbbc).
> 
> With this change, Go and Typescript RPC clients are responsible for checking RPC server readiness. The downside of this approach is that we might not find out about server initialization failures until an RPC client is created.

This approach had to be partially rolled back via 4d00b02528a96863b12d220781daa8e333ad185b. We rely on `WaitForRPCClient` to delay starting alice and bob servers (before any clients are instantiated).

PR #1655, opened 2023-09-10
Workaround for NATS flicker

### LINKED ISSUE #1645: `panic: send on a closed channel` in `TestRpcWithNats` test
https://github.com/statechannels/go-nitro/actions/runs/6102333531/job/16560598157

lalexgap:
Fixes #1645  (maybe)

I suspect this flicker may be an[ issue with NATS](https://github.com/statechannels/go-nitro/issues/1645#issuecomment-1713006067), as our subscribe function shouldn't be called by the NATS client after close is called. To work around this I've added a check in our subscribe function that checks if the connection has already been closed. This should prevent us from sending on a `closed` channel.

### DIFF
--- a/rpc/transport/nats/client.go
+++ b/rpc/transport/nats/client.go
@@ -53,7 +53,11 @@ func (c *natsTransportClient) Subscribe() (<-chan []byte, error) {
 	}
 	c.notificationChan = make(chan []byte)
 	subscription, err := c.nc.Subscribe(nitroNotificationTopic, func(msg *nats.Msg) {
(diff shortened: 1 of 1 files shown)


geoknee (approved):
Is there an issue in the upstream repo tracking this? Perhaps we should report it and link to it here.

lalexgap:
> Is there an issue in the upstream repo tracking this? Perhaps we should report it and link to it here.

https://github.com/nats-io/nats.go/issues/1396

PR #1656, opened 2023-09-11
Create channels tweaks

geoknee:
Get `create-channels` script working.

### DIFF
--- a/cmd/create-channels/main.go
+++ b/cmd/create-channels/main.go
@@ -49,7 +49,7 @@ func createChannels() error {
 			panic(err)
 		}
 
-		err = utils.WaitForRpcClient(url, 500*time.Millisecond, 5*time.Minute)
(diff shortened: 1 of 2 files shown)


cmd/create-channels/main.go
```diff
@@ -49,7 +49,7 @@ func createChannels() error {
 			panic(err)
 		}
 
-		err = utils.WaitForRpcClient(url, 500*time.Millisecond, 5*time.Minute)
+		err = utils.WaitForRpcClient("http://"+url, 500*time.Millisecond, 5*time.Minute)
```
geoknee, line 52: Open to suggestions on how to fix this in a more holistic way. It would be nice to avoid putting these prefixes in ad hoc.

Also may interact with https://github.com/statechannels/go-nitro/issues/1630

PR #1659, opened 2023-09-11
Channel class rejects old chain event updates

bitwiseguy:
The `Channel` class should prevent chain events from making updates to the `Channel` if the event is older than the last event that updated the `Channel`.

Reference: https://github.com/statechannels/go-nitro/pull/1624

### DIFF
--- a/channel/channel.go
+++ b/channel/channel.go
@@ -32,10 +32,17 @@ type Channel struct {
 
 	OnChain  OnChainData
 	OffChain OffChainData
+
+	LastChainUpdate ChainUpdateData
 }
 
-// New constructs a new Channel from the supplied state.
-func New(s state.State, myIndex uint) (*Channel, error) {
+type ChainUpdateData struct {
+	BlockNum uint64
+	TxIndex  uint
+}
+
+// NewChannel constructs a new Channel from the supplied state.
+func NewChannel(s state.State, myIndex uint) (*Channel, error) {
 	c := Channel{}
 	var err error = s.Validate()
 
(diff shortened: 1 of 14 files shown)


channel/channel.go
```diff
@@ -321,28 +328,37 @@ func (c *Channel) SignAndAddState(s state.State, sk *[]byte) (state.SignedState,
...
-	case chainservice.DepositedEvent:
-		c.OnChain.Holdings[e.Asset] = e.NowHeld
-	case chainservice.ConcludedEvent:
-		break // TODO: update OnChain.StateHash and OnChain.Outcome
-	case chainservice.ChallengeRegisteredEvent:
-		h, err := e.StateHash(c.FixedPart)
-		if err != nil {
-			return nil, err
+	if event.BlockNum() > c.LastChainUpdate.BlockNum ||
+		(event.BlockNum() == c.LastChainUpdate.BlockNum && event.TxIndex() > c.LastChainUpdate.TxIndex) {
```
geoknee, line 332: Made a suggestion here #1675

geoknee (approved):

channel/channel.go
bitwiseguy: That make sense. Merged that PR into this branch

channel/channel.go
```diff
@@ -32,10 +32,17 @@ type Channel struct {
...
+
+type ChainUpdateData struct {
+	BlockNum uint64
+	TxIndex  uint
 }
 
-// New constructs a new Channel from the supplied state.
-func New(s state.State, myIndex uint) (*Channel, error) {
+// NewChannel constructs a new Channel from the supplied state.
+func NewChannel(s state.State, myIndex uint) (*Channel, error) {
```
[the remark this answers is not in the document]
bitwiseguy: Reverted that change here: https://github.com/statechannels/go-nitro/pull/1659/commits/acd6e86b5a8965c969745e58d6ac40f5abb02ef4

PR #1660, opened 2023-09-11
Add automatic restart to cloud node workflow script

bitwiseguy:
In order to make the cloud node more resilient to crashes without requiring manual intervention, this PR adds an automatic restart policy to the github action that deploys the cloud node and starts it.

The docker restart policy options (set with the `--restart` flag in the `docker run` command) are as follows:

* `--restart=no`: Do not automatically restart the container when it exits. This is the default behavior.
* `--restart=always`: Always restart the container regardless of the exit status. If the container is manually stopped, it is restarted only after Docker daemon restarts or the container itself is manually restarted.
* `--restart=unless-stopped`: Similar to always, but if the container is stopped manually, it will not be restarted automatically even after Docker daemon restarts.
* `--restart=on-failure[:max-retries]`: Restart only if the container exits with a non-zero exit status. Optionally, limit the number of restart retries the Docker daemon attempts.

### DIFF
--- a/node/engine/chainservice/eth_chainservice.go
+++ b/node/engine/chainservice/eth_chainservice.go
@@ -345,6 +345,7 @@ out:
 				eventSub.Unsubscribe()
 			}
 
+			// Try to re-establish subscription once before failing
 			eventSub, err = ecs.chain.SubscribeFilterLogs(ecs.ctx, eventQuery, eventChan)
 			if err != nil {
 				errorChan <- fmt.Errorf("subscribeFilterLogs failed to resubscribe: " + err.Error())
@@ -380,6 +381,7 @@ out:
 				newBlockSub.Unsubscribe()
 			}
 
+			// Try to re-establish subscription once before failing
 			newBlockSub, err = ecs.chain.SubscribeNewHead(ecs.ctx, newBlockChan)
 			if err != nil {
 				errorChan <- fmt.Errorf("subscribeNewHead failed to resubscribe: %w", err)
--- a/node/engine/chainservice/eth_chainservice.go
+++ b/node/engine/chainservice/eth_chainservice.go
@@ -184,7 +184,7 @@ func (ecs *EthChainService) listenForErrors(errChan <-chan error) {
 			return
 		case err := <-errChan:
 			ecs.logger.Error("chain service error", "error", err)
-			panic(err) // Manually panic in case we're using a logger that doesn't call exit(1)
+			panic(err)
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


lalexgap:
I think `--restart=unless-stopped` makes sense to me.

node/engine/chainservice/eth_chainservice.go
```diff
@@ -344,18 +341,16 @@ out:
 
 		case err := <-eventSub.Err():
 			if err != nil {
```
bitwiseguy, line 343: This changes here means the chainservice will attempt to re-establish the websocket connection once before triggering a panic. We can come back later and refine this code if we want (e.g. add exponential back-off and ensure no missed chain events between websocket error and resubscribing)

node/engine/chainservice/eth_chainservice.go
```diff
@@ -183,11 +183,8 @@ func (ecs *EthChainService) listenForErrors(errChan <-chan error) {
 			ecs.wg.Done()
 			return
 		case err := <-errChan:
-			fmt.Println(err) // Print to STDOUT in case we're using a noop logger
 			ecs.logger.Error("chain service error", "error", err)
-
-			// Manually panic in case we're using a logger that doesn't call exit(1)
-			panic(err)
+			panic(err) // Manually panic in case we're using a logger that doesn't call exit(1)
```
lalexgap, line 187: With the switch to `slog` we know that `ecs.logger.Error` will not `panic`, so we can remove the comment.
```suggestion
			panic(err)
```

node/engine/chainservice/eth_chainservice.go
```diff
@@ -381,18 +376,16 @@ out:
...
+				newBlockSub.Unsubscribe()
 			}
 
-			// If the error is nil then the subscription was closed and we need to re-subscribe.
-			// This is a workaround for https://github.com/ethereum/go-ethereum/issues/23845
-			var sErr error
-			newBlockSub, sErr = ecs.chain.SubscribeNewHead(ecs.ctx, newBlockChan)
-			if sErr != nil {
-				errorChan <- fmt.Errorf("subscribeNewHead failed on resubscribe: %w", sErr)
+			newBlockSub, err = ecs.chain.SubscribeNewHead(ecs.ctx, newBlockChan)
```
lalexgap, line 383: Could consider adding a comment here explaining that this is the "retry" logic"

node/engine/chainservice/eth_chainservice.go
lalexgap: It could be nice to add a comment with some of that info to the code; it took me a second to understand this was the retry logic.

lalexgap (approved):
🚀

node/engine/chainservice/eth_chainservice.go
bitwiseguy: Added comment in https://github.com/statechannels/go-nitro/pull/1660/commits/56607ea190829746e6c0531e435ca948cd131e4e

lalexgap:
Fixes #1668

PR #1672, opened 2023-09-11
Fildev demo

lalexgap:
⚠️ This is a feature branch which we will merge PRs into.


geoknee:
@lalexgap I'm wondering if we should let the user "fetch" even without a channel? It might just mean they hit a 402, but we can catch that and display a hint.

It may just make configuring the deployment a bit easier (we can test client against proxy without any of the payment infrastructure).

lalexgap:
> @lalexgap I'm wondering if we should let the user "fetch" even without a channel? It might just mean they hit a 402, but we can catch that and display a hint.

Yup that could be useful! If there's no channel defined then we can make the request without attaching any voucher info.

docker/nitro/anthony.toml
```diff
@@ -11,13 +11,16 @@ pk = ""
...
+caaddress = "0xe1790ea824035184a3bf344e087fb61744992545"
 
 # RPC provider docs: https://lotus.filecoin.io/lotus/developers/glif-nodes/#testnet-endpoint
 chainurl = "wss://wss.calibration.node.glif.io/apigw/lotus/rpc/v0"
-chainstartblock = 909569
+chainstartblock = 915929
 chainauthtoken = ""
 
-bootpeers = "/ip4/67.207.88.72/tcp/3005/p2p/16Uiu2HAmDQj9XnHPD1d4GzS1K3DhmM5Z1cSGLfVTwVcNVdEsAN3Q"
+bootpeers = "/ip4/67.207.88.72/tcp/3005/p2p/16Uiu2HAmDQj9XnHPD1d4GzS1K3DhmM5Z1cSGLfVTwVcNVdEsAN3Q,/ip4/192.81.214.172/tcp/3005/p2p/16Uiu2HAmMZD9hwtMBBDpbE3qEigb4FF1YDAjePLCVR6RQZYo6RPJ"
```
geoknee, line 23: I think we should remove the second bootpeer here. Using this is sort of bypassing the DHT and amounts to hardcoding the entire phone book into each client.

docker/nitro/brad.toml
```diff
@@ -11,13 +11,16 @@ pk = ""
...
+caaddress = "0xe1790ea824035184a3bf344e087fb61744992545"
 
 # RPC provider docs: https://lotus.filecoin.io/lotus/developers/glif-nodes/#testnet-endpoint
 chainurl = "wss://wss.calibration.node.glif.io/apigw/lotus/rpc/v0"
-chainstartblock = 909569
+chainstartblock = 915929
 chainauthtoken = ""
 
-bootpeers = "/ip4/67.207.88.72/tcp/3005/p2p/16Uiu2HAmDQj9XnHPD1d4GzS1K3DhmM5Z1cSGLfVTwVcNVdEsAN3Q"
+bootpeers = "/ip4/67.207.88.72/tcp/3005/p2p/16Uiu2HAmDQj9XnHPD1d4GzS1K3DhmM5Z1cSGLfVTwVcNVdEsAN3Q,/ip4/134.122.114.102/tcp/3005/p2p/16Uiu2HAmUPifMQSzPGmt3UdTc86bMuebBJ26e2xBWus7xmQ5zhMt"
```
geoknee, line 23: Remove second boot peer.

netlify.toml
```diff
@@ -1,3 +1,2 @@
 [build]
-  ignore = "if  [ ${COMMIT_REF} == ${CACHED_COMMIT_REF} ] ; then false ; else git diff --quiet $CACHED_COMMIT_REF $COMMIT_REF -- docs mkdocs.yml ; fi"
-
+ignore = "if  [ ${COMMIT_REF} == ${CACHED_COMMIT_REF} ] ; then false ; else git diff --quiet $CACHED_COMMIT_REF $COMMIT_REF -- docs mkdocs.yml /packages/payment-proxy-client packages/nitro-rpc-client ; fi"
```
geoknee, line 2: it's awkward to have a single `netlify.toml` and ignore command for 3 separate sites which all use this repo :-/

node_test/paymentproxy_test.go
```diff
@@ -185,7 +185,7 @@ func createVoucher(t *testing.T, client rpc.RpcClientApi, channelId types.Destin
 }
 
 func expectedPaymentErrorMessage(total, numPaid int) string {
-	return fmt.Sprintf("payment of %d required, the voucher only resulted in a payment of %d", total, numPaid)
+	return fmt.Sprintf("payment of %d attoFIL required, the voucher only resulted in a payment of %d attoFIL", total, numPaid)
```
geoknee, line 188: Here is some Filecoin-specific code which doesn't belong here. We can fix that as a part of https://github.com/statechannels/go-nitro/issues/1785.

packages/nitro-rpc-client/src/transport/http.ts
```diff
@@ -60,3 +62,12 @@ export class HttpTransport {
     };
   }
 }
+
+// For testing with self-signed certs, ignore certificate errors. DO NOT use in production.
+export function unsecureHttpsAgent(): https.Agent {
+  // For testing with self-signed certs, ignore certificate errors. DO NOT use in production.
```
geoknee, line 68: Can we put this inside `internal` or otherwise make it harder to use accidentally?

packages/payment-proxy-client/src/main.tsx
```diff
@@ -1,8 +1,8 @@
 import React from "react";
 import ReactDOM from "react-dom/client";
 
+// import "./index.css";
```
geoknee, line 4: Delete this.

paymentproxy/proxy.go
```diff
@@ -126,7 +148,7 @@ func (p *PaymentProxy) handleDestinationResponse(r *http.Response) error {
 	// s.Delta is amount our balance increases by adding this voucher
 	// AKA the payment amount we received in the request for this file
 	if cost > s.Delta.Uint64() {
-		return createPaymentError(fmt.Errorf("payment of %d required, the voucher only resulted in a payment of %d", cost, s.Delta.Uint64()))
+		return createPaymentError(fmt.Errorf("payment of %d attoFIL required, the voucher only resulted in a payment of %d attoFIL", cost, s.Delta.Uint64()))
```
geoknee, line 151: We would prefer to not hardcode this here. However, the entire demo is already quite tightly coupled to Filecoin, so it is not high priority to fix.

geoknee (approved):

node_test/paymentproxy_test.go
geoknee: https://github.com/statechannels/go-nitro/issues/1816

packages/nitro-rpc-client/src/transport/http.ts
geoknee: It's `ts` code, so we don't have the option of making it an `internal` thing. It does look unused though, so could be deleted.

PR #1683, opened 2023-09-12
Add TLS to HTTP transport

kerzhner:
This PR adds TLS to websocket and http messaging between RPC client and server. TLS is needed in order to secure auth tokens. With this PR, rpc test will pass only if a valid certificate and private key exist in the `tls` directory and are signed by a valid or local certificate authority. These details have been added to the readme.


internal/tls/statechannels.org.pem
```diff
@@ -0,0 +1,26 @@
+-----BEGIN CERTIFICATE-----
+MIIEajCCAtKgAwIBAgIRALxmKKJiHaf4cgZxwFTXITEwDQYJKoZIhvcNAQELBQAw
+fzEeMBwGA1UEChMVbWtjZXJ0IGRldmVsb3BtZW50IENBMSowKAYDVQQLDCFtaXNo
```
lalexgap, line 3: This isn't our real statechannel key is it?

lalexgap: Ah nevermind, I see above that it's not.

kerzhner: This is a private key used for a self signed certificate. Self signed certificates should not be used in production. The key is checked in for testing ease.

rpc/transport/http/client.go
```diff
@@ -29,10 +30,11 @@ func NewHttpTransportAsClient(url string, retryTimeout time.Duration) (*clientHt
 		return nil, err
 	}
 
-	subscribeUrl, err := urlUtil.JoinPath("ws://", url, "subscribe")
+	subscribeUrl, err := urlUtil.JoinPath("wss://", url, "subscribe")
```
lalexgap, line 33: Right now the `start-rpc-servers` doesn't have specify a cert file, so will run over http/ws. This means that the rpc client won't be able to connect to any of the clients since it assumes https/wss.

We could probably update our CLI to take a url with a protocol(`http://localhost:4005/v1/subscribe` instead of `localhost:4005/v1/subscribe`).   `http` would mean `http/ws` and `https` would mean `https/wss`, that way the rpc clients could support either.

kerzhner: Check out https://github.com/statechannels/go-nitro/blob/327591540a862741e7fb4fe1e448f7dc57ca6d4d/tls/readme.md?plain=1#L9. Does that help?

Note that all of the tests currently pass, including the create channels Typescript test.

lalexgap: Hmm I'm still seeing the same issue.
From the root `go-nitro` directory I ran:
- `make -C tls create-cert`
- ` go run ./cmd/start-rpc-servers/main.go`
-`npx ts-node ./packages/nitro-rpc-client/scripts/client-runner.ts create-channels`

lalexgap: Oops I missed running `export NODE_EXTRA_CA_CERTS="$(mkcert -CAROOT)/rootCA.pem". That resolved it!

geoknee (approved):

PR #1687, opened 2023-09-12
Store channel in local storage

### LINKED ISSUE #1678: Storing a user’s virtual channel in browser storage
so when they come back they see the same balance

lalexgap:
Fixes #1678 

This stores the virtual channel Id in local storage. When hitting the UI we check if the channel id is already in local storage and if so, we fetch the channel details and use that channel.


geoknee (approved):

geoknee:
@lalexgap Currently there is no mechanism to clear this storage, apart from doing so manually in the browser developer tools. We probably want something smarter to deal with the situation when the nitro network has been restarted, or later on down the line if/when the virtual channel was closed.

If we find a channel in storage, but then the nitro node reports an error when we try to query it, we could clear the storage and create a new channel?

Issue #1661, opened 2023-09-11
Deploy two more cloud nodes to digital ocean


lalexgap:
One will be the multi-client `payer`. One will be the `payee` for the boost / SP node.

bitwiseguy:
Completed via https://github.com/statechannels/go-nitro/pull/1697

PR #1697, opened 2023-09-13
Deploy cloud nodes

bitwiseguy:
Creates a single, unified workflow that does the following:

* Builds docker image
* Deploys Iris
* Deploys Anthony
* Deploys Brad

The workflow includes input flags for each of those jobs that allows the user to select which one executes for a given run. The workflow file also includes conditional logic that ensures that the node(s) are deployed **AFTER** the docker image is built (if building the docker image is enabled).


.github/workflows/deploy-cloud-nodes.yml
```diff
@@ -0,0 +1,123 @@
...
+        required: false
+        default: 'true'
+
+jobs:
+  build-image:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        if: github.event.inputs.build-image == 'true'
```
geoknee, line 29: Let's not worry about this now: but I wonder if we can evaluate this condition for the whole job, rather than each step?

geoknee (approved):

.github/workflows/deploy-cloud-nodes.yml
bitwiseguy: I tried to set the conditional for the whole job first. However, it caused the `build-image` job to get skipped if the conditional returned `false`.  This is an issue because it caused the `deploy-iris` job to get skipped because it has the following line, which is necessary to prevent the node from getting deployed before the `build-image` job completes.
```
  deploy-iris:
    needs: build-image
```

bitwiseguy:
Closes https://github.com/statechannels/go-nitro/issues/1661

PR #1701, opened 2023-09-13
Fix cloud port mapping

bitwiseguy:
Allows us to access cloud GUI


Makefile
```diff
@@ -2,7 +2,7 @@ docker/cloud/build:
 	docker build -f docker/cloud/Dockerfile -t go-nitro-cloud .
 docker/cloud/start:
```
lalexgap, line 3: Outside the scope of this PR but should this make command also start the other cloud nodes?

lalexgap (approved):

PR #1702, opened 2023-09-13
Remove channel id from local storage if it cannot be fetched

### LINKED ISSUE #1694: Handle channel not existing on node
@lalexgap Currently there is no mechanism to clear this storage, apart from doing so manually in the browser developer tools. We probably want something smarter to deal with the situation when the nitro network has been restarted, or later on down the line if/when the virtual channel was closed.

If we find a channel in storage, but then the nitro node reports an error when we try to query it, we could clear the storage and create a new channel?

_Originally posted by @geoknee in https://github.com/statechannels/go-nitro/issues/1687#issuecomment-1716986435_

lalexgap:
Fixes #1694 

Handles the case where the channel id in storage no longer exists on the nitro node. We log a warning message out and remove the channel id, so the user can easily create a new one.

### DIFF
--- a/packages/payment-proxy-client/src/App.tsx
+++ b/packages/payment-proxy-client/src/App.tsx
@@ -56,14 +56,24 @@ function App() {
   }, [url]);
 
   useEffect(() => {
+    if (!nitroClient) return;
     const fetchedId = localStorage.getItem(CHANNEL_ID_KEY);
     if (fetchedId && fetchedId != "") {
       setPaymentChannelId(fetchedId);
 
-      nitroClient?.GetPaymentChannel(fetchedId).then((paymentChannel) => {
-        console.log(paymentChannel);
-        setPaymentChannelInfo(paymentChannel);
-      });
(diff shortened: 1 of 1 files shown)


packages/payment-proxy-client/src/App.tsx
```diff
@@ -56,14 +56,24 @@ function App() {
...
     const fetchedId = localStorage.getItem(CHANNEL_ID_KEY);
     if (fetchedId && fetchedId != "") {
       setPaymentChannelId(fetchedId);
 
-      nitroClient?.GetPaymentChannel(fetchedId).then((paymentChannel) => {
-        console.log(paymentChannel);
-        setPaymentChannelInfo(paymentChannel);
-      });
+      nitroClient
+        ?.GetPaymentChannel(fetchedId)
```
geoknee, line 65: Do we need the `?.` given the new L59?

geoknee (approved):

PR #1704, opened 2023-09-13
Accept a filepath to certificate file

lalexgap:
This allows the filepath to the certificate to be passed in when starting a nitro node.

### DIFF
--- a/node_test/rpc_test.go
+++ b/node_test/rpc_test.go
@@ -1,6 +1,7 @@
 package node_test
 
 import (
+	"crypto/tls"
 	"encoding/json"
 	"fmt"
 	"log/slog"
@@ -403,6 +404,11 @@ func setupNitroNodeWithRPCClient(
 		t.Fatal(err)
 	}
 
+	cert, err := tls.LoadX509KeyPair("../internal/tls/statechannels.org.pem", "../internal/tls/statechannels.org_key.pem")
+	if err != nil {
+		panic(err)
+	}
+
 	slog.Info("Initializing message service on port " + fmt.Sprint(msgPort) + "...")
 	messageService := p2pms.NewMessageService("127.0.0.1", msgPort, *ourStore.GetAddress(), pk, bootPeers)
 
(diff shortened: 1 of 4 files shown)


node_test/rpc_test.go
```diff
@@ -403,6 +404,11 @@ func setupNitroNodeWithRPCClient(
 		t.Fatal(err)
 	}
 
+	cert, err := tls.LoadX509KeyPair("../internal/tls/statechannels.org.pem", "../internal/tls/statechannels.org_key.pem")
```
lalexgap, line 407: I think relative paths are safe to use in a test file. This seems to function correctly if I run `go test ./...` from the root directory and if I run it from the `node_test` directory.

rpc/transport/http/server.go
```diff
@@ -62,24 +62,27 @@ func NewHttpTransportAsServer(port string) (*serverHttpTransport, error) {
...
-		panic(err)
-	}
+	var listener net.Listener
+	var err error
 
-	// Create a TLS config
-	tlsConfig := &tls.Config{
-		Certificates: []tls.Certificate{cert},
+	if cert == nil {
+		listener, err = net.Listen("tcp", ":"+transport.port)
```
lalexgap, line 69: This lets us continue to use http/ws if we choose to.

rpc/transport/http/server.go
```diff
@@ -62,24 +62,27 @@ func NewHttpTransportAsServer(port string) (*serverHttpTransport, error) {
...
+		}
+	} else {
+		// Create a TLS config
+		tlsConfig := &tls.Config{
+			Certificates: []tls.Certificate{*cert},
+		}
+		// Create a new TLS listener
+		listener, err = tls.Listen("tcp", ":"+port, tlsConfig)
+		if err != nil {
+			panic(err)
```
kerzhner, line 81: We should probably return the error instead of a panic.

kerzhner (approved):

PR #1708, opened 2023-09-14
Improved frontend for the Filecoin Paid Retrieval Demo

geoknee:
It is:
* mobile friendly
* dark mode friendly
* does not yet work with micropayments (that we can add back later)
Screenshot 2023-09-14 at 15 13 53
Screenshot 2023-09-15 at 13 10 54


geoknee:
Still need to wire up the "skip payment " switch, and remove the micropayments one for now.

packages/payment-proxy-client/src/App2.tsx
lalexgap, line 1: Could we rename this back to `App.tsx`? That would show the diff on the PR.

packages/payment-proxy-client/src/App.tsx
```diff
@@ -1,58 +1,107 @@
...
+  defaultNitroRPCUrl,
+  fileUrl,
+  hub,
+  initialChannelBalance,
+  provider,
+} from "./constants";
+import { fetchFile } from "./file";
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function Copyright(props: any) {
```
lalexgap, line 43: Could we type this so we don't have to supress the `any` warning?

packages/payment-proxy-client/src/main.tsx
```diff
@@ -1,8 +1,8 @@
 import React from "react";
 import ReactDOM from "react-dom/client";
 
+// import "./index.css";
```
lalexgap, line 4: Could we remove this?

lalexgap (approved):

PR #1716, opened 2023-09-14
Update Demo branch with latest changes from main and wss

lalexgap:
This merges in the latest changes from `main` as well as the latest tls changes from `wss` into our demo feature branch.

The `wss` branch contains the latest TLS changes that we need to get things working.


.github/workflows/ts-rpc-test.yml
```diff
@@ -20,6 +20,8 @@ jobs:
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@v1
 
+      - uses: ./.github/actions/mkcert
```
geoknee, line 23: 💪

geoknee:
Changes look good, but 32 commits seems a lot. We could squash and merge to tidy that up? Or do we think that will make eventually merging `fildev-payment-client` to `main` more difficult?

I am imagining we may want to rewrite the history of `fildev-payment-client` / cherry-pick some or all of it to `main` (let's say in a few weeks' time).

geoknee (approved):

PR #1719, opened 2023-09-15
add sequence diagram for demo

geoknee:


packages/payment-proxy-client/readme.md
```diff
@@ -3,3 +3,46 @@
...
+group paid retrieval
+user->payer: create payment
+user<-payer: voucher
+destroysilent payer
+user->proxy: request + voucher
+proxy->payee: validate voucher
+payee->proxy: yes / no
+destroysilent payee
+proxy->origin: request
+proxy<-origin: 200 / 402 response
```
geoknee, line 43: I think this bit needs modifying. The request is conditional on the validation step, so could be shown with a dashed arrow. I don't think the origin server ever serves 402 errors.

geoknee: Actually, the result is not even conditional on the payment step
[150c842](https://github.com/statechannels/go-
nitro/pull/1719/commits/150c842d5ed9b7de40118ddf6cbb39bfc2c67bf0)

lalexgap (approved):

PR #1724, opened 2023-09-15
Merge main into fildev-payment-client

lalexgap:
This adds the latest peer discovery changes.


.github/workflows/deploy-cloud-nodes.yml
```diff
@@ -4,21 +4,21 @@ on:
...
+        description: "Deploy Anthony"
         required: false
-        default: 'true'
+        default: "true"
       deploy-brad:
-        description: 'Deploy Brad'
+        description: "Deploy Brad"
         required: false
-        default: 'true'
+        default: "true"
```
bitwiseguy, line 21: How can we avoid all of these changes between single quote and double quote characters? Seems like a conflict in linters

bitwiseguy (approved):

.github/workflows/deploy-cloud-nodes.yml
lalexgap: I think this is my bad, I think I have an extension that is auto-formatting `yml` files. Normally I work around it by not triggering the linter but sometimes I do it accidently.

I'll track down what extension is doing that and remove it.

PR #1728, opened 2023-09-15
Fix divide by 0 error in demo UI

lalexgap:


kerzhner (approved):
Is the formula prior to this PR just incorrect? If `PaidSoFar` is causing the division by 0, then that will still be an issue after this PR.

lalexgap:
> Is the formula prior to this PR just incorrect? If `PaidSoFar` is causing the division by 0, then that will still be an issue after this PR.

Yes, good point!

PR #1731, opened 2023-09-15
Remove local storage

lalexgap:
~Currently we fetch and set the channel if there's a channel id in local storage. However you still start at step 0 of UI, meaning you're forced to create a new channel.~

Removes any use of local storage

### DIFF
--- a/packages/payment-proxy-client/src/App.tsx
+++ b/packages/payment-proxy-client/src/App.tsx
@@ -122,6 +122,9 @@ export default function App() {
           localStorage.removeItem(CHANNEL_ID_KEY);
           setPaymentChannelId("");
         });
+
+      // If we already have a channel, skip to the payment step
+      setActiveStep(2);
     }
   }, [nitroClient]);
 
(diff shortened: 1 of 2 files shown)


geoknee:
I forgot to say at standup, but I was thinking maybe to not persist the channel in storage after all...

geoknee:
Rationale being that refreshing the page would become like "resetting" the app. Which is probably more useful than having users able to pick up where they left off?

packages/payment-proxy-client/src/App.tsx
```diff
@@ -122,6 +122,9 @@ export default function App() {
           localStorage.removeItem(CHANNEL_ID_KEY);
           setPaymentChannelId("");
         });
+
+      // If we already have a channel, skip to the payment step
+      setActiveStep(2);
```
geoknee: This could be a numeric enum, perhaps? https://www.typescriptlang.org/docs/handbook/enums.html#numeric-enums

lalexgap:
> I forgot to say at standup, but I was thinking maybe to not persist the channel in storage after all...

@geoknee I've updated this PR to just remove local storage.

geoknee (approved):
I'm open minded about the best approach to preserving the channel between sessions. 

This PR does at least restore consistency 👍

geoknee:
> @geoknee I've updated this PR to just remove local storage.

OK, your call!

PR #1732, opened 2023-09-15
UI tweaks

geoknee:


packages/payment-proxy-client/src/App.tsx
```diff
@@ -59,11 +59,9 @@ function Copyright(props: any) {
 }
 
 function computePercentagePaid(info: PaymentChannelInfo): number {
-  return (
-    Number(
-      info.Balance.PaidSoFar /
-        (info.Balance.RemainingFunds + info.Balance.PaidSoFar)
-    ) * 100
+  return Number(
+    (100n * info.Balance.PaidSoFar) /
```
lalexgap, line 63: I don't think this is correct, since we're only multiplying the numerator by 100?

lalexgap, line 63: I think this means that the percentage value we return is "truncated"?
 IE: If there's `65.8941% paid` this function would return `65%`?

geoknee: I'm going to try out something on this branch which avoids the need for percentages altogether.

packages/payment-proxy-client/src/App.tsx
```diff
@@ -59,6 +59,9 @@ function Copyright(props: any) {
 }
 
 function computePercentagePaid(info: PaymentChannelInfo): number {
+  if (info.Balance.RemainingFunds <= 0) {
```
lalexgap: Do we need this? The only way we divide by 0 is if both `PaidSoFar` and `RemainingFunds` are `0`

geoknee: Yes this was wrong, good catch.

geoknee:
Screenshot 2023-09-15 at 23 20 17
Dropping the percentage approach seems to work -- the little tooltip was showing a percentage with the existing method anyway, which is a bit confusing.

packages/payment-proxy-client/src/App.tsx
```diff
@@ -331,24 +323,25 @@ export default function App() {
...
                       valueLabelDisplay="on"
                     />
                     <StorageIcon />{" "}
                     <Typography variant="body2" color="text.secondary">
-                      {paymentChannelInfo?.Balance.Payee}
+                      {paymentChannelInfo?.Balance.PaidSoFar.toString()}
                     </Typography>
                   </Stack>
-                  {paymentChannelId}
+                  {paymentChannelId.slice(0, 6) + "..."}
```
lalexgap, line 344: Could be worth throwing this in a helper function; I bet we'll need to shorten channel ids in other places too.

lalexgap (approved):

PR #1733, opened 2023-09-15
Fix percentage returned

lalexgap:
Previously we were dividing with `bignumbers`, which uses floor division, meaning we would always get back `0`. This updates the function to perform the division with `number`s to avoid this.


lalexgap:
No longer needed, see #1732

PR #1734, opened 2023-09-15
Use the proxy rul env var

lalexgap:
Updates the demo to use to the proxy url, instead of being hardcoded to the non-payment proxy url `www.w3.org`

### DIFF
--- a/packages/payment-proxy-client/src/constants.ts
+++ b/packages/payment-proxy-client/src/constants.ts
@@ -2,7 +2,7 @@
 export const QUERY_KEY = "rpcUrl";
 export const costPerByte = 1;
 
-export const proxyUrl = "https://www.w3.org";
+export const proxyUrl = import.meta.env.VITE_PROXY_URL;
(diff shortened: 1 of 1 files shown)


geoknee (approved):

packages/payment-proxy-client/src/constants.ts
```diff
@@ -2,7 +2,7 @@
 export const QUERY_KEY = "rpcUrl";
 export const costPerByte = 1;
 
-export const proxyUrl = "https://www.w3.org";
+export const proxyUrl = import.meta.env.VITE_PROXY_URL;
```
geoknee, line 5: Are we setting this anywhere?

lalexgap: That is a good point, fixed in https://github.com/statechannels/go-nitro/pull/1735/commits! I'm surprised linting doesn't catch this, since we do define our env var names in a typing file.

PR #1735, opened 2023-09-15
Use correct env var

lalexgap:
Oops, I was using a made up env var name. This updates the env var names and values to `VITE_PROXY_URL`


lalexgap:
Since this is a pretty straight-forward change and should get the demo working, I'm going to merge this PR.

PR #1737, opened 2023-09-17
Switch to archive node

lalexgap:
This switches the cloud nodes to use the archive node provided to us by filecoin. The JWT token is stored in a github secret.

TODO:
- [ ] Verify the URL works for WSS


lalexgap:
Closing as this is being handled by @bitwiseguy

PR #1738, opened 2023-09-18
Some tweaks to the UI

geoknee:
A number of small improvements (see commit messages for detail).


packages/payment-proxy-client/src/prettyPrintFIL.ts
lalexgap, line 1: Could we name this file something more generic?

lalexgap (approved):

PR #1742, opened 2023-09-18
Add multiple file support to the demo

### LINKED ISSUE #1740: Have the demo app work for a list of 10 files.

lalexgap:
Fixes #1740 
This adds the ability to run the demo with multiple files available. The `VITE_FILE_PATHS` env var takes a comma delimited list of file paths.

A radio group with the available files has been added to step 3. Since we're using lots of padding I tried to keep it pretty compact to prevent forcing the user to scroll.

image


geoknee (approved):
Looks good!

1. Could we make it so that the "next" file is auto selected when we successfully download a file?

2. I think your instinct to keep the selection of the file very compact is right. The file names are unfortunately very long on our sample files. Perhaps we can get away with using a component like this https://mui.com/material-ui/react-stepper/#dots to represent which file will be downloaded? We may get away without displaying file names because it is a sort of "lucky dip" experience.

lalexgap:
> Looks good!
> 
>     1. Could we make it so that the "next" file is auto selected when we successfully download a file?
> 
>     2. I think your instinct to keep the selection of the file very compact is right. The file names are unfortunately very long on our sample files. Perhaps we can get away with using a component like this https://mui.com/material-ui/react-stepper/#dots to represent which file will be downloaded? We may get away without displaying file names because it is a sort of "lucky dip" experience.

I'm wondering if that might be too guided and streamlined? If a user keeps clicking the same button and gets served different images; it might not be obvious what's going on, since anything to do with a "file" is hidden from the user. I I also don't think there's any benefit to getting a user to download all 10, so I don't think we necessarily need to "guide" the user onto the next file.

I think allowing the user to select the file to download makes the demo slightly more interesting as well. Instead of just clicking the same button over and over, the user is given some agency and can check out different files.

PR #1744, opened 2023-09-18
Properly log out rpc payload

lalexgap:
I noticed that on one of the cloud nodes that the `Rpc server received request` log entry always has a null params.
```
{"time":"2023-09-18T18:44:09.56610977Z","level":"DEBUG","msg":"Rpc server received request","address":"0x3E6b9c3e569e72932Cfc2A943E5497ed07FEfB9e","request":{"jsonrpc":"","id":62649479,"method":"create_voucher","params":null}}
{"time":"2023-09-18T18:44:09.700037134Z","level":"DEBUG","msg":"Rpc server received request","address":"0x3E6b9c3e569e72932Cfc2A943E5497ed07FEfB9e","request":{"jsonrpc":"","id":62649618,"method":"get_payment_channel","params":null}}
{"time":"2023-09-18T18:44:28.379842962Z","level":"DEBUG","msg":"Rpc server received request","address":"0x3E6b9c3e569e72932Cfc2A943E5497ed07FEfB9e","request":{"jsonrpc":"","id":62668135,"method":"create_voucher","params":null}}
{"time":"2023-09-18T18:44:28.523026297Z","level":"DEBUG","msg":"Rpc server received request","address":"0x3E6b9c3e569e72932Cfc2A943E5497ed07FEfB9e","request":{"jsonrpc":"","id":62668440,"method":"get_payment_channel","params":null}}
```

This PR updates the ` validateJsonrpcRequest` to set the `Params` on the `JsonRpcGeneralRequest` so we actually log out the params.

### DIFF
--- a/rpc/serde/jsonrpc.go
+++ b/rpc/serde/jsonrpc.go
@@ -146,7 +146,7 @@ type JsonRpcGeneralRequest struct {
 	Jsonrpc string      `json:"jsonrpc"`
 	Id      uint64      `json:"id"`
 	Method  string      `json:"method"`
-	Payload interface{} `json:"payload"`
+	Params  interface{} `json:"payload"`
 }
 
 type JsonRpcGeneralResponse struct {
--- a/rpc/serde/jsonrpc.go
+++ b/rpc/serde/jsonrpc.go
@@ -146,7 +146,7 @@ type JsonRpcGeneralRequest struct {
 	Jsonrpc string      `json:"jsonrpc"`
 	Id      uint64      `json:"id"`
 	Method  string      `json:"method"`
-	Params  interface{} `json:"params"`
+	Payload interface{} `json:"payload"`
 }
 
 type JsonRpcGeneralResponse struct {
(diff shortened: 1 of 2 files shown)


rpc/serde/jsonrpc.go
```diff
@@ -146,7 +146,7 @@ type JsonRpcGeneralRequest struct {
 	Jsonrpc string      `json:"jsonrpc"`
 	Id      uint64      `json:"id"`
 	Method  string      `json:"method"`
-	Params  interface{} `json:"params"`
+	Params  interface{} `json:"payload"`
```
kerzhner, line 149: Does this make sense? Is it a red flag that we are encoding a value under a json key that is different than the struct field?

lalexgap: Good catch! I originally renamed `params` to `payload` but then decided to rename it back, and I missed this.

lalexgap: Fixed in https://github.com/statechannels/go-nitro/pull/1744/commits/06e6eeb6f7bff7678f4c92262ae39d9b0cdbdc55

PR #1754, opened 2023-09-19
Support differing file sizes

lalexgap:
**Based on #1756**

This introduces an env var `VITE_FILE_SIZES` so we can specify the sizes of different files. This allows us to pay the specific amount for a file.

### DIFF
--- a/packages/payment-proxy-client/src/App.tsx
+++ b/packages/payment-proxy-client/src/App.tsx
@@ -32,9 +32,9 @@ import { NitroRpcClient } from "@statechannels/nitro-rpc-client";
 import { PaymentChannelInfo } from "@statechannels/nitro-rpc-client/src/types";
 
 import {
+  AvailableFile,
   QUERY_KEY,
   costPerByte,
-  dataSize,
   defaultNitroRPCUrl,
   files,
   hub,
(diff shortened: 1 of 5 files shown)


packages/payment-proxy-client/src/constants.ts
```diff
@@ -15,9 +13,23 @@ export const initialChannelBalance = parseInt(
   import.meta.env.VITE_INITIAL_CHANNEL_BALANCE,
   10
 );
+const ENV_VAR_SPLIT_CHAR = ";";
```
lalexgap, line 16: I switched to `;` as the delimiter as some of the image names in #1753  contain a `,`

packages/payment-proxy-client/src/App.tsx
```diff
@@ -359,20 +359,28 @@ export default function App() {
...
                           {files.map((file) => (
                             <FormControlLabel
                               value={file.url}
                               key={file.url}
                               control={<Radio />}
-                              label={file.fileName}
+                              label={
+                                file.fileName.length < 50
+                                  ? file.fileName
+                                  : "..." + file.fileName.slice(-50)
```
lalexgap, line 382: This trims the filename in case of long file names like we'll have (IE: `DALL·E 2023-09-08 15.21.55 - surrealist painting of weighing scales balanced with one very large apple and tiny small apple.png`)

lalexgap: Alternatively we could also introduce an env var for file names, so we can use some shorter names.

geoknee (approved):

PR #1755, opened 2023-09-19
Add micropayment support to the demo

lalexgap:
**Includes #1759**

This adds support for micro-payments to the demo. Check out the [deploy preview](https://deploy-preview-1755--nitro-payment-demo.netlify.app/) to see it in action.

### DIFF
--- a/packages/payment-proxy-client/src/App.tsx
+++ b/packages/payment-proxy-client/src/App.tsx
@@ -357,34 +357,35 @@ export default function App() {
                         }
                         label="Skip payment"
                       />
-                      <FormControl>
-                        <RadioGroup
-                          name="availableFiles"
-                          row={true}
-                          value={selectedFile.url}
-                          onChange={(e) => {
-                            const found = files.find(
-                              (f) => f.url == e.target.value
-                            );
-                            if (found) {
-                              setSelectedFile(found);
-                            }
-                          }}
-                        >
-                          {files.map((file) => (
-                            <FormControlLabel
-                              value={file.url}
-                              key={file.url}
-                              control={<Radio />}
-                              label={
-                                file.fileName.length < 50
-                                  ? file.fileName
-                                  : "..." + file.fileName.slice(-50)
+                      <Box>
+                        <FormControl>
+                          <RadioGroup
+                            name="availableFiles"
+                            value={selectedFile.url}
+                            onChange={(e) => {
+                              const found = files.find(
+                                (f) => f.url == e.target.value
+                              );
+                              if (found) {
+                                setSelectedFile(found);
                               }
-                            />
-                          ))}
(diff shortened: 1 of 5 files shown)


geoknee:
Looking good! 

My rapid feedback would be:
* things getting a little "busy" now with the addition of the second progress bar, I don't know how to combat that (could we use one of those very thin progress bars that sticks to the edge of the viewport... not sure. Or maybe just make it a bit thinner? Or use a circular progress bar for a bit of visual diversity. )
* be good to toggle the progress bar with the "micropayments" slider. As of now, the progress bar appears / disappears in a seemingly disjointed way (I guess when the file starts/stops fetching)
* be good to show the granularity of the payments somehow. Perhaps using graduated progress bar?

lalexgap:
> Looking good!
> 
> My rapid feedback would be:
> 
>     * things getting a little "busy" now with the addition of the second progress bar, I don't know how to combat that (could we use one of those very thin progress bars that sticks to the edge of the viewport... not sure. Or maybe just make it a bit thinner? Or use a circular progress bar for a bit of visual diversity. )

I did play around with a circular progress bar next to the download button. I did find it hard to notice the progress when the circle isn't large.
image


I wonder if keeping the progress bars consistent in their styling would help minimize how "busy" the UI feels?

>     * be good to toggle the progress bar with the "micropayments" slider. As of now, the progress bar appears / disappears in a seemingly disjointed way (I guess when the file starts/stops fetching)

If we'd like the progress bar to stick around it might  make sense to label it, since it looks kinda funny by itself without context.
image
 I kinda like that the progress bar only shows when a download is in progress, but I get that it makes the UI jump around.

>     * be good to show the granularity of the payments somehow. Perhaps using graduated progress bar?

I don't see an [out of the box](https://mui.com/material-ui/react-progress/) way to get a graduated progress bar but I'm sure it's just a matter of some css wizardy.

packages/payment-proxy-client/src/App.tsx
```diff
@@ -357,34 +384,43 @@ export default function App() {
...
+                                  file.fileName.length < 50
+                                    ? file.fileName
+                                    : "..." + file.fileName.slice(-50)
+                                }
+                              />
+                            ))}
+                          </RadioGroup>
+                        </FormControl>
+                      </Box>
+                      {useMicroPayments && payDisabled && (
```
lalexgap, line 416: If we remove `payDisabled` then the bar will always be displayed when use micropayments is checked.

geoknee: I am actually wondering now (given #1761) if we want the progress bar there all the time, and some other visual feedback to indicate the use of micropayments...

geoknee (approved):

PR #1758, opened 2023-09-19
Smaller payment channel budget for the demo

lalexgap:
Even with the larger deployed files we're using, the balance of the virtual channel is so large it's hard to see any change on the progress bars.

[See the deployment preview to see what it looks like](https://deploy-preview-1758--nitro-payment-demo.netlify.app/)


geoknee:
did I screw up my numbers somewhere 🤦

geoknee (approved):

PR #1759, opened 2023-09-19
Allow custom file names for demo

lalexgap:
Allows file names to be specified using a new env var, so we can use shorter names for hosted files. This makes things look a bit more presentable:
image


geoknee:
Is this an opportune moment to display the cost of each file on the screen?

geoknee (approved):

PR #1760, opened 2023-09-20
Allow docker containers to read config at runtime (not just build time)

### LINKED ISSUE #1752: Docker: Enable config `.toml` files to be read at runtime (not just build time)
We can use the `-v` flag to overwrite the config file to allow an image to be built once, and then run with many different configs. 

We could do that by using env vars, but that rather obviates config files which are a nice compact way of specifying config.

geoknee:
Fixes #1752

### DIFF
--- a/docker/local/Dockerfile
+++ b/docker/local/Dockerfile
@@ -14,4 +14,4 @@ COPY --from=builder /app/config.toml .
 
 EXPOSE 3005 4005
 
-CMD ["./nitro", "--config", "./config.toml"]
+CMD ./nitro --config ./config.toml
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ docker/cloud/start:
 	docker run -it -d --name go-nitro-cloud \
 	  -p 3005:3005 -p 4005:4005 -p 5005:5005 \
 		-e NITRO_CONFIG_PATH="./nitro_config/iris.toml" \
+		-v ./docker/cloud:/app/nitro_config \
 		go-nitro-cloud
 
 docker/cloud/push:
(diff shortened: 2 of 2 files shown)


docker/local/Dockerfile
```diff
@@ -14,4 +14,4 @@ COPY --from=builder /app/config.toml .
 
 EXPOSE 3005 4005
 
-CMD ["./nitro", "--config", "./config.toml"]
+CMD ./nitro --config ./config.toml
```
geoknee, line 17: @bitwiseguy I think in principle we ought to be able to have just the one Dockerfile. The only difference between cloud and local versions (I think) ought to be config. And since we can pass that at runtime, we don't need two different images?

There is one genuine "build time" decision, which is whether to include the UI (the cloud version does it, the local one does not).

bitwiseguy: Yes I agree we should use a single dockerfile and allow the user to configure them differently through the config file if they want.

There is also this [dockerfile](https://github.com/statechannels/go-nitro/blob/main/docker/Dockerfile) which I think is being unused and should be removed. It spins up three go-nitro nodes within the same docker container which seems like an abnormal use-case we probably don't have to support.

geoknee: Yes I agree your proposal is beneficial. 

Let's double check with @kerzhner that he is happy to lose the "three nodes in one container" Dockerfile.

bitwiseguy (approved):

docker/local/Dockerfile
kerzhner: The intention with "three nodes in one container" Docker image is to have a very easy to use and simple way for developers to run a sample nitro network. In practice, it is not being used. So we can remove it.

PR #1762, opened 2023-09-20
updat channel info immediately when payment leaves client

geoknee:
Especially when considering the comparison to micropayments, it will be nice if the effect of the payment is visible as soon as possible. 

Currently, we don't see the updated balance until the entire file is downloaded.

Here, I have updated the channel balance as soon as the request + voucher is launched. In principle, it could be even earlier but I think this is the most accurate thing to do.

### DIFF
--- a/packages/payment-proxy-client/src/file.ts
+++ b/packages/payment-proxy-client/src/file.ts
@@ -5,15 +5,18 @@ export async function fetchFile(
   url: string,
   paymentAmount: number,
   channelId: string,
-  nitroClient: NitroRpcClient
+  nitroClient: NitroRpcClient,
+  updateChannelCallback: () => void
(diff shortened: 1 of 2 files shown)


packages/payment-proxy-client/src/file.ts
```diff
@@ -5,15 +5,18 @@ export async function fetchFile(
   url: string,
   paymentAmount: number,
   channelId: string,
-  nitroClient: NitroRpcClient
+  nitroClient: NitroRpcClient,
+  updateChannelCallback: () => void
```
geoknee, line 9: I'm not fussed about how this is implemented just now / open to suggestions for a different approach.

lalexgap (approved):

PR #1763, opened 2023-09-20
rename paymentProgress => downloadProgress and always show download progress

geoknee:
and display progress unconditionally
without micropayments, progress will jump from 0 to 100

In future we could hook into the request to stream the progress visually, or we can leave as is to emphasize it is not micropayments

Based on #1755


geoknee:
In favour of #1778

PR #1781, opened 2023-09-21
Display png file in modal

geoknee:


cmd/sample-file-server/main.go
```diff
@@ -2,8 +2,10 @@ package main
 
 import (
 	"fmt"
+	"image"
+	"image/color"
+	"image/png"
 	"log"
-	"math/rand"
```
geoknee, line 6: Slight regression here in that all files are now identical. Probably not too hard to fix if we care a lot about that.

Issue #1767, opened 2023-09-20
Verify contract deployments on filscan


geoknee:
https://calibration.filscan.io/en/address/0xe32d4B5C5a80660710f6a2aD3cB1c11664138F34/

bitwiseguy:
Attempts so far:
* Filscan
  * produced a generic `status 500` error with no further information. Tough to troubleshoot with no further info.
* Filecoin [docs](https://docs.filecoin.io/smart-contracts/developing-contracts/verify-a-contract) (uses Remix)
  * Remix would not compile the flattened file without replacing aliased `Outcome` with `ExitFormat`
  *  `Error : Verification failed: mismatch between source code, compiler version, optimization parameters, contract constructor arguments, EVM version, etc.; the backend verification logic is incomplete and may not consider all scenarios`
* Hardhat [readme instructions](https://github.com/statechannels/go-nitro/blob/main/packages/nitro-protocol/readme.md?plain=1#L91)
  * Need to get an Etherscan and Infura API key before I can attempt this. Need for both of those service to support Filecoin Calibration testnet for this to work

geoknee:
Some other places to try: 

https://filecoin.io/blog/posts/the-fvm-imaginarium-block-explorers/#:~:text=Filfox,widely%2Dadopted%20explorers%20for%20Filecoin.

geoknee:
https://github.com/statechannels/go-nitro/issues/911

Issue #911, opened 2022-09-30
Single pathway for contract compilation


kerzhner:
From @geoknee: we have two compilation pathways in the repo. One via hardhat, and the other via solc directly (which produces the bytecode above). The two pathways are generally configured slightly differently (optimization runs, etc).

The two pathways should be collapsed into a single, canonical pathway.

geoknee:
I think it is possible to not generate bytecode (or not save it, at least) when we call out to `solc` and then `abigen`. It might break a couple of things but this would prevent us from having two sources of truth about bytecode in the repo?

PR #1789, opened 2023-09-21
Verify fvm contract

bitwiseguy:
This PR includes changes implemented in order to successfully verify a deployed contract on Filecoin Calibration testnet. This branch uses `hardhat` to deploy AND verify contracts, instead of trying to verify manually through a block explorer UI (which was a pain-staking, unsuccessful task).

The block explorer used for contract verification was [Starboard](https://fvm.starboard.ventures/calibration/explorer) because it was the only Filecoin Calibration explorer I could find with a readily accessible API and associated contract verification documentation. There is a nice etherscan hardhat plugin that can be used to verify against most common chains.

The `generate-nitro-adjudicator.sh` script was updated so that it uses the same `hardhat compile` command and settings as defined in `packages/nitro-protocol`. This prevents the repo from having two different smart contract compilation flows, which should help reduce the likelihood of errors when trying to verify contracts. This change closes https://github.com/statechannels/go-nitro/issues/911

This branch was used to deploy and verify a `NitroAdjudicator` contract on Calibration testnet: [link to contract](https://fvm.starboard.ventures/calibration/explorer/address/0xe1790Ea824035184a3Bf344e087FB61744992545). If we move the demo to that contract, this PR closes https://github.com/statechannels/go-nitro/issues/1767


lalexgap (approved):

package.json
```diff
@@ -22,5 +22,8 @@
     "lint": "yarn workspaces foreach -pt run lint",
     "test": "yarn workspaces foreach -pt run test"
   },
-  "packageManager": "yarn@3.5.1"
+  "packageManager": "yarn@3.5.1",
+  "dependencies": {
+    "ethers": "^6.7.1"
```
NiloCK, line 27: Curious: why was this added?

packages/nitro-protocol/hardhat-deploy/deploy-fvm.ts
```diff
@@ -2,24 +2,22 @@ import 'hardhat-deploy';
 import 'hardhat-deploy-ethers';
 
 import {HardhatRuntimeEnvironment} from 'hardhat/types';
-import {BigNumber} from 'ethers';
```
NiloCK, line 5: Same curiosity: was this import breaking the workflow for contract verification somehow? Or just a refactor to reduce the import of effectively duplicated type info?

NiloCK (approved):
Great :)

package.json
bitwiseguy: Good catch. Removed in https://github.com/statechannels/go-nitro/pull/1789/commits/c92fcf19641294135ba12cb8930fef100e3f62b0. It was added by mistake

packages/nitro-protocol/hardhat-deploy/deploy-fvm.ts
bitwiseguy: This was not breaking the verification workflow, rather it is more of an optimization. Instead of using `ethers`, which is a rather large package, we are using the `hardhat-ethers` package which is already imported

PR #1797, opened 2023-09-27
Add ws resubscribe retries to chain service

bitwiseguy:
Makes the `chainservice` more robust by adding a resubscribe retry loop whenever a websocket connection (either to chain events or new blocks) fails. Previously, the `chainservice` just had a single retry then cause a `panic`

### DIFF
--- a/node/engine/chainservice/eth_chainservice.go
+++ b/node/engine/chainservice/eth_chainservice.go
@@ -49,6 +49,11 @@ var topicsToWatch = []common.Hash{
 	challengeClearedTopic,
 }
 
+const (
+	MIN_BACKOFF_TIME = 1 * time.Second
+	MAX_BACKOFF_TIME = 5 * time.Minute
+)
+
 type ethChain interface {
 	bind.ContractBackend
 	ethereum.TransactionReader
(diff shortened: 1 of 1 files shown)


node/engine/chainservice/eth_chainservice.go
```diff
@@ -356,34 +363,55 @@ func (ecs *EthChainService) dispatchChainEvents(logs []ethTypes.Log) error {
...
+				ecs.logger.Warn("chain event subscription closed")
 			}
 
-			// Try to re-establish subscription once before failing
-			eventSub, err = ecs.chain.SubscribeFilterLogs(ecs.ctx, eventQuery, eventChan)
-			if err != nil {
-				errorChan <- fmt.Errorf("subscribeFilterLogs failed to resubscribe: " + err.Error())
-				break out
+			// Use exponential backoff loop to attempt to re-establish subscription
+			for backoffTime := MIN_BACKOFF_TIME; backoffTime < MAX_BACKOFF_TIME; backoffTime *= 2 {
```
bitwiseguy, line 388: Instead of exponential backoff should this attempt the re-connect at a consistent interval?

geoknee (approved):

node/engine/chainservice/eth_chainservice.go
geoknee: Seems marginal to me. I think [exponential backoff](https://en.wikipedia.org/wiki/Exponential_backoff) is fine, although I'm not sure if it is giving us much benefit over a consistent interval. It might help us avoid being rate limited by the chain endpoint.

PR #1801, opened 2023-10-03
Wire up "payment channel updated" events stream to demo

geoknee:
Updates the demo to hook the UI into events rather than using timeouts + polling.

### DIFF
--- a/packages/payment-proxy-client/src/App.tsx
+++ b/packages/payment-proxy-client/src/App.tsx
@@ -136,7 +136,7 @@ export default function App() {
       [hub],
       initialChannelBalance
     );
-    console.timeEnd("Create Payment Channel");
+    console.time("Create Payment Channel");
 
     // TODO: If the objective completes fast enough, we might start waiting after it's already done
     // await nitroClient.WaitForObjective(result.Id);
--- a/packages/payment-proxy-client/src/App.tsx
+++ b/packages/payment-proxy-client/src/App.tsx
@@ -142,11 +142,14 @@ export default function App() {
 
     setPaymentChannelId(result.ChannelId);
 
-    // nitroClient.PaymentChannelUpdated(result.ChannelId, setPaymentChannelInfo);
-    // const paymentChannel = await nitroClient?.GetPaymentChannel(
-    //   result.ChannelId
-    // );
-    // setPaymentChannelInfo(paymentChannel);
+    nitroClient.PaymentChannelUpdated(result.ChannelId, setPaymentChannelInfo);
+
+    // It's possible the channel updated before we registered the handler above, so
+    // query the channel once now to get the latest information:
+
+    setPaymentChannelInfo(
+      await nitroClient?.GetPaymentChannel(result.ChannelId)
+    );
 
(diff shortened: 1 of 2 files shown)


packages/payment-proxy-client/src/App.tsx
```diff
@@ -149,15 +141,18 @@ export default function App() {
...
     setPaymentChannelId(result.ChannelId);
-    updateChannelInfo(result.ChannelId);
-    console.timeEnd("Create Payment Channel");
 
-    // TODO: Slightly hacky but we wait a beat before querying so we see the updated balance
-    setTimeout(() => {
-      updateChannelInfo(result.ChannelId);
-    }, 1000);
-  };
+    nitroClient.PaymentChannelUpdated(result.ChannelId, setPaymentChannelInfo);
```
geoknee, line 145: Should we name it `.onPaymentChannelUpdated` ?

packages/payment-proxy-client/src/App.tsx
```diff
@@ -149,15 +141,18 @@ export default function App() {
...
+
+    // It's possible the channel updated before we registered the handler above, so
+    // query the channel once now to get the latest information:
 
+    setPaymentChannelInfo(
+      await nitroClient?.GetPaymentChannel(result.ChannelId)
+    );
```
geoknee, line 152: This is a pretty common issue and makes for pretty bad DX. Perhaps a smarter event queue could be used in place of an event emitter.

packages/payment-proxy-client/src/App.tsx
```diff
@@ -182,25 +177,17 @@ export default function App() {
             nitroClient,
             (progress) => {
               setDownloadProgress(progress);
-              updateChannelInfo(paymentChannelInfo.ID);
```
geoknee, line 185: This is the nub of the change. Instead of reaching out with http queries each time we think the channel will have changed, we just listen for events being pushed to us. This should cut down on a bit of latency, I think (round trip vs  half round trip).

packages/nitro-rpc-client/src/interface.ts
```diff
@@ -0,0 +1,140 @@
...
+  /**
+   * PaymentChannelUpdated attaches a callback which is triggered when the channel with supplied ID is updated.
+   *
+   * @param objectiveId - The id objective to wait for
+   */
+  PaymentChannelUpdated(
+    channelId: string,
+    callback: (info: PaymentChannelInfo) => void
+  ): Promise<void>;
+}
```
geoknee, line 117: Here's the new method. The rest of the changes in this file are just sugar.

geoknee:
Using the live demo, it doesn't look like the payments are reflected in the UI :-/ 

(Pretty sure I had that working locally, though)

geoknee:
> Using the live demo, it doesn't look like the payments are reflected in the UI :-/
> 
> (Pretty sure I had that working locally, though)

That's because I didn't redeploy the cloud nodes :-) 🤦

lalexgap (approved):

PR #1803, opened 2023-10-04
Add payment channel updated stream to ts rpc client

### LINKED ISSUE #1770: Add `PaymentChannelUpdated` event stream to TS rpc client
We already have this on the Go client, so it shouldn't be too much work.

geoknee:
Closes #1770 

- [x] Should we name it .onPaymentChannelUpdated ?

There is very little in the way of unit or integration tests setup at present. Therefore I have filed #1806, and we should make sure to cover this event stream in that work.

### DIFF
--- a/packages/nitro-rpc-client/src/interface.ts
+++ b/packages/nitro-rpc-client/src/interface.ts
@@ -105,6 +105,15 @@ interface syncAPI {
    * @param objectiveId - The id objective to wait for
    */
   WaitForObjective(objectiveId: string): Promise<void>;
+  /**
+   * PaymentChannelUpdated attaches a callback which is triggered when the channel with supplied ID is updated.
+   *
+   * @param objectiveId - The id objective to wait for
+   */
+  PaymentChannelUpdated(
+    channelId: string,
+    callback: (info: PaymentChannelInfo) => void
+  ): Promise<void>;
 }
 
 export interface RpcClientApi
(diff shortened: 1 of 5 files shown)


bitwiseguy (approved):

geoknee:
@bitwiseguy any opinion on this 

>  Should we name it .onPaymentChannelUpdated ?

packages/nitro-rpc-client/src/interface.ts
```diff
@@ -0,0 +1,140 @@
...
+  WaitForObjective(objectiveId: string): Promise<void>;
+  /**
+   * PaymentChannelUpdated attaches a callback which is triggered when the channel with supplied ID is updated.
+   *
+   * @param objectiveId - The id objective to wait for
+   */
+  PaymentChannelUpdated(
+    channelId: string,
+    callback: (info: PaymentChannelInfo) => void
+  ): Promise<void>;
```
geoknee, line 116: Here's what I am wondering about the naming of.

Thinking about the usage, which of the following makes the most sense
```ts
nitroClient.PaymentChannelUpdated("0xabc",()=>{dothing()})
nitroClient.onPaymentChannelUpdated("0xabc",()=>{dothing()})
```

bitwiseguy: Ok I got it. I don't have strong preference, but lean towards:
```go
nitroClient.onPaymentChannelUpdate("0xabc",()=>{dothing()})
```

geoknee: [1b547d2](https://github.com/statechannels/go-nitro/pull/1803/commits/1b547d2a15d9c3c63cb87ce79c1cf6e57f946691)

PR #1804, opened 2023-10-04
Improve ts-rpc-test config

geoknee:


.github/workflows/ts-rpc-test.yml
```diff
@@ -14,7 +16,7 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           cache: "yarn"
-          node-version: "18.15.0"
+          node-version: "20.0.0"
```
lalexgap, line 19: Could we make this change across all our github workflows, to keep things consistent?

lalexgap (approved):

.github/workflows/ts-rpc-test.yml
geoknee: Yep [fbdcc92](https://github.com/statechannels/go-nitro/pull/1804/commits/fbdcc9233a69e7cd80c97671178689d7871237d4)

Issue #1509, opened 2023-08-02
Sign DHT custom namespace messages with SC Key


lalexgap:
In #1492 we're publishing a custom DHT record that contains our peer id and SC address. The record is signed with the 
message key.

However this does mean anyone could publish a DHT record with any SC address they want.  This opens up an attack vector where a malicious actor could publish a bunch DHT records for a SC address they want to interfere with.  

To resolve this we could sign the DHT record with the SC key, so we're guaranteed that the DHT record is published by the owner of the SC address.To do this we have to consider how to pass the SC key into the message service. Ideally we don't want to tie the message service to the SC key, so it may make sense to expose some kind of Signing API, that the message service can call to sign the record with the SC key.

Issue #1508, opened 2023-08-02
Switch tests to use a seperate message key


lalexgap:
Currently in most of our tests we use the[ SC private key as the message key](https://github.com/statechannels/go-nitro/blob/f1aae3d4e5a497fc580e0710960d58177bf5746b/internal/rpc/rpc.go#L81).  However we should probably move away from that and generate a specific message key for the message service, as this is a more realistic option.

We also currently assume that the pk will be a [Secp256k1 key](https://github.com/statechannels/go-nitro/blob/f1aae3d4e5a497fc580e0710960d58177bf5746b/node/engine/messageservice/p2p-message-service/service.go#L90) but if the message key is going to be separate from the SC key then there's no reason for that limitation.

Issue #1075, opened 2023-01-27
Do not pass private key directly to p2p messaging service. 


geoknee:
This doesn't align with best practices around handling private keys:

https://github.com/statechannels/go-nitro/blob/92d6f4d92c4e567181216a0517b631e7f6f65e05/client/engine/messageservice/p2p-message-service/service.go#L76

See https://www.notion.so/statechannels/RFC-8-Securing-Private-StateChannel-Key-c4f2945172d84db2b84220d64502e38c for details. 

Apparently the motivation was to create a deterministic message key from the channel signing key.
I believe the message key is used to encrypt or sign messages. 

Alternatives are:
* pass in the hash of the private key
* pass in the corresponding public key
* ...

These retain the desired property of a deterministic message key.

lalexgap:
This is related to #1495, passing in the SC key directly allows us to recover the SC address from the peer ID.

PR #1810, opened 2023-10-05
Sign dht record with scAddr

bitwiseguy:
Closes https://github.com/statechannels/go-nitro/issues/1509

The implementation in this PR uses the following flow to sign the dht record with the state-channel address:
1. `messageservice` sends `signatureRequest` on a new channel (requests includes a dedicated response channel)
2. New case statement on `engine`'s `run` loop listens to that channel
3. `engine` processes the `signatureRequest` by retrieving the `secretKey` from the `store`, then returns the signature on the dedicated response channel

This implementation (especially the use of a one-time, dedicated response channel instead of a simple function call and return value) feels a bit hacky to me. However, it seemed like the best way to perform signatures while keeping the `messageservice` decoupled from the `engine` and `store`.

### DIFF
--- a/node/engine/engine.go
+++ b/node/engine/engine.go
@@ -3,6 +3,8 @@ package engine // import "github.com/statechannels/go-nitro/node/engine"
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"log/slog"
@@ -10,11 +12,13 @@ import (
 	"sync"
 	"time"
 
+	"github.com/ethereum/go-ethereum/crypto/secp256k1"
 	"github.com/statechannels/go-nitro/channel"
 	"github.com/statechannels/go-nitro/channel/consensus_channel"
 	"github.com/statechannels/go-nitro/internal/logging"
 	"github.com/statechannels/go-nitro/node/engine/chainservice"
 	"github.com/statechannels/go-nitro/node/engine/messageservice"
+	p2pms "github.com/statechannels/go-nitro/node/engine/messageservice/p2p-message-service"
 	"github.com/statechannels/go-nitro/node/engine/store"
 	"github.com/statechannels/go-nitro/node/query"
 	"github.com/statechannels/go-nitro/payments"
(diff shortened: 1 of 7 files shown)


node/engine/messageservice/messageservice.go
```diff
@@ -1,11 +1,16 @@
...
+	"github.com/statechannels/go-nitro/protocols"
+)
 
 type MessageService interface {
-	// Out returns a chan for receiving messages from the message service
-	Out() <-chan protocols.Message
+	// P2PMessages returns a chan for receiving messages from the message service
+	P2PMessages() <-chan protocols.Message
+	// SignRequests returns a chan for receiving signature requests from the message service
+	SignRequests() <-chan p2pms.SignatureRequest
```
geoknee, line 13: Do we want the general `messageservice` interface to depend on a type from the `p2pms` package? Maybe we should hoist this type up to this package, instead? That would also prevent `test-messageservice` depending on `p2pms`.

node/engine/engine.go
```diff
@@ -61,9 +65,10 @@ type Engine struct {
 	ObjectiveRequestsFromAPI chan protocols.ObjectiveRequest
 	PaymentRequestsFromAPI   chan PaymentRequest
 
-	fromChain  <-chan chainservice.Event
-	fromMsg    <-chan protocols.Message
-	fromLedger chan consensus_channel.Proposal
+	fromChain    <-chan chainservice.Event
+	fromMsg      <-chan protocols.Message
+	fromLedger   chan consensus_channel.Proposal
+	signRequests <-chan p2pms.SignatureRequest
```
geoknee, line 71: Do you think we will be able to reuse this interface to solve these issues:
https://github.com/statechannels/go-nitro/issues/1508
#1075

geoknee (approved):

node/engine/engine.go
bitwiseguy: ### Issue 1508
We will likely need to add another config option for the `libp2p` peerId private key (which is used to derive the peerId). We need to pass that private key to the `libp2p` config options in the following code:
```
	options := []libp2p.Option{
		libp2p.Identity(privateKey),
		...
	}
```
Right now the `messageservice` holds that private key in-memory. To improve consistency, we could put that private key in the `store` (as we do with the scAddr). We could then update the `engine.handleSignRequests` method to make both signatures on the DHT record:
1. Sign the record using the scAddr private key
2. Sign the record using the peerId private key
3. Return both signatures to the `messageservice` so that the record can be published to the DHT

### Issue 1075
I'm not sure how to resolve this issue. It seems that `libp2p` requires the peerId private key during its configuration, and the key is used within `libp2p` abstractions to sign DHT records and encrypt p2p communication. By using separate keys for the peerId and scAddr (as discussed above in issue 1508), we would at least mitigate risk by reducing responsibilities/purposes of each key

PR #1815, opened 2023-10-09
add HTLCs to ledger channel update protocol

NiloCK:
This is draft work toward including HTLC transfers as part of the base `consensus_channel`.

Strongest caveat: there is no on-chain implementation to enforce any of these data types.


geoknee:
This is looking good so far @NiloCK !

I think we probably want:
- [ ]  an ADR to accompany these changes, explaining how LN-compatible multi hop payments can be incorporated into go-nitro.
- [ ] some unit test for the functionality you built so far
- [ ] eventually an integration test involving any changes to the go-nitro API (etc). We could develop that as a feature branch and merge it all in one go

NiloCK:
:+1: 

Will probably shelve this until after the hackathon.

Issue #1798, opened 2023-09-27
Analyze polling chain instead of using websocket conns


bitwiseguy:
CC: @lalexgap 

The `chainservice` currently has two websocket subscriptions:
1. Events emitted by the `NitroAdjudicator` contract
2. New blocks mined by the chain

The websocket connections frequently (once every couple of hours?) receive errors causing the connection to break, and forcing the `chainservice` to re-establish the connection. A common error message on those connections is `i/o timeout`.

We should investigate an alternative approach where the `chainservice` polls the chain instead of relying on a constant websocket connection. The polling frequency could be set to 0.5-1x the block time to ensure the `chainservice` does not lag behind too much.

## Pros of polling
1. Simplifies `chainservice` code substantially. Polling is less prone to many of the race conditions that the current implementation is susceptible to. Switching to polling would allow us to remove lots of logic currently in place to protect against race conditions.
2. Http requests seem to have a lower failure rate compared to websocket connections. If we switch to polling we could also remove code related to a [workaround involving a known issue](https://github.com/statechannels/go-nitro/blob/main/node/engine/chainservice/eth_chainservice.go#L383) with the ethereum event subscription.
3. Http requests targeting the same RPC provider likely scale better than adding additional websocket connections to the same RPC provider

## Cons of polling
1. Less performant (events will be detected later than they are on websockets). However, if the polling frequency is set to 0.5x the block time, we can reduce the maximum delay between event emission and nitro detection to 0.5x the block time (0.5 x 10sec on Filecoin = 5 sec max delay)

PR #1819, opened 2023-10-09
Fix bug in chain event resubscribe

bitwiseguy:
The bug fixed in this PR was causing the cloud nodes to crash repeatedly because even when resubscribing to chain events was successful, an error was being sent on the `errorChan`, which triggers a panic. The fix involves the use of a `resubscribed` boolean flag to indicate if an error should be sent on the `errorChan` or not

### DIFF
--- a/node/engine/chainservice/eth_chainservice.go
+++ b/node/engine/chainservice/eth_chainservice.go
@@ -372,46 +372,49 @@ func (ecs *EthChainService) listenForEventLogs(errorChan chan<- error, eventChan
 			return
 
 		case err := <-ecs.eventSub.Err():
-			latestBlockNum := ecs.GetLastConfirmedBlockNum()
+			// Use helper function block to ensure "defer" statement is called for all exit paths
+			func() {
+				latestBlockNum := ecs.GetLastConfirmedBlockNum()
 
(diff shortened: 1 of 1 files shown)


geoknee (approved):

node/engine/chainservice/eth_chainservice.go
```diff
@@ -372,41 +372,49 @@ func (ecs *EthChainService) listenForEventLogs(errorChan chan<- error, eventChan
...
-			errorChan <- fmt.Errorf("subscribeFilterLogs failed to resubscribe")
-			return
+				if !resubscribed {
+					ecs.logger.Error("subscribeFilterLogs failed to resubscribe")
+					errorChan <- fmt.Errorf("subscribeFilterLogs failed to resubscribe")
+					return
+				}
+			}()
 
 		case <-time.After(RESUB_INTERVAL):
```
bitwiseguy, line 419: Probably out of scope for this PR, but a thought that came to mind: we are intentionally closing the websocket connection every 15 seconds, then catching that error and resubscribing. It seems like polling the chain every 15 seconds instead of setting up a 15-second long subscription would be more efficient. Further details noted in this issue: https://github.com/statechannels/go-nitro/issues/1798

PR #1829, opened 2023-10-12
Fix race condition in rpc client / test

### LINKED ISSUE #1749: RPC Client: `WaitForObjective` sometimes never resolves
This means in our demo we have to just optimistically hope the channel gets create successfully. 

```go
    const result = await nitroClient.CreatePaymentChannel(
      provider,
      [hub],
      initialChannelBalance
    );

// Wait for 1 sec before waiting for the objective to be created
    await new Promise<void>((resolve) => {
      setTimeout(() => {
        resolve();
      }, 1000); 
    await nitroClient.WaitForObjective(result.Id);
```

We should audit how this method (and the corresponding method in the Go rpc client `ObjectiveCompletedChan`) operate.
In the typescript client, there seems to be no mechanism to deal with this race condition. 

In the Go client, there is a mechanism in the sense that the client is listening for RPC events signalling objectives completing. When those events come in, it will store a closed channel internally, which can be picked up later by the user (who will immediately unblock):

https://github.com/statechannels/go-nitro/blob/1046f9862e3d6442d0b24fae16635af6d37cb671/rpc/client.go#L256-L264

This is still not perfect, since the information is only there if the objective completed _since the client was instantiated_. The real gold standard would involve handling the case where the client was instantiated afterwards, too. 

The TS client is worse: it only registers a listener for objectives completing when `WaitForObjective` is called:

https://github.com/statechannels/go-nitro/blob/e19c9fb899dd729535ff040daafbe590d666811c/packages/nitro-rpc-client/src/rpc-client.ts#L71-L87

(Side question: should we be removing the listener when the promise resolves, here?)

Perhaps the simplest solution would be for the API method to query the rpc server once and resolve immediately if the objective has already completed.

geoknee:
Closes #1749 

There are a number of ways to solve this issue (essentially that the `ObjectiveCompleted` event can fire before an event listener is registered for it). One would be to cache completed objectives in the client (which is how the go rpc client does it). 

I went another way, since that solution will only work if the event fired after the client itself is started. Here we attach the listener and perform a static query for the information straight afterward (with this being abstracted inside the rpc client). 

Further, since we do not have a query path for objectives, I have flipped over the events and queries to be **channel-centric** and not **objective-centric**. This is mostly justified by expediency, but I think also it simplifies the API to not have objectives as an additional concept for developers to understand. 

- [ ] We may want to make a similar change to the go rpc client
- [ ] at the reviewer's request I can pinch off some of the side effects of this PR: e.g. the colorized / human friendly logs

### DIFF
--- a/.github/workflows/ts-rpc-test.yml
+++ b/.github/workflows/ts-rpc-test.yml
@@ -29,16 +29,16 @@ jobs:
         run: make ui/build
 
       - name: Run go-nitro RPC servers with GUI
-        run: go run ./cmd/start-rpc-servers -ui=true &> output1.log &
+        run: go run ./cmd/start-rpc-servers -ui=true &> output.log &
 
       - name: Run Create Channels script
         # TODO: We could write a test specific script that creates channels and checks the results
-        run: npx ts-node ./scripts/client-runner.ts create-channels -w 300000 &> output2.log
+        run: npx ts-node ./scripts/client-runner.ts create-channels -w 300000
         working-directory: packages/nitro-rpc-client
 
       - name: Archive logs
         if: always()
         uses: actions/upload-artifact@v2
         with:
-          name: logs
-          path: ./**/*.log
+          name: rpc server logs
+          path: ./output.log
--- a/.github/workflows/ts-rpc-test.yml
+++ b/.github/workflows/ts-rpc-test.yml
@@ -29,11 +29,11 @@ jobs:
         run: make ui/build
 
       - name: Run go-nitro RPC servers with GUI
-        run: go run ./cmd/start-rpc-servers -ui=true &> output.log &
+        run: go run ./cmd/start-rpc-servers -ui=true &> output1.log &
 
       - name: Run Create Channels script
         # TODO: We could write a test specific script that creates channels and checks the results
-        run: npx ts-node ./scripts/client-runner.ts create-channels -w 300000 &> output.log
+        run: npx ts-node ./scripts/client-runner.ts create-channels -w 300000 &> output2.log
         working-directory: packages/nitro-rpc-client
 
       - name: Archive logs
--- a/internal/logging/logging.go
+++ b/internal/logging/logging.go
@@ -6,7 +6,9 @@ import (
 	"log/slog"
 	"os"
 	"path/filepath"
+	"time"
 
+	"github.com/lmittmann/tint"
 	"github.com/statechannels/go-nitro/protocols"
 	"github.com/statechannels/go-nitro/types"
 )
@@ -64,6 +66,9 @@ func SetupDefaultFileLogger(filename string, level slog.Level) {
 
 // SetupDefaultLogger sets up a default logger that writes to the specified writer
 func SetupDefaultLogger(w io.Writer, level slog.Level) {
-	h := slog.NewJSONHandler(w, &slog.HandlerOptions{Level: level})
+	h := tint.NewHandler(w, &tint.Options{
+		Level:      level,
+		TimeFormat: time.Kitchen,
+	})
 	slog.SetDefault(slog.New(h))
 }
--- a/packages/nitro-rpc-client/src/interface.ts
+++ b/packages/nitro-rpc-client/src/interface.ts
@@ -101,22 +101,22 @@ interface paymentApi {
 
 interface syncAPI {
   /**
-   * WaitForLedgerChannelToHaveStatus blocks until the ledger channel with the given ID to have the given status.
+   * WaitForLedgerChannelStatus blocks until the ledger channel with the given ID to have the given status.
    *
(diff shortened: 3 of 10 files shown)


internal/logging/logging.go
```diff
@@ -64,6 +66,9 @@ func SetupDefaultFileLogger(filename string, level slog.Level) {
 
 // SetupDefaultLogger sets up a default logger that writes to the specified writer
 func SetupDefaultLogger(w io.Writer, level slog.Level) {
-	h := slog.NewJSONHandler(w, &slog.HandlerOptions{Level: level})
+	h := tint.NewHandler(w, &tint.Options{
+		Level:      level,
+		TimeFormat: time.Kitchen,
+	})
```
geoknee, line 72: gives output like this: 
Screenshot 2023-10-12 at 10 52 14

packages/nitro-rpc-client/src/rpc-client.ts
```diff
@@ -57,17 +59,47 @@ export class NitroRpcClient implements RpcClientApi {
...
+    const promise = new Promise<void>((resolve) => {
       this.transport.Notifications.on(
-        "objective_completed",
-        (params: ObjectiveCompleteNotification["params"]) => {
-          if (params["payload"] === objectiveId) {
-            resolve();
+        "payment_channel_updated",
+        (payload: PaymentChannelUpdatedNotification["params"]["payload"]) => {
+          if (payload.ID === channelId) {
+            this.GetLedgerChannel(channelId).then((l) => {
```
bitwiseguy, line 92: Should this be `this.GetPaymentChannel` instead of `this.GetLedgerChannel`?

.github/workflows/ts-rpc-test.yml
```diff
@@ -33,12 +33,12 @@ jobs:
 
       - name: Run Create Channels script
         # TODO: We could write a test specific script that creates channels and checks the results
-        run: npx ts-node ./scripts/client-runner.ts create-channels -w 300000 &> output.log
```
bitwiseguy, line 36: Does this get rid of the duplicate log files that are archived by the CI?

packages/nitro-rpc-client/src/interface.ts
```diff
@@ -100,11 +101,25 @@ interface paymentApi {
...
 interface syncAPI {
   /**
-   * WaitForObjective blocks until the objective with the given ID to complete.
+   * WaitForLedgerChannelToHaveStatus blocks until the ledger channel with the given ID to have the given status.
    *
-   * @param objectiveId - The id objective to wait for
+   * @param objectiveId - The channel id to wait for
+   * @param status - The channel id to wait for (e.g. Ready or Closing)
+   */
+  WaitForLedgerChannelToHaveStatus(
```
bitwiseguy, line 109: Nit: could change the function name to `WaitForLedgerChannelStatus` to be a bit more concise

packages/nitro-rpc-client/src/rpc-client.ts
```diff
@@ -57,17 +59,47 @@ export class NitroRpcClient implements RpcClientApi {
...
+        (payload: LedgerChannelUpdatedNotification["params"]["payload"]) => {
+          if (payload.ID === channelId) {
+            this.GetLedgerChannel(channelId).then((l) => {
+              if (l.Status == status) resolve();
+            });
+          }
+        }
+      );
+    });
+    const ledger = await this.GetLedgerChannel(channelId);
```
bitwiseguy, line 78: Is there need for both this `GetLedgerChannel` call and the same one above nested inside the `Notifications` handler?

packages/nitro-rpc-client/src/rpc-client.ts
geoknee: Yes! good catch.

.github/workflows/ts-rpc-test.yml
geoknee: Yes we now upload a single log file for the servers, and print the client's output straight to the console.

packages/nitro-rpc-client/src/rpc-client.ts
geoknee: What I am trying to do here is make it impossible for there to be a race condition. So the approach is, 
Step 1: register a handler in case the event has not yet happened (if it already happened, this will never fire)
Step 2: make a static query in case the event has already happened. 

Which should result in the promise resolving whenever the event fires / was fired. 

Does that make sense?

packages/nitro-rpc-client/src/interface.ts
geoknee: [6e06584](https://github.com/statechannels/go-nitro/pull/1829/commits/6e0658447119880bdc2243e0251e422d8a408113)

PR #1831, opened 2023-10-13
feat: checkpointed event

### LINKED ISSUE #1830: Consider adding `checkpointed` event
Currently `ForceMove` has `ChallengeCleared` event, which is emitted on every `checkpoint(...)` call. However, checkpoint can be invoked with no active challenge, so event name `ChallengeCleared` is confusing.

Nevertheless, I think we should have both `ChallengeCleared` and `Checkpointed` events, as it will be convenient to have different handlers for those events.

nksazonov:
Closes #1830.

### DIFF
--- a/packages/nitro-protocol/test/contracts/ForceMove/checkpoint.test.ts
+++ b/packages/nitro-protocol/test/contracts/ForceMove/checkpoint.test.ts
@@ -93,8 +93,8 @@ describe('checkpoint', () => {
   it.each`
     description | largestTurnNum                         | support              | finalizesAt  | reason
     ${accepts1} | ${turnNumRecord + 1}                   | ${valid}             | ${undefined} | ${undefined}
-    ${accepts2} | ${turnNumRecord + 1}                   | ${valid}             | ${never}     | ${undefined}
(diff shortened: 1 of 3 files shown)


packages/nitro-protocol/test/contracts/ForceMove/checkpoint.test.ts
```diff
@@ -93,8 +93,8 @@ describe('checkpoint', () => {
   it.each`
     description | largestTurnNum                         | support              | finalizesAt  | reason
     ${accepts1} | ${turnNumRecord + 1}                   | ${valid}             | ${undefined} | ${undefined}
-    ${accepts2} | ${turnNumRecord + 1}                   | ${valid}             | ${never}     | ${undefined}
-    ${accepts3} | ${turnNumRecord + 1 + participantsNum} | ${valid}             | ${future}    | ${undefined}
+    ${accepts2} | ${turnNumRecord + 1}                   | ${valid}             | ${future}    | ${undefined}
+    ${accepts3} | ${turnNumRecord + 1 + participantsNum} | ${valid}             | ${never}     | ${undefined}
```
nksazonov, line 97: The description of `accepts2` test case states a challenge must be in progress, so `finalizesAt` should be `future`.

geoknee (approved):

PR #1837, opened 2023-11-14
Fix patsy authentication code

geoknee:


.github/workflows/deploy-patsy.yml
```diff
@@ -41,7 +41,7 @@ jobs:
     steps:
       - name: Deploy to Droplet
         env:
-          DO_API_KEY: ${{ secrets.DO_API_KEY }}}
+          DO_API_KEY: ${{ secrets.DO_API_KEY }}
```
geoknee, line 44: This typo means the deployment would never actually succeed, since the droplet would fail to authenticate with the container registry. 

In fact this bug has been here since the existence of the action! So we haven't actually been deploying any code changes for the last two months https://github.com/statechannels/go-nitro/blame/main/.github/workflows/deploy-patsy.yml#L44.

Changes to configuration _were_ being deployed, which made this easier to miss. 

To follow up, we need to make sure the github action fails when t[he registry auth fails](https://github.com/statechannels/go-nitro/actions/runs/6859634003/job/18658091029).

lalexgap (approved):

PR #1839, opened 2024-01-03
build(nitro-protocol): upgrade Solidity version

maxpushka:
**Upgrade Solidity version**

**Body**

- Upgraded and pinned version (with `^`) in `solidity pragma`
- Updated gas benchmarks
- Generated adjudicator bindings

Restricting major compiler version allow contract inheritance to compile with higher solidity versions, loosely coupling the versions. Hence, the risk of introducing breaking changes into the project is lowered. Major version increments (as per semantic versioning) usually include changes that are not backwards-compatible. Using the caret symbol helps maintain stability in the project.

**How Has This Been Tested?**

Tests were run with `yarn test`

---

#### Code quality


NiloCK:
Updating compiler target is OK, and I'm also fine with including the `^` modifier.

Q: should _all_ compiler pragmas include the `^`? Seems right now like only the "core" contracts / interfaces do.



Problem: after running the tests, this seems to have broken some dependencies.

First guess at root cause is that the simulatedbackend service used in the tests [does not implement PUSH0](https://github.com/statechannels/go-nitro/actions/runs/7399284856/job/20199741000?pr=1839#step:8:456), which was added in 0.8.20. See: https://medium.com/coinmonks/push0-opcode-a-significant-update-in-the-latest-solidity-version-0-8-20-ea028668028a

First guess at a fix would be to update the go-ethereum dependencies that the SimulatedBackend uses.

